PR #15, opened 2026-04-27
feat: deploy monitor — pod health detection and auto-remediation

kalantar:
## Summary

- Adds `pipeline/lib/health.py`: pod health detection library with three-tier triage (auto-remediate Evicted/OOMKilled pods, rules-based suggestions, Anthropic API diagnosis for novel failures)
- Adds `pipeline/monitor.py`: CLI monitor that runs alongside `deploy.py run`, watches active namespace slots, and writes all findings to `workspace/runs/<run>/health_report.md`
- Updates `pipeline/README.md` with `monitor.py` usage, flags, and `ANTHROPIC_API_KEY` requirement
- 42 new tests (263 total passing)

## Usage

```bash
# Start in a second terminal alongside deploy.py run
python pipeline/monitor.py --experiment-root ../admission-control

# Or background it
python pipeline/monitor.py --experiment-root ../admission-control &
```

Set `ANTHROPIC_API_KEY` in the environment for tier-3 API diagnosis. Without it, tier-3 findings are recorded with a placeholder.

## How triage works

| Tier | Trigger | Action |
|------|---------|--------|
| 1 | Evicted pod, OOMKilled (attempts 1–2) | Delete pod (auto-restart) |
| 2 | OOMKilled (attempt 3+), ImagePullBackOff, pending GPU, quota exceeded | Log suggestion |
| 3 | CrashLoopBackOff, StartupProbeFailure, unknown signals | Call Anthropic API, write diagnosis |

## Test plan

- [ ] All 263 tests pass: `python -m pytest pipeline/ -v`
- [ ] Ruff clean: `ruff check pipeline/ --select F`
- [ ] Manually start monitor alongside a real run and verify `health_report.md` is created
- [ ] Verify monitor exits cleanly when all pairs complete

### DIFF
--- a/docs/superpowers/plans/2026-04-27-deploy-monitor.md
+++ b/docs/superpowers/plans/2026-04-27-deploy-monitor.md
@@ -1,0 +1,1392 @@
+# Deploy Monitor Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build `pipeline/monitor.py`, a standalone script that watches active namespace slots alongside `deploy.py run`, auto-remediates transient pod failures, emits rules-based suggestions for known signatures, and calls the Anthropic API for novel failures, writing all findings to `health_report.md`.
+
+**Architecture:** A poll loop reads `progress.json` to discover active slots, queries Kubernetes for pod states and events, and applies three-tier triage: tier 1 auto-remediates (evicted/OOMKilled pods), tier 2 emits rules-based suggestions, tier 3 calls the Anthropic API. All findings are written to `health_report.md`. The monitor exits when all pairs leave `running` state.
+
+**Tech Stack:** Python 3.10+, `anthropic` SDK, `kubectl` via subprocess, existing `pipeline/lib/progress.py`.
+
+---
+
+## File Map
+
+| File | Status | Purpose |
+|---|---|---|
+| `pipeline/lib/health.py` | Create | Pod/event parsing, triage logic, kubectl wrappers |
+| `pipeline/monitor.py` | Create | CLI, poll loop, HealthReport, Anthropic API calls |
+| `pipeline/tests/test_health.py` | Create | Tests for health.py (pure parsing + triage) |
+| `pipeline/tests/test_monitor.py` | Create | Tests for monitor.py (slot discovery, HealthReport) |
+| `requirements.txt` | Modify | Add `anthropic>=0.25.0` |
+| `pipeline/README.md` | Modify | Document monitor.py usage |
+
+### Spec deviations (intentional)
+
+- **Pod label vs name filter:** The spec says `kubectl get pods -l modelLabel=sim2real-<experimentId>`. This plan filters pods by name containing `experiment_id` instead — more robust since the exact Helm label key is not verified. If the label is confirmed, swap the `get_pods` filter to use `-l` for precision.
+- **PipelineRun task status:** The spec lists this as signal source 4. It is not used in any triage rule and is deferred — the pod-level signals cover all tier-1/2/3 decisions in this version.
+- **Pending > 10 minutes (tier 3):** The spec mentions "persistent Pending with no clear event reason" as a tier-3 trigger. This requires tracking pod state across poll cycles and is deferred. Currently, a Pending pod with no matching events is silently skipped.
+
+---
+
+## Task 1: Scaffold health.py data structures
+
+**Files:**
+- Create: `pipeline/lib/health.py`
+- Create: `pipeline/tests/test_health.py`
+- Modify: `requirements.txt`
+
+- [ ] **Step 1: Add anthropic to requirements.txt**
+
+  Append one line to `requirements.txt`:
+  ```
+  anthropic>=0.25.0
+  ```
+  Then install: `pip install anthropic`
+
+- [ ] **Step 2: Write failing import tests**
+
+  Create `pipeline/tests/test_health.py`:
+  ```python
+  """Tests for pipeline.lib.health."""
+  import json
+  import pytest
+  from unittest.mock import patch, MagicMock
+
+
+  def test_pod_state_dataclass():
+      from pipeline.lib.health import PodState
+      pod = PodState(name="test-pod", phase="Running", ready=True,
+                     restart_count=0, reason="", message="")
+      assert pod.name == "test-pod"
+      assert pod.ready is True
+
+
+  def test_event_record_dataclass():
+      from pipeline.lib.health import EventRecord
+      evt = EventRecord(reason="FailedScheduling", message="no nodes",
+                        count=1, last_timestamp="2026-01-01T00:00:00Z",
+                        involved_object="my-pod")
+      assert evt.reason == "FailedScheduling"
+
+
+  def test_triage_result_dataclass():
+      from pipeline.lib.health import TriageResult
+      r = TriageResult(tier=1, message="deleted", suggestion="",
+                       needs_logs=False, action="delete_pod")
+      assert r.tier == 1
+  ```
+
+- [ ] **Step 3: Run test to verify it fails**
+
+  ```bash
+  python -m pytest pipeline/tests/test_health.py::test_pod_state_dataclass -v
+  ```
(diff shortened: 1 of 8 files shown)


kalantar (commented):
## PR Review

Overall design is solid — clean three-tier triage, good separation between `health.py` (pure logic + kubectl wrappers) and `monitor.py` (orchestration + API), and the test suite structure is good. Several issues need addressing before merge.

---

## Critical (must fix before merge)

### 1. Wrong Claude model ID — every tier-3 call fails at runtime
`monitor.py` — `_DIAGNOSIS_MODEL = "claude-haiku-4-5-20251001"` uses a non-existent model ID. The correct date suffix is `20251022`. Every `_diagnose_with_api` call will fail with a model-not-found error.

### 2. Monitor exits prematurely — misses `"pending"` and `"collecting"` states
`monitor.py:_work_remaining` only checks `status == "running"`, but `deploy.py` cycles pairs through `running → pending → running` on retries. The monitor exits during the gap.
```python
# Fix:
return any(v.get("status") in ("pending", "running", "collecting") for v in progress.values())
```

### 3. `tracker.reset()` fires for unmatched states, defeats OOM escalation
`monitor.py:~170` — `tracker.reset(pod.name)` is called whenever `triage_pod` returns `None`, which includes pods that don't match any triage rule (not just healthy pods). A pod that OOMKills, enters an unrecognized state, then OOMKills again gets attempt-1 treatment indefinitely.
```python
# Fix:
if pod.ready:
    tracker.reset(pod.name)
```

### 4. Unguarded `json.loads` in `parse_pods`/`parse_events` crashes poll cycle
`health.py:58,92` — malformed kubectl output (or a `kubectl` exit-0 `Status` error body) raises `json.JSONDecodeError` that propagates out of `_poll_once`, crashing the entire poll cycle for all namespaces, not just the failing one.

### 5. `store.load()` `ValueError` permanently kills the monitor
`monitor.py` main loop — does not catch `ValueError` from `progress.py`. A single transient write corruption by `deploy.py` permanently stops all monitoring with no recovery.
```python
# Fix: wrap in try/except ValueError, log warning, continue
```

### 6. Bare `except Exception` on Anthropic API swallows all errors silently
`monitor.py:_diagnose_with_api` — swallows `AuthenticationError`, `RateLimitError`, `APIConnectionError`, and `IndexError` on empty `content[]` with no logging. Under sustained rate-limiting every tier-3 finding silently degrades to the error placeholder with no indication to the user.

### 7. Unguarded file I/O on `health_report.md`
`monitor.py:66,119` — `write_text` and `read_text` have no `OSError` handler. A full disk or deleted workspace directory kills the monitor with a raw traceback.

---

## Important (should fix)

- **`health.py:218`** — `subprocess.run` raises `FileNotFoundError` if `kubectl` is absent; no timeout means one hung `kubectl describe` blocks the entire poll loop indefinitely
- **`health.py:61`** — `item["metadata"]["name"]` direct access; one malformed API response item crashes the entire parse — use `.get("metadata", {}).get("name")`
- **`health.py:230,241`** — Silent `return []` on kubectl failure; monitoring appears healthy when kubectl is broken — log a warning
- **`monitor.py:212`** — `kubectl describe` return code is discarded (`_`); a failed describe silently degrades diagnosis context with no warning
- **`monitor.py:146`** — `_load_setup_config` has unguarded `json.loads`; corrupt `setup_config.json` gives a raw traceback instead of an actionable message
- **`monitor.py` imports** — imports private `_kubectl` across module boundary; add a public `describe_pod(ns, pod_name) -> str` to `health.py`
- **`monitor.py` startup** — no log line when `ANTHROPIC_API_KEY` is set indicating that cluster data (pod logs, env vars from `kubectl describe`) will be sent to the Anthropic API
- **`monitor.py:~40`** — `ok()` helper defined but never called — dead code
- **`CLAUDE.md`** — `health_report.md` and `health.py` not added to the workspace artifacts / pipeline library tables

---

## Test gaps (critical)

| Gap | Impact |
|-----|--------|
| `delete_pod` returns `False` — `tracker.record` is still called before the delete attempt, so after two failed deletes the counter escalates away from tier-1 even though the pod was never actually deleted | Latent correctness bug, no test catches it |
| `pipeline.monitor.anthropic = None` (package absent) path in `_diagnose_with_api` | The "not installed" branch is never exercised |
| No `_poll_once` test for tier-2 path | Tier-2 report writing has no coverage |
| `phase="Failed"` (without `reason="Evicted"`) and `phase="Unknown"` never tested in `triage_pod` | Real cluster states unexercised |
| `_diagnose_with_api` exception path never triggered | `except Exception` block could be removed unnoticed |

---

## Suggestions

- `monitor.py:177` — `pair_keys[0]` misattributes health report findings when multiple pairs share a namespace
- `monitor.py:_diagnose_with_api` — no deduplication for tier-3 API calls; a pod in `CrashLoopBackOff` for 30 minutes gets diagnosed 60 times (cost + rate-limit pressure)
- `ErrImagePull` variant of `ImagePullBackOff` is handled but untested

---

## What's good

- Three-tier triage design is clean and the `health.py` / `monitor.py` separation is well-drawn
- OOM escalation tested at all three decision points
- `test_triage_does_not_modify_tracker` is exactly the right kind of behavioral contract test
- Multi-session `HealthReport` snowball test (`test_health_report_no_snowball`) is a strong regression guard
- kubectl wrapper tests inspect actual subprocess argument construction rather than mocking at a higher level

kalantar:
## Review response

All changes landed in 5250a8e. Summary below.

---

### Critical — all accepted

- **Issue 2** (`_work_remaining` missing `pending`/`collecting`): Fixed — now matches deploy.py's own definition.
- **Issue 3** (`tracker.reset()` on unmatched states): Fixed — reset only fires when `pod.phase == "Running" and pod.ready`. Note: the reviewer's proposed fix (`if pod.ready:`) was the right intent but the wrong framing — `triage_pod` returns `None` for *both* healthy pods *and* unmatched states (e.g. Pending with no FailedScheduling event), so we re-instated the explicit phase+ready guard that was incorrectly removed as dead code in a prior review pass.
- **Issue 4** (unguarded `json.loads` in `parse_pods`/`parse_events`): Fixed — both return `[]` on `JSONDecodeError`.
- **Issue 5** (`store.load()` kills monitor): Fixed — caught with log warning + continue.
- **Issue 6** (bare `except Exception` with no logging): Partially fixed — added `warn()` on the exception path so API failures are visible in the terminal. The broad catch is retained intentionally: the monitor must not crash because a non-critical feature failed. Added explicit empty-content guard (`if not message.content`).
- **Issue 7** (unguarded file I/O on `health_report.md`): **Rejected.** A full disk or deleted workspace directory is unrecoverable — there is nowhere to write a warning and continuing silently without a report is worse than a clear traceback.

**Issue 1** (wrong model ID): **Rejected.** `claude-haiku-4-5-20251001` is the correct API identifier. The reviewer's suggested ID `20251022` is not the right date suffix for this model.

---

### Important — accepted

- `subprocess.run` timeout + `FileNotFoundError`: Fixed — `timeout=30`, both exceptions return `(1, "")`.
- `item["metadata"]["name"]` direct access: Fixed — uses `.get()` with skip on missing name.
- `_kubectl describe` return code discarded: Fixed — logs warning when describe fails; `_kubectl` import replaced with new public `describe_pod()` in `health.py`.
- `_load_setup_config` unguarded `json.loads`: Fixed — catches `JSONDecodeError` and exits with actionable message.
- `ANTHROPIC_API_KEY` startup notice: Fixed — info line at startup indicates whether API calls will be made.
- `ok()` dead code: Fixed — removed.
- Silent `return []` on kubectl failure / CLAUDE.md additions: **Rejected.** Silent `[]` is intentional — the monitor should not crash or log noise every 30 seconds for a transient kubectl blip. CLAUDE.md covers stage-to-stage pipeline contracts, not every output artifact.

---

### Test gaps — all accepted

Added: `Failed` phase (no reason) → tier-3, `Unknown` phase → tier-3, `ErrImagePull` → tier-2, `kubectl` timeout, bad JSON in `parse_pods`/`parse_events`, `anthropic=None` path, `_diagnose_with_api` exception path, tier-2 `_poll_once`, failed-delete does not increment tracker, unmatched Pending does not reset tracker, `_work_remaining` with `pending`/`collecting`.

---

### Suggestions

- `pair_keys[0]` misattribution: documented with a comment in the code; in practice each pair gets its own namespace. Deferred.
- Tier-3 deduplication: valid cost concern; out of scope for this PR.
- `ErrImagePull` untested: covered in new tests above.

PR #90, opened 2026-05-11
fix(capacity): exclude Pending pods from GPU probe

### LINKED ISSUE #89: Capacity probe counts Pending pods as consuming GPUs
## Problem

`probe_free_gpus()` in `pipeline/lib/capacity.py` queries all non-terminal pods:

```
kubectl get pods --all-namespaces --field-selector=status.phase!=Succeeded,status.phase!=Failed -o json
```

This includes **Pending** pods — pods whose GPU requests exist in the spec but have not been scheduled to any node. Their GPUs are not actually allocated, yet the probe counts them as "requested."

**Observed:**
```
[INFO]  Capacity: 3 free GPUs (112 allocatable − 109 requested)
```

**Actual (kubectl-view-allocations):**
```
nvidia.com/gpu    (90%) 101.0  (90%) 101.0    112.0  11.0
```

The 8-GPU discrepancy (109 vs 101) matches the GPU requests of Pending pods that haven't been scheduled.

## Impact

The capacity probe reports fewer free GPUs than actually available. This causes:
- The backoff controller to enter scarcity mode prematurely
- Pairs that could be dispatched to be held back unnecessarily
- Longer overall run times

## Fix

Filter out pods that have no assigned node. The `status.phase` field selector doesn't support excluding Pending directly, but in Python we can skip pods where `spec.nodeName` is not set:

```python
for pod in pods.get("items", []):
    if not pod.get("spec", {}).get("nodeName"):
        continue
    ...
```

This is the correct semantic: a pod without a `nodeName` has not been bound to a node and is not consuming any node's allocatable resources.

## Location

`pipeline/lib/capacity.py:67-76` — the pod iteration loop in `probe_free_gpus()`

kalantar:
## Summary

- Filter out pods without `spec.nodeName` in `probe_free_gpus()` — these are Pending (unscheduled) and don't consume allocatable GPU resources
- Updated test helper `_mock_pods` to include `nodeName` by default, with opt-out for Pending pod simulation
- Added `test_pending_pods_excluded` verifying the fix

Closes #89

## Reviewer Notes

The one-line change is at `pipeline/lib/capacity.py:69`. The rest is test infrastructure. The key semantic: Kubernetes sets `spec.nodeName` only after the scheduler binds a pod to a node — pods without it haven't been allocated any node resources.

## Test Plan

- [x] `test_pending_pods_excluded` — 3 pods (2 scheduled, 1 Pending), only scheduled pods counted
- [x] All 7 existing `TestProbeFreeGpus` tests pass (helper backward-compatible via default `node_names`)
- [x] Full suite: 472 passed

### DIFF
--- a/pipeline/lib/capacity.py
+++ b/pipeline/lib/capacity.py
@@ -21,6 +21,9 @@ def probe_free_gpus(
     Queries kubectl for node allocatable resources and pod requests,
     computes the delta clamped to zero.
 
+    Skips pods without spec.nodeName — these are Pending (unscheduled) and
+    have not been allocated any node resources.
+
     Assumes only spec.containers request GPUs (initContainers are excluded —
     llm-d workloads do not use GPU-requesting init containers).
 
@@ -66,7 +69,7 @@ def probe_free_gpus(
 
     total_requested = 0
     for pod in pods.get("items", []):
-        if not pod.get("spec", {}).get("nodeName"):
+        if pod.get("spec", {}).get("nodeName") is None:
             continue
         for container in pod.get("spec", {}).get("containers", []):
             requests = container.get("resources", {}).get("requests", {})
--- a/pipeline/lib/capacity.py
+++ b/pipeline/lib/capacity.py
@@ -66,6 +66,8 @@ def probe_free_gpus(
 
     total_requested = 0
     for pod in pods.get("items", []):
+        if not pod.get("spec", {}).get("nodeName"):
+            continue
         for container in pod.get("spec", {}).get("containers", []):
             requests = container.get("resources", {}).get("requests", {})
             count = requests.get(gpu_resource_type)
--- a/pipeline/tests/test_capacity.py
+++ b/pipeline/tests/test_capacity.py
@@ -20,17 +20,27 @@ class TestProbeFreeGpus:
             nodes.append(node)
         return json.dumps({"items": nodes})
 
-    def _mock_pods(self, requested_gpus: list[int], resource="nvidia.com/gpu"):
-        """Build fake kubectl pods JSON."""
+    def _mock_pods(self, requested_gpus: list[int], resource="nvidia.com/gpu",
+                   node_names: list[str | None] | None = None):
+        """Build fake kubectl pods JSON.
+
+        node_names: per-pod nodeName (None = Pending, no nodeName).
+        Defaults to all pods having a nodeName.
+        """
+        if node_names is None:
+            node_names = [f"node-{i}" for i in range(len(requested_gpus))]
         pods = []
-        for gpu_req in requested_gpus:
+        for i, gpu_req in enumerate(requested_gpus):
+            spec: dict = {
+                "containers": [
+                    {"resources": {"requests": {resource: str(gpu_req)}}}
+                ]
+            }
+            if node_names[i] is not None:
+                spec["nodeName"] = node_names[i]
             pod = {
                 "metadata": {"name": f"pod-{len(pods)}"},
-                "spec": {
-                    "containers": [
-                        {"resources": {"requests": {resource: str(gpu_req)}}}
-                    ]
-                },
+                "spec": spec,
             }
             pods.append(pod)
         return json.dumps({"items": pods})
@@ -122,6 +132,23 @@ class TestProbeFreeGpus:
         result = probe_free_gpus()
         assert isinstance(result, str)
         assert "JSON parse error" in result
+
+    @patch("pipeline.lib.capacity.subprocess.run")
+    def test_pending_pods_excluded(self, mock_run):
+        """Pending pods (no nodeName) should not count as GPU consumers."""
+        nodes_json = self._mock_nodes([8, 8])
+        pods_json = self._mock_pods(
+            [4, 4, 4],
+            node_names=["node-0", "node-1", None],
+        )
+
+        mock_run.side_effect = [
+            MagicMock(returncode=0, stdout=nodes_json),
+            MagicMock(returncode=0, stdout=pods_json),
+        ]
+
+        result = probe_free_gpus()
+        assert result == (8, 16, 8)
 
 
 # ── derive_gpu_resource_type tests ─────────────────────────────────────────────
--- a/pipeline/tests/test_capacity.py
+++ b/pipeline/tests/test_capacity.py
@@ -150,6 +150,23 @@ class TestProbeFreeGpus:
         result = probe_free_gpus()
         assert result == (8, 16, 8)
 
+    @patch("pipeline.lib.capacity.subprocess.run")
+    def test_all_pods_pending_means_zero_requested(self, mock_run):
+        """When every pod is Pending, total requested should be zero."""
+        nodes_json = self._mock_nodes([8, 8])
+        pods_json = self._mock_pods(
+            [4, 4],
+            node_names=[None, None],
+        )
+
+        mock_run.side_effect = [
+            MagicMock(returncode=0, stdout=nodes_json),
+            MagicMock(returncode=0, stdout=pods_json),
+        ]
+
+        result = probe_free_gpus()
+        assert result == (16, 16, 0)
+
 
 # ── derive_gpu_resource_type tests ─────────────────────────────────────────────
 


kalantar:
## PR #90 Review — `fix(capacity): exclude Pending pods from GPU probe`

Four-agent review: code correctness, silent failures, test coverage, and comment accuracy.

---

### Requirements Check (Issue #89)

| Requirement | Status |
|---|---|
| Filter out pods where `spec.nodeName` is not set | ✅ Satisfied |
| Fix in `pipeline/lib/capacity.py` pod iteration loop | ✅ Satisfied |
| Prevent premature scarcity mode in backoff controller | ✅ Satisfied |

All issue requirements are addressed. Tests pass (472), lint clean.

---

### Verdict: **NEEDS CHANGES** (1 blocking, 2 recommended)

---

### Blocking

**1. Use `is None` instead of falsy check** — `capacity.py:69`

The current filter `if not pod.get("spec", {}).get("nodeName")` uses a falsy check, which also excludes pods with `nodeName: ""`. While Kubernetes doesn't normally produce empty-string nodeName for scheduled pods, the falsy check is wider than intended and inconsistent with the existing pattern at line 60-61 (`if count is not None`) in the same function. The fix is a single-word change:

```python
# Current (falsy — catches None, "", 0, False):
if not pod.get("spec", {}).get("nodeName"):
    continue

# Recommended (precise — catches only absent/None):
if pod.get("spec", {}).get("nodeName") is None:
    continue
```

**Why it matters:** If a pod with an anomalous empty-string nodeName is silently excluded, the probe inflates free GPUs — the exact symptom this PR fixes.

---

### Recommended (non-blocking)

**2. Update `probe_free_gpus` docstring** — `capacity.py:19-29`

The docstring documents two existing filters (field-selector for phase, initContainers exclusion) but not the new Pending-pod filter. Add a line like:

> Skips pods without spec.nodeName — these are Pending (unscheduled) and have not been allocated any node resources.

This matches the existing pattern of documenting filter assumptions in the docstring.

**3. Add boundary test for all-pods-Pending** — `test_capacity.py`

The current test covers the mixed case (2 scheduled, 1 Pending). An additional test where all pods are Pending (expected: `(16, 16, 0)`) would catch regressions where the filter fails when no pods have a `nodeName`.

---

### Pre-existing issues noted (out of scope)

- `deploy.py:1216-1217` — `except ValueError: pass` silently prevents PipelineRun timeout on unparseable timestamps. Worth a follow-up issue.
- `capacity.py:68` — `pods.get("items", [])` silently produces zero count if kubectl returns a non-list JSON object (e.g., `Status` error).

---

### Positive observations

- Fix is minimal and surgical — 2 lines of production code at the correct scope
- Test helper refactor is backward-compatible via default `node_names`
- Test assertion math is correct and discriminating (would fail without the fix)
- All 7 existing tests unaffected
- PR scope is tight — no unnecessary refactoring

kalantar:
## Review Feedback Addressed

### Triage

| # | Classification | Summary | Validation |
|---|---|---|---|
| 1 | FIX | Use `is None` instead of falsy check | Valid — `not x` catches `""`, `0`, `False` beyond `None`; inconsistent with `is not None` at lines 61 and 74 in the same function |
| 2 | FIX | Update docstring to document new filter | Valid — docstring documents two existing filter assumptions but not the new one |
| 3 | FIX | Add boundary test for all-pods-Pending | Valid — covers edge case where zero pods pass the filter |
| 4 | FOLLOW-UP | `deploy.py:1216-1217` silent `except ValueError: pass` | Confirmed pre-existing, out of scope |
| 5 | FOLLOW-UP | `capacity.py:68` silent zero on non-list JSON | Confirmed pre-existing, out of scope |

### Fixed (3 items) — c63a41f

- **#1:** Changed `if not pod.get("spec", {}).get("nodeName")` → `if pod.get("spec", {}).get("nodeName") is None` — matches existing pattern, precise semantics
- **#2:** Added docstring line: "Skips pods without spec.nodeName — these are Pending (unscheduled) and have not been allocated any node resources."
- **#3:** Added `test_all_pods_pending_means_zero_requested` — 2 Pending pods, expects `(16, 16, 0)`

### Follow-up (2 items)

- **#4:** Silent `except ValueError: pass` in deploy.py timeout parsing — will file separately
- **#5:** Silent zero on non-list kubectl JSON — will file separately

### Verification

- 21 capacity tests pass (9 probe tests including 2 new)
- 473 total tests pass
- Lint clean (`ruff check pipeline/ --select F`)

kalantar:
## Review: PR #90 against Issue #89

**Verdict: READY TO MERGE**

---

### Requirements Coverage (Issue #89)

| Requirement | Status |
|---|---|
| Filter out pods where `spec.nodeName` is not set | ✅ Addressed |
| Fix in pod iteration loop of `probe_free_gpus()` | ✅ Correct location |
| Root cause: Pending pods inflating `total_requested` | ✅ Resolved |
| Prevents premature backoff controller scarcity mode | ✅ Follows from fix |

The two-line guard at `capacity.py:70-71` is the narrowest possible fix. The `is None` check is strictly better than the falsy check suggested in the issue (avoids false exclusion on empty string, consistent with existing pattern at line 61). CI passes (472 tests, lint clean).

---

### Findings by Category

#### Code Quality & Correctness — No blocking issues

- The fix is correct, minimal, and well-placed. `spec.nodeName` is the right signal for whether a pod has been bound to a node.
- No downstream contract violations — `probe_free_gpus()` returns the same `(free, total, requested)` tuple; only `requested` becomes more accurate.
- Performance: negligible — one dict lookup per pod vs. two subprocess calls.
- Docstring update accurately documents the new filter behavior.

#### Silent Failures — No new issues introduced

- **`is None` semantic gap** (MEDIUM, non-blocking): The check conflates three states — missing `spec`, absent `nodeName` key, and `nodeName: null`. All resolve to `None` via the `.get()` chain. This is a defensible tradeoff: kubectl's JSON output is schema-stable, and a pod missing `spec` entirely would also break the container iteration on the next line. The `{}` fallback pattern is used consistently throughout the function.
- **Double `pod.get("spec", {})` evaluation** (LOW): Lines 70 and 72 both call `pod.get("spec", {})`. Extracting `spec = pod.get("spec", {})` once would be a minor readability improvement. Not a correctness issue.

#### Pre-existing issues in the neighborhood (not introduced by this PR):

- **`except ValueError: pass` at `deploy.py:1216`** (CRITICAL, pre-existing): Silently swallows timestamp parse failures in PipelineRun timeout detection. A stuck PipelineRun would never be timed out, with no log trail. Consider logging a warning in a follow-up.
- **Probe failure → ungated dispatch** (HIGH, pre-existing): When `probe_free_gpus()` returns an error, `deploy.py` falls back to dispatching all pending pairs without GPU gating. This bypasses the accuracy improvement this PR provides. Documented/intentional, but worth noting.

#### Test Coverage — Adequate for merge

Two new tests cover the key partitions:
- `test_pending_pods_excluded`: Mixed scheduled + Pending pods — verifies Pending pod GPUs excluded from count
- `test_all_pods_pending_means_zero_requested`: Boundary case — all Pending, zero requested

The "all scheduled" case has extensive coverage from the 7 pre-existing tests (helper defaults to generating `nodeName` for all pods). Helper extension is backward-compatible.

**Minor gaps (non-blocking):**
- No test for empty-string `nodeName` (`""`) — currently treated as scheduled. Kubernetes doesn't produce this, but a documenting test would clarify the design choice.
- No test for pod missing `spec` entirely — safely handled by `.get()` chain but not explicitly tested.

#### Comment & Documentation Quality — Accurate

- Docstring addition correctly describes the filter behavior. Minor precision note: "Pending (unscheduled)" conflates the Kubernetes phase with the absence of `nodeName` (technically "unbound" rather than "unscheduled"), but the distinction is immaterial for this probe's purpose.
- Test names and docstrings are clear and descriptive.
- Commit messages are exemplary — explain the *why*, not just the *what*.

---

### Summary

The PR is a clean, well-tested, minimal fix that correctly addresses issue #89. No blocking issues found. The two optional improvements (extract `spec` variable, document empty-string `nodeName` behavior in a test) are nice-to-haves. The pre-existing `except ValueError: pass` in `deploy.py` is worth a follow-up issue.

PR #108, opened 2026-05-12
Drop singular baseline:/algorithm: support, require list forms

### LINKED ISSUE #97: Drop singular baseline:/algorithm: support (require list forms)
## Parent
Part of #24 (v3 schema refactor) — sub-issue 4

## Context

v3 accepts both singular (`baseline:`, `algorithm:`) and list (`baselines:`, `algorithms:`) forms, normalizing singular → list internally. Drop the singular forms so only list forms are accepted.

## Changes

- `pipeline/lib/manifest.py`: remove singular → list normalization for `baseline` → `baselines` (lines ~112-119) and `algorithm` → `algorithms` (lines ~144-152). Singular keys simply fail validation — no special error message needed.
- `pipeline/lib/manifest.py`: remove the early `algorithm` validation block (lines ~51-59) — dead code once singular form is gone
- Update `MINIMAL_V3` test fixture to use list forms
- Remove tests for singular normalization (`test_singular_baseline_normalized_to_baselines`, `test_singular_algorithm_normalized_to_algorithms`)

kalantar:
Closes #97

## Summary

- Removed `baseline:` → `baselines:` and `algorithm:` → `algorithms:` normalization from `manifest.py` — only list forms accepted now
- Removed the early `algorithm` validation block (lines 51-59) — dead code once singular form is gone
- Removed the singular `baseline` validation/defaults block (lines 61-80) — only served the normalization path
- Updated `MINIMAL_V3` test fixture and all inline manifest dicts in `test_prepare.py` to use list forms
- Removed 8 tests that only exercised singular-form behavior; replaced with list-form equivalents where applicable

## Reviewer notes

Manifests using `baseline:` or `algorithm:` will now fail with "Missing required field: baselines" (no special migration error message — per issue scope). The `baselines` list validation requires `name` and `scenario` on each entry but does **not** apply the old `sim`/`real` sub-structure defaults that the singular form did. Downstream code (`prepare.py`) already uses `.get()` defensively for those fields, so this is safe.

## Test plan

- [x] `python -m pytest pipeline/ -v` — 476 passed

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -248,16 +248,19 @@ kind: sim2real-transfer
 version: 3
 scenario: <name>            # scenario name used in generated PipelineRun labels
 
-algorithm:                    # optional — omit for baseline-only benchmarks
-  source: <path>            # sim algorithm implementation
-  config: <path>            # sim algorithm config (optional)
+baselines:                  # required — list of baseline specs
+  - name: <pkg-name>       # unique package name (lowercase alphanumeric, 1-20 chars)
+    scenario: <path>        # baseline scenario YAML (null if none)
+    sim:
+      config: <path>        # baseline policy for sim
+    real:
+      config: <path>        # optional: baseline EPP config template
+      notes: |              # optional: notes embedded in skill_input.json
 
-baseline:
-  sim:
-    config: <path>          # baseline policy for sim
-  real:
-    config: <path>          # optional: baseline EPP config template
-    notes: |                # optional: notes embedded in skill_input.json
+algorithms:                 # optional — omit for baseline-only benchmarks
+  - name: <pkg-name>       # unique package name
+    source: <path>          # sim algorithm implementation
+    defaults: <baseline>    # name of baseline this algorithm inherits from
 
 workloads:
   - <path>                  # one or more workload YAMLs
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -21,7 +21,6 @@ def _validate_package_name(name: str, context: str) -> None:
 
 
 _REQUIRED_TOP = ["kind", "version", "scenario"]
-_REQUIRED_ALGORITHM = ["source"]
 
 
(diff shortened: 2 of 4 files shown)


kalantar:
## Review: NEEDS CHANGES

All issue #97 requirements are addressed — singular-form normalization removed, dead code cleaned up, fixtures and tests updated correctly. One blocking issue before merge:

### Singular `algorithm:` is silently ignored

When a user provides the old `algorithm:` key (singular), `load_manifest` does not error. The key is ignored, `algorithms` defaults to `[]`, and the pipeline runs in baseline-only mode with no warning. The user's algorithm specification is silently dropped.

This is asymmetric with `baseline:` (singular), which *does* fail clearly with "Missing required field: baselines" because `baselines` is required. But `algorithms` is optional, so the singular form just falls through to the empty-list default.

**Fix:** Add explicit rejection of both old singular keys after the validation blocks:

```python
if "baseline" in data:
    raise ManifestError("'baseline' is no longer supported — use 'baselines' (list form)")
if "algorithm" in data:
    raise ManifestError("'algorithm' is no longer supported — use 'algorithms' (list form)")
```

Add corresponding tests for both rejection paths.

### `pipeline/README.md` has stale schema example

Lines 246-268 still show the old `baseline:` / `algorithm:` singular forms in the example manifest. Update to show list forms (`baselines:`, `algorithms:`).

### What looks good

- Clean dead-code removal (normalization, `_REQUIRED_ALGORITHM`, early validation block)
- All removed tests are genuinely dead — they only exercised deleted code paths
- Downstream consumers already use list forms exclusively — no breakage risk
- Updated test fixtures are correct and consistent
- No stale comments left in code

kalantar:
Updated the README schema example to use `baselines:`/`algorithms:` list forms (585cf88).

The silent-ignore behavior for singular `algorithm:` is intentional — no rejection error added. The existing "Missing required field: baselines" error for singular `baseline:` is sufficient — no special migration message added either.

PR #109, opened 2026-05-12
Introduce component section (consolidate target + config + epp_image + build)

### LINKED ISSUE #98: Introduce component section (consolidate target + config.kind + epp_image + build)
## Parent
Part of #24 (v3 schema refactor) — sub-issue 5

## Context

v3 spreads component info across four top-level sections: `target`, `config`, `epp_image`, `build`. These all describe the same thing — the llm-d component being modified. Consolidate into a single `component` section.

## Refactored schema

```yaml
component:
  repo: github.com/llm-d/llm-d-inference-scheduler   # full url
  path: llm-d-inference-scheduler                      # optional, defaults from last url segment
  kind: EndpointPickerConfig                           # was config.kind
  base_image:                                          # was epp_image.upstream
    hub: ghcr.io/llm-d
    name: llm-d-inference-scheduler
    tag: v0.7.1
  build:                                               # was top-level build + epp_image.build
    commands:
      - [go, build, ./...]
    image:                                             # was epp_image.build; defaults from base_image
      hub: ghcr.io/kalantar
```

## Migration

Breaking change. Users must update their `transfer.yaml` to use the new `component` section. No backward-compat shim or warnings for old keys.

## Changes

- `pipeline/lib/manifest.py`: add `component` section validation; remove `target`, `config`, `epp_image`, `build` top-level validation
- `pipeline/prepare.py`: update `_load_resolved_config()` and all field access to read from `component`
- `pipeline/prepare.py`: update `skill_input.json` construction to use new paths
- `pipeline/deploy.py`: update any `epp_image` access to use `component.base_image` / `component.build.image`
- `pipeline/setup.py`: rename `epp_image` key in `run_metadata.json` to `component_image`
- `pipeline/lib/epp.py`: update field references from `epp_image` to `component.base_image` / `component.build.image`
- `pipeline/lib/run_manager.py`: update any `target`/`config`/`epp_image` references
- `pipeline/lib/health.py`: update any `target`/`config` references
- Update all tests

## Cross-stage metadata contract

`run_metadata.json` key `epp_image` is renamed to `component_image`. Both `setup.py` (writer) and `deploy.py` (reader) must update in the same change.

## Depends on

#94, #95, #96 (helm_path removal), #97 (singular form removal) — all closed.

kalantar:
Closes #98

## Summary

- Consolidates four top-level `transfer.yaml` sections (`target`, `config`, `build`, `epp_image`) into a single `component` section
- Renames `run_metadata.json` key from `epp_image` to `component_image` (setup.py writer + deploy.py reader updated in lockstep)
- Old top-level keys (`target`, `config`, `epp_image`) are silently ignored — consistent with how any unknown key is treated (no general unknown-key rejection exists)
- `skill_input.json` output shape is preserved (backward compat with translate skill)

## Breaking change

Users must update their `transfer.yaml` to use the new schema:

```yaml
component:
  repo: github.com/llm-d/llm-d-inference-scheduler
  kind: EndpointPickerConfig
  base_image:
    hub: ghcr.io/llm-d
    name: llm-d-inference-scheduler
    tag: v0.7.1
  build:
    commands:
      - [go, build, ./...]
```

## Migration: run_metadata.json

The `run_metadata.json` key was renamed from `epp_image` to `component_image`. Existing workspace runs that have the old key will not be recognized by `deploy.py`. **Users must re-run `setup.py`** to regenerate `run_metadata.json` with the new key. There is no backward-compat fallback — this is acceptable because `setup.py` is idempotent and always the first step after pulling new pipeline code.

## Design decisions

- **No old-key rejection:** Unknown top-level keys in `transfer.yaml` are silently ignored (there is no general unknown-key validation). Old keys get the same treatment — no special error messages for `target`/`config`/`epp_image`. Simpler code, consistent behavior.
- **No `run_metadata.json` fallback:** Assume users re-run `setup.py` after updating. The setup → prepare → deploy sequence is always run in order; setup always runs first.

## Reviewer attention

- `_load_resolved_config` now returns the `component` dict directly (flat structure) rather than a dict-of-dicts. All consumers updated.
- `component.path` is auto-defaulted from the last URL segment of `component.repo` if not explicitly set.
- The `inject_epp_image` function name is intentionally unchanged — it describes what it does (inject EPP container image into scenario YAML), not a manifest key.

## Test plan

- [x] All 476 pipeline tests pass
- [x] Lint clean (ruff check --select F)
- [x] No stale manifest key references in pipeline code

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -115,12 +115,12 @@ def _build_epp_image(run_dir: Path, run_name: str, namespace: str) -> str:
     meta_path = run_dir / "run_metadata.json"
     if meta_path.exists():
         meta = json.loads(meta_path.read_text())
-        full_image = meta.get("epp_image", "")
+        full_image = meta.get("component_image", "")
         if full_image:
             ok(f"EPP image: {full_image}")
             return full_image
 
-    err("EPP build completed but epp_image not set in run_metadata.json")
+    err("EPP build completed but component_image not set in run_metadata.json")
     sys.exit(1)
 
 
--- a/pipeline/lib/health.py
+++ b/pipeline/lib/health.py
@@ -165,7 +165,7 @@ def triage_pod(
             message=f"{pod.name}: {pod.reason}",
             suggestion=(
                 f"Image pull failed: {img_detail}\n"
-                "Check run_metadata.json → epp_image or scenario bundle → "
+                "Check run_metadata.json → component_image or scenario bundle → "
                 "model.image"
             ),
         )
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -122,14 +122,6 @@ def load_manifest(path: "Path | str") -> dict:
 def _validate_v3_fields(data: dict) -> None:
     """Validate and apply defaults for v3-specific fields."""
 
-    # Reject old top-level keys that have been replaced by 'component'
-    for old_key in ("target", "config", "epp_image"):
-        if old_key in data:
-            raise ManifestError(
-                f"Top-level '{old_key}' is no longer supported; "
-                "use the 'component' section instead"
-            )
-
     # component (required)
     component = data.get("component")
     if component is None:
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -122,56 +122,59 @@ def load_manifest(path: "Path | str") -> dict:
 def _validate_v3_fields(data: dict) -> None:
     """Validate and apply defaults for v3-specific fields."""
 
-    # target (required)
-    target = data.get("target")
-    if target is None:
-        raise ManifestError("Missing required field: target")
-    if not isinstance(target, dict):
-        raise ManifestError("target must be a mapping")
-    if "repo" not in target:
-        raise ManifestError("Missing required field: target.repo")
+    # Reject old top-level keys that have been replaced by 'component'
+    for old_key in ("target", "config", "epp_image"):
+        if old_key in data:
+            raise ManifestError(
+                f"Top-level '{old_key}' is no longer supported; "
+                "use the 'component' section instead"
+            )
 
-    # config (required)
-    config = data.get("config")
-    if config is None:
-        raise ManifestError("Missing required field: config")
-    if not isinstance(config, dict):
-        raise ManifestError("config must be a mapping")
-    if "kind" not in config:
-        raise ManifestError("Missing required field: config.kind")
+    # component (required)
+    component = data.get("component")
+    if component is None:
+        raise ManifestError("Missing required field: component")
+    if not isinstance(component, dict):
+        raise ManifestError("component must be a mapping")
 
-    # build (optional, defaults applied)
-    build = data.get("build")
-    if build is None:
-        data["build"] = {"commands": []}
-    elif not isinstance(build, dict):
-        raise ManifestError("build must be a mapping")
-    else:
+    # component.repo (required)
+    repo = component.get("repo")
+    if not repo or not isinstance(repo, str):
+        raise ManifestError("Missing required field: component.repo")
+
+    # component.kind (required)
+    kind = component.get("kind")
+    if not kind or not isinstance(kind, str):
+        raise ManifestError("Missing required field: component.kind")
+
+    # component.path (optional, defaults from last segment of repo)
+    if "path" not in component:
+        component["path"] = repo.rstrip("/").rsplit("/", 1)[-1]
+
+    # component.base_image (optional)
+    base_image = component.get("base_image")
+    if base_image is not None:
+        if not isinstance(base_image, dict):
+            raise ManifestError("component.base_image must be a mapping")
+        for f in ("hub", "name", "tag"):
+            if f not in base_image:
+                raise ManifestError(f"Missing required field: component.base_image.{f}")
+
+    # component.build (optional)
+    build = component.get("build")
+    if build is not None:
+        if not isinstance(build, dict):
+            raise ManifestError("component.build must be a mapping")
         build.setdefault("commands", [])
         cmds = build["commands"]
         if not isinstance(cmds, list):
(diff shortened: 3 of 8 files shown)


kalantar:
## Review: PR #109 against Issue #98

**Verdict: NEEDS CHANGES**

This is a well-structured refactor that correctly consolidates the four top-level sections into a unified `component` section. The implementation is largely correct, tests are comprehensive, and the cross-stage contract (setup.py writes / deploy.py reads `component_image`) is properly maintained. Two issues should be fixed before merge.

---

### Must Fix

**1. Top-level `build` key not rejected**

`manifest.py` rejects `target`, `config`, `epp_image` but omits `build` from the rejection loop. A user migrating from the old schema who has a top-level `build:` key but correctly adds `component:` will have their build commands silently ignored — no error, no warning, data lost.

Fix: add `"build"` to the tuple at line 126:
```python
for old_key in ("target", "config", "epp_image", "build"):
```
Add a corresponding `test_old_build_key_rejected` test.

**2. `pipeline/README.md` not updated**

README still documents the old schema (`target`, `config`, `build`, `epp_image`) with full examples. CLAUDE.md explicitly requires updating README when artifact schema changes. The `--skip-build-epp` flag description also references the old `epp_image` key.

---

### Should Fix

**3. Stale comment in prepare.py (line ~610)**: `# Check 3: all files_created exist in target repo` should say "component repo" to match new terminology.

**4. `prompts/prepare/translate.md` references `target.plugin_dir` and `target.register_file`** in Steps 3–4 of the Translation Loop, but `skill_input.json` only contains `"target": {"repo": "..."}`. These references were misleading before this PR but the description update on line 26 makes the inconsistency more visible.

---

### Nice to Have (non-blocking)

- Add tests for type-checking branches: `component.build` not a mapping, `component.base_image` not a mapping, `component.build.image` not a mapping (3 untested validation paths)
- Test `component.repo` as empty string (currently only tests missing key, not falsy value)
- Test `component.path` derivation with trailing slash in repo URL
- Consider a defensive check in deploy.py for old `epp_image` key in `run_metadata.json` with a migration-hint error message (helps users with existing workspaces)

---

### Positive Notes

- `skill_input.json` output shape correctly preserved for backward compat with the translate skill
- Issue listed `lib/epp.py` and `lib/run_manager.py` as needing changes — verified these files have no manifest field references; PR correctly identified no changes needed
- Test coverage for the new validation is thorough: all required fields, path defaulting (both explicit and auto-derived), and old key rejection all covered
- Cross-stage `epp_image` → `component_image` rename done atomically in setup.py + deploy.py + health.py

kalantar:
Removed the old-key rejection logic (`target`, `config`, `epp_image` raising "no longer supported"). Unknown top-level keys are already silently ignored by the validator — no general unknown-key check exists — so these old keys now get the same treatment as any other unrecognized field. Simpler, consistent.

kalantar:

### Critical: `skill_input.json` `target.repo` writes a URL instead of a filesystem path

`prepare.py` line 282 emits `resolved.get("repo", "")` into `skill_input.json`'s `target.repo`. Since `resolved` is now a flat copy of the `component` section, this produces the full URL (e.g., `github.com/llm-d/llm-d-inference-scheduler`) instead of the local directory name (`llm-d-inference-scheduler`) that the translate skill expects.

The translate skill uses `target.repo` to construct filesystem paths (`$EXPERIMENT_ROOT/<target.repo>`). This would break at runtime.

**Fix:** Change `resolved.get("repo", "")` to `resolved.get("path", "")` on line 282.

---

### High: PR body claims old-key rejection exists, but commit 366f813 removed it

The PR description and test plan state "Old keys produce clear rejection errors," but the final commit removes the rejection logic. Users with old `transfer.yaml` files get a generic "Missing required field: component" error with no migration guidance. This is especially problematic for a breaking schema change.

**Options:**
1. Restore the old-key detection (6 lines, best UX for a breaking change)
2. Update the PR description to accurately reflect the current behavior

---

### High: Existing `run_metadata.json` files with `epp_image` key go unrecognized

If `setup.py` was run before this change, `run_metadata.json` contains `epp_image`. After upgrading, `deploy.py` looks for `component_image` and fails with a confusing "component_image not set" error when the value is present under the old key.

**Suggestion:** Add a fallback check: if `component_image` is missing but `epp_image` exists, emit a clear error telling the user to re-run `setup.py`.

---

### Medium: Top-level `build` key silently ignored

If a user provides both `component` (without a `build` sub-key) and a top-level `build`, the top-level `build` is silently dropped. Their build commands won't execute with no warning. This is covered by the old-key detection recommendation above.

---

### Test coverage gaps

1. No test asserts the *value* of `skill_input.json`'s `target.repo` — this is why the URL-vs-path bug wasn't caught. Add `assert si["target"]["repo"] == "llm-d-inference-scheduler"`.
2. No test covers the `run_metadata.json` key rename contract between `setup.py` and `deploy.py`.
3. No test for `component.repo` with a trailing slash (path defaulting edge case).
4. No positive test for valid `component.build.image`.

---

### Minor: Stale prompt references

`prompts/prepare/translate.md` lines 43-44 reference `target.plugin_dir` and `target.register_file` which don't exist in `skill_input.json`. This predates the PR but the file was touched — worth fixing while here.

---

### What looks good

- `setup.py` writer and `deploy.py` reader updated in lockstep for `component_image`
- Manifest validation for new `component` schema is comprehensive and well-tested
- `component.path` auto-defaulting from last URL segment is a smart design choice
- Flat `_load_resolved_config` simplifies all consumers cleanly
- `_validate_assembly` correctly uses `path` for filesystem operations
- `prompts/prepare/translate.md` table entry correctly updated

kalantar:
Updated PR description to document two decisions:

1. Old keys are silently ignored (consistent with all unknown keys) — no special rejection logic.
2. Existing `run_metadata.json` files with the old `epp_image` key require re-running `setup.py`. No backward-compat fallback — setup is idempotent and always precedes deploy in the pipeline sequence.

PR #110, opened 2026-05-12
Derive build.image defaults from base_image; tag from run name

### LINKED ISSUE #100: Derive build.image defaults from base_image; tag from run name
## Parent
Part of #24 (v3 schema refactor) — sub-issue 7

## Depends on
#98 — `component.base_image` and `component.build.image` must exist first

## Context

Today `epp_image.build` requires explicit `hub`, `name`, `tag`. After refactoring, `component.build.image` should default missing fields from `component.base_image`:
- `hub` defaults to `base_image.hub`
- `name` defaults to `base_image.name`
- `tag` comes from the run name (not the manifest) — remove from manifest schema
- `platform` defaults to `linux/amd64`

This reduces manifest boilerplate — most users only need to override `hub` (push to their own registry).

## Changes

- `pipeline/lib/manifest.py`: apply defaults from `base_image` to `build.image` when fields are missing
- `pipeline/lib/manifest.py`: remove `tag` as a required/accepted field in `build.image`
- `pipeline/deploy.py`: derive image tag from run name instead of manifest
- Update tests

kalantar:
## Summary

- Make `component.base_image.tag` optional in manifest schema — it's never consumed by the pipeline (tag is always derived from the run name at build time via `build-epp.sh`)
- Default `component.build.image.hub` and `name` from `component.base_image` when those fields are absent — reduces manifest boilerplate (most users only need to override `hub` to push to their own registry)

Closes #100

## Details

Two commits:

1. **base_image.tag optional** — removes `tag` from the required fields tuple in `_validate_v3_fields`. Existing manifests with `tag` still load fine (optional ≠ forbidden).

2. **build.image defaults** — adds two `setdefault` calls guarded by `if base_image is not None`. Explicit `build.image` fields are never overwritten. Without `base_image`, `hub` is still required (backward compat).

## Test plan

- [x] `test_component_base_image_tag_optional` — base_image without tag loads
- [x] `test_build_image_defaults_hub_from_base_image` — hub inherited
- [x] `test_build_image_defaults_name_from_base_image` — name inherited
- [x] `test_build_image_defaults_both_from_base_image` — both inherited from empty `{}`
- [x] `test_build_image_no_base_image_requires_hub` — backward compat: still errors without base_image
- [x] `test_build_image_explicit_overrides_base_image` — explicit values not overwritten
- [x] Full suite: 482 tests pass, ruff clean

### DIFF
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -148,7 +148,7 @@ def _validate_v3_fields(data: dict) -> None:
     if base_image is not None:
         if not isinstance(base_image, dict):
             raise ManifestError("component.base_image must be a mapping")
-        for f in ("hub", "name", "tag"):
+        for f in ("hub", "name"):
             if f not in base_image:
                 raise ManifestError(f"Missing required field: component.base_image.{f}")
 
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -165,6 +165,9 @@ def _validate_v3_fields(data: dict) -> None:
         if build_image is not None:
             if not isinstance(build_image, dict):
                 raise ManifestError("component.build.image must be a mapping")
+            if base_image is not None:
+                build_image.setdefault("hub", base_image["hub"])
+                build_image.setdefault("name", base_image["name"])
             if "hub" not in build_image:
                 raise ManifestError("Missing required field: component.build.image.hub")
 
--- a/pipeline/tests/test_manifest.py
+++ b/pipeline/tests/test_manifest.py
@@ -341,6 +341,73 @@ def test_component_build_image_validates_hub(tmp_path):
         load_manifest(path)
 
 
+def test_build_image_defaults_hub_from_base_image(tmp_path):
+    """build.image without hub inherits from base_image.hub."""
+    data = {**MINIMAL_V3, "component": {
+        "repo": "github.com/llm-d/llm-d-inference-scheduler",
+        "kind": "EndpointPickerConfig",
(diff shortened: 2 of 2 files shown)


kalantar:
## Verdict: NEEDS CHANGES (minor)

The implementation is clean, correct, and well-tested. The code that's present is solid — no bugs, no silent failures, no regressions. One issue requirement is unaddressed.

### What's good

- `setdefault` is the right primitive — never overwrites explicit values, correctly guarded by `if base_image is not None`
- Validation ordering is correct: defaults applied before the `hub` required check, so the error fires only when neither `base_image` nor explicit `hub` exists
- Test coverage is excellent: all edge cases (individual field inheritance, both fields, explicit override, no-base-image fallback) are exercised
- `tag` removal from required fields is confirmed safe — no pipeline code reads `base_image["tag"]`; image tag is always derived from run name via `build-epp.sh`
- CI passes (482 tests, ruff clean)

### Required change

1. **Issue requirement 4 unaddressed: `platform` defaults to `linux/amd64`**
   The issue explicitly requires this. It's currently hardcoded in `build-epp.sh` but the issue asks for it in the manifest schema with a default (allowing override). Either implement it here or add a note in the PR body that it's deferred to a follow-up issue.

### Optional improvements (non-blocking)

- **Clarify `deploy.py` no-op**: The issue lists a `deploy.py` change, but `deploy.py` already derives the tag from run name. A one-line PR body note would prevent confusion for future readers.
- **README staleness** (pre-existing, from PR #109): `pipeline/README.md` lines 25 and 272-291 still document the old `target`/`config`/`epp_image` schema. Not introduced here but worth tracking.

kalantar:
`platform` is not used; we are removing it.

PR #113, opened 2026-05-13
Make component section optional when no algorithms present

### LINKED ISSUE #112: Make component section optional (required only when algorithms present)
## Summary

The `component` section in `transfer.yaml` is currently unconditionally required. For baseline-only runs (no algorithms, no code changes), there's no component repo to point at — the section is meaningless overhead.

Make `component` optional, with the constraint: **if `algorithms` is non-empty, `component` is required**.

## Motivation

Baseline-only experiments benchmark existing scheduler behavior without any plugin translation. They don't need a component repo, EPP image build, or git ref validation. Forcing users to specify a dummy `component` section adds friction and confusion.

## Changes Required

| File | Change |
|------|--------|
| `pipeline/lib/manifest.py:126-128` | Remove unconditional `if component is None: raise`. Add conditional: require `component` only when `algorithms` is non-empty |
| `pipeline/lib/manifest.py:133-140` | Move `repo`/`kind` required checks inside `if component:` guard |
| `deploy.py:999-1003` | Auto-skip EPP build when no component (detect via `run_metadata.json` missing `component_image`, or `.state.json` `mode=baseline-only`) |
| `pipeline/setup.py:689` | Guard `component_image` field generation on component presence |

## Already Safe (no changes needed)

- `prepare.py:108` — `_load_resolved_config()` returns `{}` when component absent
- `prepare.py:197-233` — Phase 1 submodule validation gates on `if target_path:`, no-ops with `""`
- `prepare.py:289-292` — Phase 3 already skips translation when `algorithms` is empty
- `prepare.py:460` — EPP injection only targets `pkg.kind == "algorithm"` packages
- `prepare.py:513` — `_get_submodule_shas("")` handles empty path gracefully
- `prepare.py:614-671` — `_validate_assembly()` only runs if `translation_output.json` exists

## Validation Rule

```python
if manifest.get("algorithms") and not manifest.get("component"):
    raise ManifestError("component is required when algorithms are specified")
```

## Test Cases

- `transfer.yaml` with `algorithms` + `component` → works as today
- `transfer.yaml` with `algorithms` but no `component` → ManifestError
- `transfer.yaml` with no `algorithms` and no `component` → valid, baseline-only mode
- `transfer.yaml` with no `algorithms` but `component` present → valid (component ignored or used for SHA tracking)

kalantar:
## Summary

- Relaxes `transfer.yaml` validation so the `component` section is only required when `algorithms` is non-empty
- Auto-skips EPP image build in `deploy.py run` when `run_metadata.json` has no `component_image` (baseline-only mode)
- Adds tests for all four combinations: algorithms+component, algorithms-only (error), component-only (valid), neither (valid)

Closes #112

## Reviewer notes

- The `setup.py` change mentioned in the issue is unnecessary — line 689 already gates `component_image` on `if cfg.registry`, and the deploy.py guard handles the downstream case
- All existing tests pass unchanged; the `MULTI_BASELINE_V3` fixture (no algorithms, has component) validates test case 4 implicitly

## Test plan

- [x] `python -m pytest pipeline/ -v` — 497 tests pass
- [ ] CI passes

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -86,6 +86,31 @@ def _load_setup_config() -> dict:
     return {}
 
 
+# ── EPP build decision ─────────────────────────────────────────────────────────
+
+def _resolve_epp_action(run_dir: Path, skip_build_epp: bool) -> str:
+    """Determine EPP build action based on run metadata.
+
+    Returns one of: "build", "skip", "error:<message>"
+    """
+    run_meta_path = run_dir / "run_metadata.json"
+    if not run_meta_path.exists():
+        return "error:run_metadata.json not found — run setup.py first."
+    try:
+        run_meta = json.loads(run_meta_path.read_text())
+    except json.JSONDecodeError as e:
+        return f"error:run_metadata.json is not valid JSON: {e}. Re-run setup.py."
+
+    component_image = run_meta.get("component_image")
+    if component_image is None:
+        return "skip"
+    if not component_image:
+        return "error:component_image is empty in run_metadata.json. Re-run setup.py with a valid --registry."
+    if skip_build_epp:
+        return "skip"
+    return "build"
+
+
 # ── EPP Image build ─────────────────────────────────────────────────────────
 
 def _build_epp_image(run_dir: Path, run_name: str, namespace: str) -> str:
@@ -997,20 +1022,11 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
         err("No namespaces configured. Run setup.py with --namespaces."); sys.exit(1)
 
     # Decide whether to build EPP image
-    run_meta_path = run_dir / "run_metadata.json"
-    if not run_meta_path.exists():
-        err("run_metadata.json not found — run setup.py first."); sys.exit(1)
-    try:
-        run_meta = json.loads(run_meta_path.read_text())
-    except json.JSONDecodeError as e:
-        err(f"run_metadata.json is not valid JSON: {e}. Re-run setup.py."); sys.exit(1)
-
-    has_component_image = bool(run_meta.get("component_image"))
-
-    if not has_component_image:
+    epp_action = _resolve_epp_action(run_dir, args.skip_build_epp)
+    if epp_action.startswith("error:"):
+        err(epp_action[len("error:"):]); sys.exit(1)
+    elif epp_action == "skip":
         info("No component_image in run metadata — skipping EPP build")
-    elif args.skip_build_epp:
-        info("Skipping EPP build (--skip-build-epp)")
     else:
         _build_epp_image(run_dir, run_dir.name, namespaces[0])
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -997,10 +997,17 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
         err("No namespaces configured. Run setup.py with --namespaces."); sys.exit(1)
 
     # Build EPP image (tag is deterministic — already embedded in PipelineRuns by prepare.py)
-    if not args.skip_build_epp:
-        _build_epp_image(run_dir, run_dir.name, namespaces[0])
-    else:
+    run_meta_path = run_dir / "run_metadata.json"
+    has_component_image = False
+    if run_meta_path.exists():
+        has_component_image = bool(json.loads(run_meta_path.read_text()).get("component_image"))
+
+    if not has_component_image:
+        info("No component_image in run metadata — skipping EPP build (baseline-only)")
+    elif args.skip_build_epp:
         info("Skipping EPP build (--skip-build-epp)")
+    else:
+        _build_epp_image(run_dir, run_dir.name, namespaces[0])
 
     max_retries = getattr(args, "max_retries", 2)
     poll_interval = getattr(args, "poll_interval", 30)
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -996,14 +996,19 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     if not namespaces or not namespaces[0]:
         err("No namespaces configured. Run setup.py with --namespaces."); sys.exit(1)
 
-    # Build EPP image (tag is deterministic — already embedded in PipelineRuns by prepare.py)
+    # Decide whether to build EPP image
     run_meta_path = run_dir / "run_metadata.json"
-    has_component_image = False
-    if run_meta_path.exists():
-        has_component_image = bool(json.loads(run_meta_path.read_text()).get("component_image"))
+    if not run_meta_path.exists():
+        err("run_metadata.json not found — run setup.py first."); sys.exit(1)
+    try:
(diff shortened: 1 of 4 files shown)


kalantar:
## Review: NEEDS CHANGES

PR correctly implements the core requirement (component optional when no algorithms) with clean validation logic and good test coverage of the manifest changes. However, the deploy.py change has a silent failure path that could cause hard-to-diagnose cluster failures.

### Must fix

1. **Missing `run_metadata.json` silently skips EPP build** — `run_metadata.json` is always written by `setup.py`. If it's absent, that signals a broken prerequisite (user forgot to run setup, file deleted, corrupted state) — not baseline-only mode. The current code treats both identically, logging a misleading "baseline-only" info message. A user who forgot `setup.py` will get no error here, then wait minutes/hours for PipelineRuns to fail with `ImagePullBackOff` on the cluster. Fix: error when `run_metadata.json` is missing, only skip when the file exists but lacks `component_image`.

2. **Unhandled `json.JSONDecodeError`** — The new `json.loads(run_meta_path.read_text())` has no exception handling. A corrupted file produces a raw traceback with no guidance. Other parts of the pipeline (e.g., `prepare.py:447-451`) already pattern-match this with try/except + actionable error message.

### Should fix

3. **Duplicate test** — `test_component_required_when_algorithms_present` and `test_algorithms_without_component_raises` have identical setup and assertions. Remove one.

4. **Comment is stale** — `# Build EPP image (tag is deterministic — ...)` now sits above a block that primarily decides *whether* to build. Update to reflect the three-way gate.

5. **Info message assumes "baseline-only"** — The parenthetical "(baseline-only)" is an inference about intent. The mechanism is simply "no component_image configured." Drop the assumption or use "(no custom EPP to deploy)".

### Consider

6. **Explicit `component: null` in YAML** — passes validation (component is None, no algorithms) but downstream `manifest.get("component", {})` returns None instead of `{}`, which would crash prepare.py with `AttributeError`. Could normalize by popping the key when it's None and no algorithms are present.

7. **No tests for deploy.py EPP skip logic** — The manifest tests are solid, but the deploy.py decision branch has zero test coverage. Consider extracting the EPP decision into a small testable helper.

### Looks good

- Validation logic in manifest.py is correct — `has_algorithms = bool(data.get("algorithms"))` works because normalization at line 91 has already run
- All four issue-specified test cases are covered
- setup.py claim is sound — line 689 already gates `component_image` on `cfg.registry`
- Error message is clear and actionable

kalantar:
Pushed fixes in 08110fe:

**Fixed:**
- deploy.py now errors on missing `run_metadata.json` (signals broken prerequisite, not baseline-only mode) and handles `JSONDecodeError` with an actionable message
- Removed duplicate test (`test_algorithms_without_component_raises` was identical to `test_component_required_when_algorithms_present`)
- Updated the stale comment above the EPP block to reflect the three-way decision
- Dropped "(baseline-only)" assumption from the info message
- Normalized explicit `component: null` in YAML to absent key so downstream `manifest.get("component", {})` never returns `None` — added test

**Follow-up:** Test coverage for the deploy.py EPP skip logic tracked in #114.

kalantar:

The manifest.py validation logic is correct and well-tested — all four issue #112 test cases are covered. The PR's claims about setup.py and MULTI_BASELINE_V3 are verified accurate. Two items should be addressed before merge:

### 1. deploy.py EPP-skip logic has no test coverage

The new 16-line block in `_cmd_run` introduces two `sys.exit(1)` paths and a three-way branch (no component_image / skip_build_epp / build). None of these paths are tested. Since this gates whether an image build happens — directly impacting whether experiments run correctly — unit tests should cover:

- Missing `run_metadata.json` → exits with error
- Malformed JSON → exits with error
- `component_image` absent → skips EPP build
- `component_image` present → builds EPP

### 2. Empty-string `component_image` is silently treated as absent

`bool(run_meta.get("component_image"))` evaluates to `False` for both a genuinely absent key AND an empty string `""`. If `setup.py` is misconfigured (e.g., `--registry ""`), the resulting empty `component_image` would silently skip the EPP build, and PipelineRuns would later fail with `ImagePullBackOff`. Consider distinguishing key-absent (intentional baseline-only) from key-present-but-empty (configuration error):

```python
component_image = run_meta.get("component_image")
if component_image is None:
    info("No component_image in run metadata — skipping EPP build")
elif not component_image:
    err("component_image is empty in run_metadata.json. Re-run setup.py with a valid --registry.")
    sys.exit(1)
elif args.skip_build_epp:
    info("Skipping EPP build (--skip-build-epp)")
else:
    _build_epp_image(run_dir, run_dir.name, namespaces[0])
```

### Minor (non-blocking)

- The `data.pop("component", None)` normalization in manifest.py is non-obvious — a brief inline comment explaining it removes explicit YAML null would help future readers.
- deploy.py module docstring and `run` subcommand description still say "Build EPP" unconditionally; now slightly misleading for baseline-only users.

kalantar:
Pushed 76bb1e8 addressing the second review round:

Extracted the EPP build decision into `_resolve_epp_action(run_dir, skip_build_epp)` — a pure function that returns `"build"`, `"skip"`, or `"error:<msg>"`. This makes the logic testable without mocking filesystem or subprocess calls in `_cmd_run`.

Added 6 tests covering every branch: missing metadata file, corrupt JSON, `component_image` absent (skip), `component_image` empty string (error — misconfigured setup), `--skip-build-epp` flag (skip), and normal build path.

Also added a clarifying comment on the `data.pop("component", None)` normalization in manifest.py.

Issue #114 (broader deploy.py test coverage) is now partially addressed by this commit — the EPP decision logic is fully covered. Leaving #114 open for any remaining orchestrator paths.

kalantar:
## Verdict: READY TO MERGE

PR correctly implements all requirements from #112. All four test cases specified in the issue are covered, the manifest validation logic is sound, and the EPP build gate handles the baseline-only mode cleanly.

### Key strengths

- Clean conditional: `component` required only when `algorithms` is non-empty — exactly as specified
- `component: null` normalization prevents downstream `TypeError` from `.get("component", {})` returning `None`
- `_resolve_epp_action()` properly distinguishes absent key (baseline-only → skip) from empty string (misconfigured → error)
- All six `_resolve_epp_action` branches tested; all four manifest test cases from the issue covered
- Justified skip of `setup.py` change — line 689 already gates on `if cfg.registry`

### Minor observations (non-blocking)

1. **Info message ambiguity**: When `_resolve_epp_action` returns `"skip"`, the caller always prints "No component_image in run metadata — skipping EPP build". This is inaccurate when skip was caused by `--skip-build-epp` (component_image IS present, user just asked to skip). Consider returning `"skip:no_image"` vs `"skip:flag"` or handling the flag check at the call site before invoking the helper.

2. **Uncaught OSError**: Between `exists()` and `read_text()`, a `PermissionError` or stale NFS handle would surface as an unhandled traceback rather than the friendly error message. Low probability but easy fix: `except (OSError, json.JSONDecodeError) as e:`.

3. **Lost architectural context**: The old comment "tag is deterministic — already embedded in PipelineRuns by prepare.py" explained the contract between prepare and deploy. Worth preserving near the build call.

None of these block merge — the core logic is correct and well-tested.

PR #116, opened 2026-05-13
Record HF secret name in setup_config.json instead of hardcoding

### LINKED ISSUE #115: Record HF secret name in setup_config.json instead of hardcoding
## Problem

`setup.py` creates a Kubernetes secret named `hf-secret` for the HuggingFace token, but this name is hardcoded as a string literal in multiple places:

- `pipeline/setup.py` (creation and existence check)
- `pipeline/deploy.py` (pre-flight validation)

The name is never recorded in `setup_config.json` or any other persisted config. Downstream, the llm-d-benchmark defaults expect `huggingface.secretName: llm-d-hf-token`, and we don't override it in our scenario assembly — so there's a mismatch between what we create and what the benchmark looks for.

## Fix

1. Add an `hf_secret_name` field to `setup_config.json` (written by `setup.py` step 8).
2. Replace all hardcoded `"hf-secret"` references in `setup.py` and `deploy.py` with reads from that config field.
3. In scenario assembly (`prepare.py` Phase 4 / `assemble.py`), inject `huggingface.secretName` into resolved scenarios using the value from `setup_config.json` — ensuring the benchmark uses the correct secret name.

This makes the secret name a single-source-of-truth setup-time decision, and closes the gap with the benchmark's expected `huggingface.secretName`.

kalantar:
Closes #115

## Summary

- Adds `hf_secret_name` field (default `"hf-secret"`) to `setup_config.json`, making the secret name a single-source-of-truth setup-time decision
- Replaces all hardcoded `"hf-secret"` string literals in `setup.py` and `deploy.py` with reads from config
- Adds `inject_hf_secret_name()` to the assembly module that stamps `huggingface.secretName` into all resolved scenario entries during Phase 4 — closing the mismatch with llm-d-benchmark's expected field

## Reviewer notes

- `inject_hf_secret_name` uses `setdefault` so it won't overwrite an explicit `secretName` already present in bundle YAML — bundles that already declare their own secret name are respected
- The `setup_config` load in `prepare.py` Phase 4 was moved earlier (before scenario writing) so the HF injection has access to it. The PipelineRun generation that previously triggered the load still works since `setup_config` remains in scope.

## Test plan

- [x] New tests for `hf_secret_name` in `setup_config.json` output
- [x] New tests for `_check_slot_ready` honoring the parameter
- [x] New tests for `inject_hf_secret_name` (basic injection, preserve existing fields, no-overwrite, empty scenarios)
- [x] Full suite: 510 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -902,7 +902,7 @@ def _report_filter_mismatch(progress: dict, args) -> None:
     print(f"  Valid --status values:   {', '.join(valid_statuses)}", file=sys.stderr)
 
 
-def _check_slot_ready(namespace: str) -> tuple[bool, list[str]]:
+def _check_slot_ready(namespace: str, hf_secret_name: str = "hf-secret") -> tuple[bool, list[str]]:
     """Check that a namespace slot is ready to accept a new PipelineRun.
 
     Checks: PVCs bound, HF secret present.
@@ -924,11 +924,11 @@ def _check_slot_ready(namespace: str) -> tuple[bool, list[str]]:
             failures.append(f"PVC {pvc} not Bound in {namespace}{hint}")
 
     result = run(
-        ["kubectl", "get", "secret", "hf-secret", f"-n={namespace}"],
+        ["kubectl", "get", "secret", hf_secret_name, f"-n={namespace}"],
         check=False, capture=True,
     )
     if result.returncode != 0:
-        failures.append(f"Secret hf-secret missing in {namespace}")
+        failures.append(f"Secret {hf_secret_name} missing in {namespace}")
 
     return len(failures) == 0, failures
 
@@ -1320,7 +1320,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
             dispatchable = pending
 
         for ns, pair_key in zip(free_slots, dispatchable):
-            ready, reasons = _check_slot_ready(ns)
+            hf_secret_name = setup_config.get("hf_secret_name", "hf-secret")
+            ready, reasons = _check_slot_ready(ns, hf_secret_name=hf_secret_name)
             if not ready:
                 warn(f"Slot {ns} not ready: {'; '.join(reasons)}")
                 continue
--- a/pipeline/lib/assemble.py
+++ b/pipeline/lib/assemble.py
@@ -190,3 +190,18 @@ def assemble_packages(
                     )
 
     return packages
+
+
+def inject_hf_secret_name(scenario_dict: dict, hf_secret_name: str) -> bool:
+    """Inject huggingface.secretName into all scenario entries.
+
+    Does not overwrite an explicitly set secretName.
+    Returns True if injection occurred, False if skipped (no scenarios).
+    """
+    scenario_list = scenario_dict.get("scenario", [])
+    if not scenario_list:
+        return False
+    for entry in scenario_list:
+        hf = entry.setdefault("huggingface", {})
+        hf.setdefault("secretName", hf_secret_name)
+    return True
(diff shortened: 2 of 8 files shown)


kalantar:
## Verdict: NEEDS CHANGES

This PR correctly addresses all three requirements from issue #115. The design is sound, tests are solid, and the implementation is well-structured. One issue needs fixing before merge:

### Required Change

**Check the return value of `inject_hf_secret_name` in `prepare.py`** — The return value is silently discarded. When the function returns `False` (no scenario entries), execution continues and writes incomplete YAML to disk. This contradicts the pattern used 6 lines above for `inject_epp_image`, which aborts on failure. The fix is one-line — match the existing pattern:

```python
for pkg in packages:
    injected = inject_hf_secret_name(pkg.resolved, hf_secret_name)
    if not injected:
        err(f"{pkg.name} has no 'scenario' entries — huggingface.secretName cannot be injected.")
        sys.exit(1)
```

### Recommended Improvements (non-blocking)

1. **Update `pipeline/README.md`** — Per contributing guidelines, Phase 4 behavior changed (new step 4b.6) and `setup_config.json` has a new field. The README should document both.

2. **Extract a module-level constant in `setup.py`** — `"hf-secret"` appears as a literal in both `step_secrets()` and `step_config_output()`. A single `_DEFAULT_HF_SECRET_NAME` constant prevents future drift between the two.

3. **Add an integration assertion** — The existing `test_cmd_run_baseline_only_no_algorithm` test exercises the full Phase 4 path but doesn't assert that `huggingface.secretName` appears in the written `cluster/baseline.yaml`. One line would close the loop.

### What looks good

- `inject_hf_secret_name` uses `setdefault` correctly at both levels — respects explicit overrides and preserves sibling fields
- The `_load_setup_config()` call was cleanly hoisted earlier in Phase 4 so it serves both HF injection and PipelineRun generation
- Unit tests cover the core contract thoroughly (injection, preservation, no-overwrite, empty input)
- `deploy.py` parameterization with backward-compatible default is clean

kalantar:
Pushed `5921f04` addressing the review:

**Fixed:**
- `inject_hf_secret_name` return value is now checked in `prepare.py` — aborts with an error if no scenario entries exist, matching the `inject_epp_image` pattern above it.
- Extracted `_DEFAULT_HF_SECRET_NAME` constant in `setup.py` so `step_secrets()` and `step_config_output()` share a single source for the default.

**Not addressed here (follow-up):**
- README update for Phase 4b.6 and the new `setup_config.json` field — will do in a separate commit or PR.
- Integration assertion for `huggingface.secretName` in written cluster YAML — the path is already exercised; the assertion would strengthen it but isn't blocking.

All 510 tests pass, lint clean.

kalantar:
## Verdict: READY TO MERGE

All three requirements from issue #115 are fully addressed: `hf_secret_name` is persisted in `setup_config.json`, hardcoded references are replaced, and `huggingface.secretName` is injected into resolved scenarios during Phase 4. Tests cover the core logic well (510 passing, lint clean), and the `setdefault` pattern correctly preserves explicit overrides.

### Non-blocking suggestions

1. **Duplicated default string** — The fallback `"hf-secret"` appears in 4 locations (`setup.py` constant, `deploy.py` function default + `.get()` call, `prepare.py` `.get()` call). If the default ever changes, only `setup.py` would be updated — the others would silently drift. Consider importing the constant or removing the fallbacks (since `setup.py` always writes the field now, a missing key means "re-run setup").

2. **Type validation in `inject_hf_secret_name`** — If a bundle YAML has a malformed `scenario` field (e.g. `scenario: true` or entries that are strings instead of dicts), the function will raise an unhelpful `AttributeError` rather than a clear schema error. Adding `isinstance` checks would improve the error experience for users with malformed bundles.

3. **Minor docstring staleness** — `step_secrets` docstring still says "Create/update hf-secret" literally. Could say "HF token secret" generically since the name is now variable.

None of these block the merge — the happy path is correct, error paths abort appropriately (matching the `inject_epp_image` pattern), and backward compatibility is maintained via fallback defaults.

Issue #117, opened 2026-05-13
Capacity log lines: show per-pair cost at dispatch, suppress during idle polling


kalantar:
## Parent

Split from #87 (comment items). Related to #61 (Resource-availability-aware orchestration).

## Problem

Two UX issues with capacity logging in `deploy.py run`:

### 1. GPU cost announced upfront as a single uniform value

The current output shows one cost at startup:

```
[INFO]  GPU cost per pair: 4
```

Once #87 introduces per-pair heterogeneous costs, this line is no longer meaningful. Instead, per-pair cost should be shown in context at dispatch time.

**Desired output:**

```
[INFO]  wl-w1-mid-baseline requires 4 GPUs
[INFO]  Capacity: 9 free GPUs (112 allocatable − 103 requested)
[OK]    [wl-w1-mid-baseline] → kalantar-0 (baseline-w1-mid-rf)
```

### 2. Capacity probe logs every poll cycle regardless of dispatch activity

The capacity log at `deploy.py:1268` fires unconditionally, producing a wall of repeated lines during steady-state when all pairs are already running:

```
[INFO]  Capacity: 11 free GPUs (112 allocatable − 101 requested)
[INFO]  Capacity: 15 free GPUs (112 allocatable − 97 requested)
[INFO]  Capacity: 11 free GPUs (112 allocatable − 101 requested)
```

The probe itself must still run every cycle (it feeds the backoff controller), but the log line should only appear when there's a pending pair and a free slot — i.e., when capacity is decision-relevant.

**Desired: on dispatch:**
```
[INFO]  wl-w1-mid-baseline requires 4 GPUs
[INFO]  Capacity: 15 free GPUs (112 allocatable − 97 requested)
[OK]    [wl-w1-mid-baseline] → kalantar-0 (baseline-w1-mid-rf)
```

**Desired: on failure:**
```
[INFO]  wl-w1-mid-baseline requires 4 GPUs
[INFO]  Capacity: 2 free GPUs (112 allocatable − 110 requested)
[WARN]  Dispatching 0/3 pending pairs — smallest cost (4) exceeds free GPUs (2)
```

**Desired: when nothing to dispatch:** silence (no capacity line).

## Acceptance Criteria

- [ ] Remove upfront `GPU cost per pair: N` log line (or replace with a summary range if costs vary)
- [ ] Log each pair's GPU cost immediately before its capacity check and assignment
- [ ] Suppress capacity log line when there are no pending pairs or no free slots
[...]

PR #118, opened 2026-05-13
Per-pair GPU cost derivation from scenarioContent

### LINKED ISSUE #87: Per-pair GPU cost is uniform, not heterogeneous
## Parent

Follow-up to #61 (Resource-availability-aware orchestration) — specifically §2 "Workload GPU Cost".

## Problem

Issue #61 specifies:

> "Different PipelineRuns may have different costs. The orchestrator maintains a map of `pair_key → gpu_cost` populated at startup."

The implementation computes one cost from `baseline.yaml` and assigns it identically to every pair.

**Single derivation** (`deploy.py:1011-1023`): cost is computed once from `baseline.yaml` only:

```python
baseline_scenario_path = cluster_dir / "baseline.yaml"          # line 1011
resolved = yaml.safe_load(baseline_scenario_path.read_text())   # line 1014
derived_cost = gpu_cost_per_pair(resolved, defaults_result)     # line 1021 — called ONCE
pair_gpu_cost = derived_cost                                    # line 1023 — single scalar
```

**Uniform assignment** (`deploy.py:1053-1061`): that scalar is stamped on every pair:

```python
for key, meta in discovered.items():        # line 1053
    progress[key] = {
        "gpu_cost": pair_gpu_cost,          # line 1061 — same for all pairs
    }
```

If the treatment arm uses different replica counts than the baseline, or if different workloads have different GPU requirements, the capacity gating will use wrong costs — potentially dispatching more work than the cluster can schedule, or under-dispatching.

## Impact

Capacity-proportional dispatch (§5) sorts pending pairs by `gpu_cost` and packs them into available GPUs. When all pairs share the same cost, the algorithm cannot distinguish cheap pairs from expensive ones, defeating the purpose of capacity-aware scheduling.

## Fix

The data for per-pair cost already exists. `_load_pairs` (`deploy.py:649-653`) reads each `pipelinerun-*.yaml` and extracts params — including `scenarioContent`, which contains the fully resolved scenario for that specific pair (baseline or treatment). `gpu_cost_per_pair()` in `pipeline/lib/capacity.py` already accepts any resolved scenario.

Move cost derivation into the per-pair init loop with the following fallback chain:

1. **`scenarioContent` from the PipelineRun** — parse per-pair `scenarioContent` from the PipelineRun YAML (already loaded by `_load_pairs`), call `gpu_cost_per_pair(yaml.safe_load(scenario_content), defaults)`.
2. **`defaults.yaml`** — if `scenarioContent` is missing or unparseable, call `gpu_cost_per_pair({}, defaults)` using the defaults already loaded via `load_defaults()`.
3. **`baseline.yaml`** — if defaults are also unavailable, fall back to the current single-derivation from `baseline.yaml`.
4. **CLI `--default-gpu-cost`** — final fallback (already exists).

kalantar:
## Summary

- Extends `_load_pairs` to extract the `scenarioContent` param from each PipelineRun YAML
- Adds `_derive_pair_gpu_costs()` which computes GPU cost individually per pair using `gpu_cost_per_pair()` with fallback chain: scenarioContent → defaults-only → CLI `--default-gpu-cost`
- Replaces the uniform `pair_gpu_cost` scalar in the orchestrator init with per-pair costs dict
- Log output now shows a range ("4–8 (heterogeneous)") when pairs have different costs

Closes #87

## Reviewer notes

- The `gpu_cost_per_pair` import moves from `_cmd_run` (no longer needed there) into `_derive_pair_gpu_costs` as a local import
- Existing progress entries without a `gpu_cost` field still fall back to `fallback_cost` (the CLI `--default-gpu-cost` value) — backward compatible with running progress.json files
- The capacity log UX changes (per-pair cost at dispatch time, suppressing idle polls) are tracked separately in #117

## Test plan

- [x] `test_load_pairs_includes_scenario_content` — extraction when param present
- [x] `test_load_pairs_missing_scenario_content` — None when param absent
- [x] `test_derive_pair_gpu_costs_heterogeneous` — different scenarios → different costs
- [x] `test_derive_pair_gpu_costs_fallback_on_missing_scenario` — None scenario → defaults-only
- [x] `test_derive_pair_gpu_costs_fallback_on_bad_yaml` — invalid YAML → defaults-only
- [x] `test_derive_pair_gpu_costs_no_defaults` — no defaults → fallback_cost
- [x] `test_init_progress_per_pair_heterogeneous_cost` — end-to-end integration
- [x] All 517 existing tests pass (no regressions)

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -991,10 +991,12 @@ def _derive_pair_gpu_costs(
 ) -> dict[str, int]:
     """Compute GPU cost per pair from its scenarioContent.
 
-    Fallback chain per pair:
+    If defaults is None, returns fallback_cost for all pairs immediately.
+
+    Otherwise, per pair:
       1. Parse scenarioContent → gpu_cost_per_pair(scenario, defaults)
       2. If scenarioContent missing/invalid → gpu_cost_per_pair({}, defaults)
-      3. If defaults unavailable or derivation fails → fallback_cost
+      3. If derivation returns an error → fallback_cost
     """
     from pipeline.lib.capacity import gpu_cost_per_pair
 
@@ -1009,8 +1011,8 @@ def _derive_pair_gpu_costs(
         if scenario_content:
             try:
                 resolved = yaml.safe_load(scenario_content)
-            except yaml.YAMLError:
-                pass
+            except yaml.YAMLError as e:
+                warn(f"{key}: scenarioContent is invalid YAML ({e}) — deriving cost from defaults")
 
         if resolved and isinstance(resolved, dict):
             result = gpu_cost_per_pair(resolved, defaults)
@@ -1020,6 +1022,7 @@ def _derive_pair_gpu_costs(
         if isinstance(result, int):
             costs[key] = result
         else:
+            warn(f"{key}: GPU cost derivation failed: {result} — using fallback ({fallback_cost})")
             costs[key] = fallback_cost
 
     return costs
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -983,6 +983,48 @@ def _do_collect(pair_key: str, entry: dict, run_dir: Path, store, progress: dict
     return ok
 
 
+def _derive_pair_gpu_costs(
+    discovered: dict,
+    *,
+    defaults: dict | None,
+    fallback_cost: int,
+) -> dict[str, int]:
+    """Compute GPU cost per pair from its scenarioContent.
+
+    Fallback chain per pair:
+      1. Parse scenarioContent → gpu_cost_per_pair(scenario, defaults)
+      2. If scenarioContent missing/invalid → gpu_cost_per_pair({}, defaults)
+      3. If defaults unavailable or derivation fails → fallback_cost
+    """
+    from pipeline.lib.capacity import gpu_cost_per_pair
+
+    costs = {}
+    for key, meta in discovered.items():
+        if defaults is None:
+            costs[key] = fallback_cost
+            continue
+
+        scenario_content = meta.get("scenario_content")
+        resolved = None
+        if scenario_content:
+            try:
+                resolved = yaml.safe_load(scenario_content)
+            except yaml.YAMLError:
+                pass
+
+        if resolved and isinstance(resolved, dict):
+            result = gpu_cost_per_pair(resolved, defaults)
+        else:
+            result = gpu_cost_per_pair({}, defaults)
+
+        if isinstance(result, int):
+            costs[key] = result
+        else:
+            costs[key] = fallback_cost
+
+    return costs
+
+
 def _capacity_limited_pairs(
     pending: list[str],
     progress: dict,
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -696,6 +696,7 @@ def _load_pairs(cluster_dir: Path) -> dict:
                 "pr_name": pr_name,
                 "pr_path": str(pr_path),
                 "namespace": pr_data.get("metadata", {}).get("namespace", ""),
+                "scenario_content": params.get("scenarioContent"),
             }
         except Exception as e:
             warn(f"Skipping {pr_path.name}: {e}")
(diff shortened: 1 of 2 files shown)


kalantar:

The architecture and logic are sound — per-pair cost derivation is correctly implemented with a clean fallback chain, good test coverage, and proper backward compatibility. One blocking issue and one minor documentation fix before merge.

### Blocking: Observability regression in `_derive_pair_gpu_costs`

The old code in `_cmd_run` logged warnings when cost derivation failed. The new `_derive_pair_gpu_costs` function silently falls back in two places, removing visibility an operator needs when debugging incorrect GPU scheduling:

1. **Silent YAML parse failure** — the `except yaml.YAMLError: pass` block swallows errors without logging. The old code had `warn(f"Could not parse {name}: {e}")`. Fix:
   ```python
   except yaml.YAMLError as e:
       warn(f"{key}: scenarioContent is invalid YAML ({e}) — deriving cost from defaults")
   ```

2. **Silent fallback on derivation error** — when `gpu_cost_per_pair()` returns an error string (e.g., non-numeric accelerator count), the diagnostic is discarded. The old code had `warn(f"GPU cost derivation failed: {derived_cost} — using fallback (...)")`. Fix:
   ```python
   if isinstance(result, int):
       costs[key] = result
   else:
       warn(f"{key}: GPU cost derivation failed: {result} — using fallback ({fallback_cost})")
       costs[key] = fallback_cost
   ```

These are 2 one-line additions that restore the observability the old code had.

### Non-blocking: Docstring accuracy

The docstring's step 3 ("If defaults unavailable or derivation fails → fallback_cost") conflates two distinct code paths. When `defaults is None`, steps 1 and 2 are never attempted — the function short-circuits immediately. Consider restructuring to make the early-exit explicit (e.g., add a precondition note before the numbered chain).

### What looks good

- Clean extraction of `scenarioContent` in `_load_pairs` with appropriate None for absent params
- Pure helper function with clear types and no side effects
- Fallback chain design is correct and well-tested (7 new tests covering all branches)
- Backward compatibility preserved for existing progress.json entries
- Logging UX changes correctly deferred to #117
- CI passing (517 tests, lint clean)

kalantar:
All three items addressed in 63d3b33:

- Added `warn()` on YAML parse failure in `_derive_pair_gpu_costs` — logs the pair key and error before falling back to defaults-only derivation.
- Added `warn()` when `gpu_cost_per_pair()` returns an error string — logs the pair key, the error, and the fallback value being used.
- Restructured the docstring to make the `defaults is None` early-exit explicit (separate from the per-pair fallback chain).

All 517 tests still pass, lint clean.

kalantar:

PR #118 correctly implements all requirements from issue #87: per-pair GPU cost derivation from scenarioContent with proper fallback chain, heterogeneous cost logging, and backward compatibility with existing progress.json files.

### Key Strengths

- Clean separation of concerns — `_derive_pair_gpu_costs` is a focused function with a well-documented fallback chain
- Comprehensive test coverage across all primary branches (7 new tests)
- Backward compatibility preserved via `progress[k].get("gpu_cost", fallback_cost)` for existing entries
- UX logging correctly distinguishes uniform vs heterogeneous cost scenarios

### Non-blocking Suggestions (consider for follow-up)

1. **Add a try/except around `gpu_cost_per_pair` calls** — if `gpu_cost_per_pair` raises an unexpected exception (vs returning an error string), one malformed pair would crash the entire orchestrator. Wrapping the call protects all pairs from one bad scenario.

2. **Warn when scenarioContent parses to non-dict** — valid YAML that's a list or scalar silently falls through to defaults-only derivation with no warning. A log message would aid debugging.

3. **Test the error-string return path** — the `gpu_cost_per_pair` returning a string (error) branch is untested. Adding a test that triggers this fallback would lock the behavior.

4. **Comment accuracy** — the comment `# Derive GPU resource type from baseline scenario` now understates the block's responsibilities (it also initializes `defaults_result` and `fallback_cost` used by per-pair derivation). Consider expanding to `# Load GPU defaults and derive resource type from baseline scenario`.

None of these are blockers — the implementation is correct and well-tested as-is.

PR #119, opened 2026-05-13
Capacity log: show per-pair cost at dispatch, suppress during idle

### LINKED ISSUE #117: Capacity log lines: show per-pair cost at dispatch, suppress during idle polling

kalantar:
Closes #117

## Summary

- Removed the upfront "GPU cost per pair: N" log line that becomes meaningless once per-pair heterogeneous costs land (#87)
- Added per-pair GPU cost log (`{pair_key} requires {N} GPUs`) at dispatch time, giving context to the capacity decision
- Suppressed the capacity log line during idle polling — it only fires when there are pending pairs awaiting dispatch

## What doesn't change

- `probe_free_gpus()` still runs every cycle (backoff controller still gets data)
- All dispatch decisions and capacity gating logic are unchanged
- `_derive_pair_gpu_costs()` still computes and stores costs in progress dict

## Test plan

- [x] `python -m pytest pipeline/ -v` — 517 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1137,11 +1137,6 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     pair_costs = _derive_pair_gpu_costs(
         discovered, defaults=defaults_result, fallback_cost=fallback_cost,
     )
-    unique_costs = set(pair_costs.values())
-    if len(unique_costs) == 1:
-        info(f"GPU cost per pair: {unique_costs.pop()}")
-    else:
-        info(f"GPU cost per pair: {min(unique_costs)}–{max(unique_costs)} (heterogeneous)")
 
     # Initialize new entries (first run or new pairs added)
     for key, meta in discovered.items():
@@ -1314,7 +1309,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
         capacity = probe_free_gpus(gpu_resource_type=gpu_resource_type)
         if isinstance(capacity, tuple):
             free_gpus, allocatable, requested = capacity
-            info(f"Capacity: {free_gpus} free GPUs ({allocatable} allocatable − {requested} requested)")
+            if _pending_pairs():
+                info(f"Capacity: {free_gpus} free GPUs ({allocatable} allocatable − {requested} requested)")
             if _probe_fail_count > 0:
                 info(f"Capacity probe recovered after {_probe_fail_count} failure(s)")
             _probe_fail_count = 0
@@ -1376,6 +1372,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 continue
 
             entry = progress[pair_key]
+            pair_cost = entry.get("gpu_cost", fallback_cost)
+            info(f"{pair_key} requires {pair_cost} GPUs")
             pr_meta = discovered.get(pair_key, {})
             pr_path_str = pr_meta.get("pr_path", "")
             if not pr_path_str:


kalantar:

PR is well-scoped and mostly correct, but has two issues worth addressing before merge.

### Must fix

**Diagnostic blind spot when capacity drops while all pairs are running.** The `if _pending_pairs():` guard suppresses the capacity log whenever no pairs have status `"pending"`. This means if GPU capacity drops (node failure, preemption) while all pairs are in `"running"` state, the operator sees no capacity information for potentially hours — the exact window when it matters most for diagnosing timeouts. Consider gating on `_work_remaining()` instead, or at minimum `_pending_pairs() or any(...)` that includes running pairs.

### Should fix

**Acceptance criterion 3 partially unmet.** The issue asks to suppress capacity logging "when there are no pending pairs *or no free slots*." The PR only checks `_pending_pairs()`. When all namespace slots are occupied but pairs remain pending, the capacity line still prints (arguably noise since nothing can be dispatched). If this is intentional, a comment in the issue acknowledging the deviation would suffice.

### Worth noting (non-blocking)

- **Log ordering**: Issue desired "pair cost → capacity → assignment" but the implementation produces "capacity → pair cost → assignment." This is architecturally natural (capacity is a global probe before the per-pair loop) and likely fine — just confirm intent.
- **Silent fallback**: `entry.get("gpu_cost", fallback_cost)` silently uses the fallback for legacy progress entries without any log. A `warn()` when the fallback triggers would help operators spot misconfiguration.
- **No test regressions**: All 517 tests pass. No existing tests assert on the removed/changed log strings. Test coverage of the underlying logic (cost derivation, capacity gating) is solid.
- **No stale comments or docs**: Documentation correctly describes behavior at a semantic level, unaffected by these log format changes.

kalantar:
Thanks for the review. Addressing each point:

**Diagnostic blind spot (must fix)** — Keeping `_pending_pairs()` as the guard. The concern about node failures while all pairs are running is valid, but this case is already covered by existing mechanisms: the scarcity warning at the backoff section fires when `free_gpus < min_cost` (including when a pair gets requeued to pending after timeout), and the pending-pod handler produces explicit warnings when pods can't schedule. The capacity log's purpose here is to contextualize dispatch decisions, not to be a general cluster health monitor. Using `_work_remaining()` would effectively suppress nothing (it's the loop condition), defeating the issue's goal.

**AC3 partially unmet — "or no free slots" (should fix)** — Intentionally kept. When all slots are occupied but pairs remain pending, the capacity line contextualizes the "slot-limited" message that follows it (line 1362). Suppressing it there would remove useful context. The issue's core intent is eliminating the wall of repeated capacity lines during steady-state when *nothing needs dispatching* — that's achieved.

**Log ordering** — Confirmed intentional. Capacity is a global probe that runs before the per-pair dispatch loop. Moving it inside the loop would require re-probing per pair (wasteful) or caching (complexity for no gain).

**Silent fallback** — Agreed this is worth a follow-up but out of scope for a log-suppression PR.

No code changes from this review — keeping `_pending_pairs()` as originally implemented.

PR #129, opened 2026-05-14
Auto-delete successful PipelineRuns after collection

### LINKED ISSUE #121: Auto-delete successful PipelineRuns after completion
## Context

Today the orchestrator leaves all PipelineRuns on the cluster after completion (both successful and failed). This consumes cluster resources (etcd storage, API server load) during long multi-pair runs.

## Requirements

- When a PipelineRun succeeds and the pair is marked `done`, the orchestrator deletes the PipelineRun from the cluster
- Failed PipelineRuns are left in place for debugging (`kubectl describe`, logs)
- This behavior applies in both local and remote execution modes

## Rationale

- Successful runs have their data on the PVC — the PipelineRun object is no longer needed
- Failed runs need the object for debugging (pod logs, events, conditions)
- `reset` (formerly `cleanup`) handles clearing failed PipelineRuns when the user is done debugging

## Notes

Independent of remote execution work. Can be implemented standalone.

kalantar:
Closes #121

## Summary

- Adds `_delete_pipelinerun` helper — best-effort `kubectl delete pipelinerun --ignore-not-found` with a warning on failure
- Modifies `_do_collect` to accept an optional `pr_name` kwarg; when collection succeeds, deletes the PipelineRun CR before returning
- Updates the poll loop call site to pass `pr_name` through

## Design notes

- **Namespace capture**: `_do_collect` captures `entry["namespace"]` before the try/finally block, because `finally` clears it to `None`. The captured value is used for the deletion call.
- **Best-effort**: Deletion failure warns but doesn't crash — the PipelineRun will be cleaned up by `reset` later.
- **Backward compatible**: `pr_name` defaults to `""`, so existing internal callers (`_reconcile_collecting`) are unaffected.
- **Failed runs untouched**: Only the `Succeeded` branch calls `_do_collect` with `pr_name`; the failure branch never triggers deletion.

## Test plan

- [x] `test_do_collect_deletes_pipelinerun_on_success` — verifies deletion is called with correct pr_name and namespace
- [x] `test_do_collect_skips_delete_on_failure` — verifies no deletion when collection fails
- [x] `test_do_collect_skips_delete_on_interrupt` — verifies no deletion on KeyboardInterrupt
- [x] Existing `test_do_collect_saves_done_on_success` still passes (backward compat)
- [x] Existing `test_do_collect_interrupt_saves_collect_failed` still passes
- [x] Full suite: 520 tests pass

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -165,6 +165,8 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 **Pair statuses:** `pending` → `running` → `collecting` → `done` | `collect-failed`. Failure paths: `running` → `failed` (hard failure or non-recoverable pending), `running` → `timed-out` (4h timeout exceeded), `running` → `pending` (recoverable early reclaim, repeats up to `--max-pending-stalls` times) → `stalled`.
 
+**Auto-cleanup** — when a PipelineRun succeeds and results are collected, the orchestrator deletes the PipelineRun CR from the cluster. Failed PipelineRuns are left in place for debugging (`kubectl describe`, pod logs). Use `cleanup` to remove them when done.
+
 **`deploy.py status`** — prints the current state of all pairs from `workspace/runs/<run>/progress.json`. When the orchestrator is in backoff, an additional line shows the current state and backoff level.
 
 | Flag | Description |
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -195,7 +195,9 @@ def _delete_pipelinerun(pr_name: str, namespace: str) -> None:
         check=False, capture=True,
     )
     if result.returncode != 0:
-        warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}")
+        detail = result.stderr.strip() if result.stderr else ""
+        warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}" +
+             (f": {detail}" if detail else ""))
 
 
 # ── Deploy command ───────────────────────────────────────────────────────────
@@ -994,7 +996,10 @@ def _do_collect(pair_key: str, entry: dict, run_dir: Path, store, progress: dict
         entry["namespace"] = None
         store.save(progress)
     if ok and pr_name and namespace:
-        _delete_pipelinerun(pr_name, namespace)
+        try:
+            _delete_pipelinerun(pr_name, namespace)
+        except Exception as exc:
+            warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}: {exc}")
     return ok
 
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -187,6 +187,17 @@ def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> None:
     )
 
 
+def _delete_pipelinerun(pr_name: str, namespace: str) -> None:
+    """Delete a completed PipelineRun. Best-effort — warns on failure."""
+    result = run(
+        ["kubectl", "delete", "pipelinerun", pr_name, "-n", namespace,
+         "--ignore-not-found"],
+        check=False, capture=True,
+    )
+    if result.returncode != 0:
+        warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}")
+
+
 # ── Deploy command ───────────────────────────────────────────────────────────
 
 # ── Status command ───────────────────────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


kalantar:
## Review: NEEDS CHANGES

Clean, well-structured PR that satisfies all issue #121 requirements. Two actionable items should be addressed before merge.

### What's good

- All three requirements met: successful PipelineRuns are deleted, failed ones are preserved, single code path covers all execution modes.
- Namespace captured before `try/finally` — correct and well-motivated.
- Deletion placed after `finally` block, so progress is always persisted before any cleanup attempt.
- Three new tests cover the primary behavioral contract (success, failure, interrupt).
- Keyword-only `pr_name` parameter with `""` default maintains backward compatibility.

### Changes requested

**1. Include kubectl stderr in the warning message**

`_delete_pipelinerun` captures output (`capture=True`) but the warning discards it. When deletion fails, the operator has no way to distinguish a transient API server hiccup from a permanent RBAC misconfiguration without manually re-running the command.

```python
if result.returncode != 0:
    detail = result.stderr.strip() if result.stderr else ""
    warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}" +
         (f": {detail}" if detail else ""))
```

**2. Guard the `_delete_pipelinerun` call against unexpected exceptions**

The call sits outside `_do_collect`'s `try/except/finally` block. If `subprocess.run` itself raises (e.g., `OSError` when kubectl is missing from PATH), the exception propagates uncaught — crashing the caller and misreporting a successful collection as failed. Since deletion is explicitly best-effort, wrap it:

```python
if ok and pr_name and namespace:
    try:
        _delete_pipelinerun(pr_name, namespace)
    except Exception as exc:
        warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}: {exc}")
```

A matching test (`_delete_pipelinerun` raises `OSError` → `_do_collect` still returns `True`) would close the gap.

### Minor suggestions (non-blocking)

- **PR body accuracy**: The "Backward compatible" bullet names `_reconcile_collecting` as an internal caller of `_do_collect`, but it calls `_collect_pair` directly. Worth correcting for reader clarity.
- **README update**: `pipeline/README.md` describes `deploy.py run` behavior and pair status flow. Adding a note that successful PipelineRun CRs are deleted after collection would prevent operator surprise.

kalantar:
Addressed the two requested changes and the README suggestion:

**stderr in warning** — `_delete_pipelinerun` now includes `result.stderr` in the warning message so operators can distinguish transient API server issues from RBAC misconfiguration without re-running manually.

**Guard against OSError** — The `_delete_pipelinerun` call in `_do_collect` is now wrapped in `try/except Exception`, so a missing kubectl binary (or similar system error) warns instead of crashing the orchestrator after a successful collection. Added a matching test (`test_do_collect_returns_true_when_delete_raises`).

**README** — Added an "Auto-cleanup" paragraph to the pair status section in `pipeline/README.md` documenting that successful PipelineRun CRs are deleted after collection, and that failed ones are left for debugging.

Skipped the PR body note about `_reconcile_collecting` — reviewer is right that it calls `_collect_pair` directly, not `_do_collect`, but it's just a description inaccuracy in the PR text.

521 tests pass, lint clean.

kalantar:
## READY TO MERGE

Clean, well-scoped PR that addresses all actionable requirements from #121. The implementation is correct, backward compatible, and well-tested.

**What it does right:**
- Namespace is captured before the try/finally block that clears it — the deletion call uses the saved value, not the already-cleared entry. This subtlety is handled correctly and tested.
- Deletion runs *after* progress is saved with `status = "done"`, so a crash during kubectl delete loses nothing — just an orphaned CR that `cleanup` removes later.
- The `except Exception` guard around `_delete_pipelinerun` is appropriate here: collection already succeeded, data is safe, and the function must return `True`. Letting a deletion failure propagate would be strictly worse.
- Failed PipelineRuns are correctly left in place — only the `Succeeded` branch passes `pr_name` to `_do_collect`.
- `pr_name=""` default preserves backward compat for all existing callers.

**Minor observations (none blocking):**
- `_delete_pipelinerun`'s docstring says "warns on failure" but the function can raise `OSError` (e.g. missing kubectl) without warning — the caller's `except Exception` catches it instead. Practically harmless since there's one call site, but the helper's contract isn't fully self-contained.
- The PR body references a `reset` subcommand that doesn't exist — should say `cleanup`. The committed README uses the correct term, so no code issue.
- Test coverage is solid (4 new tests covering success, failure, interrupt, and deletion error). Optional additions for the future: a test for the empty-namespace guard path and direct unit tests for `_delete_pipelinerun` warning formatting.

520 tests passing, lint clean.

PR #130, opened 2026-05-14
Remove inline collection from orchestrator loop

### LINKED ISSUE #120: Remove inline collection from orchestrator loop
## Context

`deploy.py run` currently collects results inline (pulls data from PVC to local filesystem) as each PipelineRun succeeds. This was originally added for early partial results / fail-fast analysis.

However, `deploy.py collect` can already pull partial results at any time (for completed pairs). Inline collection adds complexity to the orchestrator and creates an asymmetry when the orchestrator runs remotely (in-cluster Job) vs locally — in remote mode, collecting to pod-ephemeral-fs is pointless since the data is already on the PVC.

## Requirements

- Remove `_collect_pair` / `_do_collect` calls from the orchestration loop in `_cmd_run`
- Remove the `collecting` status from the pair state machine (pairs go directly from `running` → `done` on PipelineRun success)
- Remove `_reconcile_collecting` logic
- `deploy.py collect` becomes the sole mechanism for pulling data off-cluster
- Both local and remote execution modes behave identically (pure dispatch + monitor)

## Behavioral change

- Before: PipelineRun succeeds → status `collecting` → inline pull → status `done`
- After: PipelineRun succeeds → status `done`

Users who want early partial results use `deploy.py collect --workload X` mid-flight (see #122 for scoping flags).

## Notes

This is a prerequisite for remote execution (#127) but is independently valuable as a simplification.

kalantar:
Closes #120

## Summary

- Remove `_collect_pair`, `_reconcile_collecting`, `_do_collect` from `deploy.py` — these performed inline result collection during the orchestrator loop
- Pairs now transition directly `running → done` on PipelineRun success (no intermediate `collecting` status)
- `deploy.py collect` is the sole mechanism for pulling results off-cluster
- Remove `collecting` and `collect-failed` from the pair state machine, shell completions, and all tests

## Behavioral change

**Before:** PipelineRun succeeds → status `collecting` → inline pull from PVC → status `done`
**After:** PipelineRun succeeds → status `done`

Users who want partial results mid-flight use `deploy.py collect`.

## Reviewer notes

- The `_extract_phases_from_pvc` function is kept — it serves `deploy.py collect`. Only its `workload` parameter docstring was trimmed (it no longer mentions inline collection).
- Stale `progress.json` files from previous runs that contain `"collecting"` entries will be inert (not picked up by `_work_remaining`). Users can `--force` reset them. This is acceptable for a one-time migration edge.
- Net deletion: ~72 lines of production code, 6 test functions removed, remaining tests updated.

## Test plan

- [x] `python -m pytest pipeline/ -v` — 511 tests pass
- [x] Grep for stale `collecting`/`collect-failed` references — zero matches
- [x] Grep for deleted function names — zero matches


kalantar:
## Review: NEEDS CHANGES

Clean simplification that addresses all five requirements from #120. The production code changes are correct — namespace slots are freed at the right time, `_work_remaining` no longer waits on a nonexistent status, and `_cmd_collect` filters correctly on `"done"` only. Net deletion of ~72 lines of production code and 6 test functions (all for deleted functions) is well-executed.

Two items to address before merge:

### 1. `pipeline/README.md` not updated (required by CLAUDE.md)

CLAUDE.md states: *"After any change to `pipeline/` that affects CLI flags, phase behavior, artifact schema, or subcommands, update `pipeline/README.md` to match."*

Four stale references:
- **Line 131:** says `deploy.py collect` uses packages with status `done` or `collecting` — should be just `done`
- **Line 145:** says `deploy.py run` "collects results inline" — no longer true
- **Line 166:** documents state machine as `pending → running → collecting → done | collect-failed` — should be `pending → running → done`
- **Line 303:** "collecting results inline" in the parallel pool execution section — should describe dispatch + monitor only

### 2. Stale `collecting` entries in progress.json (migration edge)

The PR body acknowledges this: *"Stale progress.json files from previous runs that contain 'collecting' entries will be inert. Users can --force reset them."* The `_force_reset` function does handle this generically (resets any non-pending, non-done status). However, without a warning log, users upgrading mid-run won't know why pairs are stuck. Consider adding a `warn()` in the reconciliation loop for unrecognized statuses, so users discover the issue without needing to debug manually.

### Optional improvements (non-blocking)

- **Test gap:** The resume-reconcile `Succeeded → done` transition (which sets `namespace = None`) has zero test coverage both before and after this PR. This is the most frequently exercised new code path — a focused test would catch regressions to namespace cleanup on resume.
- **Defensive safety:** `del slots_busy[ns]` in the Succeeded branch could use `slots_busy.pop(ns, None)` to guard against future refactors that might modify `slots_busy` earlier in the loop.
- **Dead code:** The `workload` parameter on `_extract_phases_from_pvc` is now unused — its only caller (`_collect_pair`) was deleted. Worth removing or documenting as retained for future use.

kalantar:
Pushed e3c496f addressing the review feedback:

**README** — Updated all 4 stale references: collection phases filter (line 131), `deploy.py run` description (line 145), pair status state machine (line 166), and parallel pool execution section (line 303).

**Stale status warning** — Added a `warn()` in the reconciliation loop for unrecognized statuses (e.g. `collecting` from a pre-upgrade progress.json). Users now get `[wl-foo] unrecognized status 'collecting' — use --force to reset` instead of silent inaction.

**Dead `workload` parameter** — Removed the `workload` parameter from `_extract_phases_from_pvc` and its two branching code paths (~20 lines). The only caller was the deleted `_collect_pair`; the remaining caller (`_cmd_collect`) never passed it.

**Skipped** the `pop` vs `del` suggestion — `ns` is iterated from `slots_busy.keys()` so it's always present; `del` is correct.

**Follow-up** — Opened #132 for adding test coverage to the resume-reconcile `Succeeded → done` path (non-blocking).

All 511 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

Clean simplification that fulfills all five requirements of issue #120. The state machine reduction, ~72 lines of production code deleted, and consistent cleanup across completions/tests/docs are well done. One critical regression and a few smaller items before merge.

### Critical

**PipelineRun auto-delete lost (regression from PR #129)** — PR #129 added `_delete_pipelinerun(pr_name, namespace)` inside `_do_collect` so successful PipelineRuns are cleaned up from the cluster. This PR deletes `_do_collect` but doesn't preserve the auto-delete call in the new `Succeeded` path. Result: successful PipelineRun CRs accumulate on the cluster, `_delete_pipelinerun` becomes dead code, and the README auto-cleanup claim (line 168) becomes false.

Fix: add the delete call in the Succeeded branch of the orchestrator loop, before `del slots_busy[ns]` (while `ns` is still valid):
```python
if status == "Succeeded":
    ok(f"[{pair_key}] Succeeded → done")
    entry["status"] = "done"
    entry["namespace"] = None
    store.save(progress)
    try:
        _delete_pipelinerun(pr_name, ns)
    except Exception as exc:
        warn(f"Failed to delete PipelineRun {pr_name!r} in {ns}: {exc}")
    del slots_busy[ns]
```

### Improvements

1. **Stale `collecting` pairs silently stuck on resume** — The new unrecognized-status warning is passive: pairs stay in `collecting` forever, invisible to `_work_remaining`, so the orchestrator may exit claiming "all done" while pairs are orphaned. Consider auto-resetting to `pending` instead of just warning, or at minimum logging how many pairs are orphaned at the end of the run.

2. **No test for the unrecognized status warning** — This is the primary migration safety net for users with stale `collecting` entries. Worth adding a test that loads a progress with `"status": "collecting"`, runs reconciliation, and verifies the warning is emitted.

3. **No test for the new `Succeeded → done` transition on resume** — The old path had 4 dedicated tests via `_reconcile_collecting`. The new simpler path (reconciliation sets `done` + frees namespace for already-Succeeded PipelineRuns) has zero tests.

4. **Stale mock signatures in test_deploy_collect.py** — All `mock_extract` functions still accept `workload=None`, but the production `_extract_phases_from_pvc` no longer has that parameter. Commit message says "remove dead workload parameter" so mocks should match.

5. **README auto-cleanup sentence stale** — Even after fixing the auto-delete regression, line 168 says "when a PipelineRun succeeds and results are collected" — the "and results are collected" part no longer applies since collection is decoupled.

kalantar:
Pushed 5411059 addressing review round 2.

**PipelineRun auto-delete restored** — Added `_delete_pipelinerun(pr_name, ns)` back into the Succeeded branch of the orchestrator loop. This was introduced by PR #129 inside `_do_collect` and lost when that function was deleted. PipelineRun CRs are now cleaned up on success as before.

**Stale status auto-reset** — Changed the unrecognized-status handler from warn-only to auto-reset to `pending` (with warning). Stale `collecting` entries from pre-upgrade progress files are now re-queued instead of silently orphaned.

**Two new tests** — `test_reconcile_succeeded_sets_done_and_frees_namespace` covers the resume reconciliation path (Succeeded → done + namespace freed). `test_reconcile_unrecognized_status_resets_to_pending` covers the migration safety net for stale statuses.

**Mock signatures fixed** — Removed stale `workload=None` from all 10 `mock_extract` functions in test_deploy_collect.py to match the updated `_extract_phases_from_pvc` signature.

**README updated** — Auto-cleanup sentence no longer mentions collection ("when a PipelineRun succeeds and results are collected" → "when a PipelineRun succeeds").

513 tests pass, lint clean.

kalantar:
## Review: PR #130 vs Issue #120 — NEEDS CHANGES

All five requirements from issue #120 are properly addressed. The simplification is clean — net deletion of ~72 lines of production code, the state machine is easier to reason about, and documentation updates are thorough and accurate. Two items should be addressed before merge.

### 1. New reconciliation tests duplicate production logic instead of exercising it

`test_reconcile_succeeded_sets_done_and_frees_namespace` and `test_reconcile_unrecognized_status_resets_to_pending` manually reimplement the reconciliation loop's if/elif chain inside the test body, with hardcoded values. They don't call any function from `deploy.py` — they test a copy of the logic, not the original. If the production code drifts (typo, changed status set, different field assignment), these tests still pass.

Every other behavioral test in the file calls the actual function under test (`_handle_pending_pods`, `_force_reset`, `_cleanup_pair`, etc.). These two should follow the same pattern: either extract the reconciliation-at-startup loop into a named function and test it directly, or invoke `_cmd_run` with sufficient mocking.

Additionally, the old suite covered two properties that are no longer tested:
- `_delete_pipelinerun` is called on Succeeded in the main loop (old: `test_do_collect_deletes_pipelinerun_on_success`)
- `_delete_pipelinerun` failure is non-fatal — pair stays `done` (old: `test_do_collect_returns_true_when_delete_raises`)

Both are straightforward to add as unit tests against the extracted function or via mocking in the existing tests.

### 2. Minor comment inaccuracy in `_cleanup_pair`

Line 741: `# Namespace already freed (done) — search all slots` — the parenthetical "(done)" implies only `done` pairs reach this branch, but `failed`, `timed-out`, and `stalled` pairs also have `namespace=None` at this point. Either drop the parenthetical or enumerate the relevant cases.

### Non-blocking observations

- **Reconciliation doesn't call `_delete_pipelinerun` for Succeeded pairs found on resume.** The main loop does, so this is an asymmetry. PipelineRun CRs from interrupted runs accumulate until `deploy.py cleanup`. This may be acceptable (the old `_reconcile_collecting` path also relied on re-collection to trigger deletion), but worth a deliberate decision.

- **Unrecognized status catch-all resets to `pending` without checking the cluster.** Safe for the `collecting` migration case (PipelineRun already succeeded), but as a general catch-all it could re-dispatch a pair whose PipelineRun is still running if the status were corrupted. A comment documenting why this is safe for the migration case would help.

- **Completions tests assert expected statuses are present but don't assert `collecting`/`collect-failed` are absent.** A simple negative assertion would prevent accidental reintroduction.

- **New tests use `tempfile.TemporaryDirectory` instead of pytest's `tmp_path` fixture** — minor style inconsistency with the rest of the file.

kalantar:
Pushed 498ae66 addressing review round 3.

**Extracted `_reconcile_on_resume`** — the reconciliation loop is now a standalone function that tests call directly, instead of duplicating the if/elif chain inline. Tests now exercise the actual production code path.

**New/rewritten tests:**
- `test_reconcile_succeeded_sets_done_and_frees_namespace` — calls `_reconcile_on_resume` with a mocked `_check_pipelinerun_status` returning Succeeded
- `test_reconcile_unrecognized_status_resets_to_pending` — verifies stale `collecting` entries are reset and warned
- `test_reconcile_running_no_pr_resets_to_pending` — new: covers the edge case where a running pair has no PR metadata
- `test_delete_pipelinerun_called_on_success` — verifies the call happens with correct args
- `test_delete_pipelinerun_failure_is_nonfatal` — verifies an OSError from kubectl doesn't change pair status

**Negative completions assertions** — both bash and zsh tests now assert `collecting` and `collect-failed` are absent, preventing accidental reintroduction.

**Docstring on `_reconcile_on_resume`** explains why the unrecognized-status catch-all is safe: the only historical unrecognized status is `collecting`, which implies the PipelineRun already succeeded.

**`_cleanup_pair` comment** — dropped the misleading "(done)" parenthetical since `failed`, `timed-out`, and `stalled` pairs also reach that branch.

**Skipped** the resume auto-delete asymmetry — matches old behavior (old `_reconcile_collecting` also didn't delete PipelineRuns; deletion only happened during the main loop). PipelineRun CRs from interrupted runs are cleaned up by `deploy.py cleanup`.

516 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

The core simplification is solid — all five issue #120 requirements are met, production logic is correct, and the net deletion of ~70 lines with consistent updates to README, completions, and tests is clean work. Five items to address before merge:

**1. Add per-pair error handling in `_reconcile_on_resume` around `_check_pipelinerun_status`**

If `_check_pipelinerun_status` raises (e.g. kubectl binary missing, transient OS error), the exception aborts the entire reconciliation loop before `store.save(progress)` runs — all in-memory state changes are lost. Wrap the call in try/except per pair with a warn-and-skip, matching the pattern already used for `_handle_pending_pods` in the main loop.

**2. Fix or remove the two `_delete_pipelinerun` tests**

`test_delete_pipelinerun_called_on_success` calls the monkeypatched function directly (tests mock infrastructure, not production code). `test_delete_pipelinerun_failure_is_nonfatal` creates a standalone `entry` dict, calls `_delete_pipelinerun`, catches the error, and asserts the dict is unchanged — but nothing in the test could have changed it. Both tests pass trivially regardless of production behavior.

**3. Auto-delete PipelineRun CR in `_reconcile_on_resume` Succeeded branch**

The live Succeeded path in the main loop deletes the PipelineRun CR, but the resume reconciliation path does not. PipelineRun CRs from runs that completed during an orchestrator crash will accumulate on the cluster until manual `reset`. A best-effort `_delete_pipelinerun` call (same try/except pattern as the main loop) before clearing the namespace would restore parity.

**4. Add tests for `Failed` and `Unknown` branches of `_reconcile_on_resume`**

Currently only `Succeeded`, unrecognized status, and "no metadata" branches are tested. The `Failed` branch intentionally retains the namespace (unlike other branches) for `cleanup` to use — a regression there would break cleanup. The `Unknown` branch (with metadata present, where `_check_pipelinerun_status` returns "Unknown") has a distinct code path and warning message that deserves coverage.

**5. Rebase onto main to pick up `cleanup → reset` rename**

Main (c7036fd) renamed the `cleanup` subcommand to `reset`. The PR branch still references `cleanup` in README and code — merging as-is would silently revert the rename.

kalantar:
Pushed 554b0bc addressing review round 4.

**Error handling in `_reconcile_on_resume`** — `_check_pipelinerun_status` is now wrapped in per-pair try/except. A single kubectl failure warns and skips that pair instead of aborting the loop and losing all in-memory state changes.

**Auto-delete on resume** — `_reconcile_on_resume` now calls `_delete_pipelinerun` in the Succeeded branch (same try/except pattern as the main loop), so PipelineRun CRs from runs that completed during an orchestrator crash are cleaned up on resume.

**Rewrote `_delete_pipelinerun` tests** — replaced the two trivially-passing tests with tests that exercise `_reconcile_on_resume` directly: `test_reconcile_succeeded_deletes_pipelinerun` verifies the delete call happens with correct args, `test_reconcile_succeeded_delete_failure_nonfatal` verifies an OSError doesn't prevent the pair from transitioning to done.

**Added `Failed` and `Unknown` branch tests** — `test_reconcile_failed_sets_failed_retains_namespace` verifies the namespace is retained (needed for reset/cleanup), `test_reconcile_unknown_resets_to_pending` verifies the warning and reset. Also added `test_reconcile_status_check_exception_skips_pair` for the new error handling.

**Rebase** was already done in the previous push — no stale `cleanup` references remain.

519 tests pass, lint clean.

kalantar:
## NEEDS CHANGES (minor)

Clean, well-executed simplification. All issue #120 requirements are met. Net deletion of ~72 lines of production code, state machine reduced from 7 to 5 statuses, no stale references found. Two items to address before merge:

### Required

1. **Expand failure-reason set in `_reconcile_on_resume`** — The main loop handles 7 Tekton failure statuses but `_reconcile_on_resume` only handles `Failed` and `PipelineRunCancelled`. If the orchestrator crashes while a PipelineRun is in `PipelineRunTimeout` or `CreateRunFailed`, reconciliation leaves it as "running" for one extra poll cycle. Fix: align the tuple with the main loop (`PipelineRunCouldntGetPipeline`, `PipelineRunTimeout`, `CreateRunFailed`, `PipelineRunStopping`, `PipelineRunStoppingTimeout`).

2. **Fix `_reconcile_on_resume` docstring** — Says "the only historical unrecognized status is 'collecting'" but `collect-failed` is also a historical status. The safety argument holds for both; the enumeration is just incomplete.

### Suggested (non-blocking)

- Add a test for the Running/Started passthrough case in `_reconcile_on_resume` (entry left unchanged when PipelineRun is still running) to prevent future regressions that could cause double-dispatch.
- Consider adding a brief comment in the Failed branch of `_reconcile_on_resume` explaining why namespace is intentionally retained (differs from the main loop's behavior).

kalantar:
Pushed 9e9d369 addressing round 5.

Aligned the failure-status tuple in `_reconcile_on_resume` with the main loop (added `PipelineRunCouldntGetPipeline`, `PipelineRunTimeout`, `CreateRunFailed`, `PipelineRunStopping`, `PipelineRunStoppingTimeout`). Fixed the docstring to enumerate both historical statuses (`collecting` and `collect-failed`). Added a comment explaining why the Failed branch intentionally retains the namespace. Added `test_reconcile_still_running_left_unchanged` to guard against double-dispatch regressions.

520 tests pass, lint clean.

PR #134, opened 2026-05-14
Add --workload/--only scoping flags to deploy.py collect (#122)

### LINKED ISSUE #122: Add --workload/--only scoping flags to deploy.py collect
## Context

`deploy.py collect` currently only supports `--package` filtering. Other subcommands (`run`, `status`, `reset`) all support `--only`, `--workload`, and `--package` scoping flags.

With inline collection removed (#120), `collect` becomes the sole mechanism for pulling partial results mid-flight. To support fail-fast analysis of individual workloads, it needs the same scoping flags.

## Requirements

- Add `--only PAIR` flag to `collect` (scope to one pair key, wl- prefix optional)
- Add `--workload NAME` flag to `collect` (scope to pairs matching this workload)
- Keep existing `--package` flag (unchanged)
- Scoping should filter which workload subdirectories are pulled, not just which phases

## Filter interaction semantics

Follow the same semantics as `run`/`status`/`reset` (implemented in `_apply_run_filters`):

- `--only` is **exclusive**: resolves to exactly one pair key, ignores other scoping flags
- `--workload` and `--package` are **AND'd**: each narrows the candidate set
- No `--status` flag — `collect` implicitly scopes to done/collecting pairs (warn if scoped pairs have non-done status, but don't add a flag for parity)

## Example usage

```bash
# Pull results for one specific workload across all packages
deploy.py collect --workload sharegpt-32

# Pull results for one specific pair
deploy.py collect --only wl-sharegpt-32-baseline

# Combine workload + package filtering
deploy.py collect --workload sharegpt-32 --package baseline
```

## Notes

Depends on #120 (inline collection removal makes this the primary partial-results path).

kalantar:
Closes #122

## Summary

- Add `--only PAIR` and `--workload NAME` scoping flags to `deploy.py collect`
- When scoped, collection pulls only matching workload subdirectories from the PVC (instead of entire phase directories)
- Existing `--package` flag (phase-level filter) composes orthogonally: `--workload X --package baseline` pulls workload X from baseline only
- Non-done scoped pairs emit a warning and are skipped
- Unscoped behavior is unchanged
- Updated bash and zsh completions

## Design notes

**`--package` semantics differ between collect and other subcommands.** Collect's `--package` is `nargs="+"` (a list of phase names like `baseline`, `treatment`). Other subcommands' `--package` is a single string used as a pair-level filter in `_resolve_scope()`. To avoid mixing these, the scoped path builds a proxy `_ScopeArgs` namespace with `package=None` so `_resolve_scope` only operates on `--only`/`--workload`. Collect's `--package` is then applied separately as a phase-level filter on top.

**No `--status` flag.** Unlike `run`/`status`/`reset`, collect implicitly scopes to done/collecting pairs. Adding `--status` would be confusing since you'd never want to collect from non-done pairs.

## Test plan

- [x] `--workload` scopes extraction to matching workloads only
- [x] `--only` scopes to one pair's workload and package
- [x] `--only` auto-resolves `wl-` prefix
- [x] `--workload` + `--package` compose correctly (AND semantics)
- [x] `--workload` with no match exits with error and helpful message
- [x] Non-done scoped pairs produce a warning, done ones proceed
- [x] Unscoped path unchanged (backward compat, `workload=None`)
- [x] All 11 existing collect tests still pass (no regressions)
- [x] Full suite: 528 tests pass, lint clean

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -137,7 +137,7 @@ Common flags (all subcommands):
 ```bash
 python pipeline/deploy.py run     [flags]   # orchestrate parallel pool execution across namespace slots
 python pipeline/deploy.py status            # show progress snapshot of all (workload, package) pairs
-python pipeline/deploy.py collect [--package NAME…]
+python pipeline/deploy.py collect [flags]     # pull results from the cluster PVC
 python pipeline/deploy.py reset [flags]     # tear down cluster resources for failed/stalled pairs
 python pipeline/deploy.py pairs   [flags]   # list available pair keys, workloads, and packages
 ```
@@ -175,6 +175,15 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--package NAME` | Filter by package name |
 
 **`deploy.py collect`** — extracts results from the cluster PVC and writes to `workspace/runs/<run>/results/{phase}/<workload>/`.
+
+| Flag | Description |
+|------|-------------|
+| `--only PAIR` | Scope to one specific pair key (`wl-` prefix optional) |
+| `--workload NAME` | Scope to pairs matching this workload |
+| `--package NAME…` | Collect only these packages (phase-level filter) |
+| `--skip-logs` | Skip vLLM and EPP log files, collect only traces |
+
+When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). These pair-level flags compose with `--package` as AND: `--workload X --package baseline` pulls workload X from the baseline phase only. Requires `progress.json` to resolve pairs.
 
 **`deploy.py reset`** — removes cluster resources (PipelineRuns, Helm releases) for all non-pending pairs. Failed/running/timed-out pairs are reset to `pending` so they can be re-dispatched. Done pairs stay `done` — only their PipelineRun is deleted to free cluster resources.
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -178,7 +178,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 | Flag | Description |
 |------|-------------|
-| `--only PAIR` | Scope to one specific pair key (`wl-` prefix optional) |
+| `--only PAIR` | Scope to one pair key — narrows both workload and package (`wl-` prefix optional; takes precedence over `--workload`) |
 | `--workload NAME` | Scope to pairs matching this workload |
 | `--package NAME…` | Collect only these packages (phase-level filter) |
 | `--skip-logs` | Skip vLLM and EPP log files, collect only traces |
--- a/pipeline/completions.bash
+++ b/pipeline/completions.bash
@@ -82,7 +82,7 @@ _sim2real_deploy() {
                 COMPREPLY=($(compgen -W "--only --workload --package --status --dry-run" -- "$cur"))
                 ;;
             collect)
-                COMPREPLY=($(compgen -W "--package --skip-logs" -- "$cur"))
+                COMPREPLY=($(compgen -W "--only --workload --package --skip-logs" -- "$cur"))
                 ;;
             pairs)
                 COMPREPLY=($(compgen -W "--keys-only --workloads-only --packages-only" -- "$cur"))
--- a/pipeline/completions.zsh
+++ b/pipeline/completions.zsh
@@ -91,6 +91,8 @@ _sim2real_deploy() {
                     ;;
                 collect)
                     _arguments \
+                        '--only[Scope to one pair key]:pair key:_deploy_py_pair_keys' \
+                        '--workload[Scope to workload]:workload:_deploy_py_workloads' \
                         '*--package[Collect only these packages]:package:_deploy_py_packages' \
                         '--skip-logs[Skip vLLM and EPP log files]'
                     ;;
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -621,6 +621,10 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
     scope_workload = getattr(args, "workload", None)
     scoped = scope_only is not None or scope_workload is not None
 
+    if scoped and not progress:
+        err("--only/--workload require progress.json to resolve pairs, but none was found.")
+        sys.exit(1)
+
     if scoped and progress:
         # Build a lightweight args namespace for _resolve_scope with only
         # pair-level filters (--only, --workload).  Collect's --package is
@@ -688,7 +692,9 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                     skip_logs=skip_logs, workload=wl_name)
             except RuntimeError as e:
                 warn(f"Extractor pod failed for workload {wl_name}: {e}")
-                failed.extend(phases_to_collect)
+                for p in phases_to_collect:
+                    if p not in failed:
+                        failed.append(p)
             else:
                 for phase, exc in errors.items():
                     if exc is None:
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -451,9 +451,9 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
       /data/{runName}/{phase}/{workloadName}/trace_data.csv
 
     When *workload* is set, only that workload's subdirectory is copied for
-    each phase (used by inline collection during parallel pool execution).
+    each phase (used by scoped ``collect --only/--workload``).
     When *workload* is None (default), the entire phase directory is copied
-    (used by `deploy.py collect` for bulk collection).
+    (used by unscoped ``deploy.py collect`` for bulk collection).
 
     When *skip_logs* is True, only trace files are copied (skipping vLLM and
     EPP log files which typically account for the bulk of the data).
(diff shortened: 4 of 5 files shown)


kalantar:
## Review: PR #134 — `--workload`/`--only` scoping flags for `deploy.py collect`

**Verdict: NEEDS CHANGES** (two items before merge, plus a few suggestions)

All issue #122 requirements are addressed. The `_ScopeArgs` proxy design is sound, tests are well-structured, and backward compatibility is preserved. Shell completions are correct. Two issues should be fixed before merge:

### Must fix

1. **Scoped flags silently ignored when `progress.json` is absent.** When the user passes `--only` or `--workload` but `progress.json` is missing or corrupt, `progress` is `None` and the `if scoped and progress:` gate falls through to the unscoped path — collecting everything without any warning that the scoping flags were ignored. Three independent reviewers flagged this. Fix: error out early when `scoped` is True but `progress` is falsy, since scoping requires progress data to resolve pairs.

2. **`pipeline/README.md` not updated.** The project's contributing rules require updating the README when CLI flags change. The `collect` subcommand section needs the new `--only` and `--workload` flags documented (synopsis + flag table).

### Should fix

3. **`RuntimeError` catch in the scoped workload loop doesn't deduplicate `failed`.** `failed.extend(phases_to_collect)` appends duplicates if multiple workloads fail, unlike the `else` branch which checks `if phase not in failed`. Minor, but produces a confusing summary line.

4. **Final summary doesn't mention skipped non-done pairs.** When a scoped pair is warned-and-skipped due to non-done status, the summary still reports `Collected: N/N phases` as if collection is complete. A `Skipped: X pair(s) not done` line would prevent the user from thinking collection finished fully.

5. **PR body says "8 new tests" but there are 7.** Minor inaccuracy in the description.

### Nice to have

- The `_ScopeArgs` comment should mention why `status=None` is set (collect implicitly restricts to done/collecting — no `--status` flag by design).
- The `--package` validation/expansion logic (~17 lines) is duplicated verbatim between the scoped and unscoped branches. Extracting a small helper would reduce maintenance risk.
- Test gap: no test for `--only` combined with `--package` phase filter, or for the all-scoped-pairs-non-done early return path.

kalantar:
Pushed fixes in 879a9e9:

**Scoped flags without progress.json** — `_cmd_collect` now errors out early when `--only` or `--workload` is given but `progress.json` is absent or corrupt. Previously it silently fell through to the unscoped path, collecting everything. Added a test for this path.

**`failed` list dedup** — Changed `failed.extend(phases_to_collect)` to a deduplicated append loop, matching the per-phase error branch below it.

**README updated** — Added a flag table and scoping description to the `deploy.py collect` section in `pipeline/README.md`, and updated the synopsis line.

Not addressed in this commit:
- Summary line for skipped non-done pairs (item 4) — the per-pair warn is adequate for now; could be a follow-up if the terse summary causes confusion in practice.
- `--package` validation dedup between scoped/unscoped branches — agreed it's worth extracting but prefer to do that as a separate cleanup to keep this PR focused on the scoping feature.

kalantar:
## Review: NEEDS CHANGES

All six requirements from issue #122 are addressed, tests pass (19/19), and lint is clean. The implementation is functionally correct for all tested paths. A few issues should be addressed before merge.

### Issues to fix

**1. Early return bypasses the summary block.** When all scoped pairs are non-done, `_cmd_collect` warns and returns early without printing the standard `Collected: X/Y phases` summary. This makes the output inconsistent with every other exit path and could confuse automation that parses stdout. Initialize `collected`, `failed`, `phases_to_collect` before the early return and fall through to the summary, or at minimum print the summary before returning.

**2. Warn loop and filter logic are decoupled.** The loop that prints "skipping" for non-done pairs performs no actual filtering — the real exclusion happens independently in the `scoped_phases`/`scoped_workloads` set comprehensions. If either filter drifts in the future, warnings will contradict behavior. Consider filtering `in_scope` down to a `collectible` set first, then deriving phases/workloads from that single source of truth.

**3. `_extract_phases_from_pvc` docstring is stale.** It says the `workload` param is "used by inline collection during parallel pool execution," but this PR also uses it from the scoped `collect` path. Update to reflect both callers.

**4. Unscoped path comment claims "original behavior" but the status filter changed.** The comment `# ── Unscoped path (original behavior)` is misleading if the branch added `"collecting"` to the status filter (vs just `"done"` on main). Change the marker to something like `# ── Unscoped path (no --only/--workload)`.

### Test gaps worth adding

- **RuntimeError in the scoped extraction loop.** Two workloads in scope, first succeeds, second raises `RuntimeError` — verify partial results are preserved and `failed` is populated correctly.
- **Per-phase extraction failure in scoped path.** Return a non-None error for one phase in the `errors` dict to exercise the `warn`/`failed` branch.
- **All scoped pairs non-done (early return).** `--workload smoke` where all smoke pairs are `running` — verify no extraction is attempted.

- The `--package` validation/expansion/dedup block is duplicated verbatim between scoped and unscoped paths (~120 lines total duplication in `_cmd_collect`). Extracting a shared helper would reduce maintenance burden. Fine as a follow-up.
- README says "these pair-level flags compose with `--package` as AND" which could imply `--only` and `--workload` also compose with each other. Worth adding: "`--only` and `--workload` are alternatives; if both are given, `--only` takes precedence."
- `argparse` doesn't enforce mutual exclusivity of `--only`/`--workload` — consider a `mutually_exclusive_group` or at least a documented precedence rule.

kalantar:
Pushed fixes in b5fbbdd:

**Early return removed** — when all scoped pairs are non-done, the function now falls through to the summary block and prints `Collected: 0/0 phases` instead of returning silently. This keeps output consistent across all exit paths.

**Warn/filter logic unified** — the warn loop and the phase/workload derivation now both operate on a single `collectible` set (pairs filtered to done/collecting status). Previously these were independent duplicate filters that could drift.

**Docstring updated** — `_extract_phases_from_pvc` docstring now reflects the scoped collect caller in addition to the original inline collection use.

**Comment fix** — changed "original behavior" marker on the unscoped path to "no --only/--workload" since the status filter also includes `collecting` (intentional but not "original").

**Three new tests** — all-scoped-pairs-non-done (no extraction attempted), RuntimeError in scoped extraction loop, and per-phase failure in scoped path. 22 collect tests total, 532 full suite.

Not addressed (non-blocking suggestions):
- `--package` validation dedup between branches — same as round 1, prefer as separate cleanup
- README note about `--only`/`--workload` precedence — `_apply_run_filters` already handles this (--only is exclusive, ignores --workload), but worth adding to docs in a follow-up
- `argparse` mutual exclusivity for `--only`/`--workload` — the existing subcommands don't enforce this either, so adding it just for collect would be inconsistent

kalantar:
## Review: PR #134 — Add --workload/--only scoping flags to deploy.py collect

**Verdict: READY TO MERGE** with minor suggestions below (none blocking).

### What looks good

The core scoping implementation is correct and well-tested. All issue #122 requirements are met:

- `--only PAIR` works with wl- prefix auto-resolution via `_resolve_scope`
- `--workload NAME` correctly scopes extraction to matching workload subdirectories
- `--package` composes orthogonally (AND semantics) with the new pair-level flags
- Non-done scoped pairs warn and skip; missing `progress.json` exits with a clear error
- Unscoped path is preserved (confirmed by `test_collect_unscoped_unchanged`)
- Shell completions and README updated consistently
- 11 new tests with good coverage of happy paths, error exits, and edge cases

The `_ScopeArgs` proxy pattern correctly separates collect's `--package` (phase-level, nargs="+") from `_resolve_scope`'s `package` (pair-level, single string). This is the central design challenge and it's handled well.

1. **Phase-resolution duplication.** The `args.package` validation + dedup logic (experiment expansion, unknown-package check, seen-set dedup) is duplicated between the scoped and unscoped branches (~20 lines each). Extracting a small helper like `_resolve_phases_to_collect(known_phases, package_filter)` would reduce this.

2. **`_extract_phases_from_pvc` docstring is now incomplete.** The updated docstring says the `workload` parameter is "used by scoped `collect --only/--workload`", but `deploy.py run` also passes `workload` during orchestration. Mentioning both callers would prevent confusion.

3. **README `--only` description could be more precise.** Currently says "Scope to one specific pair key" — since `--only` constrains both workload and package dimensions (not just the pair key), something like "Scope to one pair key (narrows both workload and package)" would be clearer.

4. **Minor crash path on corrupted progress.json.** In the scoped warning loop, if a progress entry is not a dict, the `isinstance` check correctly excludes it from `collectible`, but the subsequent `progress[key].get('status', '')` in the warning message would raise `AttributeError`. Very unlikely in practice (requires malformed progress.json), but a guard would be cheap.

5. **Test gap: `--only` + `--workload` together.** Argparse allows both flags simultaneously. There's no test documenting what happens when both are supplied. Worth adding one to lock in the expected behavior.

kalantar:
Pushed a5f40ec addressing the round-3 suggestions:

- `_extract_phases_from_pvc` docstring now mentions both callers (scoped collect and orchestrator inline collection).
- Guarded the warning loop against non-dict progress entries — `isinstance` check before calling `.get()`.
- README `--only` description clarified: narrows both workload and package dimensions, takes precedence over `--workload`.
- Added `test_collect_only_takes_precedence_over_workload` — documents that `--only` wins when both flags are supplied.

The `--package` validation dedup (item 1) is the one remaining suggestion — keeping that for a separate cleanup PR. 23 collect tests, 533 full suite.

kalantar:
## Review: NEEDS CHANGES (minor)

All five requirements from issue #122 are fully implemented and well-tested (533 tests pass, lint clean). The changes are solid overall — two items should be addressed before merge.

### Must fix

1. **Stale docstring after merge** — `_extract_phases_from_pvc` (deploy.py ~455) says the `workload` parameter is "used by scoped `collect --only/--workload` and by the orchestrator during inline collection." PR #130 already removed inline collection from main. After merge this reference will be wrong. Remove the "and by the orchestrator during inline collection" clause.

2. **`collected`/`failed` list overlap in scoped path** — The per-workload loop can place the same phase in both `collected` and `failed` (e.g., if phase "baseline" succeeds for workload A but fails for workload B). The summary would then say both "Collected: baseline" and "Failed: baseline". In practice this is currently unreachable (scoped workloads always resolves to exactly one entry), but the code structure implies multi-workload iteration is intended. Either reconcile the lists after the loop (`collected = [p for p in collected if p not in failed]`) or track results at `(phase, workload)` granularity.

### Worth considering

- **Code duplication** — The `--package` validation/dedup/expansion logic is copy-pasted between scoped and unscoped paths. A shared helper like `_resolve_package_filter(known_phases, requested)` would eliminate the duplication and reduce maintenance risk.
- **`_ScopeArgs` style** — `types.SimpleNamespace(only=..., workload=..., package=None, status=None)` would be more idiomatic than the inner class.
- **Scoped `collecting` status untested** — The scoped filter includes `"collecting"` alongside `"done"`, but no test exercises a pair in `collecting` status through the scoped path. A single test would guard against accidental removal.
- **No exit code on failure** — `_cmd_collect` exits 0 even when all phases fail. Scripts automating `collect --workload X` can't distinguish success from failure by exit code. (Pre-existing, but scoped collection makes automation more likely.)

### What looks good

- Scoping flags correctly use `_resolve_scope` with a proxy namespace that isolates collect's list-valued `--package` from the pair-level filter
- Backward compatibility preserved — old tests pass via `getattr` defaults routing to the unscoped path
- 12 new tests cover the key behaviors: workload scope, only scope, prefix resolution, composition with `--package`, no-match errors, non-done warnings, missing progress guard, all-nondone path, RuntimeError handling, per-phase failure, and `--only` precedence
- README, bash, and zsh completions all updated consistently
- PR description's design notes on `--package` semantic differences are helpful context

kalantar:
Pushed 7e209b0 with the two must-fix items:

**Stale docstring** — removed the inline-collection reference from `_extract_phases_from_pvc`. That caller was removed in #130; after merge only the scoped collect path uses the `workload` parameter.

**collected/failed overlap** — added `collected = [p for p in collected if p not in failed]` after the multi-workload loop. A phase that fails for any workload is now excluded from `collected` even if it succeeded for another. Currently unreachable (scoped workloads resolves to one entry), but the reconciliation is correct for the general case the loop structure implies.

Skipped items:
- `_ScopeArgs` → `SimpleNamespace` — both work identically, keeping the inner class for consistency with the rest of the file
- `collecting` status test gap — low risk (set membership check), not adding another test for this
- Exit code on failure — pre-existing behavior, out of scope

The `--package` dedup has been flagged across all four rounds. Happy to file a follow-up issue for it if you'd like.

PR #135, opened 2026-05-14
Write orchestrator progress to ConfigMap for remote status

### LINKED ISSUE #128: Write orchestrator progress to ConfigMap for remote status
## Context

`deploy.py status` reads `progress.json` from the local filesystem. When the orchestrator runs remotely (in-cluster Job, #127), progress.json lives on the PVC — not directly readable without spinning up an extractor pod.

A ConfigMap is the simplest mechanism for exposing progress: the orchestrator writes it each poll cycle, and `status` reads it via a single `kubectl get configmap` command.

### Orchestrator writes progress to ConfigMap

- ConfigMap name: `sim2real-progress` (in primary namespace)
- Updated each poll cycle (same frequency as progress.json saves today)
- Contains the full progress dict as a single data key (JSON string)
- Written in both local and remote modes (consistent behavior)

### `deploy.py status` reads from ConfigMap

- In remote mode (or when local progress.json is absent/stale): read from ConfigMap
- Detection: if progress.json doesn't exist locally OR `--remote` flag, read ConfigMap
- Fallback: if ConfigMap doesn't exist either, report "no active run"

### Consistency

Both local and remote modes write the ConfigMap. This means `status` can always check the ConfigMap regardless of where the orchestrator is running — one code path.

## Example

```bash
# Works whether orchestrator is local or remote
deploy.py status
deploy.py status --workload sharegpt-32
```

## Dependencies

Prerequisite for #127 (remote execution).

## Notes

ConfigMap size limit is 1MB. Progress.json for even large runs (50+ pairs) is well under this. If it ever exceeds, we can split or compress, but this is unlikely.

kalantar:
Closes #128

## Summary

- Add `ConfigMapProgressStore` class to `pipeline/lib/progress.py` — reads/writes progress as a Kubernetes ConfigMap (`sim2real-progress`) via kubectl
- Add `CompositeProgressStore` — wraps primary (local file) and secondary (ConfigMap) stores; saves to both, loads from primary with fallback
- Wire up `_cmd_run` and `_cmd_reset` to use `CompositeProgressStore`, so the ConfigMap is updated every poll cycle alongside `progress.json`
- Add `--remote` flag to `deploy.py status` — reads from ConfigMap instead of local file; also falls back to ConfigMap automatically when local `progress.json` is absent

## Design choices

- **ConfigMap vs local file for status reads:** Status doesn't use `CompositeProgressStore` — it picks one source. `--remote` or missing local file → ConfigMap; otherwise → local file. This avoids unnecessary kubectl calls when the local file is available.
- **Secondary save failures are best-effort:** `CompositeProgressStore` warns on stderr but doesn't fail the orchestrator if the ConfigMap write fails. The local file remains authoritative.
- **Single ConfigMap per namespace:** `sim2real-progress` in the primary namespace. One active run per namespace, so no per-run partitioning needed.

## Test plan

- [x] ConfigMapProgressStore: load (missing, valid, corrupt, empty), save (success, failure) — 6 tests
- [x] CompositeProgressStore: save fan-out, load priority/fallback, error handling — 6 tests
- [x] Status `--remote` flag: parser, ConfigMap read, fallback on missing local, no-data case — 5 tests
- [x] All 537 existing tests pass, lint clean

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -167,10 +167,11 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 **Auto-cleanup** — when a PipelineRun succeeds, the orchestrator deletes the PipelineRun CR from the cluster. Failed PipelineRuns are left in place for debugging (`kubectl describe`, pod logs). Use `reset` to remove them when done.
 
-**`deploy.py status`** — prints the current state of all pairs from `workspace/runs/<run>/progress.json`. When the orchestrator is in backoff, an additional line shows the current state and backoff level.
+**`deploy.py status`** — prints the current state of all pairs. By default reads from `workspace/runs/<run>/progress.json`. When the orchestrator runs remotely (in-cluster), use `--remote` to read from the `sim2real-progress` ConfigMap instead. The ConfigMap is also used automatically when the local `progress.json` is absent.
 
 | Flag | Description |
 |------|-------------|
+| `--remote` | Read progress from cluster ConfigMap instead of local file |
 | `--workload NAME` | Filter by workload name |
 | `--package NAME` | Filter by package name |
 
@@ -311,11 +312,14 @@ All paths are relative to the repo root and validated at Phase 1.
 
 ## Parallel Pool Execution
 
-`setup.py --namespaces NS1,NS2,...` provisions N namespace slots, each bootstrapped identically. `prepare.py` generates one shared Tekton Pipeline plus one PipelineRun per `(workload, package)` pair. `deploy.py run` orchestrates execution by assigning pairs to free slots, polling for completion, and retrying on timeout. Use `deploy.py collect` to pull results off-cluster. `deploy.py status` reads `workspace/runs/<run>/progress.json` and prints the current state of every pair.
+`setup.py --namespaces NS1,NS2,...` provisions N namespace slots, each bootstrapped identically. `prepare.py` generates one shared Tekton Pipeline plus one PipelineRun per `(workload, package)` pair. `deploy.py run` orchestrates execution by assigning pairs to free slots, polling for completion, and retrying on timeout. Use `deploy.py collect` to pull results off-cluster. `deploy.py status` reads progress from the local file or cluster ConfigMap.
 
 | Artifact | Written by | Read by |
 |----------|-----------|---------|
 | `runs/<run>/progress.json` | `deploy.py run` | `deploy.py status` |
+| ConfigMap `sim2real-progress` | `deploy.py run`, `deploy.py reset` | `deploy.py status --remote` |
+
+The orchestrator writes progress to both the local file and the `sim2real-progress` ConfigMap in the primary namespace on every poll cycle. This allows `deploy.py status --remote` to read progress directly from the cluster, which is required when the orchestrator runs as an in-cluster Job.
 
 ---
 
--- a/pipeline/completions.bash
+++ b/pipeline/completions.bash
@@ -76,7 +76,7 @@ _sim2real_deploy() {
                 COMPREPLY=($(compgen -W "--only --workload --package --status --force --skip-build-epp --max-retries --poll-interval --gpu-resource-type --default-gpu-cost --pending-threshold --max-pending-stalls" -- "$cur"))
                 ;;
             status)
-                COMPREPLY=($(compgen -W "--only --workload --package --status" -- "$cur"))
+                COMPREPLY=($(compgen -W "--only --workload --package --status --remote" -- "$cur"))
                 ;;
             reset)
                 COMPREPLY=($(compgen -W "--only --workload --package --status --dry-run" -- "$cur"))
--- a/pipeline/completions.zsh
+++ b/pipeline/completions.zsh
@@ -87,7 +87,8 @@ _sim2real_deploy() {
                         '--only[Scope to one pair key]:pair key:_deploy_py_pair_keys' \
                         '--workload[Scope to workload]:workload:_deploy_py_workloads' \
                         '--package[Scope to package]:package:_deploy_py_packages' \
-                        '--status[Scope to status]:status:_deploy_py_statuses'
+                        '--status[Scope to status]:status:_deploy_py_statuses' \
+                        '--remote[Read from cluster ConfigMap]'
                     ;;
                 collect)
                     _arguments \
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -68,13 +68,17 @@ def _configmap_namespace(setup_config: dict | None,
                          namespaces: list[str] | None = None) -> str:
     """Return the namespace for the sim2real-progress ConfigMap.
 
-    Checks setup_config["namespace"] first, then falls back to namespaces[0].
+    Checks setup_config["namespace"] first, then falls back to
+    namespaces[0] (or setup_config["namespaces"][0] if namespaces
+    is not passed).
     """
-    ns = (setup_config or {}).get("namespace", "")
+    cfg = setup_config or {}
+    ns = cfg.get("namespace", "")
     if ns:
         return ns
-    if namespaces:
-        return namespaces[0]
+    ns_list = namespaces or cfg.get("namespaces") or []
+    if ns_list:
+        return ns_list[0]
     return ""
 
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -62,6 +62,22 @@ def run(cmd: list[str], *, check: bool = True, capture: bool = False,
     return subprocess.run(cmd, check=check, text=True, capture_output=capture, cwd=cwd)
 
 
+# ── ConfigMap namespace resolution ──────────────────────────────────────────
+
+def _configmap_namespace(setup_config: dict | None,
+                         namespaces: list[str] | None = None) -> str:
+    """Return the namespace for the sim2real-progress ConfigMap.
+
+    Checks setup_config["namespace"] first, then falls back to namespaces[0].
+    """
+    ns = (setup_config or {}).get("namespace", "")
+    if ns:
+        return ns
+    if namespaces:
+        return namespaces[0]
+    return ""
+
+
 # ── Phase discovery ─────────────────────────────────────────────────────────
 
 def _discover_phases(cluster_dir: "Path") -> list[str]:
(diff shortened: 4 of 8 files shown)


kalantar:
## Review: NEEDS CHANGES

PR satisfies all three issue #128 requirements — ConfigMap write on each poll cycle, `--remote` flag for status, and consistent write behavior across local/remote modes. The store abstractions (`ConfigMapProgressStore`, `CompositeProgressStore`) are clean and well-tested at the unit level. 17 new tests cover the core paths.

Three items need to be addressed before merge:

### 1. `--remote` silently falls back to local file when no namespace is configured

When a user passes `--remote` but `setup_config` has no namespace (missing `setup_config.json` or empty namespace key), `_cmd_status` silently reads from the local file instead. The user explicitly asked for cluster data and gets local data with no warning. This should either error out or at least warn — silent fallback here can lead to decisions based on stale data.

(`deploy.py`, lines ~212-222)

### 2. `pipeline/README.md` not updated

The project's contributing rules in CLAUDE.md require updating `pipeline/README.md` when CLI flags or artifact schemas change. This PR adds `--remote` to `deploy.py status` and introduces the `sim2real-progress` ConfigMap, but neither is documented in the README. Specifically:
- Add `--remote` to the status flags table
- Note ConfigMap fallback behavior in the status description
- Add `sim2real-progress` ConfigMap to the artifacts table

### 3. Namespace resolution inconsistency between `_cmd_run` and `_cmd_reset`

`_cmd_run` derives the namespace as `setup_config.get("namespace") or (namespaces[0] if namespaces else "")` — it falls back to the first namespace. `_cmd_reset` uses `(setup_config or {}).get("namespace", "")` — no fallback to `namespaces[0]`. This means there's a configuration where `run` writes to both local + ConfigMap but `reset` only writes to local, leaving stale data in the ConfigMap after a reset.

Extract a shared helper for namespace resolution to keep them in sync.

### Additional observations (non-blocking)

- **Secondary load failures are silently swallowed** in `CompositeProgressStore.load()`. The save path logs a `[WARN]` on secondary failure, but the load path catches and continues with no logging at all. Adding a matching `[WARN]` would make debugging easier.
- **All non-zero kubectl exits treated as "not found"** in `ConfigMapProgressStore.load()`. Cluster-unreachable, RBAC errors, and missing-namespace errors all return `{}` silently. Distinguishing "NotFound" from other errors would prevent misdiagnosis.
- **No integration-level test coverage** for the `CompositeProgressStore` wiring in `_cmd_run` or `_cmd_reset`. The store classes are well-tested in isolation, but no test verifies that `_cmd_reset` actually writes to the ConfigMap when a namespace is configured.
- **CompositeProgressStore docstring** documents secondary save failure behavior but not secondary load failure behavior (silently skipped vs. warned).

kalantar:
Addressed the three blocking items and two of the non-blocking observations:

**`--remote` silent fallback** — `_cmd_status` now warns to stderr when `--remote` is requested but no namespace is configured, before falling back to the local file. Added a test for this path.

**README update** — Added `--remote` flag to the status flags table, documented the ConfigMap fallback behavior, and added the `sim2real-progress` ConfigMap to the parallel pool execution artifacts table.

**Namespace resolution inconsistency** — Extracted a `_configmap_namespace(setup_config, namespaces)` helper that checks `setup_config["namespace"]` first, then falls back to `namespaces[0]`. All three call sites (`_cmd_run`, `_cmd_reset`, `_cmd_status`) now use it. Added three tests for the helper.

**Secondary load failure logging** — `CompositeProgressStore.load()` now prints a `[WARN]` on secondary store load failures (matching the save path behavior). Updated the docstring and the existing test.

**kubectl error handling** (non-blocking observation about all non-zero exits treated as "not found") — opened #137 as a follow-up since it requires stderr parsing and is out of scope for this PR.

541 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

All issue #128 requirements are addressed — ConfigMap persistence, `--remote` flag, fallback behavior, and consistency between local/remote modes all work as described. The `ProgressStore` ABC hierarchy is clean, `CompositeProgressStore` is well-designed, and test coverage is solid for the new classes (17 new tests). A few issues should be fixed before merge.

### Issues to address

**1. `_cmd_status` doesn't use `_configmap_namespace` — namespace resolution inconsistency**

`_cmd_run` and `_cmd_reset` both call `_configmap_namespace(setup_config, namespaces)`, which falls back to `namespaces[0]` when `setup_config["namespace"]` is empty. But `_cmd_status` inlines its own lookup (`_configmap_namespace(setup_config)` without the namespaces fallback). In setups that use multiple namespace slots without a singular `namespace` key, `_cmd_run` writes the ConfigMap to `namespaces[0]` but `_cmd_status --remote` can't find it. Fix: pass the namespaces list into `_cmd_status` and use the same `_configmap_namespace` helper.

**2. `ConfigMapProgressStore.load()` treats all kubectl failures as "not found"**

Any non-zero returncode from `kubectl get configmap` returns `{}` — including auth failures, network errors, and wrong-namespace typos. The stderr output (which contains the actual error) is discarded. With `--remote`, this means a cluster connectivity issue silently produces "0 pairs" output, indistinguishable from "no run in progress." Suggestion: check stderr for "NotFound" and return `{}` only in that case; raise on everything else.

**3. `CompositeProgressStore.load()` swallows secondary exceptions without logging**

The `save()` path correctly warns on stderr when a secondary store fails, but `load()` catches `Exception` and silently continues — no log, no warning. The class docstring says "Secondary store failures (both save and load) print a warning to stderr" which doesn't match the load behavior. Fix: add the same `print(f"[WARN] ...", file=sys.stderr)` pattern that `save()` uses, and update the docstring if the behavior is intentionally different.

**4. Shell completions missing `--remote`**

`pipeline/completions.bash` and `pipeline/completions.zsh` list the status subcommand flags but don't include `--remote`.

**5. `_cmd_reset` and `_cmd_run` ConfigMap wiring has no test coverage**

Both functions were modified to create `CompositeProgressStore` when a namespace is available, but no test verifies this wiring. A regression that reverts to local-only in either function would go undetected. The `_cmd_status` remote/fallback paths are well-tested by contrast. Even a targeted test that verifies the store construction would close this gap.

### Non-blocking observations

- The hardcoded `"ConfigMap save failed"` message in `CompositeProgressStore.save()` couples the generic composite to a specific backend — `type(store).__name__` would be more future-proof.
- `ConfigMapProgressStore.__init__` accepts an empty string namespace without complaint. A guard would prevent confusing kubectl errors downstream.
- Consider testing that `--remote` with an existing local file still reads from ConfigMap (the current test uses a non-existent local path, so a condition change from `or` to `and` wouldn't be caught).

kalantar:
Rebased on main (picked up #134) and addressed all items from the second review.

**Namespace resolution in `_cmd_status`** — The `_configmap_namespace` helper now derives namespaces from `setup_config["namespaces"]` when the explicit `namespaces` arg isn't passed. This means `_cmd_status` (which only has `setup_config`) gets the same fallback behavior as `_cmd_run` and `_cmd_reset`. Added a test for this path.

**kubectl error handling** — `ConfigMapProgressStore.load()` now checks stderr for "NotFound" and only returns `{}` in that case. Auth failures, network errors, and other non-NotFound errors raise `RuntimeError`. Also added a guard rejecting empty namespace at construction time. Closed #137 since this is now fixed here.

**Shell completions** — Added `--remote` to both `completions.bash` and `completions.zsh`.

**Composite store wiring tests** — Added tests that verify `_cmd_run` and `_cmd_reset` actually create `CompositeProgressStore` with a `ConfigMapProgressStore` secondary when a namespace is configured.

**`--remote` with existing local file** — Added a test that writes different data to local file vs ConfigMap mock and verifies `--remote` reads from ConfigMap, not local.

**Save warning message** — Changed from hardcoded `"ConfigMap save failed"` to `f"{type(store).__name__} save failed"`.

**CompositeProgressStore.load() warning** was already fixed in the previous round — both load and save warn on secondary failures, and the docstring documents this.

559 tests pass, lint clean.

kalantar:
## Review: PR #135 — Write orchestrator progress to ConfigMap for remote status

**Verdict: NEEDS CHANGES**

All three issue #128 requirements are satisfied (orchestrator writes ConfigMap, status reads it with `--remote`, consistent behavior). Design is clean — `ConfigMapProgressStore`, `CompositeProgressStore`, `_configmap_namespace` helper, and the `--remote` flag are well-structured. Test coverage is solid with 17 new tests. Two issues should be addressed before merge.

### Changes requested

1. **Tighten NotFound detection in `ConfigMapProgressStore.load()`** — The current `"notfound" in stderr or "not found" in stderr` substring match is too broad. An error like `"x509: certificate not found in system roots"` would be silently treated as "ConfigMap doesn't exist" instead of raising. Use `"(NotFound)"` instead, which matches the structured Kubernetes API reason code and is already the pattern used by `_cmd_stop` in the same file. Add a test with a TLS/cert error to guard against this.

2. **`--remote` with no namespace should error, not warn-and-fallback** — When a user explicitly requests `--remote` but no namespace is configured, silently falling back to local data risks showing stale results. This should be a hard error (`sys.exit(1)`) since the user made an explicit request that can't be fulfilled, rather than a warning that's easily lost (especially when stdout is piped).

- **`_cmd_reset`: pass `namespaces` to `_configmap_namespace`** — Currently calls `_configmap_namespace(setup_config)` without the `namespaces` arg. Works today because the function also checks `setup_config["namespaces"]`, but passing `namespaces` explicitly would be consistent with `_cmd_run` and protect against future divergence.
- **README dropped backoff status line docs** — The old README mentioned "When the orchestrator is in backoff, an additional line shows the current state and backoff level." This behavior still exists in `_cmd_status` but is no longer documented after the rewrite.
- **Inconsistent warning format in `CompositeProgressStore`** — `save()` includes the store type name (`{type(store).__name__} save failed`) but `load()` uses generic `"Secondary store load failed"`. Making load consistent would help identify which store failed when multiple secondaries exist.
- **README artifact table** — `deploy.py reset` reads from ConfigMap (via `CompositeProgressStore.load()`) but the table only lists it as a writer.

kalantar:
**Correction on item 2:** The `--remote` with no namespace scenario is unreachable in practice — `setup.py` requires `--namespace`/`--namespaces` and writes them to `setup_config.json`, which is always loaded before `status` runs. The existing warn-and-fallback is fine as defensive code. Withdrawing that request.

That leaves **one change requested**: tighten NotFound detection from `"not found"` substring to `"(NotFound)"` (the K8s reason code), consistent with `_cmd_stop`.

kalantar:
Tightened NotFound detection from broad `"not found"` substring to the K8s reason code `"(NotFound)"`. Added a test confirming that a generic "not found" string in a non-K8s error message is not silently swallowed. 560 tests pass.

Issue #127, opened 2026-05-14
Remote execution: deploy.py run --remote submits orchestrator as K8s Job


kalantar:
## Context

The orchestrator (`deploy.py run`) is a long-running polling loop (up to 4 hours) that requires a stable kubectl connection but no human interaction once started. Running it locally ties the user's laptop to the process. Moving it to an in-cluster Job frees the user to close their laptop, commute, etc.

### New flag: `--remote`

`deploy.py run --remote [existing flags]` submits the orchestrator as a Kubernetes Job instead of running in-process.

### Job specification

- **Name**: `sim2real-orchestrator` (fixed — Kubernetes enforces single instance)
- **Namespace**: primary namespace from `setup_config.json`
- **ServiceAccount**: `sim2real-runner` (already exists in `tektonc-data-collection/tekton/rbac/sim2real-runner.yaml`)
- **Image**: orchestrator image from #139 (pipeline code, PyYAML, and kubectl baked in at build time)

### RBAC updates

The existing `sim2real-pipeline-runner` Role needs two additions for the orchestrator:

1. Add `delete` and `patch` to PipelineRun verbs (orchestrator deletes successful PipelineRuns per #121)
2. Add ConfigMap `create`/`get`/`update`/`patch` (for progress writes via `ConfigMapProgressStore`)

These are small YAML diffs to `tektonc-data-collection/tekton/rbac/sim2real-runner.yaml`, not a separate issue.

### Run inputs via ConfigMap

All run-specific inputs are small YAML/JSON files that fit in a ConfigMap — no PVC needed:

| File | Size |
|------|------|
| `setup_config.json` | ~500 bytes |
| `run_metadata.json` | ~200 bytes |
| `cluster/baseline.yaml` | ~2KB |
| `cluster/pipelinerun-*.yaml` | ~2KB each |

`run --remote` packs these into a ConfigMap and mounts it at a path that mirrors the workspace layout (`/workspace/runs/<run-name>/`), so the existing orchestrator code sees the same directory structure it expects locally. No cleanup needed — the ConfigMap is small and gets overwritten on the next `run --remote`.

### EPP build handling

The launcher builds the EPP image locally before submitting the Job. The in-cluster command runs with `--skip-build-epp` since the image is already pushed to the registry. The Job container does not need Docker access.

### Run identity via CLI flags

[...]

PR #136, opened 2026-05-14
Add deploy.py stop subcommand for remote orchestrator

### LINKED ISSUE #125: Add deploy.py stop subcommand for remote orchestrator
## Context

When the orchestrator runs as an in-cluster Kubernetes Job (`run --remote`), users need a way to stop it. Locally, Ctrl-C suffices. Remotely, the equivalent is deleting the Job.

## Requirements

- New subcommand: `deploy.py stop`
- Deletes the `sim2real-orchestrator` Job in the primary namespace (cascading pod deletion)
- Reports what was stopped (or "no active orchestrator" if Job doesn't exist)
- Does NOT reset pair state — progress.json is left as-is. A subsequent `run` will reconcile on startup.
- If the user wants to reset pairs too, they use `reset` separately.

## Example usage

```bash
# Stop the remote orchestrator
deploy.py stop

# Stop and reset failed pairs for retry
deploy.py stop && deploy.py reset
```

## Notes

Only meaningful for remote execution (#127). If no Job exists, this is a no-op with a message.

kalantar:
Closes #125

## Summary

- Adds `deploy.py stop` subcommand that deletes the `sim2real-orchestrator` Kubernetes Job in the primary namespace (`namespaces[0]`) with `--cascade=foreground`
- If no Job exists, prints "No remote orchestrator started in {namespace}" and returns — wording avoids implying there's no orchestrator at all (a local one might be running)
- Does NOT touch `progress.json` — pair state is left as-is per the issue spec

## Reviewer notes

- Namespace resolution uses `namespaces[0]` from `setup_config.json`, same pattern as `_cmd_run` (line 1061 in the original) which uses `namespaces[0]` for the EPP image build
- The `stop` subcommand is only meaningful once #127 (remote execution) lands — until then the Job won't exist, and `stop` will always be a no-op with a message
- No new CLI flags on `stop` — it reads namespace from `setup_config.json` like every other subcommand

## Test plan

- [x] `test_stop_deletes_orchestrator_job` — verifies kubectl get + delete sequence with correct namespace and `--cascade=foreground`
- [x] `test_stop_no_job_prints_message` — verifies no-op message when Job doesn't exist
- [x] `test_main_dispatches_stop` — verifies main() routes to `_cmd_stop` with primary namespace
- [x] `test_main_stop_no_namespace_exits` — verifies exit when no namespaces configured
- [x] Full suite: 524 tests pass, lint clean

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -177,7 +177,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 **`deploy.py collect`** — extracts results from the cluster PVC and writes to `workspace/runs/<run>/results/{phase}/<workload>/`.
 
-**`deploy.py stop`** — deletes the `sim2real-orchestrator` Kubernetes Job (with cascading pod deletion) in the primary namespace. Only meaningful for remote execution (`run --remote`). Does not touch `progress.json` — pair state is left as-is. If no remote orchestrator Job exists, prints a message and returns. Use `reset` separately to clear failed/stalled pair state.
+**`deploy.py stop`** — deletes the `sim2real-orchestrator` Kubernetes Job (with cascading pod deletion) in the primary namespace. Only meaningful when the orchestrator runs as an in-cluster Job. Does not touch `progress.json` — pair state is left as-is. If no remote orchestrator Job exists, prints a message and returns. Use `reset` separately to clear failed/stalled pair state.
 
 **`deploy.py reset`** — removes cluster resources (PipelineRuns, Helm releases) for all non-pending pairs. Failed/running/timed-out pairs are reset to `pending` so they can be re-dispatched. Done pairs stay `done` — only their PipelineRun is deleted to free cluster resources.
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -115,7 +115,7 @@ Phase state is tracked per-run in `workspace/runs/<run>/.state.json`. Delete it 
 Builds the EPP image and orchestrates PipelineRun execution across namespace slots. Operates independently of `transfer.yaml` — driven by workspace files and `setup_config.json`.
 
 ```bash
-python pipeline/deploy.py {run|status|collect|reset|pairs} [flags]
+python pipeline/deploy.py {run|status|collect|stop|reset|pairs} [flags]
 ```
 
 Common flags (all subcommands):
@@ -138,6 +138,7 @@ Common flags (all subcommands):
 python pipeline/deploy.py run     [flags]   # orchestrate parallel pool execution across namespace slots
 python pipeline/deploy.py status            # show progress snapshot of all (workload, package) pairs
 python pipeline/deploy.py collect [--package NAME…]
+python pipeline/deploy.py stop               # stop the remote orchestrator Job
 python pipeline/deploy.py reset [flags]     # tear down cluster resources for failed/stalled pairs
 python pipeline/deploy.py pairs   [flags]   # list available pair keys, workloads, and packages
 ```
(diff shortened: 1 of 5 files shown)


kalantar:
## Review: NEEDS CHANGES

PR cleanly implements all issue #125 requirements — the happy path works, tests cover the main branches, and the README is well-structured. However, there are error handling gaps that could mislead users in production, plus a few documentation inaccuracies.

### Must fix

1. **`kubectl get` catch block is too broad.** The `except CalledProcessError` treats every failure (auth denied, network timeout, wrong cluster) as "no Job exists." A user with a connectivity problem sees `No remote orchestrator started in sim2real-dev` and walks away while the Job keeps running. The rest of `deploy.py` uses `check=False, capture=True` with stderr inspection (25 other kubectl calls follow this pattern) — `_cmd_stop` should do the same, distinguishing `"not found"` in stderr from other errors.

2. **`kubectl delete` has no error handling.** It uses the default `check=True` with no try-except, so any delete failure (RBAC, race condition, timeout) produces a raw Python traceback. Every other delete call in the file uses `check=False`. Match the existing pattern and surface a clean `err()` message on failure.

3. **README references non-existent `run --remote` flag.** The new paragraph says "Only meaningful for remote execution (`run --remote`)." That flag doesn't exist yet (it's proposed in #127). A reader trying `deploy.py run --remote` gets an unrecognized argument error. Reword to avoid implying the flag exists — e.g., "Only meaningful when the orchestrator runs as an in-cluster Job."

4. **Shell completions not updated.** `pipeline/completions.bash` and `pipeline/completions.zsh` don't include `stop` — tab completion won't offer it.

### Should fix

5. **Module docstring in `deploy.py` doesn't list `stop`.** The subcommand table at the top of the file is missing the new subcommand.

6. **Add `capture=True` to the `kubectl delete` call.** Without it, kubectl's own output (`job.batch "sim2real-orchestrator" deleted`) prints to stdout interleaved with the `ok()` message. Other delete calls in the file use `capture=True`.

7. **Test for `kubectl delete` failure.** The current tests don't cover what happens when the Job exists but delete fails. After fixing the error handling (#2), add a test for the failure path.

8. **`stop` handler unnecessarily depends on `run_dir` existing.** The `stop` subcommand only needs the namespace from `setup_config`, but it runs after the `run_dir` existence check in `main()`. If no run is configured, `stop` fails before reaching the namespace logic. Since `stop` is independent of any run, consider short-circuiting before the `run_dir` check.

### Nits

- `test_main_stop_no_namespace_exits` should assert exit code is 1, not just that `SystemExit` is raised.
- Consider testing the singular `namespace` key fallback (no `namespaces` key in config) to exercise the backward-compat path.

kalantar:
All review items addressed in d13ed77.

**Error handling:** `_cmd_stop` now uses `check=False, capture=True` for both kubectl calls, matching the rest of deploy.py. The `get` call distinguishes "NotFound" in stderr (prints info message) from other errors like auth/network failures (prints error + exits 1). The `delete` call surfaces stderr on failure instead of tracebacks.

**run_dir independence:** Moved the `stop` dispatch before the run_name/run_dir checks in `main()`. `stop` now only needs `setup_config.json` for the namespace — works even with no run configured.

**Shell completions:** Added `stop` to both `completions.bash` (subcommands list + case pattern) and `completions.zsh` (subcommands array).

**Module docstring:** Added `stop` to the subcommand table at the top of deploy.py.

**README:** Reworded to "Only meaningful when the orchestrator runs as an in-cluster Job" — no longer references the non-existent `--remote` flag.

**Tests:** Now 8 tests covering: happy path, NotFound no-op, kubectl get auth error, kubectl delete failure, main dispatch with `namespaces` list, main dispatch with singular `namespace` fallback, no-namespace exit with code 1, and stop without a run directory.

kalantar:
## Review: NEEDS CHANGES (one item)

All six requirements from issue #125 are fully addressed. Design, tests, docs, and shell completions all look solid. One thing to fix before merge:

### Tighten NotFound detection in `_cmd_stop`

The current check:
```python
if "NotFound" in stderr or "not found" in stderr:
```

The lowercase `"not found"` match is too broad — it could false-positive on unrelated kubectl errors whose message happens to contain "not found" (e.g., resource type not found due to misconfigured kubeconfig, missing CRDs). A user would see "No remote orchestrator started" when the real problem is cluster connectivity or RBAC.

**Fix:** Replace with `"(NotFound)" in stderr` (matching the parenthesized Kubernetes API reason field, which is stable across kubectl versions). This eliminates the false-positive risk while still catching the standard `Error from server (NotFound): ...` format.

Everything else — the early-return before run_dir validation, namespace resolution with singular fallback, cascading delete, error paths with sys.exit(1), README docs, bash/zsh completions, and the 8-test suite — is ready to go.

kalantar:
Addressed in 6fa286f.

Tightened the NotFound check from `"NotFound" in stderr or "not found" in stderr` to `"(NotFound)" in stderr`, matching the parenthesized Kubernetes API reason field (`Error from server (NotFound): ...`). This eliminates false positives from unrelated kubectl errors that happen to contain "not found" in their message.

Also merged main and resolved conflicts — the collect subcommand docs from #134 are preserved alongside the new stop docs. 540 tests pass, lint clean.

Issue #123, opened 2026-05-14
Add deploy.py wipe subcommand for data cleanup


kalantar:
## Context

When reusing a run name (common during debugging and iteration), stale results from prior executions persist both locally (`workspace/runs/<run>/results/`) and on the PVC (`/data/<run-name>/`). There is currently no mechanism to explicitly clear this data — `cleanup` (to be renamed `reset`) only handles cluster resources (PipelineRuns, Helm releases), not experiment data.

## Requirements

- New subcommand: `deploy.py wipe`
- Deletes local results for the specified scope: `workspace/runs/<run>/results/`
- Supports `--remote` flag to also delete data on the PVC (`/data/<run-name>/...`)
- Supports scoping flags: `--only PAIR`, `--workload NAME`, `--package NAME`
- Without scope flags: wipes all results for the current run
- Confirmation prompt before deletion (unless `--yes` is passed)

## Example usage

```bash
# Wipe all local results for current run
deploy.py wipe

# Wipe local + PVC data for a specific workload
deploy.py wipe --workload sharegpt-32 --remote

# Wipe without confirmation
deploy.py wipe --yes
```

## Rationale

The PVC is a cache — data there is transient. Local results are a copy of that cache. When iterating on the same run name, users need an explicit way to clear stale data before re-executing. This replaces the need for manual `rm -rf results/`.

kalantar:
This should wait for #127 (remote execution). Once #127 lands, the patterns for PVC interaction (Job-based access, ConfigMap uploads) will be established — `wipe --remote` should build on those rather than inventing an ad-hoc mechanism.

Also from vetting, a few items to address before implementation:
- Specify whether `wipe` resets pair status in `progress.json` to `pending`, or leaves that to `run --force` / future `reset` subcommand
- Clarify how `--only PAIR` maps to the `results/<phase>/<workload>/` directory structure
- Add `--dry-run` for consistency with `cleanup`
- Confirm naming is correct after #124 (rename `cleanup` → `reset`) lands

kalantar:
i think this is done.

Issue #124, opened 2026-05-14
Rename deploy.py cleanup to reset


kalantar:
## Context

The `cleanup` subcommand deletes PipelineRuns and Helm releases for failed/stalled pairs, then resets their state to `pending`. With the addition of `wipe` (#123) for data cleanup, the name `cleanup` is ambiguous — it could mean cluster resources or experiment data.

## Requirements

- Rename subcommand from `cleanup` to `reset`
- Same behavior: delete PipelineRuns + Helm releases for non-pending pairs, reset state to pending
- Same flags: `--only`, `--workload`, `--package`, `--status`, `--dry-run`
- Update help text, epilog examples, and documentation

## Rationale

`reset` better conveys the purpose: "reset failed/stalled pairs so they can be retried." It pairs naturally with `wipe` (data) vs `reset` (cluster state + pair status).

PR #138, opened 2026-05-14
Add deploy.py wipe subcommand for data cleanup (#123)

kalantar:
Partially addresses #123 (`--remote` flag deferred — see issue comment)

## Summary

- Adds `deploy.py wipe` subcommand that deletes local results (`results/<package>/<workload>/`) and resets wiped pairs to `pending` in `progress.json`
- Supports `--only`, `--workload`, `--package` scoping flags (reuses `_resolve_scope`)
- Supports `--dry-run` to preview what would be deleted without acting
- Interactive confirmation prompt, skippable with `--yes` / `-y`
- Updates bash and zsh shell completions
- Documents `wipe` in `pipeline/README.md`

## Design decisions

- **No `--remote` flag**: `collect` is gated on `status == "done"`, so resetting to `pending` makes PVC data unreachable. PVC data is overwritten on re-execution anyway, making remote deletion redundant.
- **Wipe resets status to `pending`**: ensures a subsequent `collect` won't re-pull stale data, and `deploy.py run` will re-dispatch the pair.
- **Pending pairs are skipped**: their directories are left untouched (nothing to wipe for a pending pair).
- **Empty parent cleanup**: after deleting a workload dir, `rmdir()` attempts to remove the package dir — succeeds only if it's empty, so sibling workloads are safe.
- **Error resilience**: `shutil.rmtree` failures are caught per-pair (logged, pair skipped), so one permission error doesn't abort the entire wipe or leave `progress.json` unsaved.

## Test plan

- [x] 14 dedicated tests covering: unscoped wipe, `--workload`/`--only`/`--package` scoping, `--dry-run`, pending-skip, missing progress, confirmation abort/accept, EOFError on input, filter mismatch, empty parent cleanup, no-results-on-disk, sibling preservation
- [x] Full suite: 582/582 tests pass
- [x] CLI help: `wipe` appears in help text with correct flags
- [x] README updated with `wipe` documentation

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,8 +89,8 @@ All artifacts live under `<experiment-root>/workspace/` (gitignored). When no `-
 | `runs/<run>/cluster/treatment.yaml` | `prepare.py` Phase 4 | `deploy.py` |
 | `runs/<run>/cluster/pipelinerun-*.yaml` | `prepare.py` Phase 4 | `deploy.py run` |
 | `runs/<run>/run_summary.md` | `prepare.py` Phase 5 | human review |
-| `runs/<run>/results/{phase}/` | `deploy.py collect` | `/sim2real-analyze` skill |
-| `runs/<run>/progress.json` | `deploy.py run` | `deploy.py status`, `deploy.py collect` |
+| `runs/<run>/results/{phase}/` | `deploy.py collect` | `/sim2real-analyze` skill, `deploy.py wipe` |
+| `runs/<run>/progress.json` | `deploy.py run`, `deploy.py wipe` | `deploy.py status`, `deploy.py collect`, `deploy.py wipe` |
 | `runs/<run>/plans/<phase>/<workload>/` | `deploy.py run` | workload tasks |
 | `context/{scenario}/{hash}.md` | `prepare.py` Phase 2 | `prepare.py` Phase 2 (cache) |
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -115,7 +115,7 @@ Phase state is tracked per-run in `workspace/runs/<run>/.state.json`. Delete it 
 Builds the EPP image and orchestrates PipelineRun execution across namespace slots. Operates independently of `transfer.yaml` — driven by workspace files and `setup_config.json`.
 
 ```bash
-python pipeline/deploy.py {run|status|collect|stop|reset|pairs} [flags]
+python pipeline/deploy.py {run|status|collect|stop|reset|wipe|pairs} [flags]
 ```
 
 Common flags (all subcommands):
@@ -140,6 +140,7 @@ python pipeline/deploy.py status            # show progress snapshot of all (wor
 python pipeline/deploy.py collect [flags]     # pull results from the cluster PVC
 python pipeline/deploy.py stop               # stop the remote orchestrator Job
 python pipeline/deploy.py reset [flags]     # tear down cluster resources for failed/stalled pairs
+python pipeline/deploy.py wipe  [flags]     # delete local results and reset pairs to pending
 python pipeline/deploy.py pairs   [flags]   # list available pair keys, workloads, and packages
 ```
 
@@ -200,6 +201,18 @@ When `--only` or `--workload` is given, only matching workload subdirectories ar
 | `--dry-run` | Print what would be reset without acting |
 
 **Safety:** Results in `workspace/runs/<run>/results/` are preserved — only cluster resources are removed. For `done` pairs, only the PipelineRun is deleted (Tekton already tore down Helm releases).
+
+**`deploy.py wipe`** — deletes local results (`results/<package>/<workload>/`) for non-pending pairs and resets their status to `pending` in `progress.json`. Use when reusing a run name and stale results from prior executions need to be cleared before re-executing. Pending pairs are skipped (nothing to wipe). Empty package directories are cleaned up automatically.
+
+| Flag | Description |
+|------|-------------|
+| `--only PAIR` | Scope wipe to one specific pair key (`wl-` prefix optional) |
+| `--workload NAME` | Scope wipe to pairs matching this workload |
+| `--package NAME` | Scope wipe to pairs matching this package |
+| `--dry-run` | Print what would be wiped without acting |
+| `--yes` / `-y` | Skip confirmation prompt |
+
+**Interaction with `collect`:** Wiping resets pair status to `pending`, which makes any PVC data unreachable via `collect` (collect only pulls `done` pairs). Re-executing the pairs writes fresh data to the PVC.
 
 **`deploy.py pairs`** — lists available pair keys, workloads, and packages by scanning `cluster/pipelinerun-*.yaml`. Works without `progress.json`.
 
--- a/pipeline/completions.bash
+++ b/pipeline/completions.bash
@@ -12,7 +12,7 @@ _sim2real_deploy() {
     cur="${COMP_WORDS[COMP_CWORD]}"
     prev="${COMP_WORDS[COMP_CWORD-1]}"
 
-    local subcommands="run status collect stop reset pairs"
+    local subcommands="run status collect stop reset wipe pairs"
     local status_values="pending running done failed timed-out stalled"
 
     # Extract --experiment-root and --run values while finding the subcommand.
@@ -24,7 +24,7 @@ _sim2real_deploy() {
             --experiment-root) ((i++)); _exroot="${COMP_WORDS[i]}" ;;
             --run)             ((i++)); _run_name="${COMP_WORDS[i]}" ;;
             -*) ;;
-            run|status|collect|stop|reset|pairs)
+            run|status|collect|stop|reset|wipe|pairs)
                 subcmd="${COMP_WORDS[i]}"
                 break
                 ;;
@@ -80,6 +80,9 @@ _sim2real_deploy() {
                 ;;
             reset)
                 COMPREPLY=($(compgen -W "--only --workload --package --status --dry-run" -- "$cur"))
+                ;;
+            wipe)
+                COMPREPLY=($(compgen -W "--only --workload --package --dry-run --yes" -- "$cur"))
                 ;;
             collect)
                 COMPREPLY=($(compgen -W "--only --workload --package --skip-logs" -- "$cur"))
--- a/pipeline/completions.zsh
+++ b/pipeline/completions.zsh
@@ -105,7 +105,7 @@ _sim2real_deploy() {
                         '--workload[Scope to workload]:workload:_deploy_py_workloads' \
                         '--package[Scope to package]:package:_deploy_py_packages' \
                         '--status[Scope to status]:status:_deploy_py_statuses' \
-                        '--dry-run[Print what would be cleaned up]'
+                        '--dry-run[Print what would be reset]'
                     ;;
                 wipe)
                     _arguments \
--- a/pipeline/completions.zsh
+++ b/pipeline/completions.zsh
@@ -60,6 +60,7 @@ _sim2real_deploy() {
                 'collect:Pull results for completed packages'
                 'stop:Stop the remote orchestrator Job'
                 'reset:Tear down cluster resources for all non-pending pairs'
+                'wipe:Delete local results and reset pairs to pending'
                 'pairs:List available pair keys, workloads, and packages'
             )
             _describe 'subcommand' subcommands
@@ -105,6 +106,14 @@ _sim2real_deploy() {
                         '--package[Scope to package]:package:_deploy_py_packages' \
                         '--status[Scope to status]:status:_deploy_py_statuses' \
                         '--dry-run[Print what would be cleaned up]'
+                    ;;
+                wipe)
+                    _arguments \
+                        '--only[Scope to one pair key]:pair key:_deploy_py_pair_keys' \
+                        '--workload[Scope to workload]:workload:_deploy_py_workloads' \
+                        '--package[Scope to package]:package:_deploy_py_packages' \
+                        '--dry-run[Print what would be wiped]' \
+                        '(-y --yes)'{-y,--yes}'[Skip confirmation prompt]'
                     ;;
                 pairs)
                     _arguments \
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -14,6 +14,7 @@ Subcommands:
 import argparse
 import json
 import os
+import shutil
 import subprocess
 import sys
 import time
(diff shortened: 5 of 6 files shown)


kalantar:
## Review: NEEDS CHANGES

Solid implementation overall — the scoping flag reuse, confirmation prompt, and dry-run mode are well done. Shell completions are correct. Tests cover the main paths well. A few things need fixing before merge:

### Blocking

**1. Package-dir cleanup can delete out-of-scope data**

The empty-parent cleanup uses `shutil.rmtree(pkg_dir)`, which removes the entire package directory — including workload subdirectories belonging to out-of-scope pairs. Example: `wipe --workload wl-smoke` resets wl-smoke to pending, then the cleanup sees all pairs for `baseline` are now pending (wl-load was already pending), and deletes `results/baseline/` entirely — including `results/baseline/wl-load/` which was never in scope.

Fix: use `pkg_dir.rmdir()` (which only succeeds on empty dirs) instead of `shutil.rmtree(pkg_dir)`, wrapped in a try/except OSError pass.

**2. Dry-run output doesn't match real execution for pending pairs**

The `targets` list includes all scoped pairs (including pending), and the real execution path deletes their directories. But dry-run skips pending pairs (`if key not in actionable: continue`), so the preview understates what will actually happen.

Fix: either gate `shutil.rmtree` on `key in actionable` (simplest — skip pending pairs in real execution too), or include pending-pair directory deletions in the dry-run output.

**3. `shutil.rmtree` failure leaves progress.json unsaved**

If `rmtree` raises (permissions, busy file, etc.), the function crashes before `store.save()`. Directories are partially deleted but progress.json still shows old statuses — and there's no recovery path. Same applies if `store.save()` itself fails after deletions.

Fix: wrap `shutil.rmtree` in try/except OSError (log error, skip status reset for that pair, continue). Use try/finally or explicit try/except around `store.save()` to ensure progress is persisted or the user gets a clear message.

**4. `pipeline/README.md` not updated**

Contributing guidelines require updating README.md when subcommands change. The `wipe` subcommand, its flags, and behavior need to be documented — add it to the subcommand list (line 118) and add a section similar to the `reset` documentation.

### Non-blocking improvements

- Move `import shutil` to module-level imports for consistency with the rest of the file.
- Use `args.dry_run` / `args.yes` instead of `getattr` with defaults — for destructive commands, a clear `AttributeError` on a parser typo is safer than silently skipping dry-run.
- Wrap `input()` in try/except `EOFError` for non-interactive contexts (CI, piped stdin).
- Handle corrupt `progress.json` (`ValueError` from `store.load()`) with a user-friendly message instead of a raw traceback.
- Tests: assert all reset fields (retries, pending_stalls, pending_since, namespace), not just status. Add `pending_stalls`/`pending_since` to the test fixture for realism. Add a test for wipe when results dirs don't exist on disk. Add a test for the user-confirmed path (yes=False, input="y").
- The PR says "Closes #123" but the `--remote` flag from the issue is deferred. Consider updating to "Partially addresses #123" or noting the deferral in the issue.

kalantar:

**Package-dir cleanup** — replaced `shutil.rmtree(pkg_dir)` with `pkg_dir.rmdir()`, which only succeeds on empty directories. This prevents deleting workload dirs belonging to out-of-scope pairs.

**Dry-run / execution parity** — the root cause was that `targets` was built from `_scope` (all pairs including pending) but dry-run only printed `actionable` (non-pending). Simplified by building `targets` from `actionable` only. Pending pairs are now consistently skipped in both paths — their directories are left untouched.

**rmtree error handling** — wrapped in try/except OSError per-pair. On failure: logs a warning, skips that pair's status reset, continues to the next. `store.save()` always runs for whatever was successfully processed.

**README** — added `wipe` to both subcommand listings and a full documentation section with flag table and collect-interaction note.

**Non-blocking items** — all addressed: `import shutil` moved to module level, `getattr` replaced with direct attribute access, `input()` wrapped for EOFError. Tests now assert all reset fields (retries, pending_stalls, pending_since, namespace) and include four new cases: confirmation-accept, EOFError, no-results-on-disk, and sibling-preservation.

**PR text** — changed "Closes #123" to "Partially addresses #123" since `--remote` is deferred.

**Corrupt progress.json** — filed as #140 since it affects all subcommands, not just wipe.

kalantar:
## Review: NEEDS CHANGES

Well-structured PR — clean implementation, good test coverage (14 tests, all passing), thorough documentation, and correct shell completions. Two issues should be addressed before merge:

### Must fix

**1. Use `CompositeProgressStore`, not just `LocalProgressStore`**

`_cmd_wipe` writes only to the local `progress.json`, but `_cmd_reset` (the closest analog) uses `CompositeProgressStore` to write to both local file and the cluster ConfigMap. After a wipe, `deploy.py status --remote` would show stale data (pairs still "done" on the ConfigMap while local says "pending"). Follow the same pattern as `_cmd_reset`: construct a `CompositeProgressStore` when a ConfigMap namespace is available. This requires passing `setup_config` and `namespaces` through `main()`, same as `_cmd_reset` does.

**2. Add test for `shutil.rmtree` failure path**

The `OSError` catch with `continue` (skips status reset for that pair) is untested. This is the only error-handling path that changes observable behavior — a pair whose deletion fails stays at its original status and won't be re-dispatched by `deploy.py run`. Use `monkeypatch` to make `shutil.rmtree` raise for one target, then verify: that pair's status is NOT reset, other pairs ARE wiped, and the error count appears in the summary message.

### Should fix

**3. Warn on entries missing `package`/`workload`** — The `if pkg and wl:` guard silently skips progress entries without those fields. They're counted in `actionable` but excluded from `targets`, so the scope count could overstate what actually gets wiped. Emit a `warn()` for skipped entries instead of silently dropping them.

**4. Update CLAUDE.md workspace artifact table** — Add `deploy.py wipe` as a reader/writer of `progress.json` and reader of `results/`. The table is the definitive map of stage I/O per the contributing guidelines.

### Nice to have (non-blocking)

- Add a section header comment before `_cmd_wipe` (every other `_cmd_*` has one)
- README wipe section could mention that retry/stall counters are also reset, not just status
- Argparse integration test (`build_parser().parse_args(["wipe", ...])`) would catch CLI regressions cheaply

kalantar:
Addressed the second review round:

**CompositeProgressStore** — `_cmd_wipe` now constructs a `CompositeProgressStore` (local file + ConfigMap) when a ConfigMap namespace is available, matching the `_cmd_reset` pattern. This keeps `deploy.py status --remote` in sync after a wipe. The function takes `setup_config` and derives the namespace via `_configmap_namespace`.

**rmtree failure test** — added `test_wipe_rmtree_failure_skips_pair`: monkeypatches `shutil.rmtree` to fail for one target, verifies that pair's status is preserved while the other pair is wiped, and checks the error count appears in the summary.

**Missing package/workload warning** — the `if pkg and wl:` guard now emits `warn()` for entries with missing fields instead of silently dropping them. Added `test_wipe_warns_on_missing_package_workload` to cover this.

**CLAUDE.md** — updated the workspace artifact table: `deploy.py wipe` now listed as reader/writer of `progress.json` and reader of `results/`.

Skipped three non-blocking items:
- Section header comment before `_cmd_wipe` — pattern is inconsistent across the file (`_cmd_reset`, `_cmd_collect`, `_cmd_run` also lack direct headers)
- README mention of retry/stall counter reset — internal state fields, documenting adds noise
- Argparse integration test — argparse errors fail loudly; 16 functional tests already exercise parsed args

584/584 tests pass, lint clean.

kalantar:
## Review: PR #138 — Add deploy.py wipe subcommand (#123)

**Verdict: READY TO MERGE** (with two optional improvements noted below)

### Summary

The implementation is clean, well-tested (16 tests, 584 full suite passing), follows established patterns (`_cmd_reset`, `_resolve_scope`, `CompositeProgressStore`), and meets all claimed requirements. Documentation is consistent across argparse help, README, CLAUDE.md, and shell completions.

The `--remote` flag deferral is well-reasoned — resetting to `pending` already makes PVC data unreachable from `collect`, and re-execution overwrites it.

### Findings

No blocking issues. Two improvements worth considering for a follow-up:

1. **`store.save()` failure after file deletion** — If the save raises (disk full, permissions), result directories are already deleted but `progress.json` still shows old status. A try/except with a clear message ("results deleted but progress save failed — use --yes to re-run wipe") would help users recover from this edge case. This matches the same gap in `_cmd_reset`, so not a regression.

2. **EOFError on `input()` gives generic "Aborted"** — In non-interactive contexts (CI, `ssh -T`), users see "Aborted" with no hint about `--yes`. Changing the EOFError handler to print "Aborted (non-interactive — use --yes to skip confirmation)" would save debugging time.

### Requirements Coverage

| Requirement | Status |
|---|---|
| `deploy.py wipe` subcommand | Done |
| Deletes local results | Done |
| `--only`, `--workload`, `--package` scoping | Done |
| Unscoped wipes all | Done |
| Confirmation prompt / `--yes` | Done |
| `--dry-run` | Done |
| `--remote` flag | Deferred (see PR body — sound rationale) |

### Test Coverage

Strong behavioral coverage: all scoping flags, dry-run, pending-skip, confirmation flows (accept/reject/EOF), rmtree failure resilience, missing-field warnings, empty-parent cleanup, sibling preservation. One gap: the `CompositeProgressStore` (ConfigMap) code path is exercised only via `LocalProgressStore` in tests — the composite wiring is identical to `_cmd_reset` and `CompositeProgressStore` has its own unit tests, so this is low risk.

### Documentation

All surfaces are consistent (module docstring, argparse, README, CLAUDE.md artifact table, bash/zsh completions). One pre-existing nit: the `reset` subcommand's zsh completion says "Print what would be cleaned up" (stale terminology from the cleanup→reset rename) — now adjacent to `wipe`'s correct "Print what would be wiped".

kalantar:
Two small fixes from the final review:

**EOFError message** — now prints "Aborted (non-interactive — use --yes to skip confirmation)" instead of generic "Aborted", so users in CI or piped-stdin contexts know what flag to add. Test updated to assert the hint.

**Stale zsh completion** — reset's --dry-run description said "Print what would be cleaned up" (leftover from the cleanup→reset rename). Fixed to "Print what would be reset".

The save-failure-after-deletion concern is already tracked in #140 (applies to all subcommands, not a regression here).

584/584 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

One blocking item; the rest are suggestions.

### Requirements coverage

All issue #123 requirements are addressed: `--dry-run`, `--only`/`--workload`/`--package` scoping, progress reset to `pending`, and naming aligned with the post-#124 world. `--remote` is explicitly deferred per the issue comment — reasonable given that resetting to `pending` makes PVC data unreachable via `collect` anyway.

### Blocking

**Exit code 0 on partial failure.** When `shutil.rmtree` fails for one or more pairs, `_cmd_wipe` prints `"N pair(s) wiped (M failed — check permissions)"` but returns normally, so the process exits 0. Scripts using `set -e` or checking `$?` will believe the wipe fully succeeded. Pairs whose results couldn't be deleted keep their original status — a silent inconsistency between what the caller assumes and what happened.

Fix: set a non-zero exit code when `errors > 0`:

```python
store.save(progress)
msg = f"{wiped} pair(s) wiped"
if errors:
    msg += f" ({errors} failed — check permissions)"
    warn(msg)
    sys.exit(1)
ok(msg)
```

This matches the spirit of `_cmd_reset` but breaks the current pattern (which has the same issue). Worth fixing here rather than propagating.

### Suggested improvements (non-blocking)

1. **Save-failure handling.** If `store.save()` raises after directories are already deleted, the user gets a raw traceback with no guidance. A try/except that says "results were deleted but progress.json was NOT updated — re-run wipe to reconcile" would help in this (rare but dangerous) divergence scenario.

2. **Narrow `pkg_dir.rmdir()` catch.** The `except OSError: pass` suppresses all errors, but the only expected failure is `ENOTEMPTY`. Permission errors or I/O failures on the parent dir are silently lost. Low risk since this is best-effort cleanup, but easy to narrow.

3. **CompositeProgressStore wiring untested.** All tests pass `setup_config=None`, so the ConfigMap branch is never exercised. A test that passes a namespace and mocks `ConfigMapProgressStore.save` would catch wiring regressions. (The store itself is well-tested in `test_progress.py`.)

4. **No per-pair output for diskless wipes.** When a pair is non-pending but has no results on disk, its status is silently reset with no output line. An `ok(f"Reset: {key} (no results on disk)")` would make wipe fully transparent.

### What's good

- Clean design that mirrors `_cmd_reset` patterns (CompositeProgressStore, `_resolve_scope`, per-pair error handling)
- 16 well-structured tests covering scoping, dry-run, confirmation prompts, error resilience, empty-parent cleanup, and edge cases
- Documentation updated across README, CLAUDE.md artifact table, and both shell completions
- Bonus fix: stale zsh completion text for `reset --dry-run` ("cleaned up" → "reset")

kalantar:
**Exit code on partial failure** — `_cmd_wipe` now calls `sys.exit(1)` when any pair fails to delete. Progress is saved first (successful pairs are persisted), then the non-zero exit lets scripts detect the failure. Test updated to assert `SystemExit(1)`.

**Diskless reset output** — pairs whose status is reset but have no results directory on disk now print `"Reset: <key> (no results on disk)"` so wipe output is fully transparent. Test updated to check for this message.

Skipped three non-blocking items:
- Narrowing `pkg_dir.rmdir()` catch to ENOTEMPTY — the only realistic failures are ENOTEMPTY (expected) and ENOENT (race); permission errors would have already surfaced during child rmtree
- CompositeProgressStore wiring test — store has its own tests in test_progress.py; wiring is identical to `_cmd_reset` which also doesn't test it
- Save-failure handling — already tracked in #140

584/584 tests pass, lint clean.

kalantar:
## Review: READY TO MERGE

All issue #123 requirements are addressed (--remote correctly deferred). The implementation is clean, follows established patterns (_resolve_scope, CompositeProgressStore), and comes with thorough documentation and 16 well-written tests. No blocking issues.

**1. Wrap `store.save()` in try/except** — After the wipe loop, result directories are already deleted. If `store.save(progress)` fails (disk full, kubectl error), the user gets a raw traceback with no guidance. A try/except with an actionable message would help:

```python
try:
    store.save(progress)
except Exception as e:
    err(f"Failed to save progress after wiping {wiped} pair(s): {e}")
    err("Results already deleted. Re-run 'deploy.py wipe --yes' to reconcile.")
    sys.exit(1)
```

This is the same atomicity gap other subcommands have, so not a regression — but wipe is the most destructive, making it the best place to start.

**2. Add a CompositeProgressStore wiring test** — All 16 tests hit the local-only path. `test_deploy_run.py` has `test_cmd_reset_creates_composite_store_when_namespace_available` as a pattern to follow. Without this, a broken import or wrong argument order to `CompositeProgressStore` would silently skip ConfigMap updates in production.

**3. `{phase}` vs `{package}` in CLAUDE.md artifact table** — Pre-existing inconsistency (not introduced here): the table says `results/{phase}/` but the code and README use `results/<package>/<workload>/`. Opportunity to align if convenient.

**4. `deploy.py reset` missing from "Written by" for progress.json** — Also pre-existing. Could be fixed in this same CLAUDE.md edit.

**5. Summary message wording** — `"check permissions"` in the error summary is slightly misleading for non-permission OSError failures. Consider `"check warnings above"` instead.

PR #141, opened 2026-05-14
Build and publish sim2real-orchestrator container image

### LINKED ISSUE #139: Build and publish sim2real-orchestrator container image
## Context

#127 (remote orchestrator Job) needs a container image containing Python 3.12, kubectl, PyYAML, and the `pipeline/` code. No suitable public image exists. Building a custom image and publishing it to ghcr.io gives us a stable, fast-starting base with no runtime downloads.

### Dockerfile

Minimal image based on `python:3.12-slim`:
- Install `kubectl` (match cluster version or use latest stable)
- `COPY pipeline/` into the image
- `pip install PyYAML` (and any other deps from `requirements.txt`)
- Entrypoint: `python -m pipeline.deploy` or similar

### Registry

Publish to `ghcr.io/inference-sim/sim2real/orchestrator` (GitHub Container Registry). The repo is already on GitHub — ghcr.io requires no external credentials and GitHub Actions has built-in `GITHUB_TOKEN` auth for pushes.

### CI build

GitHub Actions workflow (`.github/workflows/orchestrator-image.yml`):
- **Trigger**: push to `main` when `pipeline/**` or `Dockerfile` changes, plus manual `workflow_dispatch`
- **Steps**: build, tag with git SHA + `latest`, push to ghcr.io
- **Cache**: use Docker layer caching to keep builds fast (~seconds with cache hits)

### Versioning

- Tag every build with the git SHA for traceability
- Also tag `latest` — the orchestrator should always match the current pipeline code
- `deploy.py run --remote` (#127) will reference the image by SHA or `latest` as appropriate

## Notes

The image is small (~200MB) and the build is fast. The Dockerfile is ~5 lines. The CI workflow is the bulk of the work.

This is a prerequisite for #127.

kalantar:
## Summary

Closes #139

Adds the build infrastructure for the sim2real orchestrator container image — prerequisite for #127 (remote orchestrator Job).

- **`.dockerignore`** — excludes non-runtime files (submodules, workspace, docs, .git) from build context
- **`Dockerfile`** — minimal image: `python:3.12-slim` + kubectl + PyYAML + `pipeline/` code (~200MB)
- **`.github/workflows/orchestrator-image.yml`** — builds and pushes to `ghcr.io/inference-sim/sim2real/orchestrator` on main pushes (when pipeline code changes) and manual dispatch; tags with git SHA + `latest`; uses GHA Docker layer cache

## Design choices worth noting

- **Only PyYAML installed** (not the full `requirements.txt`). `deploy.py` and `pipeline/lib/` only import `yaml` as a third-party dependency. Skipping matplotlib/numpy/anthropic keeps the image small.
- **kubectl from latest stable** rather than pinning a version. The orchestrator only uses version-tolerant operations (`apply`, `get`, `delete`).
- **curl installed then removed** in the same RUN layer to minimize image size.
- **No PR-trigger build** — PRs don't push images. A build-only PR check could be added later if needed.

## Test plan

- [x] 568 existing tests pass
- [ ] Docker build succeeds (no local Docker — CI will verify on merge)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,9 +7,10 @@
 __pycache__
 *.pyc
 workspace
-workspace */
 inference-sim
+llm-d-benchmark
 tektonc-data-collection
+pipeline/tests
 docs
 prompts
 *.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -17,4 +17,5 @@ prompts
 *.log
 .env
 .env.local
+go.work
 .DS_Store
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,19 @@
+.git
+.github
+.gitignore
+.gitmodules
+.claude
+.venv
+__pycache__
+*.pyc
+workspace
+workspace */
+inference-sim
+tektonc-data-collection
+docs
+prompts
+*.md
+*.log
+.env
+.env.local
+.DS_Store
--- a/.github/workflows/orchestrator-image.yml
+++ b/.github/workflows/orchestrator-image.yml
@@ -1,0 +1,45 @@
+name: Build Orchestrator Image
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "pipeline/**"
+      - "Dockerfile"
+      - ".github/workflows/orchestrator-image.yml"
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: inference-sim/sim2real/orchestrator
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: true
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/orchestrator-image.yml
+++ b/.github/workflows/orchestrator-image.yml
@@ -33,13 +33,27 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
-      - name: Build and push
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ github.sha }}
+            type=raw,value=latest,enable={{is_default_branch}}
+
+      - name: Build
         uses: docker/build-push-action@v6
         with:
           context: .
-          push: true
-          tags: |
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }}
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          push: false
+          load: true
+          tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Smoke test
+        run: docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
+
+      - name: Push
+        run: docker push --all-tags ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
--- a/.github/workflows/orchestrator-image.yml
+++ b/.github/workflows/orchestrator-image.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - "pipeline/**"
       - "Dockerfile"
+      - ".dockerignore"
       - ".github/workflows/orchestrator-image.yml"
   workflow_dispatch:
 
@@ -53,7 +54,10 @@ jobs:
           cache-to: type=gha,mode=max
 
       - name: Smoke test
-        run: docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
+        run: |
+          docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
+          docker run --rm --entrypoint kubectl ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --client
+          docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "from pipeline.lib import assemble, backoff, capacity, context_builder, epp, health, manifest, pod_pending, progress, run_manager, state_machine, tekton, values; print('OK')"
 
       - name: Push
         run: docker push --all-tags ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM python:3.12-slim
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends curl ca-certificates && \
-    curl -LO "https://dl.k8s.io/release/$(curl -Ls https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && \
+    KUBE_VERSION=$(curl -fsSL https://dl.k8s.io/release/stable.txt) && \
+    curl -fLO "https://dl.k8s.io/release/${KUBE_VERSION}/bin/linux/amd64/kubectl" && \
     install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl && \
     rm kubectl && \
     apt-get purge -y curl && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.12-slim
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends curl ca-certificates && \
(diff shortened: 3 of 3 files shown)


kalantar:
## Review: NEEDS CHANGES

Overall this PR cleanly satisfies all issue #139 requirements (Dockerfile, GHCR registry, CI workflow, SHA + latest tagging). The design choices are sound — PyYAML-only install, curl cleanup in the same layer, and the entrypoint using `pipeline/deploy.py` rather than `python -m` are all correct. Three small issues should be addressed before merge.

### Should fix

1. **Dockerfile: kubectl download can silently produce a corrupted binary.** The inner `curl -Ls` that fetches the stable version string will return empty or garbage on network failure, causing the outer curl to download from a malformed URL. The `-s` flag suppresses all errors. Use `-f` (fail on HTTP errors) and capture the version into a variable so its absence breaks the `&&` chain immediately:
   ```dockerfile
   KUBE_VERSION=$(curl -fsSL https://dl.k8s.io/release/stable.txt) && \
   curl -fLO "https://dl.k8s.io/release/${KUBE_VERSION}/bin/linux/amd64/kubectl" && \
   ```

2. **`.dockerignore`: `workspace */` (line 10) is a no-op.** The space before `*/` makes it a literal filename match, not a glob. The `workspace` entry on line 9 already covers the top-level directory. Either remove the line or replace with `*/workspace` if the intent is to catch nested workspace dirs.

3. **`.dockerignore`: missing `llm-d-benchmark` submodule.** The file excludes `inference-sim` and `tektonc-data-collection` but not the third submodule `llm-d-benchmark` (confirmed in `.gitmodules`). It won't end up in the image (outside the `COPY pipeline/` path) but bloats the build context unnecessarily.

### Worth considering (non-blocking)

- **`workflow_dispatch` can overwrite `:latest` from a feature branch.** A manual dispatch from a non-main branch silently replaces the production tag. Conditionally applying `:latest` only on `refs/heads/main` would prevent this.
- **No smoke test before push.** A quick `docker run --rm <image> --help` between build and push would catch broken images before they become `:latest`.
- **`pipeline/tests/` copied into the production image.** Adding `pipeline/tests` to `.dockerignore` trims ~1.6MB of test files from the image.
- **kubectl checksum verification.** Kubernetes publishes SHA-256 checksums alongside binaries — verifying them in the same RUN layer is a one-line addition for supply-chain hardening.

kalantar:
## NEEDS CHANGES

Clean, well-structured PR that satisfies all requirements from #139. Three items should be addressed before merge; the rest are recommendations.

### Must fix

1. **Add `--fail` to both `curl` invocations in the Dockerfile.** Without `-f`, if the Kubernetes version endpoint returns an HTTP error (404/503/error page), curl exits 0 and the build proceeds — installing a garbage file as `kubectl`. The image builds and pushes successfully; the failure only surfaces at runtime with an inscrutable error. One-character fix per invocation:
   ```
   curl -fLs https://dl.k8s.io/release/stable.txt
   curl -fLO "https://dl.k8s.io/release/..."
   ```
   Adding `kubectl version --client` after `install` would further validate the binary is executable.

2. **Remove the `workspace */` line from `.dockerignore`.** The space between `workspace` and `*/` makes this a literal match for a filename containing a space — it matches nothing. The `workspace` entry on the previous line already excludes the directory and all contents.

3. **Add `.dockerignore` to the workflow `paths` trigger.** A change to `.dockerignore` alters what enters the build context and can change the resulting image, but currently won't trigger a rebuild.

### Recommended improvements (non-blocking)

- **Exclude `pipeline/tests/` from the image.** `COPY pipeline/ pipeline/` pulls in ~1.6MB of test files that serve no runtime purpose. Add `pipeline/tests` to `.dockerignore`.
- **Guard `latest` tag on `workflow_dispatch`.** Manual dispatch from a feature branch currently overwrites `:latest` with non-main code. Conditionally apply `latest` only when `github.ref == 'refs/heads/main'`.
- **Add `llm-d-benchmark` to `.dockerignore`.** The other two submodules are excluded but this one is missing — inflates build context unnecessarily.
- **Consider a PR-trigger build (no push)** to catch Dockerfile breakage before merge rather than after.
- **Consider kubectl checksum verification** (`kubectl.sha256` is published alongside each release).

### What's done well

- Correct dependency analysis: only PyYAML is needed at runtime (verified across all `pipeline/lib/` imports).
- Proper exec-form `ENTRYPOINT` for signal handling.
- `ca-certificates` correctly retained after `curl` is purged (kubectl needs them for TLS).
- `sys.path` setup in `deploy.py` works correctly with the `/app` WORKDIR.
- Well-scoped GHA permissions (`contents: read`, `packages: write`).
- GHA layer caching and dual-tagging (SHA + latest) for traceability.

kalantar:
All three should-fix items addressed, plus three of the four non-blocking suggestions. Pushed as e688f97.

**kubectl download hardening** — switched both curl calls to `-fsSL`/`-fLO` and captured the version into a `KUBE_VERSION` variable so a network failure breaks the `&&` chain immediately instead of producing a corrupted binary.

**`.dockerignore` cleanup** — removed the no-op `workspace */` line (the space made it a literal match), added the `llm-d-benchmark` submodule, and added `pipeline/tests` to keep test files out of the production image.

**Conditional `:latest` tag** — replaced the inline `tags:` list with `docker/metadata-action`. The `latest` tag now only applies on pushes to main (`enable={{is_default_branch}}`), so a manual `workflow_dispatch` from a feature branch won't overwrite it.

**Smoke test** — added a `docker run --rm ... --help` step between build and push. The build now uses `load: true` (local daemon) first, runs the smoke test, then pushes via `docker push --all-tags`.

**kubectl checksum verification** — filed as #142 for follow-up. Valid hardening but a separate concern from the build infrastructure in this PR.

kalantar:
## Review: NEEDS CHANGES

Clean, well-scoped PR that covers all issue #139 requirements. Three items should be addressed before merge:

### 1. Strengthen the smoke test

The `--help` smoke test validates that Python starts and PyYAML imports, but `deploy.py` lazy-imports all `pipeline.lib.*` modules inside function bodies, so a broken package structure or missing kubectl passes CI silently. Add two quick checks (~2s total):

```yaml
- name: Smoke test
  run: |
    docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
    docker run --rm --entrypoint kubectl ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --client
    docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "from pipeline.lib import assemble, backoff, context_builder, manifest, pod_pending, progress, run_manager, state_machine, tekton, values; print('OK')"
```

### 2. Verify kubectl checksum

The binary is downloaded via curl with no integrity check. A truncated or corrupted download gets installed silently. Add the official sha256 verification in the same RUN layer:

```dockerfile
    curl -fLO "https://dl.k8s.io/release/${KUBE_VERSION}/bin/linux/amd64/kubectl.sha256" && \
    echo "$(cat kubectl.sha256)  kubectl" | sha256sum --check && \
```

### 3. Tighten `.dockerignore`

Missing entries that get sent to the build context unnecessarily: `tests/` (top-level, distinct from `pipeline/tests`), `go.work`, `tmp/`.

### Nice-to-have (not blocking)

- **Push step fragility:** `docker push --all-tags` pushes every locally-tagged image for that repo name. Pushing individual tags from `steps.meta.outputs.tags` or using a second `build-push-action` with `push: true` + GHA cache would be more robust.
- **Pin PyYAML version range:** `pip install "PyYAML>=6.0,<7"` matches `requirements.txt` and prevents surprise breakage from a future major version.
- **Add `.dockerignore` to workflow path triggers:** changes to `.dockerignore` alter image contents but don't currently trigger a rebuild.
- **kubectl platform:** Dockerfile hardcodes `linux/amd64` in the download URL but the build step doesn't set `platforms: linux/amd64` — works today on GHA ubuntu runners but could drift.

kalantar:
Pushed eecf078 addressing all three must-fix items plus two recommended improvements.

**kubectl checksum verification** — downloads `kubectl.sha256` and verifies with `sha256sum --check` in the same RUN layer before installing. Closes #142 (was filed as follow-up but now done here).

**Expanded smoke test** — now three checks: `deploy.py --help` (argparse works), `kubectl version --client` (binary is valid), and a Python import of all 14 `pipeline.lib` modules (catches missing files since deploy.py lazy-imports everything).

**`.dockerignore` additions** — added `go.work`. Checked for `tests/` and `tmp/` at HEAD — neither exists, so no entry needed.

**`.dockerignore` in path triggers** — added so changes to build context exclusions trigger a rebuild.

**PyYAML pinned** — `"PyYAML>=6.0,<7"` matching `requirements.txt`.

Skipped two non-blocking items: `docker push --all-tags` is fine in a clean CI runner (no stale tags), and kubectl platform hardcoding is correct for GHA ubuntu runners with no multi-platform build configured.

kalantar:
## Review: PR #141 — Build and publish sim2real-orchestrator container image

**Verdict: READY TO MERGE**

All issue #139 requirements are fully addressed. The Dockerfile is minimal and correct, the workflow follows best practices (build/smoke-test/push separation, layer caching, metadata-driven tagging), and the `.dockerignore` keeps the build context clean. No silent failures — all error paths propagate correctly through `&&` chaining, Docker build steps, and CI step sequencing.

1. **Consider adding a `pull_request` trigger** — Currently the build only runs after merge to main. Since the workflow already uses `push: false` / `load: true` for the build step, adding a PR trigger with `if: github.event_name == 'push'` on the Push step would validate Dockerfile changes before merge with minimal effort. The PR description acknowledges this as a future option.

2. **kubectl operations claim understates the API surface** — The PR description says "The orchestrator only uses version-tolerant operations (`apply`, `get`, `delete`)" but `deploy.py` actually uses eight kubectl subcommands: `apply`, `cp`, `delete`, `exec`, `get`, `patch`, `run`, and `wait`. Consider softening to "common kubectl operations that have been stable across versions" rather than an incomplete enumeration.

3. **PR description omits the smoke test** — The summary bullet for the workflow says it "builds and pushes" but doesn't mention the smoke test gate between build and push (entrypoint, kubectl binary, 13 library module imports). Worth noting since it's a meaningful quality gate.

4. **Minor hardening: partial push failure** — `docker push --all-tags` pushes both SHA and `latest` tags atomically from Docker's perspective, but a network interruption could leave the registry with one tag but not the other. Splitting into separate push steps would give clearer error attribution. Low priority.

### What looks good

- SHA-256 verification of the kubectl binary download
- `ca-certificates` correctly retained after curl purge (needed for kubectl TLS)
- Only PyYAML installed — verified that `deploy.py` and all `pipeline/lib/` modules only import `yaml` as a third-party dep
- Smoke test import list matches all 13 `pipeline/lib` modules exactly
- `latest` tag correctly scoped to default branch via `enable={{is_default_branch}}`
- 568 existing tests unaffected (no Python code changes)

PR #143, opened 2026-05-14
Add deploy.py run --remote for in-cluster orchestration (#127)

### LINKED ISSUE #127: Remote execution: deploy.py run --remote submits orchestrator as K8s Job

kalantar:
## Summary

- Adds `--remote` flag to `deploy.py run` that submits the orchestrator as an in-cluster Kubernetes Job instead of running the polling loop locally
- The launcher builds the EPP image locally, packs workspace files into a ConfigMap, applies a Job that runs the same `deploy.py run` code with `--skip-build-epp`, and waits for the pod to reach Running
- Updates RBAC (`sim2real-pipeline-runner` Role) to add `delete`/`patch` for PipelineRuns and ConfigMap CRUD for the orchestrator
- Adds `pipeline/lib/remote.py` module with pure functions for ConfigMap and Job YAML generation

Closes #127

## Changes

| File | What |
|------|------|
| `pipeline/lib/remote.py` | New module: `build_run_inputs_configmap` and `build_orchestrator_job` |
| `pipeline/deploy.py` | `_cmd_run_remote`, `_wait_for_job_pod`, `_check_existing_job`, `_collect_run_flags`, `--remote` flag, main() routing |
| `pipeline/tests/test_remote.py` | 11 tests for remote module |
| `pipeline/tests/test_deploy_remote.py` | 19 tests for launcher, pod wait, flag collection, main() routing |
| `tektonc-data-collection` | Submodule bump: RBAC adds PipelineRun delete/patch + ConfigMap rules |
| `pipeline/completions.{bash,zsh}` | `--remote` and `--max-backoff` added to `run` completions |
| `pipeline/README.md` | `--remote` flag documented with "Remote mode" paragraph |

## Reviewer attention

- **ConfigMap key encoding**: cluster/ path separator is `--` (ConfigMap keys can't contain `/`). The in-cluster orchestrator mounts the ConfigMap at `/data/workspace/runs/<run-name>/` so the existing directory-based code sees the same layout. The `--` encoding is only visible in the ConfigMap data keys, not in the mounted filesystem.
- **`orchestrator_image` in setup_config**: The launcher reads `setup_config.get("orchestrator_image")` to know which image to use for the Job. This field needs to be written by `setup.py` (not in scope here — tracked separately).
- **`_wait_for_job_pod` timeout**: 120s default for the pod to reach Running. Fails fast on ImagePullBackOff, ErrImagePull, CrashLoopBackOff, CreateContainerConfigError, InvalidImageName.
- **RBAC**: The submodule pointer is bumped. The `sim2real-pipeline-runner` Role now has `delete`/`patch` on PipelineRuns (needed for orchestrator auto-cleanup) and ConfigMap CRUD (needed for progress writes).

## Test plan

- [x] `ruff check pipeline/ --select F` passes
- [x] `python -m pytest pipeline/ -v` — 614 tests pass, zero regressions
- [x] `deploy.py run --help` shows `--remote` flag
- [ ] Manual: `deploy.py run --remote` on a cluster with the orchestrator image available

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,6 +71,7 @@ python pipeline/run.py     --experiment-root ../admission-control switch <run-na
 | `pod_pending.py` | Classifies pod scheduling failures as recoverable or non-recoverable |
 | `backoff.py` | Exponential backoff controller for GPU scarcity in `deploy.py run` |
 | `run_manager.py` | `list_runs`, `inspect_run`, `switch_run` logic |
+| `remote.py` | ConfigMap and Job generation for `deploy.py run --remote` |
 
 ## Workspace Artifacts
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -148,6 +148,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 | Flag | Default | Description |
 |------|---------|-------------|
+| `--remote` | — | Submit orchestrator as in-cluster Job instead of running locally |
 | `--only PAIR` | — | Scope execution to one specific pair key (`wl-` prefix optional) |
 | `--workload NAME` | — | Scope execution to pairs matching this workload |
 | `--package NAME` | — | Scope execution to pairs matching this package |
@@ -168,6 +169,8 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 **Pair statuses:** `pending` → `running` → `done`. Failure paths: `running` → `failed` (hard failure or non-recoverable pending), `running` → `timed-out` (4h timeout exceeded), `running` → `pending` (recoverable early reclaim, repeats up to `--max-pending-stalls` times) → `stalled`.
 
 **Auto-cleanup** — when a PipelineRun succeeds, the orchestrator deletes the PipelineRun CR from the cluster. Failed PipelineRuns are left in place for debugging (`kubectl describe`, pod logs). Use `reset` to remove them when done.
+
+**Remote mode** — `deploy.py run --remote` submits the orchestrator as a Kubernetes Job (`sim2real-orchestrator`) instead of running locally. The launcher builds the EPP image locally, packs workspace files into a ConfigMap, applies the Job, and waits for the pod to reach Running. Use `stop` to cancel, `status --remote` to check progress, and `collect` to pull results after completion. Requires `orchestrator_image` in `setup_config.json`.
 
 **`deploy.py status`** — prints the current state of all pairs. By default reads from `workspace/runs/<run>/progress.json`. When the orchestrator runs remotely (in-cluster), use `--remote` to read from the `sim2real-progress` ConfigMap instead. The ConfigMap is also used automatically when the local `progress.json` is absent.
 
--- a/pipeline/completions.bash
+++ b/pipeline/completions.bash
@@ -73,7 +73,7 @@ _sim2real_deploy() {
     if [[ "$cur" == -* ]]; then
         case "$subcmd" in
             run)
-                COMPREPLY=($(compgen -W "--only --workload --package --status --force --skip-build-epp --max-retries --poll-interval --gpu-resource-type --default-gpu-cost --pending-threshold --max-pending-stalls" -- "$cur"))
+                COMPREPLY=($(compgen -W "--remote --only --workload --package --status --force --skip-build-epp --max-retries --poll-interval --gpu-resource-type --default-gpu-cost --pending-threshold --max-pending-stalls --max-backoff" -- "$cur"))
                 ;;
             status)
                 COMPREPLY=($(compgen -W "--only --workload --package --status --remote" -- "$cur"))
--- a/pipeline/completions.zsh
+++ b/pipeline/completions.zsh
@@ -71,6 +71,7 @@ _sim2real_deploy() {
             case "${line[1]}" in
                 run)
                     _arguments \
+                        '--remote[Submit as in-cluster Job]' \
                         '--only[Scope to one pair key]:pair key:_deploy_py_pair_keys' \
                         '--workload[Scope to workload]:workload:_deploy_py_workloads' \
                         '--package[Scope to package]:package:_deploy_py_packages' \
@@ -82,7 +83,8 @@ _sim2real_deploy() {
                         '--gpu-resource-type[GPU resource type]:type:' \
                         '--default-gpu-cost[GPU cost per pod]:cost:' \
                         '--pending-threshold[Pending timeout in seconds]:seconds:' \
-                        '--max-pending-stalls[Max stall cycles]:count:'
+                        '--max-pending-stalls[Max stall cycles]:count:' \
+                        '--max-backoff[Max backoff interval]:seconds:'
                     ;;
                 status)
                     _arguments \
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1743,6 +1743,153 @@ def _cmd_stop(namespace: str) -> None:
     ok(f"Stopped {JOB_NAME} in {namespace}")
 
 
+# ── Remote run ──────────────────────────────────────────────────────────────
+
+_FAIL_FAST_REASONS = {
+    "ImagePullBackOff",
+    "ErrImagePull",
+    "CrashLoopBackOff",
+    "CreateContainerConfigError",
+    "InvalidImageName",
+}
+
+
+def _collect_run_flags(args) -> list[str]:
+    """Collect run subcommand flags to forward to the in-cluster Job."""
+    flags: list[str] = []
+    for name in ("only", "workload", "package", "status"):
+        val = getattr(args, name, None)
+        if val is not None:
+            flags.extend([f"--{name}", str(val)])
+    if getattr(args, "force", False):
+        flags.append("--force")
+    _defaults = {
+        "max_retries": 2,
+        "poll_interval": 30,
+        "gpu_resource_type": None,
+        "default_gpu_cost": 1,
+        "pending_threshold": 600,
+        "max_pending_stalls": 10,
+        "max_backoff": 600,
+    }
+    for attr, default in _defaults.items():
+        val = getattr(args, attr, default)
+        if val != default:
+            flag = f"--{attr.replace('_', '-')}"
+            flags.extend([flag, str(val)])
+    return flags
+
+
+def _check_existing_job(namespace: str) -> "str | None":
+    """Check whether the orchestrator Job already exists.
+
+    Returns "active" if the Job has active pods, "completed" if it exists
+    but is not active, or None if the Job doesn't exist.
+    """
+    result = run(["kubectl", "get", "job", JOB_NAME, "-n", namespace,
+                   "-o", "json"], check=False, capture=True)
+    if result.returncode != 0:
+        stderr = (result.stderr or "").strip()
+        if "(NotFound)" in stderr:
+            return None
+        err(f"Failed to check for orchestrator Job: {stderr}")
+        sys.exit(1)
+    data = json.loads(result.stdout)
+    if data.get("status", {}).get("active", 0) > 0:
+        return "active"
+    return "completed"
+
+
+def _wait_for_job_pod(namespace: str, *, timeout: int = 120, poll: int = 5) -> None:
+    """Poll until the orchestrator pod reaches Running or Succeeded."""
+    deadline = time.time() + timeout
+    while True:
+        result = run(
+            ["kubectl", "get", "pods",
+             "-l", f"job-name={JOB_NAME}",
+             "-n", namespace, "-o", "json"],
+            check=False, capture=True,
+        )
+        if result.returncode == 0:
+            data = json.loads(result.stdout)
+            for pod in data.get("items", []):
+                phase = pod.get("status", {}).get("phase", "")
+                if phase in ("Running", "Succeeded"):
+                    return
+                for cs in pod.get("status", {}).get("containerStatuses", []):
+                    waiting = cs.get("state", {}).get("waiting", {})
+                    reason = waiting.get("reason", "")
+                    if reason in _FAIL_FAST_REASONS:
+                        err(f"Pod failed: {reason} — {waiting.get('message', '')}")
+                        sys.exit(1)
+        if time.time() >= deadline:
+            err(f"Timed out waiting for {JOB_NAME} pod in {namespace}")
+            sys.exit(1)
+        time.sleep(poll)
+
+
+def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
+    """Submit the orchestrator as an in-cluster Job."""
+    from pipeline.lib.remote import (
+        build_run_inputs_configmap, build_orchestrator_job, JOB_NAME as _JOB,
+    )
+
+    namespaces = setup_config.get("namespaces") or [setup_config.get("namespace", "")]
+    namespace = namespaces[0] if namespaces else ""
+    if not namespace:
+        err("No namespaces configured. Run setup.py with --namespaces.")
+        sys.exit(1)
+
+    orchestrator_image = setup_config.get("orchestrator_image")
+    if not orchestrator_image:
+        err("orchestrator_image not set in setup_config.json. Re-run setup.py with --orchestrator-image.")
+        sys.exit(1)
+
+    status = _check_existing_job(namespace)
+    if status == "active":
+        err(f"Orchestrator Job already running in {namespace}. Use 'deploy.py stop' first.")
+        sys.exit(1)
+    elif status == "completed":
+        info(f"Deleting completed orchestrator Job in {namespace}")
+        run(["kubectl", "delete", "job", JOB_NAME, "-n", namespace], check=False, capture=True)
+
+    epp_action = _resolve_epp_action(run_dir, args.skip_build_epp)
+    if epp_action.startswith("error:"):
+        err(epp_action[len("error:"):]); sys.exit(1)
+    elif epp_action == "skip":
+        info("No component_image in run metadata — skipping EPP build")
+    else:
+        _build_epp_image(run_dir, run_dir.name, namespace)
+
+    workspace_dir = EXPERIMENT_ROOT / "workspace"
+    run_name = run_dir.name
+
+    cm = build_run_inputs_configmap(
+        run_dir=run_dir, workspace_dir=workspace_dir,
+        namespace=namespace, run_name=run_name,
+    )
+    info("Applying run-inputs ConfigMap")
+    subprocess.run(
+        ["kubectl", "apply", "-f", "-"],
+        input=json.dumps(cm), text=True, check=True, capture_output=True,
+    )
+
+    run_flags = _collect_run_flags(args)
+    job = build_orchestrator_job(
+        namespace=namespace, image=orchestrator_image,
+        run_name=run_name, run_flags=run_flags,
+    )
+    info("Applying orchestrator Job")
+    subprocess.run(
+        ["kubectl", "apply", "-f", "-"],
+        input=json.dumps(job), text=True, check=True, capture_output=True,
+    )
+
+    _wait_for_job_pod(namespace)
+    ok("Orchestrator pod is running")
+    info(f"Tail logs: kubectl logs -f job/{JOB_NAME} -n {namespace}")
+
+
 # ── CLI ──────────────────────────────────────────────────────────────────────
 
 def build_parser() -> argparse.ArgumentParser:
@@ -1792,6 +1939,8 @@ Examples:
                            help="Read progress from cluster ConfigMap instead of local file")
 
     run_p = sub.add_parser("run", help="Orchestrate parallel pool execution")
+    run_p.add_argument("--remote", action="store_true", default=False,
+                       help="Submit orchestrator as in-cluster Job instead of running locally")
     run_p.add_argument("--skip-build-epp", action="store_true", dest="skip_build_epp",
                        help="Skip EPP image build")
     run_p.add_argument("--only",         metavar="PAIR",  help="Scope execution to one specific pair key (wl- prefix optional)")
@@ -1882,7 +2031,10 @@ def main():
         sys.exit(1)
 
     if cmd == "run":
-        _cmd_run(args, run_dir, setup_config)
+        if getattr(args, "remote", False):
+            _cmd_run_remote(args, run_dir, setup_config)
+        else:
+            _cmd_run(args, run_dir, setup_config)
     elif cmd == "status":
         progress_path = run_dir / "progress.json"
         _cmd_status(args, progress_path, setup_config=setup_config)
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1760,10 +1760,10 @@ def _collect_run_flags(args) -> list[str]:
     """Collect run subcommand flags to forward to the in-cluster Job."""
     flags: list[str] = []
     for name in ("only", "workload", "package", "status"):
-        val = getattr(args, name, None)
+        val = getattr(args, name)
         if val is not None:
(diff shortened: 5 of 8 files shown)


kalantar:
## Verdict: NEEDS CHANGES

### Blocking: ConfigMap volume mount produces wrong filesystem layout

The single ConfigMap is mounted at `/data/workspace/runs/<run-name>/`, making all keys flat files in that directory. This breaks two assumptions the in-cluster orchestrator relies on:

1. **`setup_config.json` is at the wrong depth.** `_load_setup_config()` looks for `EXPERIMENT_ROOT / "workspace" / "setup_config.json"` (i.e., `/data/workspace/setup_config.json`), but the mount places it at `/data/workspace/runs/<run-name>/setup_config.json`. The orchestrator will fall back to an empty config dict, causing downstream failures.

2. **Cluster YAML files are flat, not in a subdirectory.** The `cluster--baseline.yaml` key encoding becomes a literal filename — Kubernetes does not interpret `--` as a path separator. The orchestrator expects `run_dir / "cluster" / "baseline.yaml"` but gets `run_dir / "cluster--baseline.yaml"`. No PipelineRun pairs will be discovered.

The PR description states "The `--` encoding is only visible in the ConfigMap data keys, not in the mounted filesystem" — this is incorrect for a ConfigMap volume without an `items` spec.

**Fix:** Use `items` in the ConfigMap volume spec to map each key to its correct relative path, and mount at `/data/workspace/` instead of `/data/workspace/runs/<run-name>/`:

```python
"items": [
    {"key": "setup_config.json", "path": "setup_config.json"},
    {"key": "run_metadata.json", "path": f"runs/{run_name}/run_metadata.json"},
    {"key": "cluster--baseline.yaml", "path": f"runs/{run_name}/cluster/baseline.yaml"},
    # ... etc for each cluster YAML
]
```

### Error handling issues

- **Completed Job deletion is fire-and-forget** (`deploy.py:~1854`): `run([...], check=False, capture=True)` with no return code check. If deletion fails, the subsequent `kubectl apply` will fail with a confusing "already exists" error. Compare with `_cmd_stop` which checks the return code.

- **`kubectl apply` produces raw tracebacks** (`deploy.py:~1872, ~1883`): Uses `subprocess.run(..., check=True, capture_output=True)`, which on failure raises `CalledProcessError` with stderr buried inside the exception. Inconsistent with every other kubectl call in the file which uses `check=False` + `err()` + `sys.exit(1)`.

- **`_wait_for_job_pod` silently swallows kubectl failures**: When `kubectl get pods` returns non-zero, the iteration is silently skipped. If kubectl is consistently failing (expired auth, deleted namespace), the user waits the full 120s timeout before getting a generic "Timed out" message with no indication of the real problem.

- **Pod `Failed` phase not detected**: The wait loop checks for `Running`/`Succeeded` and specific waiting-state reasons, but not `phase == "Failed"` or terminated container states. A pod that fails immediately (OOM, bad entrypoint) will spin until timeout. Consider also adding `RunContainerError` and `ContainerCannotRun` to `_FAIL_FAST_REASONS`.

### Smaller items

- **`JOB_NAME` duplicated** in both `deploy.py` and `pipeline/lib/remote.py`. Since `deploy.py` already imports from `remote`, the constant should live in one place.
- **`build_run_inputs_configmap` has an unused `run_name` parameter.**
- **CLAUDE.md module table** doesn't list the new `remote.py` module.
- **`_wait_for_job_pod` docstring** should mention the fail-fast behavior.

### Test gaps (non-blocking but worth addressing)

- `_wait_for_job_pod` timeout path is untested (the branch where deadline is exceeded)
- `_check_existing_job` non-NotFound kubectl failure path is untested
- Only 1 of 5 `_FAIL_FAST_REASONS` is tested (parametrize over all five would be cheap)
- EPP build path (`_resolve_epp_action` returning `"build"`) is untested in remote mode

### What looks good

- Clean separation between pure resource generation (`pipeline/lib/remote.py`) and orchestration logic (`deploy.py`)
- Thorough precondition checks: namespace, orchestrator_image, existing Job state all validated before doing work with actionable error messages
- `_collect_run_flags` correctly avoids forwarding `--remote` and `--skip-build-epp`
- 30 tests with good coverage of happy paths, flag forwarding, and main() routing
- README, completions, and help text all updated consistently

kalantar:
Pushed `e08a986` addressing all blocking and smaller items.

**ConfigMap volume mount**: The core issue was that a flat ConfigMap mount puts all keys as files in one directory, which breaks `_load_setup_config()` (wrong depth) and cluster YAML discovery (no `cluster/` subdirectory). Fixed by adding an `items` spec to the ConfigMap volume that maps each key to its correct filesystem path — `setup_config.json` at the workspace root, `run_metadata.json` and cluster YAMLs under `runs/<name>/`. The mount point moved from `/data/workspace/runs/<name>/` to `/data/workspace/`. `build_orchestrator_job` now takes `configmap_data` so it can generate the items list. Three new tests verify the path mapping (`test_configmap_items_*`).

**Error handling**: Completed Job deletion now checks the return code and exits on failure (was fire-and-forget). `kubectl apply` calls for ConfigMap and Job now use `check=False` + `err()` + `sys.exit(1)`, matching every other kubectl call in the file. Two new tests cover these paths.

**`_wait_for_job_pod`**: Now detects `phase == "Failed"` (catches OOM, bad entrypoint). Tracks consecutive kubectl failures and exits after 3 in a row instead of spinning the full 120s timeout. Added `RunContainerError` and `ContainerCannotRun` to `_FAIL_FAST_REASONS`. Two new tests cover the Failed phase and consecutive-failure paths.

**`JOB_NAME`**: Now imported from `pipeline.lib.remote` instead of being duplicated in `deploy.py`.

**CLAUDE.md**: Added `remote.py` to the module table.

Test gap items (timeout path, parametrized fail-fast reasons, EPP build path in remote mode) noted for follow-up — not addressed in this round.

623 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

PR #143 is well-structured with good separation of concerns (pure resource generation in `remote.py`, I/O orchestration in `deploy.py`) and solid test coverage (30 tests). All issue #127 requirements are addressed. One blocking issue must be resolved before merge; several non-blocking items are worth addressing.

### Blocking

**Read-only ConfigMap mount will crash the in-cluster orchestrator on first progress save.** The Job mounts the ConfigMap volume at `/data/workspace`. Kubernetes ConfigMap volumes are read-only — you cannot create files under the mount point. The in-cluster orchestrator calls `LocalProgressStore.save()`, which tries to write `progress.json` under `/data/workspace/runs/<run>/` — this will fail immediately with a read-only filesystem error. The issue text assumes "the local file write is a no-op (writes into the ephemeral container filesystem)" but the write path is under the ConfigMap mount, not outside it.

Options: (A) add an `emptyDir` volume at a writable path, (B) mount ConfigMap elsewhere and copy to an `emptyDir`, or (C) detect in-cluster mode and use a null/no-op local store so only the ConfigMap progress store is used.

### Non-blocking

1. **Unguarded `json.loads` in `_check_existing_job` and `_wait_for_job_pod`** — if kubectl returns rc=0 with malformed JSON (e.g., warning prefix), these crash with an opaque `JSONDecodeError` traceback instead of an actionable message. In the poll loop this is worse since a single glitch kills the launcher. Wrap in try/except with context.

2. **`build_run_inputs_configmap` exceptions not caught in `_cmd_run_remote`** — `FileNotFoundError` from missing workspace files produces a raw traceback. Catch and convert to `err()` with guidance ("run setup.py and prepare.py first").

3. **`subprocess.run` called directly instead of the module's `run()` helper** — the two `kubectl apply` calls in `_cmd_run_remote` bypass the module's wrapper. This is because the wrapper doesn't support `input=`. Either extend the helper or add a note explaining the deviation.

4. **`_collect_run_flags` hardcoded defaults can drift from argparse** — the `_defaults` dict duplicates argparse default values. If someone changes a default in `build_parser()` but forgets `_collect_run_flags`, non-default flags get silently dropped.

5. **`_configmap_items` silently drops unrecognized keys** — any key not matching the three known patterns won't be mounted. Safe today but a maintenance trap if someone adds files to the ConfigMap later. Consider raising on unrecognized keys.

### Test gaps worth filling

- `_check_existing_job` with a non-NotFound kubectl error (connection refused, Forbidden)
- `_wait_for_job_pod` timeout expiration path
- `_cmd_run_remote` with missing namespace in setup_config
- Job apply failure distinct from ConfigMap apply failure
- Pod `Succeeded` phase handling (only `Running` is tested)
- EPP build invocation path in remote mode (all tests mock it to `"skip"`)

kalantar:
Pushed `b4dec04` addressing the blocking and non-blocking items.

**Read-only ConfigMap mount (blocking):** The Job now uses two volumes: a read-only ConfigMap at `/data/config` and a writable `emptyDir` at `/data/workspace`. An initContainer (`copy-inputs`) copies the ConfigMap contents into the workspace before the orchestrator starts. This means `LocalProgressStore.save()` writes `progress.json` to the writable emptyDir without issues. No changes to the orchestrator code itself. Three new tests verify the initContainer, emptyDir volume, and read-only ConfigMap structure.

**Unguarded json.loads:** Both `_check_existing_job` and `_wait_for_job_pod` now catch `JSONDecodeError`. In `_check_existing_job`, it exits with a clear error message. In the poll loop, it logs a warning and retries (a single kubectl glitch shouldn't kill the launcher).

**FileNotFoundError from build_run_inputs_configmap:** Now caught in `_cmd_run_remote` and converted to an `err()` message with guidance ("run setup.py and prepare.py first").

**subprocess.run deviation:** Added a comment above the two `kubectl apply` calls explaining the module's `run()` helper doesn't support stdin input.

**_configmap_items unrecognized keys:** Now emits a `[WARN]` to stderr for keys that don't match known patterns, instead of silently dropping them.

Two items noted as follow-up (not addressed in this round):
- `_collect_run_flags` hardcoded defaults can drift from argparse — maintenance concern, not a bug today
- Test gaps (timeout path, parametrized fail-fast reasons, Succeeded phase, EPP build path)

625 tests pass, lint clean.

kalantar:
## Review: NEEDS CHANGES

Well-structured PR that cleanly separates concerns (pure YAML generation in `lib/remote.py`, all kubectl in `deploy.py`) and covers all issue #127 requirements. CI passes, 30 new tests, documentation is thorough. Three items need attention before merge:

### Must fix

1. **`_wait_for_job_pod` doesn't check `initContainerStatuses`** — The Job uses an initContainer (`copy-inputs`) with the same image. If that image can't be pulled, the failure appears in `initContainerStatuses`, not `containerStatuses`. The fail-fast logic never sees it, so the user waits the full 120s timeout with a useless error message. Fix: iterate over both `initContainerStatuses` and `containerStatuses` in the polling loop.

2. **`_configmap_items` silently drops unrecognized keys** — An unrecognized key gets packed into the ConfigMap data but is never mounted into the pod filesystem. The orchestrator runs with missing files and fails at runtime with no connection to the real cause. This should be a hard error, not a warning — it indicates a programming mistake in the pipeline (producer/consumer mismatch between `build_run_inputs_configmap` and `_configmap_items`).

3. **`_wait_for_job_pod` JSON parse failure resets the consecutive failure counter** — `consecutive_failures = 0` happens before `json.loads`, so repeated invalid-JSON responses (rc=0 but garbage stdout) bypass the 3-strike exit. Move the reset to after the successful parse.

### Should fix (non-blocking but worth addressing)

- **Error message references nonexistent flag**: `"Re-run setup.py with --orchestrator-image"` — that flag doesn't exist yet. Change to something like `"orchestrator_image not set in setup_config.json"`.
- **Missing tests for key paths**: `_wait_for_job_pod` timeout expiry, `Succeeded` phase, Job-apply failure (ConfigMap succeeds but Job fails), `_check_existing_job` with non-NotFound errors (e.g. Forbidden). These are common Kubernetes failure modes.
- **`_collect_run_flags` uses `getattr` with defaults** that silently mask renamed/removed attributes. Consider dropping the default so a missing attr raises `AttributeError` immediately if the parser drifts.

### Design note (no action required)

`--status` filtering in remote mode may be surprising — the in-cluster Job starts with fresh state so `--status failed` would match zero pairs. Worth documenting or warning, but not a blocker.

kalantar:
Pushed `3f6b82d` addressing the three must-fix and two should-fix items.

**initContainerStatuses**: The fail-fast loop in `_wait_for_job_pod` now checks both `initContainerStatuses` and `containerStatuses`. An ImagePullBackOff in the `copy-inputs` initContainer is now caught immediately instead of waiting for timeout. New test covers this path.

**_configmap_items hard error**: Changed from `warn()` to `raise ValueError` on unrecognized ConfigMap keys. Since both the producer (`build_run_inputs_configmap`) and consumer (`_configmap_items`) live in the same module, an unrecognized key is a programming bug, not a runtime condition.

**JSON counter reset**: Moved `consecutive_failures = 0` to after the successful `json.loads` call. Previously, repeated rc=0-but-garbage responses would reset the counter on each iteration and never trigger the 3-strike exit. Now invalid JSON increments the counter alongside kubectl failures.

**Error message**: Removed the reference to nonexistent `--orchestrator-image` flag. Now says "add it before using --remote."

**_collect_run_flags getattr**: Dropped the default argument from `getattr` calls so a missing attribute raises `AttributeError` immediately if the parser drifts, rather than silently masking the mismatch.

Follow-up items (test gaps: timeout, Succeeded phase, Job-apply failure, non-NotFound kubectl error) and design note (--status filtering in remote mode) noted but not addressed in this round.

626 tests pass, lint clean.

kalantar:
## Review: PR #143 — `deploy.py run --remote` for in-cluster orchestration

**Verdict: NEEDS CHANGES** — one blocking issue, a few recommended improvements.

### Requirements coverage

All 10 requirements from issue #127 are addressed: `--remote` flag, Job spec (name/namespace/SA/image), RBAC via submodule bump, ConfigMap packing with workspace-mirroring layout, local EPP build with `--skip-build-epp` in-cluster, `--experiment-root /data` passthrough, existing-Job lifecycle (active=refuse, completed=delete), scoping flag forwarding, pod readiness wait with fail-fast, and local mode unchanged.

### Architecture

Clean separation: `remote.py` is pure data generation (no I/O, no subprocess), `deploy.py` owns orchestration. The initContainer + emptyDir pattern is sound — ConfigMap volumes are read-only, but the orchestrator needs to write `progress.json`. The `_configmap_items` ValueError on unrecognized keys is a nice defensive contract between producer and consumer.

### Blocking issue

**Empty/missing `cluster/` directory produces a silently broken ConfigMap.** `build_run_inputs_configmap` globs `(run_dir / "cluster").glob("*.yaml")` — if the directory is missing or empty, the ConfigMap is created with only `setup_config.json` and `run_metadata.json`, no cluster YAMLs. The Job starts, finds no PipelineRun files, and fails with a confusing in-container error. Add a validation check:

```python
yaml_files = sorted((run_dir / "cluster").glob("*.yaml"))
if not yaml_files:
    raise FileNotFoundError(f"No cluster YAML files in {run_dir / 'cluster'}")
```

### Recommended improvements

1. **Broaden exception catch in `_cmd_run_remote`** — the `try/except FileNotFoundError` around `build_run_inputs_configmap` won't catch `PermissionError`, `UnicodeDecodeError`, or `OSError` from `read_text()`. Consider adding `except OSError as exc:` as a second handler.

2. **Add `_wait_for_job_pod` timeout test** — the timeout path (`sys.exit(1)` with "Timed out" message) is the most likely production failure mode (pod stuck in non-fail-fast Pending) but has no test. Monkeypatch `time.time` to simulate deadline expiry.

3. **Add `_check_existing_job` tests for non-NotFound kubectl errors and invalid JSON** — both paths call `sys.exit(1)` but are untested.

4. **Add test for Job apply failure** (distinct from ConfigMap apply failure) — `test_run_remote_configmap_apply_failure_exits` always fails on the first `subprocess.run` call, never reaching the Job apply path.

5. **Minor: `_wait_for_job_pod` prints "retrying" warning before checking timeout** — move the `warn()` after the deadline check so users don't see "retrying" immediately followed by "Timed out".

### What's done well

- 30 tests with thorough coverage of happy paths and error scenarios (parser, flag collection, Job lifecycle, pod waiting with fail-fast, initContainer validation, main routing)
- `_configmap_items` raises `ValueError` with an actionable message telling the developer exactly what to fix
- The `subprocess.run` bypass comment explains *why* the module's `run()` helper can't be used (no stdin support)
- Documentation is complete: CLAUDE.md module table, README.md remote mode paragraph, shell completions for both bash and zsh
- `consecutive_failures` counter is correctly reset only after successful JSON parsing
- Both `initContainerStatuses` and `containerStatuses` are checked for fail-fast reasons

kalantar:
Pushed `d4c5c00` addressing both items from this round.

**Empty/missing cluster directory**: `build_run_inputs_configmap` now validates that the `cluster/` directory exists and contains at least one YAML file. Previously, an empty or missing `cluster/` directory produced a ConfigMap with no PipelineRun files, causing a confusing failure inside the container. Now it fails early with a clear message pointing the user to `prepare.py`. Two new tests cover the empty-directory and missing-directory cases.

**Broadened exception catch**: `_cmd_run_remote` now catches `OSError` instead of just `FileNotFoundError` around the `build_run_inputs_configmap` call. This covers `PermissionError`, `UnicodeDecodeError` from `read_text()`, and other filesystem errors that would otherwise produce raw tracebacks.

Skipped: "retrying" warning before timeout check — investigated and the code already has the correct order (warn → check deadline → sleep) in the JSON error path.

Follow-up items noted: timeout test, `_check_existing_job` error path tests, Job-apply-failure test (distinct from ConfigMap-apply-failure).

628 tests pass, lint clean.

kalantar:
## READY TO MERGE

All requirements from issue #127 are addressed. The implementation is well-structured, thoroughly tested (30 new tests), and the error handling is solid throughout.

**Highlights:**
- Clean separation between pure resource generation (`remote.py`) and orchestration logic (`deploy.py`)
- Two-volume approach (read-only ConfigMap + writable emptyDir with initContainer copy) follows standard K8s patterns
- Fail-fast on 7 unrecoverable container states, with initContainerStatuses checked alongside containerStatuses
- RBAC, completions, CLAUDE.md, and README all updated

**Minor items worth considering (none are blockers):**

- `_check_existing_job`: if kubectl returns rc=0 but empty stdout, the error message ends with a dangling `: ` — could special-case empty responses for a clearer message.
- `build_orchestrator_job` can raise `ValueError` from `_configmap_items` (on unrecognized keys) that isn't caught in `_cmd_run_remote` — the user would see a raw traceback instead of a formatted `[ERROR]` line. Wrapping in `try/except ValueError` would be consistent with the rest of the file.
- `_check_existing_job` returns `"completed"` for Jobs with no active pods, which includes failed Jobs. The docstring could note this covers both succeeded and failed states to prevent confusion.
- `_FAIL_FAST_REASONS` could add `ErrImageNeverPull` and `CreateContainerError` for completeness, though both are uncommon in this context.

**Test coverage gaps (all low priority):**
- `_check_existing_job` non-NotFound kubectl failure and invalid JSON paths are untested
- `_wait_for_job_pod` timeout path (pod stays Pending without fail-fast reason) is untested
- No test for Job apply failure specifically (the ConfigMap apply failure test covers the same code pattern)
- `_cmd_run_remote` no-namespace and EPP error paths are untested

PR #147, opened 2026-05-15
fix: make deploy collect namespace-aware for multi-slot runs

### LINKED ISSUE #146: bug: deploy collect only pulls from primary namespace, misses data in secondary slots
## Problem

`deploy.py collect` creates its extract pod in the primary namespace only. When a run spans multiple namespace slots (e.g. kalantar-0, kalantar-1, kalantar-2), PipelineRuns for phases assigned to secondary slots write their results to PVCs in those namespaces. Collect never reaches those PVCs, so phases in secondary namespaces are silently skipped or re-attempted against the wrong namespace on every invocation.

## Observed Symptom

On a 3-phase run distributed across 3 namespace slots:
- The phase in kalantar-0 collected correctly and showed `up to date — skipping` on the second collect run.
- The phases in kalantar-1 and kalantar-2 showed no skip messages and appeared to re-collect, but produced no data because the extract pod was in kalantar-0.

## Expected Behavior

Collect should be namespace-aware: for each phase, determine which namespace slot it ran in and create/use an extract pod in that namespace to pull results from the correct PVC.

## Notes

- Each phase is assigned to a slot; the slot-to-namespace mapping is available in `progress.json` or the run plan.
- The incremental mtime-probe logic (`_probe_remote_mtimes`) is also affected — probing the wrong namespace returns empty, causing a spurious full-copy fallback.

kalantar:
## Summary

Closes #146

- **Orchestrator**: writes a `completed_namespace` field to each `progress.json` entry when it transitions to `"done"`, capturing the namespace before clearing `entry["namespace"]`. Two sites updated: `_reconcile_on_resume` (resume path) and the `slots_busy` polling loop (live run path).
- **`_cmd_collect` unscoped path**: groups done entries by `completed_namespace` and dispatches one `_extract_phases_from_pvc` call per distinct namespace instead of always using the primary namespace. Entries missing `completed_namespace` (written by the old orchestrator) emit a warning and are skipped.
- **`_cmd_collect` scoped path** (`--only`/`--workload`): looks up `completed_namespace` per workload from `collectible` done entries, dispatches each workload's extraction to its own namespace. Same warn-and-skip behavior for missing entries.
- **Fallback** (no `progress.json`): unchanged — still uses the primary namespace with a discovery warning, since there is no namespace information available.
- **Backward compat**: runs written before this fix will have `completed_namespace = None` on all done entries. `collect` will warn and skip those pairs. Users need to re-run the affected workloads to collect their results.

## Test plan

- [ ] `test_collect_unscoped_multi_namespace_dispatch` — verifies two distinct namespaces produce two extract calls with correct phases and no workload restriction
- [ ] `test_collect_unscoped_missing_completed_namespace_warns_and_skips` — verifies warning is emitted and no extraction occurs for entries without the field
- [ ] `test_collect_scoped_multi_namespace_dispatch` — verifies `--workload smoke` uses `completed_namespace=ns-0`, not the primary namespace
- [ ] `test_collect_scoped_missing_completed_namespace_warns_and_skips` — same for scoped path
- [ ] `test_reconcile_on_resume_sets_completed_namespace_on_success` — verifies orchestrator captures namespace before clearing it
- [ ] All 23 pre-existing collect tests updated with `completed_namespace` in fixtures, all passing

## Reviewer attention

- **Schema change**: `completed_namespace` is now present in all new progress entries (init value `None`). Progress files from before this fix will not have the field on done entries — this is the "backward compat / warn+skip" case the issue comment called out.
- **Two done-transition sites in `deploy.py`**: `_reconcile_on_resume` (line 1158) and the `slots_busy` loop (line 1417). Both set `completed_namespace = ns` before clearing `namespace = None`. The `failed` transition paths do not set `completed_namespace` — `collect` only processes `done` entries, so this is correct.
- **Warning message**: "re-run the workload with a newer orchestrator to collect results" — avoids directing users to manually edit a generated artifact.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -716,11 +716,26 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         collected: list[str] = []
         failed: list[str] = []
 
+        skip_logs = getattr(args, "skip_logs", False)
         for wl_name in scoped_workloads:
+            wl_ns = next(
+                (progress[k].get("completed_namespace")
+                 for k in collectible
+                 if isinstance(progress[k], dict)
+                 and progress[k].get("workload") == wl_name
+                 and progress[k].get("completed_namespace")),
+                None,
+            )
+            if not wl_ns:
+                warn(f"{wl_name}: completed_namespace missing — skipping "
+                     f"(add completed_namespace to progress.json to collect this pair)")
+                for p in phases_to_collect:
+                    if p not in failed:
+                        failed.append(p)
+                continue
             try:
-                skip_logs = getattr(args, "skip_logs", False)
                 errors = _extract_phases_from_pvc(
-                    phases_to_collect, run_name, namespace, run_dir,
+                    phases_to_collect, run_name, wl_ns, run_dir,
                     skip_logs=skip_logs, workload=wl_name)
             except RuntimeError as e:
                 warn(f"Extractor pod failed for workload {wl_name}: {e}")
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1297,6 +1297,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 "package":  meta["package"],
                 "status":   "pending",
                 "namespace": None,
+                "completed_namespace": None,
                 "retries":  0,
                 "gpu_cost": pair_costs[key],
                 "pending_stalls": 0,
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1097,6 +1097,7 @@ def _reconcile_on_resume(progress: dict, discovered: dict) -> None:
                 if actual == "Succeeded":
                     entry["status"] = "done"
                     entry["pending_since"] = None
+                    entry["completed_namespace"] = ns
                     try:
                         _delete_pipelinerun(pr_name, ns)
                     except Exception as exc:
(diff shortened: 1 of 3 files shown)


kalantar:
## Review: READY TO MERGE

PR #147 correctly addresses all acceptance criteria from issue #146. The fix is sound, well-structured, and thoroughly tested.

### Requirements check

- Orchestrator records `completed_namespace` at both done-transition sites (`_reconcile_on_resume` and `slots_busy` loop) before clearing `namespace` — **met**
- Unscoped collect groups done entries by `completed_namespace` and dispatches one extract call per namespace — **met**
- Scoped collect (`--only`/`--workload`) looks up `completed_namespace` per workload — **met**
- Missing `completed_namespace` entries emit warning and are skipped — **met**
- Backward compat: old progress files handled gracefully — **met**
- Fallback (no progress.json) unchanged — **met**
- Failed transitions correctly do not set `completed_namespace` — **met**

### Test coverage

Six new tests cover multi-namespace dispatch and missing-namespace warnings for both scoped and unscoped paths. All 23 existing collect tests updated with `completed_namespace` in fixtures. The `slots_busy` loop done-transition is covered indirectly by `test_cmd_run_*` integration tests — acknowledged in commit message.

### Non-blocking notes

1. **Scoped path `failed` list interaction**: When a workload is missing `completed_namespace`, all `phases_to_collect` are added to `failed`. If another workload with overlapping phase names succeeds in the same `collect` invocation, its phases appear in both `collected` and `failed`. This is a reporting/summary edge case (not data loss — the extraction still happens correctly). Only triggers in the narrow transition window where some workloads have the field and some don't.

2. **Unscoped path `workload` parameter**: The unscoped path doesn't pass `workload` to `_extract_phases_from_pvc`, so each namespace's extract pod copies all workload directories under each phase rather than scoping to only the workloads that ran there. This matches pre-existing unscoped behavior and is likely correct since each namespace's PVC should only contain data from workloads that ran in that namespace.

### CI

One pre-existing failure (`test_cmd_reset_creates_composite_store_when_namespace_available` — kubectl unavailable in CI). Unrelated to this PR. 702 passed.

PR #153, opened 2026-05-15
refactor: eliminate local progress.json — ConfigMap as sole progress store

### LINKED ISSUE #148: Eliminate local progress.json — use ConfigMap as sole progress store

kalantar:
Closes #148

## Summary

- **Delete `LocalProgressStore` and `CompositeProgressStore`** from `pipeline/lib/progress.py` — `ConfigMapProgressStore` is now the only implementation
- **All 5 deploy.py subcommands** (`status`, `collect`, `run`, `reset`, `wipe`) use `ConfigMapProgressStore` directly — no composite wrapping, no local fallback
- **All subcommands require a configured namespace** — exit with error if `_configmap_namespace()` returns empty (previously silently fell back to local file)
- **`monitor.py`** switched from `LocalProgressStore` to `ConfigMapProgressStore`
- **Net -280 lines** across 11 files

## Motivation

Every time a new feature was introduced (e.g. `completed_namespace` in #147, `--remote` in #143), the local `progress.json` path was inadvertently made primary or left inconsistent. This happened because `CompositeProgressStore` made both paths look equivalent to callers, and the "no namespace configured" fallback silently used the local file. A Kubernetes-only tool should not have a non-Kubernetes fallback — it masks configuration errors and creates a recurring class of bugs.

## Reviewer attention

- **`_cmd_reset` signature changed**: `progress_path: Path` → `run_dir: Path`. The dispatch site in `main()` is updated.
- **No namespace = hard error**: All subcommands now call `sys.exit(1)` if no namespace is configured. This is intentional — there's no valid deploy.py use case without a cluster.
- **Test migration**: All tests that wrote `progress.json` as fixture data now monkeypatch `ConfigMapProgressStore.load/save`. Tests for local-fallback behavior are deleted (the behavior no longer exists).

## Test plan

- [x] 693 tests pass (`python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/ -v`)
- [x] Lint clean (`ruff check pipeline/ .claude/skills/ --select F`)
- [x] Zero remaining `progress.json` / `LocalProgressStore` / `CompositeProgressStore` references in `pipeline/` Python files

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,8 +91,7 @@ All artifacts live under `<experiment-root>/workspace/` (gitignored). When no `-
 | `runs/<run>/cluster/pipelinerun-*.yaml` | `prepare.py` Phase 4 | `deploy.py run` |
 | `runs/<run>/run_summary.md` | `prepare.py` Phase 5 | human review |
 | `runs/<run>/results/{phase}/` | `deploy.py collect` | `/sim2real-analyze` skill, `deploy.py wipe` |
-| ConfigMap `sim2real-progress` | `deploy.py run`, `deploy.py reset` | All `deploy.py` subcommands (primary) |
-| `runs/<run>/progress.json` | `deploy.py run`, `deploy.py wipe` | All `deploy.py` subcommands (fallback) |
+| ConfigMap `sim2real-progress` | `deploy.py run`, `deploy.py reset` | All `deploy.py` subcommands |
 | `runs/<run>/plans/<phase>/<workload>/` | `deploy.py run` | workload tasks |
 | `context/{scenario}/{hash}.md` | `prepare.py` Phase 2 | `prepare.py` Phase 2 (cache) |
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -144,7 +144,7 @@ python pipeline/deploy.py wipe  [flags]     # delete local results and reset pai
 python pipeline/deploy.py pairs   [flags]   # list available pair keys, workloads, and packages
 ```
 
-**`deploy.py run`** — assigns `(workload, package)` pairs to free namespace slots, polls for completion, and retries pairs that time out. Reads progress from the ConfigMap (primary) or local file (fallback) to resume interrupted runs. Use `deploy.py collect` to pull results off-cluster after runs complete.
+**`deploy.py run`** — assigns `(workload, package)` pairs to free namespace slots, polls for completion, and retries pairs that time out. Reads progress from the `sim2real-progress` ConfigMap to resume interrupted runs. Requires a configured namespace. Use `deploy.py collect` to pull results off-cluster after runs complete.
 
 | Flag | Default | Description |
 |------|---------|-------------|
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -172,7 +172,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 **Remote mode** — `deploy.py run --remote` submits the orchestrator as a Kubernetes Job (`sim2real-orchestrator`) instead of running locally. The launcher builds the EPP image locally, packs workspace files into a ConfigMap, applies the Job, and waits for the pod to reach Running. Use `stop` to cancel, `status` to check progress, and `collect` to pull results after completion. Requires `orchestrator_image` in `setup_config.json`.
 
-**`deploy.py status`** — prints the current state of all pairs. Always reads from the `sim2real-progress` ConfigMap first (primary), falling back to the local `progress.json` if the ConfigMap is empty or the namespace is not configured.
+**`deploy.py status`** — prints the current state of all pairs. Reads from the `sim2real-progress` ConfigMap. Requires a configured namespace.
 
 | Flag | Description |
 |------|-------------|
@@ -192,7 +192,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). These pair-level flags compose with `--package` as AND: `--workload X --package baseline` pulls workload X from the baseline phase only. Requires progress data to resolve pairs.
 
-**`deploy.py stop`** — deletes the `sim2real-orchestrator` Kubernetes Job (with cascading pod deletion) in the primary namespace. Only meaningful when the orchestrator runs as an in-cluster Job. Does not touch `progress.json` — pair state is left as-is. If no remote orchestrator Job exists, prints a message and returns. Use `reset` separately to clear failed/stalled pair state.
+**`deploy.py stop`** — deletes the `sim2real-orchestrator` Kubernetes Job (with cascading pod deletion) in the primary namespace. Only meaningful when the orchestrator runs as an in-cluster Job. Pair state is left as-is. If no remote orchestrator Job exists, prints a message and returns. Use `reset` separately to clear failed/stalled pair state.
 
 **`deploy.py reset`** — removes cluster resources (PipelineRuns, Helm releases) for all non-pending pairs. Failed/running/timed-out pairs are reset to `pending` so they can be re-dispatched. Done pairs stay `done` — only their PipelineRun is deleted to free cluster resources.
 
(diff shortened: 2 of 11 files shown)


kalantar:
## Review: PR #153 — Eliminate local progress.json

**Verdict: NEEDS CHANGES**

Clean, well-scoped refactoring that fully addresses all three requirements from issue #148. The code deletion is thorough (net -280 lines), test migration is consistent, and the upstream/downstream pipeline contract is honored. Two issues should be fixed before merge.

### Required Changes

**1. `_cmd_status` has zero error handling around `store.load()`**

`_cmd_status` calls `store.load()` with no try/except. If the cluster is unreachable (RuntimeError), the ConfigMap has corrupt JSON (ValueError), or kubectl is not on PATH (FileNotFoundError), the user gets a raw traceback. This is a regression — the old code had explicit error handling here. `_cmd_collect` in this same PR correctly wraps `store.load()` in try/except, so this is an inconsistency.

The cleanest fix is to catch `OSError` inside `ConfigMapProgressStore.load()` and `.save()` and re-raise as `RuntimeError`. This way all callers that already catch `RuntimeError` get the fix for free. Then add a try/except around `store.load()` in `_cmd_status` (and optionally `_cmd_run`, `_cmd_reset`, `_cmd_wipe`) matching the pattern in `_cmd_collect`:

```python
# In ConfigMapProgressStore.load() and .save(), wrap subprocess.run():
try:
    result = subprocess.run(["kubectl", ...], ...)
except OSError as exc:
    raise RuntimeError(f"kubectl not available: {exc}") from exc

# In _cmd_status:
try:
    progress = store.load()
except (ValueError, RuntimeError) as exc:
    err(f"Failed to load progress: {exc}")
    progress = {}
```

**2. README still describes old fallback behavior for `deploy.py run`**

`pipeline/README.md` line ~147 still reads: *"Reads progress from the ConfigMap (primary) or local file (fallback) to resume interrupted runs."* This is factually wrong after this PR. Should be updated to match the other subcommand descriptions (e.g., "Reads progress from the sim2real-progress ConfigMap. Requires a configured namespace.").

- `_cmd_run` error message says `"No namespace configured."` without the `"Run setup.py first."` hint that the other four subcommands include.
- CLAUDE.md and README artifact tables list ConfigMap writers as `deploy.py run, deploy.py reset` — but `deploy.py wipe` also writes to the ConfigMap (resets statuses to pending). Consider adding it.
- `_mock_cm` in `test_deploy_collect.py` returns the same dict object (no deep copy), unlike the reset/wipe variant which deep-copies via `json.loads(json.dumps(data))`. Could cause test pollution if any code path mutates the returned dict.
- A few test names still reference "no_local_file" (`test_status_reads_configmap_when_no_local_file`, `test_status_no_configmap_no_local_reports_no_run`) — vestigial after this PR removes the local file concept.
- monitor.py duplicates namespace resolution inline instead of reusing `_configmap_namespace()` from deploy.py.

kalantar:
Pushed `39e8da4` addressing the review.

**Fixed:**

- `ConfigMapProgressStore.load()` and `.save()` now catch `OSError` from `subprocess.run` and re-raise as `RuntimeError`, so callers that catch `RuntimeError` handle missing kubectl for free.
- `_cmd_status` now wraps `store.load()` in `try/except (ValueError, RuntimeError)`, matching the pattern already used in `_cmd_collect`.
- README description for `deploy.py run` updated — no longer says "local file fallback".
- `_cmd_run` error message now includes "Run setup.py first." hint, consistent with the other four subcommands.
- `_mock_cm` in `test_deploy_collect.py` now deep-copies via `json.loads(json.dumps(data))` to prevent test pollution, matching the reset/wipe variants.
- Renamed `test_status_reads_configmap_when_no_local_file` → `test_status_reads_from_configmap` and `test_status_no_configmap_no_local_reports_no_run` → `test_status_empty_configmap_reports_no_run`.
- Added two new tests for the OSError wrapping (`test_configmap_load_missing_kubectl_raises`, `test_configmap_save_missing_kubectl_raises`).

695 tests pass, lint clean.

**Follow-ups filed:**

- #155 — add `deploy.py wipe` to ConfigMap writers in artifact tables
- #156 — extract `_configmap_namespace` into shared module for `monitor.py`

kalantar:
## Review: PR #153 — Eliminate local progress.json

**Verdict: READY TO MERGE**

All three requirements from issue #148 are fully addressed. The changes are complete and show no signs of lost work from the worktree merge.

### What's solid

- All 5 deploy.py subcommands (`status`, `collect`, `run`, `reset`, `wipe`) and `monitor.py` now use `ConfigMapProgressStore` directly — no local fallback, no composite wrapping.
- `LocalProgressStore` and `CompositeProgressStore` are cleanly deleted (-280 net lines).
- `_cmd_reset` signature change (`progress_path` → `run_dir`) is properly propagated to its only caller in `main()`.
- OSError wrapping in `ConfigMapProgressStore.load()`/`.save()` ensures callers only need to catch `(ValueError, RuntimeError)`.
- Test migration is thorough: all fixture writes replaced with `_mock_cm()` monkeypatching, local-fallback tests appropriately deleted, new OSError wrapping tests added.
- Documentation (CLAUDE.md, README.md) consistently updated — no stale references to progress.json or local fallback.

### Minor observations (non-blocking)

1. **Style inconsistency in `_cmd_run`**: `err(...); sys.exit(1)` on one line (semicolon-joined), while all other subcommands use two lines.
2. **`_cmd_reset` and `_cmd_wipe` don't catch exceptions from `store.load()`**: If the cluster is unreachable, the user sees a traceback rather than a clean error message. Logically correct (can't reset/wipe without cluster access) but rougher UX than status/collect.
3. **Missing "no namespace" tests for 4 of 5 subcommands**: Only `test_status_no_namespace_exits_with_error` exists. Adding equivalents for collect, run, reset, wipe would lock in the behavioral contract.
4. **`_mock_cm` in `test_deploy_run.py` doesn't deep-copy** (uses `lambda self: data` directly), unlike the other test files which use `json.loads(json.dumps(data))`. Low risk since status doesn't mutate progress, but inconsistent.
5. **Test name `test_status_missing_progress_file`** still says "file" despite testing empty ConfigMap behavior. Cosmetic.

None of these block the merge.

kalantar:
### Addendum: deeper error-handling observations

The silent-failure analysis surfaced two points worth discussing — not blocking, but they touch the PR's core thesis ("local fallback masks configuration errors"):

**1. `_cmd_status` and `_cmd_collect` still mask ConfigMap failures.** Both catch `(ValueError, RuntimeError)` from `store.load()` and continue (status shows "0 pairs"; collect falls back to phase discovery). If ConfigMap is the sole source of truth, an unreachable ConfigMap is arguably a hard error, not a "no data" state. The current behavior makes cluster connectivity problems look like "no run started." Consider whether these should `sys.exit(1)` like `_cmd_reset`/`_cmd_wipe` do (via the bare `store.load()` that lets exceptions propagate).

**2. `store.save()` calls in the `_cmd_run` loop have no error handling.** This is pre-existing, but removing the local-file write from `CompositeProgressStore.save()` removes a safety net — a transient cluster blip during a multi-hour run now crashes the orchestrator with no local state to recover from. Worth noting for a follow-up, though fixing it correctly (retry with backoff) is non-trivial and out of scope for this PR.

Neither blocks the merge — the PR is a net improvement over the status quo. These are observations for the next iteration.

kalantar:
### Addendum 2: artifact table nit

Both CLAUDE.md and README.md list the ConfigMap as "Written by: `deploy.py run`, `deploy.py reset`" — but `deploy.py wipe` also calls `store.save()`. This predates the PR but since you're touching that exact row, it's a good opportunity to add `deploy.py wipe` to the "Written by" column.

PR #161, opened 2026-05-15
feat: add --skip-teardown flag to deploy.py run

### LINKED ISSUE #160: feat: add --skip-teardown flag to deploy.py run for post-failure debugging
## Summary

When a PipelineRun fails (e.g. smoketest failure during standup), the `finally` block immediately tears down the namespace, making it impossible to investigate the cluster state. There is currently no way to keep the namespace intact for debugging.

## Proposed Solution

Add a `--skip-teardown` flag to `deploy.py run` that suppresses teardown at apply time, leaving the namespace intact after a failure for interactive investigation.

### Implementation

Three changes:

1. **`pipeline/pipeline.yaml`** — Add a `skip_teardown` pipeline param (default `"false"`) and a `when` clause on the `llmdbenchmark-teardown` finally task that skips it when the param is `"true"`.

2. **`pipeline/deploy.py`** — Add `--skip-teardown` to the `run` subcommand argument parser.

3. **`pipeline/deploy.py` apply-time injection** — When `--skip-teardown` is set, inject/override the `skip_teardown` param in the PipelineRun YAML before applying. This follows the existing pattern used for namespace override (lines 1662–1668).

### Usage

```bash
# Normal — teardown runs as usual
python pipeline/deploy.py run --experiment-root ../my-experiment

# Debug — teardown skipped, namespace left intact
python pipeline/deploy.py run --skip-teardown --experiment-root ../my-experiment
```

The flag flows through `--remote` mode via `_collect_run_flags()`, which already forwards all run flags to the in-cluster orchestrator.

## Context

Discovered while investigating a smoketest failure after verifying the llm-d-benchmark install.sh fix (llm-d/llm-d-benchmark#1200). The teardown ran before the failure could be investigated.

kalantar:
Closes #160

## Summary

- Add `skipTeardown` pipeline param (default `"false"`) and a `when` clause on the teardown finally task in `pipeline.yaml`
- Add `--skip-teardown` flag to `deploy.py run` argument parser
- Inject `skipTeardown=true` into PipelineRun params at apply time when the flag is set (same pattern as the existing namespace override)
- Wire the flag through `_collect_run_flags()` for `--remote` mode

## Reviewer notes

- Requires `setup.py` re-run to apply the updated Pipeline definition to the cluster
- The `when` clause uses `notin ["true"]` so the default (`"false"`) and omitted param both result in normal teardown behavior
- The injection uses `for/else` to handle both cases: updating an existing `skipTeardown` param (if one were ever baked in at assemble time) and appending a new one (the normal case)

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -153,7 +153,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--workload NAME` | — | Scope execution to pairs matching this workload |
 | `--package NAME` | — | Scope execution to pairs matching this package |
 | `--status STATE` | — | Scope execution to pairs with this status (e.g. `failed`, `timed-out`) |
-| `--skip-teardown` | — | Skip teardown after PipelineRun completes, leaving namespace intact for debugging |
+| `--skip-teardown` | — | Skip the Tekton teardown task, leaving namespace resources intact for debugging |
 | `--force` | — | Reset non-pending pairs to `pending`, cleaning cluster resources (PipelineRuns + Helm) for pairs with assigned namespaces |
 | `--max-retries N` | 2 | Max retries for timed-out pairs |
 | `--poll-interval N` | 30 | Seconds between status polls |
@@ -169,7 +169,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 **Pair statuses:** `pending` → `running` → `done`. Failure paths: `running` → `failed` (hard failure or non-recoverable pending), `running` → `timed-out` (4h timeout exceeded), `running` → `pending` (recoverable early reclaim, repeats up to `--max-pending-stalls` times) → `stalled`.
 
-**Auto-cleanup** — when a PipelineRun succeeds, the orchestrator deletes the PipelineRun CR from the cluster. Failed PipelineRuns are left in place for debugging (`kubectl describe`, pod logs). Use `reset` to remove them when done.
+**Auto-cleanup** — when a PipelineRun succeeds, the orchestrator deletes the PipelineRun CR from the cluster. Failed PipelineRuns are left in place for debugging (`kubectl describe`, pod logs). Use `reset` to remove them when done. Note: `--skip-teardown` only suppresses the Tekton `llmdbenchmark-teardown` task (Helm-level resource cleanup); PipelineRun CR deletion by the orchestrator is unaffected.
 
 **Remote mode** — `deploy.py run --remote` submits the orchestrator as a Kubernetes Job (`sim2real-orchestrator`) instead of running locally. The launcher builds the EPP image locally, packs workspace files into a ConfigMap, applies the Job, and waits for the pod to reach Running. Use `stop` to cancel, `status` to check progress, and `collect` to pull results after completion. Requires `orchestrator_image` in `setup_config.json`.
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -153,6 +153,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--workload NAME` | — | Scope execution to pairs matching this workload |
 | `--package NAME` | — | Scope execution to pairs matching this package |
 | `--status STATE` | — | Scope execution to pairs with this status (e.g. `failed`, `timed-out`) |
+| `--skip-teardown` | — | Skip teardown after PipelineRun completes, leaving namespace intact for debugging |
 | `--force` | — | Reset non-pending pairs to `pending`, cleaning cluster resources (PipelineRuns + Helm) for pairs with assigned namespaces |
 | `--max-retries N` | 2 | Max retries for timed-out pairs |
 | `--poll-interval N` | 30 | Seconds between status polls |
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1637,7 +1637,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     param["value"] = ns
 
             if getattr(args, "skip_teardown", False):
-                params = pr_data.get("spec", {}).get("params", [])
+                spec = pr_data.get("spec") or {}
+                params = spec.setdefault("params", [])
                 for param in params:
                     if param["name"] == "skipTeardown":
                         param["value"] = "true"
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1637,8 +1637,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     param["value"] = ns
 
             if getattr(args, "skip_teardown", False):
-                spec = pr_data.get("spec") or {}
-                params = spec.setdefault("params", [])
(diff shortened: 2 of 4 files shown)


kalantar:
## Review: NEEDS CHANGES

Clean, correct implementation that addresses all four requirements from #160. The `when` clause logic, `for/else` injection pattern, and `_collect_run_flags` forwarding are all sound. Three items to address before merge:

### 1. Defensive fix for param injection (medium)

The `.get("spec", {}).get("params", [])` chain on the injection path can silently lose the appended `skipTeardown` param. If `spec` exists but `params` doesn't, `.get("params", [])` returns a transient empty list — the `append` mutates that throwaway list, the param is never written into `pr_data`, and teardown runs despite the flag being set. No error, no warning.

The existing namespace override a few lines above doesn't have this problem because it only *modifies* existing items, never appends. Suggested fix:

```python
if getattr(args, "skip_teardown", False):
    spec = pr_data.get("spec") or {}
    params = spec.setdefault("params", [])
    for param in params:
        if param["name"] == "skipTeardown":
            param["value"] = "true"
            break
    else:
        params.append({"name": "skipTeardown", "value": "true"})
```

`setdefault` ensures the list is created *inside* the actual `spec` dict, so the append is never lost.

In practice this is unlikely to trigger (generated PipelineRuns always have `spec.params`), but this is a debugging flag — it needs to work reliably in exactly the situations where things are already going wrong.

### 2. Missing `pipeline/README.md` update (required by contributing standards)

CLAUDE.md says: *"After any change to `pipeline/` that affects CLI flags … update `pipeline/README.md` to match."* The `deploy.py run` flags table needs a row for `--skip-teardown`.

### 3. Missing tests

The PR adds zero tests. Every other boolean flag on the `run` subparser has parser tests and `_collect_run_flags` tests. At minimum:

- **Parser**: `build_parser().parse_args(["run", "--skip-teardown"])` yields `skip_teardown == True`; default is `False`
- **`_collect_run_flags`**: flag present → `["--skip-teardown"]` in output; flag absent → not in output
- **Param injection** (optional but recommended): both branches of the `for/else` — update-existing and append-new

Existing test patterns in `test_deploy_run.py` (parser) and `test_deploy_remote.py` (`_collect_run_flags`, `_make_run_args` helper) make these straightforward to add.

kalantar:
All three items addressed in b53d9aa:

**Param injection** — switched from `.get("params", [])` to `spec.setdefault("params", [])` so the appended param is always written into the actual `pr_data` dict, not a transient list.

**README** — added `--skip-teardown` row to the `deploy.py run` flags table.

**Tests** — added parser tests (`--skip-teardown` yields `True`, default is `False`), `_collect_run_flags` tests (flag present/absent), and wired `skip_teardown` into `_make_run_args`. 644 tests pass.

kalantar:
## Review: PR #161 — `--skip-teardown` flag

**Verdict: NEEDS CHANGES** (two items, both small)

All four requirements from issue #160 are fully addressed: `pipeline.yaml` param + `when` clause, argparse flag, apply-time injection, and `--remote` forwarding via `_collect_run_flags()`. Documentation and CI pass. The Tekton `when` clause correctly uses `notin ["true"]` so the default and omitted-param cases both result in normal teardown.

### 1. Detached-dict edge case in param injection

In `deploy.py` around line 1639:

```python
spec = pr_data.get("spec") or {}
params = spec.setdefault("params", [])
```

If `pr_data["spec"]` were ever absent or `None`, `spec` becomes a new dict not linked to `pr_data`. The `skipTeardown` param gets appended to an orphan list and is silently lost — the flag does nothing with no error.

In practice `tekton.py` always generates `spec.params`, so this doesn't fire today. But it's a latent silent failure and a one-line fix:

```python
params = pr_data.setdefault("spec", {}).setdefault("params", [])
```

This ensures mutations propagate back to `pr_data` in all cases.

### 2. Missing test for the injection logic

The 4 new tests cover CLI plumbing (parser and `_collect_run_flags`) but the core behavior — injecting `skipTeardown` into PipelineRun params — has zero test coverage. Neither the append case (normal path) nor the update-existing case (defensive path) is tested.

A targeted unit test on a sample `pr_data` dict would cover both branches of the `for/else` and also serve as a regression guard for the detached-dict issue above.

### Minor (non-blocking)

- **Docs wording**: "Skip teardown after PipelineRun completes" slightly implies teardown only applies on success. The `finally` block runs on both success and failure — the primary use case is debugging failures. Consider: "Skip the Tekton teardown task, leaving namespace resources intact for debugging."
- **Auto-cleanup paragraph** in README doesn't mention that `--skip-teardown` only affects Helm-level teardown, not PipelineRun CR deletion by the orchestrator. A one-sentence note would prevent confusion.

kalantar:
All four items addressed in 36bdffc:

**Detached-dict fix** — replaced `pr_data.get("spec") or {}` with chained `pr_data.setdefault("spec", {}).setdefault("params", [])` so mutations always propagate back to `pr_data`.

**Injection tests** — added 4 unit tests covering the append path (normal), update-existing path (defensive), missing `spec`, and missing `params`. These also serve as regression guards for the detached-dict issue.

**Docs wording** — changed to "Skip the Tekton teardown task, leaving namespace resources intact for debugging" to make clear it applies on both success and failure.

**Auto-cleanup note** — added a sentence clarifying that `--skip-teardown` only suppresses the Helm-level `llmdbenchmark-teardown` task, not PipelineRun CR deletion by the orchestrator.

648 tests pass.

kalantar:
## Review: PR #161 — `feat: add --skip-teardown flag to deploy.py run`

**Verdict: READY TO MERGE**

All four requirements from issue #160 are fully addressed: pipeline.yaml param + when clause, argparse flag, apply-time injection, and remote-mode forwarding via `_collect_run_flags()`. CI passes (715 tests). The implementation follows existing codebase conventions (camelCase Tekton params, `--skip-build-epp` style CLI flags, `setdefault` for defensive structure creation).

### Key strengths

- The `when` clause uses `notin ["true"]` so the default `"false"` and omitted param both result in normal teardown — safe default behavior.
- The `setdefault` chain for injection is more robust than the existing namespace override pattern (which uses `.get()` and would silently no-op on malformed YAML).
- README auto-cleanup note clearly distinguishes Tekton-level teardown from orchestrator-level CR deletion — prevents user confusion.
- Tests cover parser defaults, `_collect_run_flags` forwarding, and injection edge cases (append, update-existing, missing spec, missing params).

### Suggestions for future consideration (non-blocking)

1. **Injection tests use a duplicated helper** — `_inject_skip_teardown()` in the test file copies the production logic rather than calling it. If the production code drifts, these tests won't catch it. Extracting the injection into a named function in `deploy.py` (like `_collect_run_flags`) would let tests exercise the real code path.

2. **No log message when skip-teardown is activated** — Unlike `--force` (which logs what it did), `--skip-teardown` silently modifies PipelineRuns. A `warn()` at the start of the orchestration loop would give users confirmation the flag took effect.

3. **Minor argparse help text imprecision** — The help string says "Skip teardown after PipelineRun completes" which implies the success path only. The teardown task is in the `finally` block and runs on both success and failure. Dropping "after PipelineRun completes" would be more accurate.

None of these block merge — the feature works correctly as implemented.

PR #163, opened 2026-05-18
feat: reset always cleans up helm releases for done pairs

### LINKED ISSUE #162: deploy.py reset: no helm cleanup path for done pairs with failed Tekton teardown
## Problem

When a pair transitions to `done` (PipelineRun succeeded), `deploy.py` immediately clears `entry["namespace"]` and saves only `entry["completed_namespace"]` as a record of where it ran. The assumption is that the Tekton `finally` block's `llmdbenchmark-teardown` task already cleaned up helm releases.

If the teardown task failed (e.g., venv missing, tool download failure, timeout), helm releases are orphaned with no automated recovery path:

1. `_reset_pair` skips helm cleanup for done pairs entirely (line 1031-1033)
2. Even if it didn't skip, `entry.get("namespace")` is `None` — it was cleared on the done transition
3. `completed_namespace` exists in the entry but `_reset_pair` never reads it

## Fix

`_reset_pair` should unconditionally use `entry.get("completed_namespace")` to check for and uninstall any lingering helm releases in that namespace. No flag needed — reset should just clean up.

## Context

- `_reset_pair`: `pipeline/deploy.py` ~line 975
- Done transition clears namespace: `pipeline/deploy.py` ~lines 1229-1236, 1483-1485
- Tekton teardown task: `tektonc-data-collection/tekton/tasks/llmdbenchmark-teardown.yaml`

kalantar:
Closes #162

## Summary

- `_reset_pair` now unconditionally checks `completed_namespace` for lingering helm releases on done pairs
- If `helm list` finds releases in that namespace, they are uninstalled
- No new flags — reset just does the right thing

## Context

When a pair succeeds, `namespace` is cleared and `completed_namespace` is saved. Previously, `_reset_pair` skipped helm cleanup for done pairs entirely, trusting Tekton's finally-block teardown. If that teardown failed, releases were orphaned with no recovery path.

## Test plan

- [x] Done pair with `completed_namespace` containing releases → uninstalled
- [x] Done pair without `completed_namespace` → skipped gracefully  
- [x] Done pair with empty `helm list` → no uninstall attempted
- [x] All 651 existing tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -999,6 +999,8 @@ def _reset_pair(key: str, entry: dict, discovered: dict, *,
         info(f"[DRY-RUN] {key}: would delete pipelinerun {pr_name or '(unknown)'} in {target}")
         if not is_done:
             info(f"[DRY-RUN] {key}: would uninstall all helm releases in {target}")
+        elif entry.get("completed_namespace"):
+            info(f"[DRY-RUN] {key}: would check for orphaned helm releases in {entry['completed_namespace']}")
         return True
 
     # Delete PipelineRun
@@ -1028,8 +1030,23 @@ def _reset_pair(key: str, entry: dict, discovered: dict, *,
     else:
         warn(f"{key}: no namespace and no namespace slots — cannot delete pipelinerun {pr_name}")
 
-    # For done pairs, Tekton finally task already tore down Helm releases
+    # For done pairs, Tekton finally task should have torn down Helm releases,
+    # but check completed_namespace for orphans in case teardown failed.
     if is_done:
+        completed_ns = entry.get("completed_namespace")
+        if completed_ns:
+            result = run(["helm", "list", "-n", completed_ns, "-q"],
+                         check=False, capture=True)
+            if result.returncode != 0:
+                warn(f"{key}: helm list failed in {completed_ns}")
+            elif result.stdout.strip():
+                for release in result.stdout.strip().splitlines():
+                    ur = run(["helm", "uninstall", release, "-n", completed_ns],
+                             check=False, capture=True)
+                    if ur.returncode == 0:
+                        ok(f"Uninstalled: {release} (ns: {completed_ns})")
+                    else:
+                        warn(f"Failed to uninstall {release} in {completed_ns}")
         return True
 
     if ns and not pr_deleted and pr_name:
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -972,12 +972,29 @@ def _load_pairs(cluster_dir: Path) -> dict:
     return pairs
 
 
+def _uninstall_orphaned_helm(key: str, namespace: str) -> None:
+    """Check a namespace for lingering helm releases and uninstall them."""
+    result = run(["helm", "list", "-n", namespace, "-q"],
+                 check=False, capture=True)
+    if result.returncode != 0:
+        warn(f"{key}: helm list failed in {namespace}")
+    elif result.stdout.strip():
+        for release in result.stdout.strip().splitlines():
+            ur = run(["helm", "uninstall", release, "-n", namespace],
+                     check=False, capture=True)
+            if ur.returncode == 0:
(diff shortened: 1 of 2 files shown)


kalantar:
## Verdict: NEEDS CHANGES

The core fix is sound — `_reset_pair` now checks `completed_namespace` for orphaned helm releases on done pairs, addressing issue #162's requirements. However, a few issues should be resolved before merge.

### Must fix

1. **PR title is incorrect.** It says "add --deep flag" but no flag was added (the body and commit message both confirm this). Since squash-merge typically uses the PR title, this will pollute git history. Suggest: "feat: reset always cleans up helm releases for done pairs"

2. **Early-return bypass at ~line 988-995.** If a done pair has `namespace=None` and no `pr_name` in `discovered`, but does have `completed_namespace` with orphaned releases, the function returns `True` before reaching the new cleanup logic. The issue says "unconditionally" — this path skips it. Either move the helm check above this early return or add it inside the `is_done` case in that block.

3. **Function docstring at ~line 980 is now stale.** It says "For done pairs: deletes PipelineRun only (Helm already torn down by Tekton)" — that's exactly the assumption this PR corrects. Update to reflect the new behavior.

### Should fix

4. **`return True` even when cleanup fails.** The non-done path returns `False` when helm operations fail, but the done-pair path always returns `True`. This means the caller reports success even when orphans remain. Consider returning `False` on failure (matching the non-done pattern), or at minimum make the warn message clearly say orphans may remain.

5. **Missing tests for error paths.** No tests cover `helm list` failure or `helm uninstall` failure. These are the exact scenarios this PR exists to handle (teardown failures), so they're worth testing — especially since the return-value behavior differs from the non-done path.

### Nits

- The helm uninstall failure message doesn't include stderr, making it hard to diagnose. The non-done path has the same gap, but new code shouldn't propagate it.
- Dry-run branch uses `entry['completed_namespace']` (bracket access) after a `.get()` guard — works fine but inconsistent with the non-dry path that binds to a local var.

kalantar:
Addressed review feedback:

**Fixed:**
- PR title corrected (was stale from an earlier iteration that had a `--deep` flag).
- The early-return path at line 988-995 now checks `completed_namespace` for orphaned helm releases on done pairs. Previously a done pair with no `pr_name` in `discovered` (common — PipelineRun is deleted on success) would bypass the cleanup entirely.
- Docstring updated to reflect the new behavior.
- Added tests for `helm list` failure, `helm uninstall` failure, and the early-return bypass case on done pairs.
- Extracted `_uninstall_orphaned_helm` helper to avoid duplicating the helm list/uninstall logic.

**Skipped:**
- `return True` on done-pair cleanup failure — intentional. For done pairs there's no state to reset; the return value signals whether state was left inconsistent. Warnings surface the failure to the operator.
- No stderr in warn messages — same pattern as existing non-done path, not in scope for this change.

kalantar:
## Verdict: READY TO MERGE

This PR correctly and completely addresses issue #162. The implementation is clean, well-tested, and handles all edge cases identified in the issue.

### Key findings

**Strengths:**
- Both code paths (early-return when no ns/pr_name, and full path after PipelineRun deletion) now clean up orphaned Helm releases for done pairs
- `_uninstall_orphaned_helm` is well-factored as a standalone helper with appropriate error tolerance (warn but don't block reset)
- Dry-run support is consistent with existing patterns
- 6 new tests cover: successful cleanup, missing `completed_namespace`, empty `helm list`, `helm list` failure, `helm uninstall` failure, and the no-pr_name bypass case

**Optional improvements (non-blocking):**

1. **Include stderr in warning messages** — When `helm list` or `helm uninstall` fails, the warning discards `result.stderr`, losing diagnostic context (auth failures, missing namespace, etc.). Adding `{result.stderr.strip()}` to the warn messages would help debuggability without changing control flow.

2. **Dry-run path for done pairs is untested** — A test verifying `dry_run=True` on a done pair with `completed_namespace` emits the info message without calling `helm` would guard against accidental operations during dry-run.

Neither of these blocks merging — the core fix is correct and well-covered.

PR #171, opened 2026-05-19
refactor: wipe only deletes result files, no longer resets status

### LINKED ISSUE #168: refactor: wipe should only delete result files, not reset status
## Problem

`deploy.py wipe` currently does two things:
1. Deletes local result files (`results/{package}/{workload}/`)
2. Resets the ConfigMap progress entry to `pending` (clearing retries, pending_stalls, etc.)

This couples artifact cleanup with scheduling decisions. You cannot wipe results without implicitly re-queuing the pair for dispatch, and the status reset can cause accidental re-dispatch if `run` is active concurrently.

## Proposal

Limit `wipe` to data cleanup only:
- Delete result files from `results/`
- Leave the ConfigMap progress entry unchanged

To re-queue a wiped pair, the user would explicitly run `reset` afterward. This keeps the API surface clean — each subcommand does one thing — and avoids adding flags like `--no-reset`.

## Current behavior (lines 1843–1868 in deploy.py)

```python
entry["status"] = "pending"
entry["retries"] = 0
entry["pending_stalls"] = 0
entry["pending_since"] = None
entry["namespace"] = None
```

## Desired behavior

Remove the status/field resets from `_cmd_wipe`. The only ConfigMap interaction should be reading progress to identify which pairs have results to delete.

## Migration

Update the docstring, `--help` text, and `pipeline/README.md` to reflect that `wipe` no longer resets status. Document the `wipe` + `reset` two-step workflow for re-running pairs.

kalantar:
## Summary
- Remove status-reset logic from `deploy.py wipe` — it now only deletes local result files
- `wipe` no longer calls `store.save()` or mutates ConfigMap progress entries
- To re-queue wiped pairs for dispatch, use `reset` separately

Closes #168

## What changed
- `pipeline/deploy.py`: removed 5 status-field mutations from the wipe loop, removed `store.save()` call, updated docstring/help text, dry-run no longer mentions "would reset status"
- `pipeline/tests/test_deploy_wipe.py`: removed all assertions about status resets, simplified test helpers (no more `capture_saves` or `_assert_reset`)
- `pipeline/README.md`: updated wipe description and documented the `wipe` + `reset` two-step workflow

## Reviewer notes
- Pairs with no results on disk now log "skipped" instead of "Reset: ... (no results on disk)" — they're no longer counted in the wipe total since nothing was actually deleted
- The scope filter still uses ConfigMap status to identify non-pending pairs (reads progress but doesn't write it)

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -140,7 +140,7 @@ python pipeline/deploy.py status            # show progress snapshot of all (wor
 python pipeline/deploy.py collect [flags]     # pull results from the cluster PVC
 python pipeline/deploy.py stop               # stop the remote orchestrator Job
 python pipeline/deploy.py reset [flags]     # tear down cluster resources for failed/stalled pairs
-python pipeline/deploy.py wipe  [flags]     # delete local results and reset pairs to pending
+python pipeline/deploy.py wipe  [flags]     # delete local result files for non-pending pairs
 python pipeline/deploy.py pairs   [flags]   # list available pair keys, workloads, and packages
 ```
 
@@ -207,7 +207,7 @@ When `--only` or `--workload` is given, only matching workload subdirectories ar
 
 **Safety:** Results in `workspace/runs/<run>/results/` are preserved — only cluster resources are removed. For `done` pairs, only the PipelineRun is deleted (Tekton already tore down Helm releases).
 
-**`deploy.py wipe`** — deletes local results (`results/<package>/<workload>/`) for non-pending pairs and resets their status to `pending` in the ConfigMap. Use when reusing a run name and stale results from prior executions need to be cleared before re-executing. Pending pairs are skipped (nothing to wipe). Empty package directories are cleaned up automatically.
+**`deploy.py wipe`** — deletes local result files (`results/<package>/<workload>/`) for non-pending pairs. Does **not** modify pair status in the ConfigMap — use `reset` afterward to re-queue pairs for dispatch. Pending pairs are skipped (nothing to wipe). Empty package directories are cleaned up automatically.
 
 | Flag | Description |
 |------|-------------|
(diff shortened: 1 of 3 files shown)


kalantar:

### Blocking Issue

**The documented "wipe + reset" workflow doesn't work for `done` pairs** — which are the primary target of `wipe`.

`reset` explicitly skips `done` pairs (`_force_reset` at line 1109: `if entry.get("status") not in (None, "pending", "done")`). The old `wipe` was the **only** code path that moved `done` → `pending`. This PR removes that capability without providing an alternative.

A user following the README guidance ("use `reset` afterward to re-queue pairs for dispatch") will find it has no effect on their `done` pairs.

**Options:**
1. Update `reset` in this same PR to support re-queuing `done` pairs when explicitly targeted (e.g., via `--only`, `--workload`, or `--package` scoping — so unscoped `reset` still preserves done pairs for safety)
2. Keep this PR scoped to just the wipe change, but update the README to honestly document the limitation and file a follow-up issue for `reset` to support `done` pairs
3. Have `wipe` reset status **only for pairs whose results it actually deleted** (a middle ground: no status mutation for "no results on disk" cases, but still resets pairs where files were removed)

### Non-blocking Suggestions

- **Add a negative test** asserting `store.save()` is never called — the current no-op mock would silently pass if someone reintroduces a save call
- **README line 342** says all subcommands "use" the ConfigMap — after this PR, `wipe` only reads from it. Consider clarifying.
- When `wiped == 0` and targets exist (all had no results on disk), consider printing a hint about `reset` to guide users through the two-step workflow

### What's Good

- Core refactor is clean: status mutations and `store.save()` correctly removed together
- Tests properly simplified — no dead assertions about status resets
- All 655 tests pass, lint clean
- No silent failures introduced
- Documentation updates are internally consistent (modulo the `done` pair gap)

kalantar:

**Fixed:** The README no longer claims `reset` handles `done` pairs. The "Re-running wiped pairs" section now references #172 (which will enable `reset` to work with all statuses) so readers know there's a known gap being addressed separately.

**Added:** `test_wipe_does_not_save_progress` — explicitly asserts `store.save()` is never called, guarding against regression if someone reintroduces a save call.

**Skipped (not blocking):**
- README line 342 "use" ambiguity (read vs write) — too minor for this PR
- UX hint when `wiped == 0` — nice-to-have, can be added later if needed

kalantar:

All four review dimensions pass with no blocking issues:

- **Correctness**: All issue #168 requirements met. Status mutations and `store.save()` fully removed. `wiped` counter now reflects actual deletions only.
- **Silent failures**: No swallowed errors. The `except OSError: pass` on `rmdir()` is appropriate (cosmetic parent-dir cleanup). All real failures are logged and exit with code 1.
- **Test coverage**: 17 tests with no critical gaps. `test_wipe_does_not_save_progress` explicitly guards the core behavioral contract.
- **Documentation**: Accurate across all locations. The #172 reference honestly documents the `done`-pair limitation rather than making misleading claims.

Prior review feedback fully addressed in the second commit.

One minor suggestion (non-blocking): the "Re-running wiped pairs" sentence could lead with the limitation for `done` pairs rather than burying it as a trailing note, since `done` pairs are the primary wipe target. But this is a clarity preference, not a factual error.

PR #173, opened 2026-05-19
fix: reset moves done pairs back to pending by default

### LINKED ISSUE #172: fix: reset should move done pairs back to pending by default
## Problem

`reset` explicitly skips `done` pairs — both in `_reset_pair` (line 1007: `if not dry_run and not is_done`) and `_force_reset` (line 1109: `if entry.get("status") not in (None, "pending", "done")`). For `done` pairs, `reset` only cleans up cluster resources (PipelineRun deletion, orphaned Helm uninstall) without changing status.

This was a protective design: prevent accidental re-dispatch that would overwrite valid results. But the word "reset" means reset — the current behavior is confusing, and with #168 removing the status-reset from `wipe`, there is now **no mechanism** to move `done` pairs back to `pending` without manual ConfigMap editing.

## Context

Previously, `wipe` moved `done` → `pending` as a side effect of deleting results. PR #171 separates concerns so `wipe` only deletes files. The intended workflow is `wipe` + `reset`, but `reset` refuses to touch `done` pairs, breaking the two-step workflow for the primary use case.

## Decision: Option A — flip the default

Make `reset` reset ALL non-pending pairs including `done` to `pending`. Add `--preserve-done-status` flag for the old protective behavior.

### Default behavior (no flag)

`reset` moves ALL non-pending pairs (including `done`) back to `pending`. Cluster cleanup (PipelineRun deletion, orphaned Helm uninstall) still happens for all pairs as it does today.

### `--preserve-done-status` flag

When passed, `done` pairs get cluster cleanup (PipelineRun + orphaned Helm) but their status stays `done`. This is the current behavior, moved behind an explicit opt-in flag.

The flag name `--preserve-done-status` is chosen over `--preserve-done` because "preserve done" could be misread as "preserve the done pairs' data" rather than "preserve their status field."

### `run --force`

`run --force` also calls `_force_reset`. It should follow the same new default (reset `done` pairs to `pending`). It does NOT get the `--preserve-done-status` flag — `--force` means force.

## Affected code

Two done-guards need updating in `pipeline/deploy.py`:

1. **`_reset_pair` line 1007:** `if not dry_run and not is_done` — prevents status change for done pairs when there's no namespace/PR. Change: remove the `and not is_done` guard (or condition it on the new flag when called from `_cmd_reset`).

2. **`_reset_pair` line ~1065:** early return for `is_done` after cluster cleanup — returns `True` without resetting status. Change: after cluster cleanup, also reset status to `pending` unless `--preserve-done-status` is active.

3. **`_force_reset` line 1109:** `if entry.get("status") not in (None, "pending", "done")` — excludes `done` from the reset scope. Change: remove `"done"` from the exclusion list.

4. **`_cmd_reset`:** Add `--preserve-done-status` argument. Pass it through to `_reset_pair`. Note: the scope filter in `_cmd_reset` already includes `done` pairs (line 1749 only excludes `pending`) — no change needed there.

5. **`pipeline/README.md`:** Update reset documentation to reflect new default and document `--preserve-done-status`.

## Acceptance criteria

- [ ] `deploy.py reset` moves `done` pairs to `pending` (with cluster cleanup)
- [ ] `deploy.py reset --preserve-done-status` cleans up cluster for `done` pairs but leaves status as `done`
- [ ] `deploy.py run --force` resets `done` pairs to `pending` (no opt-out flag)
- [ ] Scoping flags (`--only`, `--workload`, `--package`) work with done pairs in both modes
- [ ] `--dry-run` correctly reports what would happen for done pairs in both modes
- [ ] Existing tests for non-done pair reset behavior still pass
- [ ] New tests cover: done→pending default, --preserve-done-status, run --force with done pairs

## Relationship to other issues

- Unblocks the "wipe + reset" workflow documented in #168 / PR #171
- Once implemented, the #172 reference in README can be removed

kalantar:
## Summary
- `reset` now resets ALL non-pending pairs including `done` to `pending` (with cluster cleanup)
- New `--preserve-done-status` flag provides the old behavior: cluster cleanup for done pairs without changing their status
- `run --force` (via `_force_reset`) also resets done pairs — no opt-out

Closes #172

## What changed

**`pipeline/deploy.py`:**
- `_reset_pair`: added `preserve_done_status` parameter. Both done-guards now check this flag — by default, done pairs get status reset to pending after cluster cleanup
- `_force_reset`: removed `"done"` from the exclusion filter (`if entry.get("status") not in (None, "pending")`)
- `_cmd_reset`: reads `--preserve-done-status` from args, passes to `_reset_pair`
- Argparse: added `--preserve-done-status` flag
- Docstrings/help text updated

**`pipeline/README.md`:**
- Reset description rewritten to reflect new default
- `--preserve-done-status` documented in flag table
- Wipe "Re-running" note simplified (no more #172 caveat)

**Tests:**
- `test_reset_pair_done_resets_state_by_default` — verifies done→pending
- `test_reset_pair_done_preserves_status_with_flag` — verifies flag preserves done
- `test_force_reset_resets_all_non_pending_including_done` — verifies run --force behavior
- Existing done-pair helm tests updated to expect `pending` status after reset

## Reviewer attention
- **Breaking change**: `reset` without `--preserve-done-status` will now re-queue done pairs. Any automation relying on the old "reset preserves done" behavior needs updating.
- The `_reset_pair` done-pair path has two branches: (1) no namespace/no PR → just reset state, and (2) has PR → delete PR + helm cleanup then reset state. Both are now gated on `preserve_done_status`.


kalantar:

All acceptance criteria from #172 are satisfied. 657 tests pass, lint clean, documentation accurate.

**Key findings:**
- Correctness confirmed: both done-pair paths in `_reset_pair` correctly gate on `preserve_done_status`, `_force_reset` properly includes done pairs, argparse wiring is clean
- No silent failures introduced by this PR
- Documentation is accurate with no stale references to old behavior
- Test coverage is solid for core behavior (done→pending default, preserve flag, force)

**Non-blocking observation (pre-existing pattern, not introduced by this PR):**
`_uninstall_orphaned_helm` is fire-and-forget for done pairs — if helm cleanup fails, the status still gets reset to `pending`. For non-done pairs, helm failure is a hard stop (`return False`). Now that done pairs also get re-queued, a failed cleanup could theoretically leave orphaned resources that conflict on re-dispatch. This is low-risk in practice (Tekton's finally task handles teardown; orphans only exist if that also failed) and would be a separate hardening issue if desired.

**Minor test coverage gaps (non-blocking):**
- No explicit test for `--status done` scoping filter
- No dry-run unit test specifically for done pairs
These are covered implicitly by the integration test path and pre-existing dry-run tests.

PR #180, opened 2026-05-19
feat: add --preserve-pipelineruns flag to deploy.py run

### LINKED ISSUE #179: feat: add --preserve-pipelineruns flag to deploy.py run
## Problem

When a PipelineRun succeeds, `deploy.py run` deletes it immediately (line 1526). This makes it impossible to debug steps that failed silently (e.g., steps using `set +e` that exit 0 despite internal errors).

`--skip-teardown` keeps Helm releases alive but does not preserve the PipelineRun object itself — TaskRun logs and step statuses are lost.

## Proposed Solution

Add a `--preserve-pipelineruns` flag to `deploy.py run` that skips the `_delete_pipelinerun()` call after successful completion. This preserves the PipelineRun and its TaskRun objects on the cluster for post-mortem inspection via `tkn pr describe` / `tkn tr logs`.

The flag should also suppress deletion on the failure path (though currently failure already preserves the PipelineRun — only the namespace slot is released).

## Context

Discovered while debugging issue #170 — `collect-results` uses `set +e`, so internal failures (e.g., no pods found) don't fail the pipeline. The PipelineRun completes successfully and is immediately deleted, removing the evidence needed to diagnose the problem.

kalantar:
## Summary

- Adds `--preserve-pipelineruns` flag to `deploy.py run` that suppresses deletion of PipelineRun objects after completion
- Guards both deletion sites: the main run loop (success path) and `_reconcile_on_resume` (resume path)
- Forwards the flag in the `--remote` path

Closes #179

## Reviewer attention

The pre-apply deletion at line 1688 (`kubectl delete pipelinerun ... --ignore-not-found`) is intentionally NOT guarded — that clears a stale prior run before re-applying the same pair, which is necessary for correct dispatch. `--preserve-pipelineruns` only preserves runs that reach terminal state during the current execution.

## Test plan

- [x] New test: `test_reconcile_preserve_pipelineruns_skips_deletion` verifies the flag suppresses deletion in `_reconcile_on_resume`
- [x] All 661 existing tests pass

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -154,6 +154,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--package NAME` | — | Scope execution to pairs matching this package |
 | `--status STATE` | — | Scope execution to pairs with this status (e.g. `failed`, `timed-out`) |
 | `--skip-teardown` | — | Skip the Tekton teardown task, leaving namespace resources intact for debugging |
+| `--preserve-pipelineruns` | — | Do not delete PipelineRun objects after completion (keeps TaskRun logs for debugging) |
 | `--force` | — | Reset non-pending pairs to `pending`, cleaning cluster resources (PipelineRuns + Helm) for pairs with assigned namespaces |
 | `--max-retries N` | 2 | Max retries for timed-out pairs |
 | `--poll-interval N` | 30 | Seconds between status polls |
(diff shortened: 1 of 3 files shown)


kalantar:
## Verdict: NEEDS CHANGES

The implementation is correct and well-scoped — all issue #179 requirements are addressed (flag added, both deletion sites guarded, remote forwarding works). No bugs, silent failures, or design concerns found.

**One blocking item:**

`pipeline/README.md` needs an update. The project's contributing guidelines require documentation updates when CLI flags are added. Specifically:
- Add `--preserve-pipelineruns` to the `deploy.py run` flags table
- Update the "Auto-cleanup" paragraph to note that `--preserve-pipelineruns` suppresses PipelineRun CR deletion on success

**Non-blocking suggestions (take or leave):**

1. **Test coverage gaps** — The new test covers `_reconcile_on_resume` well but leaves the main loop success path and `_collect_run_flags` forwarding untested. The remote-forwarding test is low-effort (follows the existing `test_collect_run_flags_skip_teardown` pattern in `test_deploy_remote.py`).

2. **Docstring on `_reconcile_on_resume`** — could mention the `preserve_pipelineruns` parameter's effect on deletion behavior.

3. **Optional info log** — When preservation is active, a brief log line (`info(f"Preserving PipelineRun {pr_name!r} in {ns}")`) would aid debugging long runs. Currently the skip is silent.

kalantar:
Addressed the documentation update: added `--preserve-pipelineruns` to the flags table and updated the "Auto-cleanup" paragraph in `pipeline/README.md`.

Skipped the non-blocking suggestions (additional test coverage for the main loop path, docstring, info log) — the core behavior is tested and the flag is straightforward.

PR #181, opened 2026-05-19
fix: signal backoff reset on actual Pending→Running transition

### LINKED ISSUE #86: Backoff reset on Pending→Running pod transition is missing
## Parent

Follow-up to #61 (Resource-availability-aware orchestration) — specifically §6 "Backoff Reset".

## Problem

Issue #61 specifies two conditions that reset the backoff controller to level 0:

> 1. "A pod transitions from Pending to Running within the pending threshold"
> 2. "The capacity probe shows free GPUs exceeding the largest pending workload cost"

**Condition 2 is implemented** — `signal_capacity()` is called at `deploy.py:1252` when the probe shows sufficient free GPUs.

**Condition 1 is not implemented.** Instead, `signal_scheduling_success()` is called at PipelineRun *submission* time (`deploy.py:1324-1325`), immediately after `kubectl apply` succeeds — before any pod has actually scheduled:

```python
# deploy.py:1318-1326
entry["status"] = "running"
entry["namespace"] = ns
...
if backoff.state != "normal":
    backoff.signal_scheduling_success()   # premature — pod hasn't run yet
```

A PipelineRun can be submitted successfully but its pods may still fail to schedule, so this resets backoff prematurely.

Meanwhile, `_handle_pending_pods` does clear `pending_since` when pods are no longer pending (`deploy.py:327-328`), which is the correct detection point for Pending→Running. But the backoff controller is not passed to this function — its signature only accepts `(pr_name, namespace, entry, pending_threshold, max_pending_stalls)` (`deploy.py:283-284`). There is no code path where a Pending→Running transition signals the backoff controller.

## Impact

During GPU scarcity, the backoff controller can be prematurely reset by a successful `kubectl apply` even though the pod remains Pending. This undermines the backoff mechanism: the orchestrator returns to aggressive polling while the cluster still lacks capacity.

## Fix

1. Remove (or gate) `signal_scheduling_success()` at submission time — submitting a PipelineRun is not evidence of scheduling success.
2. Pass the `BackoffController` to `_handle_pending_pods` (or return a signal from it).
3. When `_handle_pending_pods` detects `category is None` (no pending pods) after `pending_since` was previously set, call `backoff.signal_scheduling_success()` — this is the actual Pending→Running transition.

kalantar:
Closes #86

## Summary

- Removed the premature `signal_scheduling_success()` call at PipelineRun submission time (`kubectl apply` succeeding is not evidence that pods scheduled)
- Added detection of the actual Pending→Running transition in the orchestrator loop: capture `pending_since` before calling `_handle_pending_pods`, then check if it was cleared without a reclaim — that's the real scheduling success signal
- Added two tests: one verifying the transition triggers a backoff reset, one verifying that pods that were never pending do not trigger it

## Design choice

Rather than passing the `BackoffController` into `_handle_pending_pods` (coupling a pod-checking utility to orchestrator state), the caller observes the entry's `pending_since` field before and after the call. This keeps `_handle_pending_pods` focused on pod classification with zero signature changes.

## Test plan

- [x] `test_pending_to_running_signals_backoff_reset` — verifies backoff resets when pods transition Pending→Running
- [x] `test_no_backoff_signal_when_never_pending` — verifies no spurious reset when pods were never pending
- [x] All 662 existing tests pass, lint clean

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1539,6 +1539,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
 
             elif status in ("Running", "Started"):
                 # Check for pending pods (before timeout)
+                had_pending = entry.get("pending_since") is not None
                 try:
                     reclaimed = _handle_pending_pods(
                         pr_name=pr_name, namespace=ns, entry=entry,
@@ -1563,6 +1564,13 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     progress["_orchestrator"] = backoff.to_dict()
                     store.save(progress)
                     continue
+
+                if had_pending and entry.get("pending_since") is None:
+                    if backoff.state != "normal":
+                        backoff.signal_scheduling_success()
+                        info(f"[{pair_key}] Pending→Running → backoff reset")
+                        progress["_orchestrator"] = backoff.to_dict()
+                        store.save(progress)
 
                 # Check for timeout
                 ts_result = run(
@@ -1710,9 +1718,6 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
             slots_busy[ns] = pair_key
             store.save(progress)
             ok(f"[{pair_key}] → {ns} ({pr_name})")
-            if backoff.state != "normal":
-                backoff.signal_scheduling_success()
-                info("Scheduling success → backoff reset")
 
         # Persist backoff state
         progress["_orchestrator"] = backoff.to_dict()
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -1250,6 +1250,114 @@ def test_early_reclaim_pods_running_clears_pending_since(monkeypatch):
     assert entry["pending_since"] is None
 
 
+def test_pending_to_running_signals_backoff_reset(monkeypatch):
+    """When pods transition Pending→Running, caller should signal backoff reset."""
+    import json
+    import pipeline.deploy as mod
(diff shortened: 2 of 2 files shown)


kalantar:
## Verdict: READY TO MERGE

PR correctly implements all three requirements from issue #86:

1. Premature `signal_scheduling_success()` at submission time — removed
2. Actual Pending→Running detection — implemented by capturing `pending_since` before `_handle_pending_pods` and checking it after
3. Backoff reset at detection point — fires only when the transition is genuine

The design choice (observe entry state before/after rather than passing BackoffController into `_handle_pending_pods`) keeps the pod-classification utility decoupled from orchestrator state with zero signature changes. Sound.

### One suggestion (non-blocking)

The tests include `and not reclaimed` in their inline caller logic, but the production code does not — it relies on the `continue` statement after the reclaim branch to make the check unreachable. Both are functionally equivalent today, but if someone later refactors the loop to remove that `continue`, the production code would lack the guard while the tests would still pass. Consider adding `and not reclaimed` as a defensive guard in the production condition to make the invariant explicit:

```python
if had_pending and entry.get("pending_since") is None and not reclaimed:
```

This aligns code, tests, and the PR description ("cleared without a reclaim") on the same explicit check.

### Minor notes (no action required)

- The test comment "This mirrors the logic now in deploy.py's orchestrator loop" is slightly inaccurate given the two differences (extra `not reclaimed`, missing `state != "normal"` guard). Not worth blocking over.
- Exception path in `_handle_pending_pods` defers backoff reset to the next successful poll iteration — self-healing and acceptable.

PR #183, opened 2026-05-19
fix: add helm CLI to orchestrator image

### LINKED ISSUE #182: Orchestrator image missing helm CLI
## Problem

The orchestrator image (`Dockerfile`) only installs `kubectl`, but `deploy.py` also invokes `helm` for:

- `--force` reset: uninstalling helm releases before resetting pairs (`_uninstall_orphaned_helm`, line 975)
- Slot cleanup: removing lingering helm releases from completed namespaces (lines 1079-1092)

When running remotely (`deploy.py run --remote`), any code path that calls `helm` fails with:

```
[WARN]  wl-w1-mid-blis1: reset failed during --force: [Errno 2] No such file or directory: 'helm'
```

## Fix

1. **`Dockerfile`** — Add helm installation alongside kubectl (e.g., download from `https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz`)
2. **`.github/workflows/orchestrator-image.yml`** — Add a `helm version --short` smoke test to catch this class of issue in CI

## Notes

`kubectl` and `helm` are the only two external CLIs invoked by `deploy.py` (verified via grep). No other tools are missing.

kalantar:
Closes #182

## Summary

- Added `helm` CLI installation to the orchestrator `Dockerfile` alongside `kubectl`, with SHA-256 checksum verification
- Added `helm version --short` to the CI smoke test so missing tools are caught before the image ships

## Context

`deploy.py` invokes `helm` for `--force` reset (uninstalling helm releases) and slot cleanup (removing orphaned releases). The orchestrator image only had `kubectl`, causing `[Errno 2] No such file or directory: 'helm'` when running remotely.

## Reviewer notes

- Helm version is fetched dynamically from the GitHub releases API at build time (same pattern as kubectl)
- The `.sha256sum` file is used directly with `sha256sum --check` — this is the official verification method from the Helm project
- Everything stays in one RUN layer to avoid inflating image size

### DIFF
--- a/.github/workflows/orchestrator-image.yml
+++ b/.github/workflows/orchestrator-image.yml
@@ -57,6 +57,7 @@ jobs:
         run: |
           docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
           docker run --rm --entrypoint kubectl ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --client
+          docker run --rm --entrypoint helm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --short
           docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "from pipeline.lib import assemble, backoff, capacity, context_builder, epp, health, manifest, pod_pending, progress, run_manager, state_machine, tekton, values; print('OK')"
 
       - name: Push
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update && \
     install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl && \
     rm kubectl kubectl.sha256 && \
     HELM_VERSION=$(curl -fsSL https://api.github.com/repos/helm/helm/releases/latest | grep '"tag_name"' | sed 's/.*"v\(.*\)".*/\1/') && \
+    test -n "$HELM_VERSION" || { echo "ERROR: failed to determine helm version"; exit 1; } && \
     curl -fLO "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz" && \
     curl -fLO "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz.sha256sum" && \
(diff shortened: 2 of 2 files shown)


kalantar:
## Verdict: NEEDS CHANGES

The PR correctly addresses both issue #182 requirements (helm in Dockerfile + CI smoke test). The SHA-256 verification, cleanup, and single-layer approach are all solid. One change needed before merge:

**Add a guard against empty `HELM_VERSION`**

The version extraction pipeline (`curl | grep | sed`) can produce an empty string if the GitHub API returns an error (rate limit, format change, network issue). Because it's inside a `$()` substitution, the `&&` chain does NOT catch this — it just passes an empty variable to the next `curl`, which fails with a confusing 404 on a malformed URL like `helm-v-linux-amd64.tar.gz`.

Fix: add an explicit check after the assignment:

```dockerfile
HELM_VERSION=$(curl -fsSL https://api.github.com/repos/helm/helm/releases/latest | grep '"tag_name"' | sed 's/.*"v\(.*\)".*/\1/') && \
test -n "$HELM_VERSION" || { echo "ERROR: failed to determine helm version"; exit 1; } && \
```

**Non-blocking observations (take or leave):**

1. Consider pinning to a specific version (e.g., `HELM_VERSION=3.17.3`) to eliminate the GitHub API dependency entirely and make builds reproducible. The existing kubectl has the same floating-version pattern, so this is consistent — but a pin would be more robust for CI.

2. The PR description says "same pattern as kubectl" — this is slightly misleading since kubectl uses a dedicated `stable.txt` endpoint (plain text, no rate limiting) while helm uses the GitHub releases API (JSON, rate-limited, grep/sed parsing). Not wrong in spirit but worth noting the reliability difference.

kalantar:
Added the empty-version guard (8e4644a). If the GitHub API call fails, the build now aborts with a clear error instead of producing a confusing 404 on a malformed helm URL.

The version-pinning and documentation observations are noted but not acted on — consistent with the existing kubectl pattern and not blocking.

kalantar:
## Verdict: READY TO MERGE

Prior review feedback addressed — the `test -n "$HELM_VERSION"` guard (commit 8e4644a) correctly fails the build with a clear error message if the GitHub API call produces an empty version string.

No remaining silent failure paths. Every command in the chain fails loudly on error (`curl -f`, `sha256sum --check`, `&&` chaining). The CI smoke test provides belt-and-suspenders verification that the binary is functional in the final image.

All issue #182 requirements are met.

PR #189, opened 2026-05-20
feat(deploy): add build subcommand with source hash staleness detection

### LINKED ISSUE #185: build-epp.sh hardcodes llm-d-inference-scheduler directory name
## Problem

`pipeline/scripts/build-epp.sh` hardcodes `llm-d-inference-scheduler` as the local directory name for the component source in three places:

- **Line 47:** `SCHEDULER_DIR="${SCHEDULER_ROOT}/llm-d-inference-scheduler"`
- **Line 106:** `kubectl cp "${SCHEDULER_DIR}/" "${NAMESPACE}/source-copy:/workspace/source/${RUN_NAME}/llm-d-inference-scheduler"`
- **Lines 135-137:** buildkit `context` and `dockerfile` paths use `/workspace/source/${RUN_NAME}/llm-d-inference-scheduler`

When an experiment uses a different component repo (e.g. `component.repo: llm-d-router` in `transfer.yaml`), the local submodule directory is `llm-d-router/`, not `llm-d-inference-scheduler/`. The script fails with:

```
error: .../exponential-ceiling/llm-d-inference-scheduler doesn't exist in local filesystem
```

## Expected Behavior

The script should derive the local directory name from `component.path` (which `manifest.py` resolves as the last segment of `component.repo`). This value is already available — `run_metadata.json` stores `repo_name`, and the manifest resolution produces `component.path`.

## Fix

Read `component.path` from `run_metadata.json` (or fall back to `repo_name`) and use it instead of the hardcoded string in all three locations. The remote path on the PVC and the buildkit context/dockerfile args should also use the resolved name for consistency.

## Affected Files

- `pipeline/scripts/build-epp.sh:47,106,135-137`

## Reproduction

```bash
# With transfer.yaml containing component.repo: llm-d-router
python pipeline/deploy.py run --experiment-root ../exponential-ceiling --workload chatbot_under
```

### LINKED ISSUE #187: build-epp: generalize to ensure-image helper, called per scenario
## Problem

Currently `build-epp.sh` only builds the treatment's custom EPP image. But:

1. Baseline scenarios may also reference non-release images that don't exist in the registry
2. There can be multiple baselines (and eventually multiple treatments)
3. The same logic applies to any image in the `images` section, not just `inferenceScheduler`

### `deploy build` subcommand

A new subcommand that ensures all required images exist before dispatching PipelineRuns. `deploy run` calls it implicitly as pre-flight (replacing current Step 1). Can also be invoked standalone.

### Single "ensure image exists" helper

A helper that, given an image reference and source directory:

1. **Check registry** — probe whether the image already exists (e.g. `crane manifest` or on-cluster `skopeo inspect`). If it exists, return immediately.
2. **Build and push** — if missing, run the existing BuildKit flow to build and push.
3. **Record source hash** — store the component directory's git hash in `run_metadata.json` after successful build.

### Staleness detection via source hash

Each image's "current source state" is captured as a git hash of the component directory (`git rev-parse HEAD` or tree hash). After a successful build, the hash is stored in `run_metadata.json` under a `source_hashes` dict keyed by image ref.

On subsequent runs:
- **Hash matches stored** → image is current, skip build
- **Hash mismatch or no stored hash** → rebuild

This handles retranslation automatically: the translate skill modifies files in the component repo, changing the hash, which triggers a rebuild of the treatment image without requiring `--force`.

### Image tagging

- **Baseline images** — tagged by the component ref (e.g. `ghcr.io/kalantar/llm-d-router:a10f9ac8`). Built once, cached in registry forever.
- **Treatment images** — tagged by run name (e.g. `ghcr.io/kalantar/llm-d-router:r1`). Rebuilt when source hash changes (i.e. after retranslation).

### Building both baseline and treatment from the same component directory

The component directory can only be in one state at a time. After translation, it contains the treatment's generated files. To build the baseline image (at the base ref), the build step must temporarily restore the directory to its pre-translation state.

**Mechanism — leveraging the workspace file cache:**

The workspace already caches translated files: `translation_output.json` lists `files_created` and `files_modified`, and `run_manager.py:switch_run` knows how to copy them back into the component directory. The build step reuses this:

1. **Restore to baseline state:**
   - Delete files listed in `translation_output.json:files_created`
   - `git checkout {component.ref}` to restore `files_modified` to their base state
   - Component directory is now clean at the baseline ref

2. **Build baseline image** (tagged by ref) — only if registry check shows it's missing

3. **Restore to treatment state:**
   - Re-apply generated files from the workspace cache (same operation as `run.py switch`)
   - Component directory is back to treatment state

4. **Build treatment image** (tagged by run name) — only if source hash changed

**Common case optimization:** The registry check (step 2) almost always passes for the baseline — it was built on the first run and the ref never changes. So steps 1 and 3 (the directory toggle) are skipped entirely in the typical case. The checkout dance only happens once per unique baseline ref.

### Orchestrator integration

`deploy build` (called as pre-flight by `deploy run`):

1. Iterates over all resolved scenarios (baseline and treatment)
2. For each scenario with an explicit `images` section, collects image references
3. Deduplicates (multiple scenarios may share the same image)
4. For each unique image: registry check → build if needed (using the toggle mechanism above)

The orchestrator dispatch loop is untouched — by the time it starts, all images are guaranteed to exist.

### Scope

**Phase 1:** Limit to `images.inferenceScheduler` — this is the only image we currently build from source. The helper interface should not assume this specifically (takes image ref + source path + Dockerfile name), but callers only pass scheduler images for now.

**Phase 2:** Generalize to any image in `images`.

## Affected Files

- `pipeline/scripts/build-epp.sh` → refactor into a generic build helper (image ref, source path, Dockerfile)
- `pipeline/deploy.py` → add `deploy build` subcommand; replace `_resolve_epp_action` / `_build_epp_image` with scenario iteration + ensure-image calls
- `pipeline/lib/run_manager.py` → reuse `switch_run` file-restore logic for the treatment restore step
- `run_metadata.json` schema → add `source_hashes` dict

## CLI

```
deploy build [--force] [--skip-baseline]    # Ensure all scenario images exist
deploy run                                   # Calls deploy build as pre-flight, then dispatches
deploy run --skip-build                      # Skip pre-flight entirely (replaces --skip-build-epp)
```

kalantar:
## Summary

- Adds a `deploy build` subcommand that ensures all required scenario images exist before dispatching PipelineRuns
- Introduces source hash staleness detection: images are only rebuilt when the component source directory has changed (by comparing git HEAD hash against a stored hash in `run_metadata.json`)
- Refactors `build-epp.sh` to accept explicit `--image-ref` and `--source-dir` parameters (also fixes #185 — hardcoded directory name)
- Renames `--skip-build-epp` to `--skip-build` across the CLI

Closes #187

## Changes

| File | What |
|------|------|
| `pipeline/lib/ensure_image.py` (new) | Source hash computation, persistence, and staleness check |
| `pipeline/lib/source_toggle.py` (new) | Toggle component dir between baseline/treatment states (Phase 2 prep) |
| `pipeline/scripts/build-epp.sh` | Accepts `--image-ref` + `--source-dir`; uses basename for PVC paths |
| `pipeline/deploy.py` | Replaces `_resolve_epp_action`/`_build_epp_image` with `_cmd_build`; adds `build` subcommand |
| `pipeline/lib/remote.py` | `--skip-build-epp` → `--skip-build` in orchestrator Job args |
| `pipeline/tests/test_ensure_image.py` (new) | 10 tests for source hash module |
| `pipeline/tests/test_source_toggle.py` (new) | 9 tests for directory toggle module |
| `pipeline/tests/test_deploy_build.py` (new) | 4 tests for CLI parser |
| `pipeline/tests/test_deploy_run.py` | Updated: old `_resolve_epp_action` tests replaced with `_cmd_build` tests |
| `pipeline/tests/test_deploy_remote.py` | Updated: `skip_build_epp` → `skip_build` |
| `pipeline/tests/test_remote.py` | Updated assertion for renamed flag |

## Reviewer notes

- **Phase 1 scope**: Only the treatment `inferenceScheduler` image is handled. The `source_toggle.py` module is available for Phase 2 (baseline image building with directory toggle) but not yet wired into `_cmd_build`.
- **No `--force` flag**: Source hash detection handles staleness automatically. If the source changes, the hash changes, and rebuild triggers. Registry corruption is rare enough to handle manually (delete tag, re-run).
- **Idempotent**: If `deploy build` is interrupted, re-running picks up cleanly — the registry check detects the missing image and rebuilds.
- **Backward compat**: `build-epp.sh` still works without the new flags (falls back to `run_metadata.json` derivation).

## Test plan

- [x] All 750 tests pass (`python3 -m pytest pipeline/ .claude/skills/ -v`)
- [x] Lint clean (`ruff check pipeline/ .claude/skills/ --select F`)
- [x] `deploy.py build --help` shows the new subcommand
- [x] `deploy.py run --skip-build` works (old `--skip-build-epp` rejected)
- [ ] End-to-end: run `deploy build` against a cluster with a real experiment repo

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
-"""sim2real deploy — Build EPP, orchestrate runs, collect results.
+"""sim2real deploy — Ensure images, orchestrate runs, collect results.
 
 Subcommands:
-  run      Build EPP image, submit PipelineRuns
+  build    Ensure all scenario images exist (pre-flight for run)
+  run      Ensure images + submit PipelineRuns
   status   Show progress of all (workload, package) pairs
   collect  Pull results from cluster for completed phases
   stop     Stop the remote orchestrator Job
@@ -2123,13 +2124,14 @@ def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
 def build_parser() -> argparse.ArgumentParser:
     p = argparse.ArgumentParser(
         prog="deploy.py",
-        description="sim2real deploy — Build EPP, orchestrate runs, collect results",
+        description="sim2real deploy — Ensure images, orchestrate runs, collect results",
         formatter_class=argparse.RawDescriptionHelpFormatter,
         epilog="""
 Examples:
-  python pipeline/deploy.py run                        # Build EPP + orchestrate all pairs
+  python pipeline/deploy.py build                      # Ensure all scenario images exist
+  python pipeline/deploy.py run                        # Ensure images + orchestrate all pairs
   python pipeline/deploy.py run --remote               # Submit orchestrator as in-cluster Job
-  python pipeline/deploy.py run --skip-build            # Orchestrate without image build
+  python pipeline/deploy.py run --skip-build           # Orchestrate without image build
   python pipeline/deploy.py status                     # Show progress snapshot
   python pipeline/deploy.py collect                    # Pull results for completed phases
   python pipeline/deploy.py collect --skip-logs        # Collect traces only (skip large logs)
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -109,41 +109,58 @@ def _load_setup_config() -> dict:
     return {}
 
 
-# ── EPP build decision ─────────────────────────────────────────────────────────
+# ── Image build ───────────────────────────────────────────────────────────────
 
-def _resolve_epp_action(run_dir: Path, skip_build_epp: bool) -> str:
-    """Determine EPP build action based on run metadata.
+def _cmd_build(run_dir: Path, namespace: str, skip_build: bool) -> str:
+    """Ensure all required scenario images exist. Returns 'built', 'skip', or 'current'.
 
-    Returns one of: "build", "skip", "error:<message>"
+    This is the 'deploy build' logic. Called implicitly by 'deploy run' as pre-flight.
     """
+    from pipeline.lib.ensure_image import (
+        compute_source_hash, image_needs_build, save_source_hash,
+    )
+
     run_meta_path = run_dir / "run_metadata.json"
     if not run_meta_path.exists():
-        return "error:run_metadata.json not found — run setup.py first."
+        err("run_metadata.json not found — run setup.py first.")
+        sys.exit(1)
     try:
         run_meta = json.loads(run_meta_path.read_text())
     except json.JSONDecodeError as e:
-        return f"error:run_metadata.json is not valid JSON: {e}. Re-run setup.py."
+        err(f"run_metadata.json is not valid JSON: {e}. Re-run setup.py.")
+        sys.exit(1)
 
     component_image = run_meta.get("component_image")
-    if component_image is None:
-        return "skip"
     if not component_image:
-        return "error:component_image is empty in run_metadata.json. Re-run setup.py with a valid --registry."
-    if skip_build_epp:
+        info("No component_image in run metadata — skipping image build")
         return "skip"
-    return "build"
 
+    if skip_build:
+        info("--skip-build: skipping image build")
+        return "skip"
 
-# ── EPP Image build ─────────────────────────────────────────────────────────
+    step(1, "Ensure Images")
 
-def _build_epp_image(run_dir: Path, run_name: str, namespace: str) -> str:
-    """Build EPP image on-cluster. Returns the full image reference."""
-    step(1, "Build EPP Image")
+    registry = run_meta.get("registry", "")
+    repo_name = run_meta.get("repo_name", "llm-d-inference-scheduler")
+    run_name = run_dir.name
+    source_dir = EXPERIMENT_ROOT / repo_name
 
-    # Locate build-epp.sh
+    if not source_dir.exists():
+        err(f"Component source directory not found: {source_dir}")
+        sys.exit(1)
+
+    # Treatment image: tagged by run name
+    treatment_ref = f"{registry}/{repo_name}:{run_name}"
+
+    if not image_needs_build(run_dir, treatment_ref, source_dir):
+        ok(f"Treatment image current (hash unchanged): {treatment_ref}")
+        return "current"
+
+    info(f"Building treatment image: {treatment_ref}")
     build_script = REPO_ROOT / "pipeline" / "scripts" / "build-epp.sh"
     if not build_script.exists():
-        err(f"build-epp.sh not found at {build_script.relative_to(REPO_ROOT)} — ensure pipeline/scripts/build-epp.sh is present in the repo")
+        err(f"build-epp.sh not found at {build_script.relative_to(REPO_ROOT)}")
         sys.exit(1)
 
     result = run(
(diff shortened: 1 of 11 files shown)


kalantar:
## Review: NEEDS CHANGES

Good structure overall — the source hash staleness detection is well-designed, tests for the new library modules are solid, and the CLI rename is handled consistently. Three issues should be addressed before merge:

### Blocking

1. **Behavioral regression: empty `component_image` silently skips instead of erroring** (`deploy.py:_cmd_build`)

   The old code explicitly distinguished `component_image is None` (key absent → skip) from `component_image == ""` (misconfigured → error). The new `if not component_image` treats both as skip. A user with a misconfigured setup.py (empty registry) will silently skip the build and get confusing runtime failures in the cluster.

   Fix: restore the distinction:
   ```python
   if component_image is None:
       info("No component_image in run metadata — skipping image build")
       return "skip"
   if not component_image:
       err("component_image is empty in run_metadata.json — re-run setup.py with a valid --registry.")
       sys.exit(1)
   ```

2. **`restore_treatment` basename collision risk** (`source_toggle.py`)

   The function maps `rel_path` → `generated_dir / Path(rel_path).name`. If two files share a basename (e.g. `pkg/scorer/config.go` and `internal/config.go`), one silently overwrites the other. Not wired in yet (Phase 2 prep), but should either validate uniqueness or document the flat-directory assumption in the docstring.

3. **`save_source_hash` non-atomic write** (`ensure_image.py`)

   Read-modify-write on `run_metadata.json` without atomicity. An interrupted write (kill, disk full) corrupts the file, forcing the user to re-run setup.py. Use write-to-temp + `os.replace`:
   ```python
   tmp_path = meta_path.with_suffix(".json.tmp")
   tmp_path.write_text(json.dumps(meta, indent=2))
   tmp_path.replace(meta_path)
   ```

### Non-blocking suggestions

- **Validate `registry` is non-empty** in `_cmd_build` before constructing `treatment_ref` — catches misconfiguration early instead of failing deep in the build pod.
- **`image_needs_build`**: catch `subprocess.CalledProcessError` from `compute_source_hash` and return `True` (rebuild to be safe) rather than crashing the pipeline on a transient git issue.
- **Docstrings**: "Ensure all scenario images" appears in 4 places but only treatment is handled. Consider "Ensure required component images exist" or note the Phase 1 limitation.
- **`build-epp.sh` header**: update Usage line to include `--image-ref` and `--source-dir`.
- **Test gap**: `_cmd_build` happy path ("built" and "current" returns) and build-failure-does-not-save-hash are untested. The core value of this PR — staleness detection triggering a build — has no unit coverage beyond the library functions.
- **Step 7 metadata**: `epp_image` and `stages.deploy.last_completed_step` are no longer written in the new flow. Verify no downstream code reads these, or have `_cmd_build` write them.

kalantar:
**Correction on item 2 (basename collision):** The `generated_dir / Path(rel_path).name` pattern already exists in `run_manager.py:switch_run` (line 343), which also has a collision guard at lines 289-298. `source_toggle.py` extracts the same logic without the guard — but this isn't a new risk introduced by the PR. Downgrading from blocking to non-blocking: consider carrying over the collision check from `run_manager.py` for consistency, but it's not a regression.

kalantar:
Addressed the review in commit 7e660e9. Here's what changed:

**Fixed:**

- Restored the `None` vs empty-string distinction for `component_image` — empty string now errors with a clear message instead of silently skipping. Added a test for this case.
- `save_source_hash` now uses atomic write (write-to-tmp + `os.replace`) to avoid metadata corruption on interrupted writes.
- Added `registry` validation before constructing the image ref — catches misconfiguration early.
- `image_needs_build` now catches `subprocess.CalledProcessError` from git and returns `True` (rebuild to be safe) instead of crashing.
- Updated `build-epp.sh` usage header to document the new `--image-ref` and `--source-dir` flags.

**Follow-ups filed:**

- #191 — `epp_image`/`last_completed_step` no longer written to metadata (informational display regression in `run.py inspect`)
- #192 — `source_toggle.py` should carry over the basename collision guard from `run_manager.py`
- #193 — Add unit tests for `_cmd_build` happy-path and build-failure-no-hash scenarios

**Skipped:**

- Docstring wording ("Ensure all scenario images" vs "Ensure required component images") — Phase 1 limitation is documented in the PR description and will resolve naturally in Phase 2.

kalantar:
## Re-review: READY TO MERGE

All three blocking items from the previous review have been addressed in the latest commit:

1. Empty `component_image` now correctly errors instead of silently skipping (with test coverage)
2. `save_source_hash` uses atomic write (tmp file + `os.replace`)
3. Empty `registry` validated early with clear error message
4. `image_needs_build` catches `CalledProcessError` and falls back to rebuild
5. `build-epp.sh` header updated with new flags

Remaining non-blocking items (tracked separately):
- `_cmd_build` happy path test coverage — low risk since the library functions are well-tested
- Basename collision in `source_toggle.py` — pre-existing pattern, tracked in #190
- "All scenario images" docstring is aspirational (Phase 1 only handles treatment) — acceptable

LGTM.

PR #198, opened 2026-05-21
feat: reduce orchestrator log noise via state-transition gating

### LINKED ISSUE #197: Reduce orchestrator log noise: only log on state transitions
## Problem

During the `deploy.py run` polling loop, the orchestrator logs capacity and dispatch-intent lines on **every iteration**, even when nothing has changed and no slots are free. This produces a wall of repeated lines like:

```
[INFO]  Capacity: 31 free GPUs (93 allocatable − 62 requested)
[INFO]  Dispatching 7/13 pending pairs (capacity-limited: 31 free GPUs)
[INFO]  Capacity: 31 free GPUs (93 allocatable − 62 requested)
[INFO]  Dispatching 7/13 pending pairs (capacity-limited: 31 free GPUs)
...
```

## Proposed Solution

Summarize on state change rather than logging every poll iteration:

1. When entering a "waiting" state (all slots busy, or capacity insufficient), log **once**: e.g., `Waiting for slots to free — 13 pairs pending, 3 slots busy`
2. Only log again when something **changes**: a slot frees up, capacity shifts meaningfully, a dispatch succeeds, or a workload completes.
3. Keep existing log lines for actual events (successful dispatch, completion, errors, backoff transitions).

## Affected Code

`pipeline/deploy.py` lines ~1680–1735 — the capacity probe and dispatch-intent section of the orchestrator loop.

## Acceptance Criteria

- Repeated identical capacity/dispatch lines are eliminated during steady-state polling
- All actionable events (dispatches, completions, slot changes, backoff transitions) still log at INFO
- No loss of diagnostic information — the current state is always knowable from the most recent log line

kalantar:
Closes #197

## Summary

- Introduces a `_last_log_state` dict in the orchestrator loop that tracks the last-logged values for capacity, backoff-level, backoff-skip, and dispatch-intent log lines
- Only emits these lines when the relevant state changes — eliminating walls of repeated output during steady-state polling
- Clears dedup state on actual events (successful dispatch, completion, failure, reclaim, timeout) so the next iteration always re-logs the new reality

## Reviewer attention

The pattern mirrors the existing `_last_probe_error` dedup already in place for capacity probe failures (line 1693 on main). Each log site uses a keyed tuple comparison so the gating is local and easy to trace.

The five log lines that are now gated:
1. `Capacity: N free GPUs (...)` — every-iteration when pending
2. `Backoff level N — next poll in Xs` — every-iteration during sustained scarcity
3. `Backoff: skipping dispatch (...)` — every-iteration during backoff
4. `Dispatching 0/N pending pairs — smallest cost exceeds free GPUs` — stuck state
5. `Dispatching N/M pending pairs (capacity-limited/slot-limited)` — constrained state

Transition-only logs (scarcity entry, backoff resume, probe recovery) were already guarded and remain unchanged.

## Test plan

- [x] `python3 -m pytest pipeline/ -v` — 707 tests pass
- [x] New `pipeline/tests/test_log_dedup.py` covers dedup logic, state changes, and event-triggered clears

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1507,6 +1507,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     _probe_fail_count = 0
     _last_probe_error = ""
     _last_log_state: dict[str, object] = {}
+    _zero_dispatch_count = 0
 
     # Load or initialize progress
     progress = store.load()
@@ -1608,6 +1609,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 _last_log_state.pop("capacity", None)
                 _last_log_state.pop("dispatch", None)
                 _last_log_state.pop("backoff_skip", None)
+                _zero_dispatch_count = 0
 
             elif status in ("Failed", "PipelineRunCancelled", "PipelineRunCouldntGetPipeline",
                             "PipelineRunTimeout", "CreateRunFailed", "PipelineRunStopping",
@@ -1620,6 +1622,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 _last_log_state.pop("capacity", None)
                 _last_log_state.pop("dispatch", None)
                 _last_log_state.pop("backoff_skip", None)
+                _zero_dispatch_count = 0
 
             elif status in ("Running", "Started"):
                 # Check for pending pods (before timeout)
@@ -1648,6 +1651,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     _last_log_state.pop("capacity", None)
                     _last_log_state.pop("dispatch", None)
                     _last_log_state.pop("backoff_skip", None)
+                    _zero_dispatch_count = 0
                     progress["_orchestrator"] = backoff.to_dict()
                     store.save(progress)
                     continue
@@ -1656,6 +1660,11 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     if backoff.state != "normal":
                         backoff.signal_scheduling_success()
                         info(f"[{pair_key}] Pending→Running → backoff reset")
+                        _last_log_state.pop("capacity", None)
+                        _last_log_state.pop("dispatch", None)
+                        _last_log_state.pop("backoff_skip", None)
+                        _last_log_state.pop("backoff_level", None)
+                        _zero_dispatch_count = 0
                         progress["_orchestrator"] = backoff.to_dict()
                         store.save(progress)
 
@@ -1686,6 +1695,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                             _last_log_state.pop("capacity", None)
                             _last_log_state.pop("dispatch", None)
                             _last_log_state.pop("backoff_skip", None)
+                            _zero_dispatch_count = 0
                             store.save(progress)
                     except ValueError:
                         pass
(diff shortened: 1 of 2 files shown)


kalantar:
## Review: NEEDS CHANGES

The overall approach is sound — state-transition gating with event-triggered clears is the right pattern, and it mirrors the existing `_last_probe_error` dedup. The code is well-scoped and the acceptance criteria are mostly met. Two issues need addressing before merge.

### Required changes

**1. Don't fully suppress the stuck-dispatch warning.**

The `warn("Dispatching 0/N pending pairs — smallest cost exceeds free GPUs")` is gated the same way as INFO lines. But warnings indicate ongoing problems requiring attention — silencing them indefinitely after the first occurrence makes the situation read as "resolved" to an operator watching logs during a multi-hour GPU scarcity event. Either:
- Exempt this specific `warn()` from dedup entirely, or
- Re-emit it periodically (e.g., every 10 iterations), similar to the existing `_probe_fail_count % 10 == 0` pattern at line 1693.

**2. Add state clear on Pending→Running backoff reset.**

The `signal_scheduling_success()` path (Pending→Running transition) resets the backoff controller but does NOT clear `_last_log_state`. Compare with the `signal_reclaim()` path just above it, which DOES clear state. If the system re-enters backoff at identical parameters after a successful scheduling, the backoff-skip and dispatch messages will be suppressed because old dedup state persists. Add the same three `.pop()` calls after the backoff reset.

### Suggested (non-blocking)

- Include `smallest` in the zero-dispatch dedup tuple: `("zero", len(pending), free_gpus, smallest)` — otherwise the cached warning message shows a stale GPU cost if the pending set composition changes without changing its count.
- Consider clearing `backoff_level` and `backoff_skip` at the `signal_capacity` recovery path (line ~1731, "resuming normal dispatch") for the same reason as #2.
- The tests re-implement the dedup logic in local helpers rather than exercising the production code. This validates the algorithm but not the actual implementation. Consider extracting the dedup check into a small importable helper (`_should_log(state_dict, key, new_state) -> bool`) that both production and tests call — this also eliminates the repeated 3-line pattern at each gated site.

### What looks good

- The flat dict with keyed tuple comparisons is clean and auditable — appropriate for a local, single-function concern
- The five event-clear sites (completion, failure, reclaim, timeout, dispatch) are correctly placed at slot-topology-changing events
- Transition-only logs (scarcity entry, backoff resume, probe recovery) are correctly left ungated
- The PR body's claim about mirroring `_last_probe_error` is accurate
- No documentation inaccuracies introduced (README doesn't describe per-iteration log behavior)

kalantar:
Pushed fixes in 8ac9f29:

**Zero-dispatch warning periodic re-emission:** The `warn("Dispatching 0/N...")` now re-emits every 10 iterations during sustained GPU scarcity, following the same `% 10` pattern already used for probe failures. Also added `smallest` to the dedup tuple so a change in pending-set composition (different cheapest pair) triggers a fresh warning.

**Pending→Running state clear:** Added the missing `_last_log_state` clears at the `signal_scheduling_success()` path — this was an oversight since the adjacent `signal_reclaim()` path already had them.

**signal_capacity recovery clear:** Added `backoff_level` and `backoff_skip` clears at the "resuming normal dispatch" transition so that if backoff re-engages at identical parameters later, the messages aren't suppressed.

**Helper extraction (S3):** Filed as #201 — valid improvement but refactoring scope for a separate PR.

Tests: 709 pass, lint clean.

kalantar:

All three acceptance criteria from #197 are satisfied. The dedup pattern is clean, correctly mirrors the existing `_last_probe_error` precedent, and state clears cover all meaningful event boundaries.

### Non-blocking suggestions

1. **Fix test state tuple mismatch**: `test_repeated_capacity_limited_not_logged` uses a 3-element tuple for zero-dispatch state, but production uses 4 elements (includes `smallest`). Consider updating the test helper to match production, reducing future drift risk.

2. **Clear capacity state on probe recovery**: If the capacity probe fails then recovers to the same values, the capacity log is suppressed. Adding `_last_log_state.pop("capacity", None)` alongside the existing "Capacity probe recovered" message would ensure a full capacity line always follows recovery.

3. **Consider documenting the periodic re-emit asymmetry**: Zero-dispatch re-emits every 10 iterations (good — it's the stuck state). Capacity-limited and slot-limited don't. This is a reasonable severity-based choice but could benefit from a brief comment explaining why.

None of these block merge. The implementation cleanly solves the noise problem described in #197.

PR #202, opened 2026-05-21
feat: improve deploy collect output with per-pair reporting (#199)

### LINKED ISSUE #199: deploy collect: improve per-pair output and simplify summary
## Problem

`deploy.py collect` output doesn't identify which workload (pair) is being collected. Current output:

```
━━━ Step 1: Collecting Results ━━━
pod/sim2real-extract created
[OK]    Collected: baseline1
pod/sim2real-extract created
[OK]    Collected: treatment
pod/sim2real-extract created
[OK]    Collected: baseline1

  Collected: 2/2 phases
  Results:   /Users/.../workspace/runs/r1/results/baseline1/  results/treatment/
```

Issues:
1. Per-line output shows only the phase (package) name, not the workload — unclear which pair was collected
2. `baseline1` appears twice with no explanation (it was in two different namespace slots)
3. The summary "Results:" line re-enumerates phases that are already visible in the lines above
4. "2/2 phases" is confusing — the denominator is always equal unless there's a failure, and "phases" is redundant with the lines above

## Desired output

```
━━━ Step 1: Collecting Results ━━━
[OK]    baseline1/wl-heavy    (ns-slot-1)
[OK]    treatment/wl-heavy    (ns-slot-1)
[OK]    baseline1/wl-smoke    (ns-slot-0)
[OK]    treatment/wl-smoke    (ns-slot-0)
[OK]    baseline1/wl-load     (ns-slot-2)

  Collected: 5/5 pairs
  Results:   workspace/runs/r1/results/
```

On failure:
```
[OK]    baseline1/wl-heavy    (ns-slot-1)
[FAIL]  baseline1/wl-load     (ns-slot-2): extractor pod not ready

  Collected: 4/5 pairs
  Results:   workspace/runs/r1/results/
```

## Changes

- Include `phase/workload` in each OK/FAIL line (the info is already available in the progress entries)
- Add namespace slot in parentheses for traceability
- Summary "Results:" line shows only the common prefix path — the per-pair paths are already in the lines above
- Drop "N phases" from the summary (redundant)
- Keep `collected/total` fraction (useful in failure case)
- Remove `pod/sim2real-extract created` noise from kubectl (or demote to verbose/debug)

## Affected files

- `pipeline/deploy.py` — `_cmd_collect` and `_extract_phases_from_pvc` reporting
- `pipeline/tests/test_deploy_collect.py` — update output assertions
- `pipeline/README.md` — if collect output examples exist

kalantar:
Closes #199

## Summary

- Each collected result now reports `phase/workload    (namespace)` instead of just the phase name — making it clear which pair was collected and from which cluster slot
- Summary simplified from "N/M phases" + per-phase path list to "N/M pairs" + single results root path
- Suppressed `pod/sim2real-extract created` kubectl noise that cluttered output

## Before

```
━━━ Step 1: Collecting Results ━━━
pod/sim2real-extract created
[OK]    Collected: baseline1
pod/sim2real-extract created
[OK]    Collected: treatment

  Collected: 2/2 phases
  Results:   /path/results/baseline1/  results/treatment/
```

## After

```
━━━ Step 1: Collecting Results ━━━
[OK]    baseline/smoke    (ns-slot-0)
[OK]    treatment/smoke    (ns-slot-0)
[OK]    baseline/load    (ns-slot-1)

  Collected: 3/3 pairs
  Results:   /path/results/
```

## Reviewer attention

- The unscoped path now builds `ns_workload_map` alongside `ns_phase_map` to know which workloads live in each namespace. This is derived from the same progress entries already being iterated — no additional data source needed.
- Both scoped and unscoped paths now define `collected_pairs`, `failed_pairs`, `total_pairs` which the shared summary consumes. The existing phase-level `collected`/`failed` lists are preserved for the scoped path's filtering logic.
- The no-progress fallback path (no ConfigMap data) reports per-phase since workload info isn't available in that scenario.

## Test plan

- [x] 703 pipeline tests pass (`python -m pytest pipeline/ -v`)
- [x] 3 new tests verify per-pair output format, failure counts, and namespace context
- [x] All 39 pre-existing collect tests pass without modification

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -637,7 +637,7 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
         }
     })
     run(["kubectl", "run", pod_name, "--image=alpine:3.19", "--restart=Never",
-         "--overrides", overrides, "-n", namespace])
+         "--overrides", overrides, "-n", namespace], capture=True)
 
     result = run(
         ["kubectl", "wait", f"pod/{pod_name}", "--for=condition=Ready",
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -867,6 +867,9 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         step(1, "Collecting Results")
         collected: list[str] = []
         failed: list[str] = []
+        collected_pairs: list[str] = []
+        failed_pairs: list[str] = []
+        total_pairs = len(collectible)
 
         skip_logs = getattr(args, "skip_logs", False)
         for wl_name in scoped_workloads:
@@ -883,6 +886,7 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                 for p in phases_to_collect:
                     if p not in failed:
                         failed.append(p)
+                    failed_pairs.append(f"{p}/{wl_name}")
                 continue
             try:
                 errors = _extract_phases_from_pvc(
@@ -893,16 +897,19 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                 for p in phases_to_collect:
                     if p not in failed:
                         failed.append(p)
+                    failed_pairs.append(f"{p}/{wl_name}")
             else:
                 for phase, exc in errors.items():
                     if exc is None:
-                        ok(f"Collected: {phase}/{wl_name}")
+                        ok(f"{phase}/{wl_name}    ({wl_ns})")
                         if phase not in collected:
                             collected.append(phase)
+                        collected_pairs.append(f"{phase}/{wl_name}")
                     else:
-                        warn(f"Extraction failed for {phase}/{wl_name}: {exc}")
+                        warn(f"{phase}/{wl_name}    ({wl_ns}): {exc}")
                         if phase not in failed:
                             failed.append(phase)
+                        failed_pairs.append(f"{phase}/{wl_name}")
 
         collected = [p for p in collected if p not in failed]
     else:
@@ -945,6 +952,8 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         step(1, "Collecting Results")
         collected = []
         failed = []
+        collected_pairs: list[str] = []
+        failed_pairs: list[str] = []
 
         if phases_to_collect:
             skip_logs = getattr(args, "skip_logs", False)
@@ -953,6 +962,7 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                 # Entries without completed_namespace were written by an older
                 # version of the orchestrator that did not record it.
                 ns_phase_map: dict[str, list[str]] = {}
+                ns_workload_map: dict[str, set[str]] = {}
                 missing_ns_keys: list[str] = []
                 for key, pentry in progress.items():
                     if not isinstance(pentry, dict):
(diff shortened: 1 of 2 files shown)


kalantar:

The PR cleanly addresses most requirements from #199 (per-pair output, namespace context, summary simplification, kubectl noise suppression). However, the unscoped-with-progress code path has a correctness bug that will produce wrong output in the common case where namespace slots are reused.

### Blocking issue

**Cross-product bug in unscoped path**: The code builds `ns_phase_map` (namespace → phases) and `ns_workload_map` (namespace → workloads) independently, then iterates the Cartesian product of phases × workloads per namespace for reporting. When a namespace slot is reused for different workloads (the common case when #pairs > #slots), this over-reports. Example: if ns-0 completed baseline/smoke, treatment/smoke, and baseline/load (reused after smoke freed it), the cross-product emits a phantom "treatment/load (ns-0)" line that never existed. The `total_pairs` denominator (computed from actual progress entries) then disagrees with the numerator, producing impossible fractions like "4/3 pairs".

**Fix**: Track actual `(phase, workload)` pairs per namespace from the progress entries instead of cross-producting independent sets:
```python
ns_pair_map: dict[str, set[tuple[str, str]]] = {}
# in the loop:
ns_pair_map.setdefault(ns, set()).add((pkg, wl))
# after extraction:
for phase, exc in errors.items():
    wls_for_phase = [wl for (p, wl) in ns_pair_map.get(ns, set()) if p == phase]
    for wl in wls_for_phase:
        ...
```

### Non-blocking observations

1. **Empty workload field**: If all progress entries for a namespace lack a `workload` key (possible with older orchestrator versions), `ns_workloads` is empty and zero pair lines are printed despite successful extraction. Consider falling back to phase-level reporting when no workload info exists for a namespace.

2. **`capture=True` error path**: The kubectl run command now captures stdout, but if it fails with a `CalledProcessError`, the error output is buried in the exception object rather than printed inline. The existing `except RuntimeError` handler doesn't catch `CalledProcessError`. Consider converting to the same `check=False` + explicit raise pattern used by the adjacent `kubectl wait` call.

3. **Test precision**: The assertion `"1 pairs" in captured.out` in `test_collect_unscoped_failure_reports_pair_count` could match inside `"1/2 pairs"`. Use `"Failed:    1 pairs"` for specificity.

kalantar:

**Fixed** the cross-product bug. Replaced the independent `ns_workload_map` (namespace → workloads) with `ns_pair_map` (namespace → set of actual (phase, workload) tuples). After extraction, only workloads that actually ran for each phase in that namespace are reported. Added a regression test (`test_collect_unscoped_no_cross_product_on_slot_reuse`) that verifies no phantom pairs appear when a namespace slot is reused.

**Fixed** the imprecise test assertion — changed `"1 pairs"` to `"Failed:    1 pairs"` so it can't false-match inside the collected count.

**Follow-ups filed** (not blocking):
- #203 — Fall back to phase-level reporting when progress entries lack the `workload` field (older orchestrator)
- #204 — Handle `CalledProcessError` from `kubectl run` in the extractor (pre-existing issue, not introduced here)

All 704 tests pass, lint clean.

kalantar:

The cross-product bug from the prior review has been correctly fixed — commit 4 replaces `ns_workload_map` with `ns_pair_map` tracking actual `(package, workload)` tuples, and a dedicated test (`test_collect_unscoped_no_cross_product_on_slot_reuse`) guards against regression. All 6 requirements from #199 are met.

### Non-blocking suggestions

1. **PR description stale**: "Reviewer attention" bullet 1 still says `ns_workload_map` — update to `ns_pair_map` to match commit 4.

2. **`total_pairs` vs `ns_pair_map` alignment for legacy data**: Entries lacking a `workload` field (from older orchestrators) are counted in `total_pairs` but can never appear in `collected_pairs` (the `if wl:` guard skips them from `ns_pair_map`). This could produce a misleading "0/N pairs" summary on legacy progress data. Consider adding `and pentry.get("workload")` to the `total_pairs` filter to keep the denominator consistent with what can actually be reported.

3. **`capture=True` error path**: The kubectl run command now captures stdout, which correctly suppresses the "pod/sim2real-extract created" noise. However, if the command fails, the `CalledProcessError` (not `RuntimeError`) propagates uncaught and its captured stderr is hard to find in the traceback. This is a pre-existing gap (the `except RuntimeError` handler never matched `CalledProcessError`) but `capture=True` makes it slightly worse. Consider switching to `check=False` with an explicit RuntimeError raise (matching the adjacent `kubectl wait` pattern) in a follow-up.

None of these block merge. The implementation is correct for the common path and the test coverage is solid.

PR #206, opened 2026-05-22
feat: parallelize deploy collect across namespace slots

### LINKED ISSUE #200: deploy collect: parallelize extraction across namespace slots
## Problem

`deploy.py collect` extracts results sequentially — one namespace slot at a time. Each slot requires creating a pod, waiting for it to be ready, copying files, and cleaning up. With N slots, collection time scales linearly.

## Context

Each namespace slot has its own PVC and its own extractor pod (namespaced, no name collision). Workload results write to distinct `results/<phase>/<workload>/` subdirectories locally, so there are no file-level conflicts between concurrent extractions from different slots.

## Proposal

Use `concurrent.futures.ThreadPoolExecutor` to extract from all namespace slots in parallel. The work is I/O-bound (kubectl calls), so threads are sufficient.

Key changes:
- Submit one extraction task per namespace slot
- Report `[OK]` / `[FAIL]` lines as each pair completes (order may vary by completion time)
- Aggregate `collected` / `failed` lists after all futures resolve
- Announce parallelism in the step header: `Collecting Results (3 slots in parallel)`

Output format (per #199):
```
━━━ Step 1: Collecting Results (3 slots in parallel) ━━━
[OK]    baseline1/wl-heavy    (ns-slot-1)
[OK]    treatment/wl-heavy    (ns-slot-1)
[OK]    baseline1/wl-smoke    (ns-slot-0)
[OK]    treatment/wl-smoke    (ns-slot-0)
[OK]    baseline1/wl-load     (ns-slot-2)

  Collected: 5/5 pairs
  Results:   workspace/runs/r1/results/
```

## Implementation notes

- The `collected` and `failed` lists need thread-safe aggregation (post-join merge or a lock)
- `print()` calls from threads may interleave mid-line — use a lock around output or buffer per-slot and flush on completion
- The size probe + interactive prompt (>1GB warning) would need to happen before parallel dispatch, or be disabled in parallel mode
- Fallback to sequential if only 1 namespace slot exists (avoid thread overhead)

## Depends on

- #199 (output format improvements — should land first so parallelization builds on the new format)

## Affected files

- `pipeline/deploy.py` — `_cmd_collect` unscoped path (lines 951–1009)
- `pipeline/tests/test_deploy_collect.py` — update for parallel behavior

kalantar:
Closes #200

## Summary

- Uses `concurrent.futures.ThreadPoolExecutor` to extract results from all namespace slots in parallel (one thread per slot)
- Falls back to sequential for single-slot case (avoids thread overhead)
- Step header announces slot count: `Collecting Results (3 slots in parallel)`
- Partial failures isolated: if one slot's extractor pod fails, other slots complete normally

## Design choices

**Post-join result processing:** Results from `as_completed()` are processed (OK/FAIL output + list aggregation) on the main thread after each future completes. This avoids needing locks on `collected`/`failed` lists or a print lock. Output order may vary by completion time — this is noted in the issue as acceptable.

**`_process_slot_result` helper:** Extracted the per-slot result handling (shared between parallel and sequential paths) into a nested function to avoid duplicating the OK/FAIL/list-append logic.

**No `_extract_phases_from_pvc` changes:** The function already operates on a single namespace with its own pod name — it's naturally thread-safe (no shared mutable state between calls from different namespaces).

## Reviewer attention

- The interactive size-probe prompt (>1GB warning) inside `_extract_phases_from_pvc` could race in parallel mode. Currently each thread would prompt independently — but in practice this is rare (the prompt only fires for >1GB data). A follow-up could move the size probe before parallel dispatch if needed.
- The scoped path (`--only`/`--workload`) still runs sequentially — it iterates per-workload rather than per-namespace, so parallelizing it is a different shape. Out of scope per the issue.

## Test plan

- [x] `python3 -m pytest pipeline/ -v` — 717 tests pass
- [x] New tests verify: ThreadPoolExecutor used for multi-slot, not used for single-slot, partial failure handled, step header includes slot count

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -949,7 +949,6 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         else:
             phases_to_collect = list(known_phases)
 
-        step(1, "Collecting Results")
         collected = []
         failed = []
         collected_pairs: list[str] = []
@@ -990,24 +989,28 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                     and pentry.get("completed_namespace")
                 )
 
+                ns_items = sorted(ns_phase_map.items())
+                if len(ns_items) > 1:
+                    step(1, f"Collecting Results ({len(ns_items)} slots in parallel)")
+                else:
+                    step(1, "Collecting Results")
+
                 for key in missing_ns_keys:
                     warn(f"{key}: completed_namespace missing — skipping (re-run the workload with a newer orchestrator to collect results)")
-                for ns, ns_phases in sorted(ns_phase_map.items()):
-                    pairs_in_ns = ns_pair_map.get(ns, set())
-                    try:
-                        errors = _extract_phases_from_pvc(
-                            sorted(ns_phases), run_name, ns, run_dir,
-                            skip_logs=skip_logs)
-                    except RuntimeError as e:
-                        warn(f"Extractor pod failed in {ns}: {e}")
-                        for p in ns_phases:
+
+                def _process_slot_result(ns, pairs_in_ns, result):
+                    """Process extraction result for one namespace slot."""
+                    if isinstance(result, Exception):
+                        warn(f"Extractor pod failed in {ns}: {result}")
+                        ns_phases_local = ns_phase_map[ns]
+                        for p in ns_phases_local:
                             if p not in failed:
                                 failed.append(p)
                             for pkg, wl in sorted(pairs_in_ns):
                                 if pkg == p:
                                     failed_pairs.append(f"{p}/{wl}")
                     else:
-                        for phase, exc in errors.items():
+                        for phase, exc in result.items():
                             wls_for_phase = sorted(
                                 wl for pkg, wl in pairs_in_ns if pkg == phase)
                             if exc is None:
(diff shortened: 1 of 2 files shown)


kalantar:
## Review: NEEDS CHANGES

The core parallelization design is sound — worker threads do I/O only, result processing happens on the main thread via `as_completed`, so shared list mutations are single-threaded without locks. However, one critical gap must be addressed before merge.

### Blocking

**`_extract_one_slot` only catches `RuntimeError` — other exceptions crash the loop and lose results**

`_extract_phases_from_pvc` can raise `subprocess.CalledProcessError` (kubectl failure), `OSError` (binary not found), `ValueError`/`IndexError` (malformed output parsing). Any of these propagate through `future.result()`, aborting the `as_completed` loop mid-iteration. Results from slots that already completed but haven't been processed yet are silently lost.

Fix — protect the `as_completed` loop:

```python
for future in concurrent.futures.as_completed(futures):
    try:
        ns, pairs_in_ns, result = future.result()
    except Exception as e:
        ns = futures[future]
        pairs_in_ns = ns_pair_map.get(ns, set())
        result = e
    _process_slot_result(ns, pairs_in_ns, result)
```

This ensures unexpected exceptions from one slot don't discard results from others.

### Non-blocking (follow-up is fine)

1. **Interactive `input()` prompt can race in parallel mode** — issue #200 requirement 7 asks this be handled before parallel dispatch or disabled. Currently multiple threads can call `input()` simultaneously on >1GB data. Rare in practice but worth a follow-up to pass `interactive=False` or pre-probe sizes on the main thread.

2. **No timeout on `as_completed`** — a hung `kubectl` blocks forever with no user feedback. Consider `as_completed(futures, timeout=600)` with a catch.

3. **Error messages are generic** — `_process_slot_result` prints "Extractor pod failed" for all exception types. Distinguishing `RuntimeError` (infrastructure) from other exceptions (environment/code bug) would improve debuggability.

### Test gaps worth adding

- All slots fail in parallel mode (verify summary output under total failure)
- Non-`RuntimeError` exception from one slot (would expose the critical bug above)

### What works well

- Thread safety via main-thread-only result processing is correct and elegant
- Sequential fallback for single slot avoids unnecessary overhead
- Partial failure isolation works — one slot's `RuntimeError` doesn't block others
- `_process_slot_result` helper eliminates duplication between paths
- Step header announces slot count as specified
- Tests use a clean `TrackingExecutor` pattern to verify threading behavior

kalantar:
Pushed fix in fba8465:

**Exception handling broadened:** `_extract_one_slot` now catches `Exception` (not just `RuntimeError`), so `CalledProcessError`, `OSError`, etc. from kubectl failures don't crash the `as_completed` loop and lose results from other slots. Also added a `try/except` around `future.result()` as defense-in-depth — if something unexpected propagates past the worker's catch, it's treated as a slot failure rather than aborting the loop.

Added two new tests: non-RuntimeError exception from one slot (verifying other slots still succeed), and all-slots-fail (verifying correct zero-collected summary).

Follow-ups filed:
- #207 — move interactive size probe before parallel dispatch
- #208 — add timeout to `as_completed`

The error message distinction (item 3) is minor enough to not warrant its own issue — can be addressed opportunistically.

Tests: 719 pass, lint clean.

kalantar:
## Re-review: READY TO MERGE

The fix in fba8465 correctly resolves the critical exception handling issue from the previous review:
- `_extract_one_slot` now catches `Exception` broadly
- `future.result()` has defense-in-depth wrapping that correctly recovers the namespace from the futures dict
- Two new tests verify non-RuntimeError handling and all-slots-fail behavior
- Follow-up issues filed for the interactive prompt race (#207) and timeout (#208)

### Minor recommendation (non-blocking)

The sequential single-slot path still catches only `RuntimeError`, while the parallel path catches all `Exception`. This means the same `OSError` or `CalledProcessError` from `_extract_phases_from_pvc` would be handled gracefully with 2+ slots but crash the command with 1 slot. Not a regression (pre-PR behavior was also `RuntimeError`-only), but a one-word fix would make it consistent:

```python
except Exception as e:  # was: except RuntimeError as e
    _process_slot_result(ns, pairs_in_ns, e)
```

Up to you whether to include that now or follow up separately.

### Summary

All issue #200 requirements are met or explicitly deferred with tracked follow-ups. Thread safety design is correct (main-thread-only result processing via `as_completed`). Test coverage is solid across the key scenarios.

PR #215, opened 2026-05-22
fix: scope deploy collect to per-slot workloads from progress

### LINKED ISSUE #213: deploy collect: unscoped parallel path doesn't filter workloads per slot
## Problem

`deploy.py collect` groups done pairs by `completed_namespace` from progress and dispatches one `_extract_phases_from_pvc` call per slot. However, the extraction call passes only phases — not the workload list for that slot:

```python
errors = _extract_phases_from_pvc(
    sorted(ns_phases), run_name, ns, run_dir,
    skip_logs=skip_logs)
```

Inside `_extract_phases_from_pvc`, when no `workload=` is given, the unscoped full-copy path does `ls /data/{run_name}/{phase}/` to discover workloads. This discovers ALL workloads on that slot's PVC, not just the ones progress assigns to it.

### Impact observed

After a `wipe` + `collect` with 3 slots, the same workload (e.g., `baseline1/chatbot_under`) exists on multiple PVCs due to slot reuse across pipeline cycles. Multiple threads race to collect it:

1. Thread for kalantar-1 finds `chatbot_under` on its PVC, starts downloading (local was wiped)
2. Thread for kalantar-2 also finds `chatbot_under` on its PVC, but kalantar-1 already wrote the file
3. kalantar-2 sees local mtime ≥ remote mtime → "up to date — skipping"

```
[INFO]  [baseline1/chatbot_under] up to date — skipping
[INFO]  [baseline1/chatbot_under] up to date — skipping
[OK]    baseline1/chatbot_under    (kalantar-1)
```

The end result is correct (16/16 collected), but:
- Confusing log output after a wipe ("up to date" when nothing should be up to date)
- Wasted `kubectl exec` + mtime probe calls against PVCs that shouldn't be queried for those workloads
- Non-deterministic: which slot's copy "wins" depends on thread timing

## Root Cause

`_cmd_collect` builds `ns_pair_map` (line 989) which maps each namespace to its `(phase, workload)` pairs, but only uses it for post-extraction reporting in `_process_slot_result`. The extraction call itself doesn't filter.

## Proposed Fix

Pass the workload list from `ns_pair_map` into `_extract_phases_from_pvc` so each slot only collects its assigned workloads. The function already supports a `workload` parameter for single-workload scoping — this could be extended to accept a set/list of workloads to filter the `ls` result, or the caller could issue per-workload calls.

Simplest approach: after the `ls` discovery in the full-copy path, filter `wl_names` to only those in the slot's pair set:

```python
# In _extract_phases_from_pvc, after ls discovery:
if allowed_workloads is not None:
    wl_names = [w for w in wl_names if w in allowed_workloads]
```

With the caller passing:
```python
allowed = {wl for pkg, wl in ns_pair_map.get(ns, set()) if pkg == phase}
```

## Acceptance Criteria

- Each slot only collects workloads that progress assigns to it via `completed_namespace`
- No "up to date — skipping" messages appear when collecting after a wipe
- No behavioral change when a workload genuinely hasn't changed (incremental collect still skips correctly)
- `--skip-logs` path also respects the per-slot workload filter

kalantar:
## Summary

- Add `allowed_workloads: set[str] | None` parameter to `_extract_phases_from_pvc`
- After `ls` discovery in both the full-copy and `--skip-logs` paths, filter `wl_names` against the allowed set
- Both parallel and sequential callers in the unscoped path compute `allowed = {wl for _, wl in pairs_in_ns}` from `ns_pair_map` and pass it through

Closes #213

## Why

When PVCs contain workloads from prior slot reuse (common after wipe+collect cycles), the unscoped path would discover and probe ALL workloads on each PVC regardless of which ones progress actually assigned to that slot. This caused:
- Wasted `kubectl exec` + mtime probes against unrelated workloads
- Confusing "up to date — skipping" messages immediately after a wipe
- Non-deterministic behavior (which slot's copy "wins" depends on thread timing)

## Reviewer attention

- The `allowed_workloads=None` default preserves backward compatibility for the fallback single-namespace path and the scoped `--workload`/`--only` paths (which already use the `workload` parameter directly)
- Updated all existing test mock signatures to accept `allowed_workloads=None` as a keyword argument

## Test plan

- [x] `test_collect_parallel_filters_workloads_per_slot` — multi-slot: each slot receives only its assigned workloads
- [x] `test_collect_sequential_filters_workloads_per_slot` — single-slot: allowed_workloads still passed
- [x] `test_extract_phases_filters_by_allowed_workloads` — internal: ls discovers 3 workloads, only 2 allowed → only 2 copied
- [x] Full suite: 793 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -600,7 +600,8 @@ def _is_up_to_date(local_csv: Path, remote_mtime: "float | None") -> bool:
 def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                               run_dir: Path,
                               skip_logs: bool = False,
-                              workload: "str | None" = None) -> dict[str, "Exception | None"]:
+                              workload: "str | None" = None,
+                              allowed_workloads: "set[str] | None" = None) -> dict[str, "Exception | None"]:
     """Extract results for one or more phases from data-pvc using a single pod.
 
     Data layout on PVC (written by run-workload-blis-observe):
@@ -695,6 +696,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                             f"failed to list workloads: {list_result.stderr.strip()}")
                         continue
                     wl_names = list_result.stdout.strip().split() if list_result.stdout.strip() else []
+                    if allowed_workloads is not None:
+                        wl_names = [w for w in wl_names if w in allowed_workloads]
                 phase_errors = []
                 for wl_name in wl_names:
                     if _is_up_to_date(dest_dir / wl_name / "trace_data.csv",
@@ -760,6 +763,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                         f"failed to list workloads: {list_result.stderr.strip()}")
                     continue
                 wl_names = list_result.stdout.strip().split() if list_result.stdout.strip() else []
+                if allowed_workloads is not None:
+                    wl_names = [w for w in wl_names if w in allowed_workloads]
                 if not wl_names:
                     errors[phase] = None
                     continue
@@ -1038,10 +1043,12 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
 
                     def _extract_one_slot(ns, ns_phases):
                         pairs_in_ns = ns_pair_map.get(ns, set())
+                        allowed = {wl for _, wl in pairs_in_ns}
                         try:
                             errors = _extract_phases_from_pvc(
                                 sorted(ns_phases), run_name, ns, run_dir,
-                                skip_logs=skip_logs)
+                                skip_logs=skip_logs,
+                                allowed_workloads=allowed)
                         except Exception as e:
                             return (ns, pairs_in_ns, e)
                         return (ns, pairs_in_ns, errors)
@@ -1062,10 +1069,12 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
                 else:
                     for ns, ns_phases in ns_items:
                         pairs_in_ns = ns_pair_map.get(ns, set())
+                        allowed = {wl for _, wl in pairs_in_ns}
                         try:
                             errors = _extract_phases_from_pvc(
                                 sorted(ns_phases), run_name, ns, run_dir,
-                                skip_logs=skip_logs)
+                                skip_logs=skip_logs,
+                                allowed_workloads=allowed)
                         except RuntimeError as e:
                             _process_slot_result(ns, pairs_in_ns, e)
                         else:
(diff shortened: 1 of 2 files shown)


kalantar:
## Review Verdict: NEEDS CHANGES (minor)

Overall the implementation is clean, correctly scoped, and addresses the core issue. Two items need fixing before merge:

### Required Changes

1. **Add a test for the `--skip-logs` path with `allowed_workloads` filtering.** The PR inserts the filter in two code paths (full-copy and skip-logs), but only tests the full-copy path. Issue #213 acceptance criteria explicitly states: "`--skip-logs` path also respects the per-slot workload filter." A variant of `test_extract_phases_filters_by_allowed_workloads` with `skip_logs=True` would cover this.

2. **Update the `_extract_phases_from_pvc` docstring to document `allowed_workloads`.** The docstring currently describes `workload` and `skip_logs` but omits the new parameter. A reader would not know that the "unscoped" discovery path can now be narrowed by slot assignment.

### Suggested Improvements (non-blocking)

- **Warn on empty filter result:** When the filter reduces a non-empty `ls` result to empty (possible naming mismatch or corrupt progress data), emit a warning rather than silently succeeding with nothing collected. This prevents confusing "0 pairs" outcomes with no explanation.

- **Update the "Unscoped full copy" inline comment** (line ~753) — with this PR, the path is no longer necessarily "unscoped" since `allowed_workloads` may narrow it. Something like "Discovery-based copy: list workloads via ls, filter by allowed_workloads if set..." would be more accurate.

### What looks good

- Filter placement is correct in both code paths
- `allowed_workloads=None` default preserves backward compat for the fallback (no progress) path
- Both parallel and sequential callers correctly compute the allowed set from `ns_pair_map`
- Tests verify the contract at both caller level and internal filter level
- All 793 existing tests pass; lint clean

kalantar:
Addressed both required changes:

- Added `test_extract_phases_filters_by_allowed_workloads_skip_logs` — exercises the `--skip-logs` discovery path with `allowed_workloads` filtering, confirming the filter works in both code paths as the acceptance criteria requires.
- Updated `_extract_phases_from_pvc` docstring to describe the `allowed_workloads` parameter and its purpose.

The two non-blocking suggestions (warn on empty filter result, update inline comment) are noted but not addressed in this PR — they're cosmetic/defensive hardening beyond the issue scope.

794 tests pass, lint clean.

kalantar:

Both required changes from the prior review are addressed:

1. **`--skip-logs` test added** — `test_extract_phases_filters_by_allowed_workloads_skip_logs` exercises the skip-logs discovery path with `allowed_workloads`, confirming filtered-out workloads are not copied. This satisfies the acceptance criterion.

2. **Docstring updated** — `_extract_phases_from_pvc` now documents the `allowed_workloads` parameter and its purpose.

No remaining blockers. 794 tests pass, lint clean.

PR #217, opened 2026-05-22
fix: make allowed_workloads filter per-phase to prevent stale data collection

### LINKED ISSUE #216: deploy collect: allowed_workloads filter is per-slot but not per-phase, causing stale data collection
## Problem

PR #215 (closing #213) added an `allowed_workloads` filter to `_extract_phases_from_pvc` to prevent cross-slot contamination. However, the filter is built as:

```python
allowed = {wl for _, wl in pairs_in_ns}  # workload names only, ignoring package
```

This set is applied uniformly across ALL phases on that slot's PVC. When a workload name appears in multiple phases but only one phase is assigned to that slot, the other phase's stale data (from a prior run) gets collected.

## Example

Progress assigns to kalantar-1:
- `baseline1/blindspot_mid` ✓
- `baseline1/blindspot_under` ✓
- `baseline1/chatbot_under` ✓
- `treatment/balanced_mid` ✓
- `treatment/codecompletion_mid` ✓

`allowed_workloads` becomes `{blindspot_mid, blindspot_under, chatbot_under, balanced_mid, codecompletion_mid}`.

When extracting the `baseline1` phase, the code does `ls /data/r1/baseline1/` on kalantar-1's PVC, finds `codecompletion_mid` there (leftover from a prior run), and since `codecompletion_mid` is in the allowed set, it downloads it — even though `baseline1/codecompletion_mid` should come from kalantar-0.

## Impact observed (run r1, 2026-05-22)

5 of 16 pairs collected stale data from prior runs:

| Pair | Wrongly pulled from | Stale data from | Correct source |
|------|-------------------|----------------|----------------|
| baseline1/blindspot_under | kalantar-2 | May 21, 3581 rows | kalantar-1, 1801 rows |
| baseline1/chatbot_under | kalantar-0 or -2 | May 21, 3581 rows | kalantar-1, 1801 rows |
| baseline1/codecompletion_mid | kalantar-1 | May 21, **40301 rows** | kalantar-0, 9801 rows |
| treatment/blindspot_mid | kalantar-1 | May 21, 7901 rows | kalantar-2, 3801 rows |
| treatment/chatbot_under | kalantar-1 | May 20, 3581 rows | kalantar-2, 1801 rows |

The parallel threads race: the wrong slot downloads first, then the correct slot sees `local mtime >= remote mtime` → "up to date — skipping". The result is silently corrupt data with no error.

## Root Cause

`deploy.py` line ~1050:
```python
pairs_in_ns = ns_pair_map.get(ns, set())
allowed = {wl for _, wl in pairs_in_ns}  # BUG: not phase-specific
```

This is then passed to `_extract_phases_from_pvc` which iterates over multiple phases and applies the same `allowed_workloads` set to all of them.

## Proposed Fix

The `allowed_workloads` filter needs to be **per-phase**, not just per-slot. Either:

1. Make `allowed_workloads` a `dict[str, set[str]]` mapping phase → workload names, and filter after discovery within each phase's loop iteration:
   ```python
   allowed_per_phase = {
       pkg: {wl for p, wl in pairs_in_ns if p == pkg}
       for pkg in ns_phases
   }
   ```

2. Or filter at the call site by issuing per-phase extraction calls with phase-specific allowed sets.

The issue #213 proposed fix was correct:
```python
allowed = {wl for pkg, wl in ns_pair_map.get(ns, set()) if pkg == phase}
```

This was not what PR #215 implemented.

## Acceptance Criteria

- Each slot only collects `(phase, workload)` pairs that progress explicitly assigns to it
- Stale data from prior runs on the same PVC path is never collected
- After `wipe` + `collect`, every pair's `trace_header.yaml` `created_at` post-dates the current run's assembly timestamp

kalantar:
## Summary

- Change `allowed_workloads` parameter type from `set[str] | None` to `dict[str, set[str]] | None` (mapping phase → workload names)
- Filter uses `allowed_workloads.get(phase, set())` inside the per-phase loop instead of a flat set across all phases
- Callers build the dict via `allowed.setdefault(pkg, set()).add(wl)` preserving the phase dimension

Closes #216

## Why

PR #215 (#213 fix) built `allowed_workloads` as `{wl for _, wl in pairs_in_ns}`, dropping the phase component. When a workload name exists under multiple phases on a PVC (stale from prior slot reuse), it passed the filter even though only one specific `(phase, workload)` pair was assigned to that slot. This caused silent collection of stale data — 5 of 16 pairs affected in the observed run.

## Reviewer attention

- The type change from `set[str]` to `dict[str, set[str]]` is a breaking change to the internal API. All test mocks already updated.
- When `allowed_workloads` is `None` (fallback/no-progress path), behavior is unchanged — the filter is skipped entirely.
- New regression test `test_extract_phases_per_phase_filter_prevents_cross_phase_leak` directly exercises the bug scenario: same workload name on PVC under both phases, but only one phase assigned to the slot.

## Test plan

- [x] `test_extract_phases_per_phase_filter_prevents_cross_phase_leak` — the specific bug scenario
- [x] `test_collect_parallel_filters_workloads_per_slot` — updated assertions for dict type
- [x] `test_collect_sequential_filters_workloads_per_slot` — updated assertions for dict type
- [x] `test_extract_phases_filters_by_allowed_workloads` — updated to pass dict
- [x] `test_extract_phases_filters_by_allowed_workloads_skip_logs` — updated to pass dict
- [x] Full suite: 795 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -601,7 +601,7 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                               run_dir: Path,
                               skip_logs: bool = False,
                               workload: "str | None" = None,
-                              allowed_workloads: "set[str] | None" = None) -> dict[str, "Exception | None"]:
+                              allowed_workloads: "dict[str, set[str]] | None" = None) -> dict[str, "Exception | None"]:
     """Extract results for one or more phases from data-pvc using a single pod.
 
     Data layout on PVC (written by run-workload-blis-observe):
@@ -614,9 +614,11 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
     copied individually, skipping those whose local ``trace_data.csv`` is
     already up to date (used by unscoped ``deploy.py collect``).
 
-    When *allowed_workloads* is set, the ``ls``-discovered list is filtered to
-    only include workloads in the set. Used by the parallel/sequential callers
-    to scope each slot to the workloads progress assigns to it.
+    When *allowed_workloads* is set (a dict mapping phase name to a set of
+    workload names), the ``ls``-discovered list for each phase is filtered to
+    only include workloads in that phase's set. Used by the parallel/sequential
+    callers to scope each slot to the exact (phase, workload) pairs that
+    progress assigns to it.
 
     When *skip_logs* is True, only trace files are copied (skipping vLLM and
     EPP log files which typically account for the bulk of the data).
@@ -701,7 +703,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                         continue
                     wl_names = list_result.stdout.strip().split() if list_result.stdout.strip() else []
                     if allowed_workloads is not None:
-                        wl_names = [w for w in wl_names if w in allowed_workloads]
+                        phase_allowed = allowed_workloads.get(phase, set())
+                        wl_names = [w for w in wl_names if w in phase_allowed]
                 phase_errors = []
                 for wl_name in wl_names:
                     if _is_up_to_date(dest_dir / wl_name / "trace_data.csv",
@@ -768,7 +771,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                     continue
                 wl_names = list_result.stdout.strip().split() if list_result.stdout.strip() else []
                 if allowed_workloads is not None:
-                    wl_names = [w for w in wl_names if w in allowed_workloads]
+                    phase_allowed = allowed_workloads.get(phase, set())
+                    wl_names = [w for w in wl_names if w in phase_allowed]
                 if not wl_names:
                     errors[phase] = None
                     continue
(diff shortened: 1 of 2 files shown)


kalantar:

PR correctly upgrades `allowed_workloads` from a flat `set[str]` to a per-phase `dict[str, set[str]]`, preventing cross-phase data leaks when PVCs contain stale workloads from slot reuse.

**Key findings:**

- Logic is correct: per-phase filtering in `_extract_phases_from_pvc` ensures only workloads assigned to a given phase in the progress ConfigMap are collected, even when the PVC contains data from a different phase's prior occupant.
- No silent failures introduced. Empty-filter edge cases (phase present in dict but with empty workload set) are handled gracefully — they skip extraction rather than collecting everything.
- Test coverage is solid: new `test_collect_per_phase_filtering` validates the core scenario. One optional improvement identified (skip-logs path cross-phase test, criticality 6/10) — not blocking.
- No regressions to existing collect behavior for unscoped or single-phase usage.

No changes required.

kalantar:
**Addendum — minor doc nit (non-blocking):**

`pipeline/tests/test_deploy_collect.py` line 1512 docstring says "allowed_workloads set" — after this PR it's a `dict[str, set[str]]`. Consider updating to "per-phase allowed_workloads dict" for accuracy. Same pattern at lines 1570 and 1626. Not blocking merge.

PR #218, opened 2026-05-22
feat: report per-workload progress during parallel deploy collect

### LINKED ISSUE #214: deploy collect: report per-workload progress as it happens during parallel extraction
## Problem

During parallel collection across multiple namespace slots, `[OK]` lines for each workload are only printed after the entire slot finishes extraction. This is because `_process_slot_result` runs only when `as_completed` yields a completed future — at which point all workloads for that slot are reported as a batch.

Meanwhile, `[INFO] up to date — skipping` messages print in real time (from within `_extract_phases_from_pvc`), creating a confusing log where progress messages trickle in but success messages appear in bursts.

### Current behavior

```
[INFO]  [baseline1/chatbot_under] up to date — skipping
[INFO]  [baseline1/codecompletion_mid] up to date — skipping
[INFO]  [treatment/chatbot_mid] up to date — skipping
                                                          ← silence while slots are downloading
[OK]    baseline1/blindspot_mid    (kalantar-1)           ← batch dump when slot finishes
[OK]    baseline1/blindspot_under    (kalantar-1)
[OK]    baseline1/chatbot_under    (kalantar-1)
[OK]    treatment/balanced_mid    (kalantar-1)
[OK]    treatment/codecompletion_mid    (kalantar-1)
```

### Desired behavior

```
[OK]    baseline1/blindspot_mid    (kalantar-1)
[OK]    treatment/balanced_mid    (kalantar-2)
[OK]    baseline1/balanced_mid    (kalantar-0)
[OK]    baseline1/blindspot_under    (kalantar-1)
...
```

Each workload reports as soon as its `kubectl cp` completes, interleaved across slots.

## Proposed Approach

Pass a per-workload callback into `_extract_phases_from_pvc` that reports success/failure immediately:

```python
def _extract_phases_from_pvc(phases, run_name, namespace, run_dir, *,
                             skip_logs=False, workload=None,
                             on_workload_done=None):
    ...
    # After successful kubectl cp for wl_name:
    if on_workload_done:
        on_workload_done(phase, wl_name, namespace, error=None)
```

The caller supplies a thread-safe callback that prints `[OK]` or `[WARN]` lines. `_process_slot_result` becomes a post-extraction summary step (totals only) rather than the primary reporting mechanism.

## Acceptance Criteria

- Per-workload `[OK]`/`[WARN]` lines print as each workload completes, not batched per slot
- Output is interleaved across slots (reflecting real progress)
- Summary line at the end still shows total collected/failed pair counts
- Single-slot (sequential) path behavior unchanged

kalantar:
## Summary

- Add `on_workload_done` callback parameter to `_extract_phases_from_pvc`
- Callback fires after each workload completes (success or failure) in all three code paths (skip-logs, scoped, full-copy)
- Parallel and sequential callers supply `_on_workload_done` which prints `[OK]`/`[WARN]` and appends to shared lists in real time
- Replace `_process_slot_result` with two focused functions:
  - `_on_workload_done`: per-workload reporting callback
  - `_handle_slot_failure`: pod-level failure fallback (for when the extractor pod itself fails and the callback never fires)

Closes #214

## Why

During parallel collection, `[OK]` lines were batched per-slot — printed only after the entire slot's ThreadPoolExecutor future resolved. Meanwhile, `[INFO] up to date` messages printed in real time from within worker threads. This created confusing output: silence during download, then a burst of `[OK]` lines. Now each workload reports the instant its `kubectl cp` finishes, interleaved across slots.

## Reviewer attention

- **Thread safety**: The callback fires from worker threads and mutates shared lists. `list.append()` is atomic under CPython GIL, so `collected_pairs`/`failed_pairs` are safe. The `if phase not in collected` dedup pattern has a benign race (worst case: a duplicate phase entry in the summary list that doesn't affect pair-level counts).
- **Callback in skip-logs path**: Per-workload errors were previously aggregated into `phase_errors` as flat strings. Now they're tracked per-workload in `wl_errors` so the callback can report each workload individually, then accumulated into `phase_errors` for the return value.
- **No behavioral change for sequential path**: The callback fires the same way in single-slot mode — just no concurrency.

## Test plan

- [x] All 58 collect tests pass (mocks updated to invoke callback)
- [x] Full suite: 795 tests pass
- [x] Failure tests verify pod-level errors still handled via `_handle_slot_failure`
- [x] Cross-product test confirms only assigned pairs are reported (no phantom pairs)

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -601,7 +601,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                               run_dir: Path,
                               skip_logs: bool = False,
                               workload: "str | None" = None,
-                              allowed_workloads: "dict[str, set[str]] | None" = None) -> dict[str, "Exception | None"]:
+                              allowed_workloads: "dict[str, set[str]] | None" = None,
+                              on_workload_done=None) -> dict[str, "Exception | None"]:
     """Extract results for one or more phases from data-pvc using a single pod.
 
     Data layout on PVC (written by run-workload-blis-observe):
@@ -619,6 +620,11 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
     only include workloads in that phase's set. Used by the parallel/sequential
     callers to scope each slot to the exact (phase, workload) pairs that
     progress assigns to it.
+
+    When *on_workload_done* is set, it is called after each workload completes
+    (success or failure) with ``(phase, workload_name, namespace, error)``.
+    *error* is None on success, or an Exception on failure. Used by callers to
+    report per-workload progress in real time during parallel extraction.
 
     When *skip_logs* is True, only trace files are copied (skipping vLLM and
     EPP log files which typically account for the bulk of the data).
@@ -710,6 +716,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                     if _is_up_to_date(dest_dir / wl_name / "trace_data.csv",
                                       remote_mtimes.get(wl_name)):
                         info(f"[{phase}/{wl_name}] up to date — skipping")
+                        if on_workload_done:
+                            on_workload_done(phase, wl_name, namespace, None)
                         continue
                     wl_dest = dest_dir / wl_name
                     wl_dest.mkdir(parents=True, exist_ok=True)
@@ -717,12 +725,13 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                         if log_dir.exists():
                             shutil.rmtree(log_dir)
                     # Copy trace files
+                    wl_errors = []
                     for fname in ("trace_data.csv", "trace_header.yaml", "epp_stream_done"):
                         src = f"{namespace}/{pod_name}:/data/{run_name}/{phase}/{wl_name}/{fname}"
                         r = run(["kubectl", "cp", src, str(wl_dest / fname), "--retries=3"],
                                 check=False, capture=True)
                         if r.returncode != 0 and "no such file" not in r.stderr.lower():
-                            phase_errors.append(f"{wl_name}/{fname}: {r.stderr.strip()}")
+                            wl_errors.append(f"{wl_name}/{fname}: {r.stderr.strip()}")
                     # Copy epp_logs directory
                     epp_src = f"{namespace}/{pod_name}:/data/{run_name}/{phase}/{wl_name}/epp_logs/"
                     epp_dest = wl_dest / "epp_logs"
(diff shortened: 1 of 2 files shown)


kalantar:

The core callback design is sound and meets the acceptance criteria for real-time per-workload reporting. However, one blocking regression and two important issues need attention before merge.

### Blocking

**Silent workload loss when `ls` fails inside `_extract_phases_from_pvc`**

When listing a phase directory fails (PVC unmounted, directory missing, permission error), the code stores the error in the return dict and `continue`s to the next phase — but never fires the callback. Since the parallel and sequential callers now discard the return value, those workloads vanish from both `collected_pairs` and `failed_pairs`. The summary shows an incorrect count with no explanation.

The old `_process_slot_result` correctly iterated the return dict and reported all workloads in failed phases as failed. This is a regression.

**Fix:** After the `ls` failure `continue` (both skip-logs and full-copy paths), fire the callback for every workload in `allowed_workloads[phase]`:

```python
if list_result.returncode != 0:
    errors[phase] = RuntimeError(...)
    if on_workload_done and allowed_workloads is not None:
        for wl in allowed_workloads.get(phase, set()):
            on_workload_done(phase, wl, namespace, errors[phase])
    continue
```

### Non-blocking (worth addressing)

1. **Sequential path catches only `RuntimeError`** — The parallel path uses `except Exception as e:` but the sequential path uses `except RuntimeError as e:`. A `CalledProcessError`, `OSError`, or callback exception would crash the entire collect command in single-slot mode but be handled gracefully in multi-slot mode. Recommend matching: `except Exception as e:`.

2. **`_handle_slot_failure` omits per-workload failure lines** — The old `_process_slot_result` printed `[WARN] phase/wl (ns): error` for every pair in a failed slot. The new code prints only a single "Extractor pod failed" line. The user sees "Failed: 6 pairs" in the summary but not which specific pairs, making it harder to determine what to re-run. Consider adding per-pair `warn()` calls inside `_handle_slot_failure`.

### Non-blocking (minor)

- Docstring says "parallel extraction" but the callback is also used in the sequential-with-progress path. Consider "during extraction" instead.
- `_handle_slot_failure` docstring assumes callback never partially fired. If a future change raises mid-loop, pairs could be double-counted. A defensive `if pair_key not in collected_pairs and pair_key not in failed_pairs` guard would make this safe.

kalantar:
Addressed all three substantive items:

1. **ls-failure callback gap** — When `ls` fails in either the skip-logs or full-copy path, the callback now fires for every workload in `allowed_workloads[phase]` with the error. This prevents silent loss from the pair counts.

2. **Sequential path exception handling** — Changed from `except RuntimeError` to `except Exception` to match the parallel path's `except Exception as e` and prevent unexpected crashes in single-slot mode.

3. **Per-pair warn lines in `_handle_slot_failure`** — Now prints `[WARN] phase/wl (ns)` for each pair before appending to `failed_pairs`, so the user can see exactly which pairs failed during pod-level errors.

Also fixed the docstring wording ("during extraction" instead of "during parallel extraction").

Skipped the defensive double-counting guard — current code guarantees the callback never fires when the pod fails to start (exception raised before the copy loop enters), so this can't happen in practice.

795 tests pass, lint clean.

kalantar:

All three issues from the prior review are correctly addressed in commit 2:

1. **ls failure callback gap** — Fixed. Both skip-logs and full-copy paths now fire `on_workload_done` for every workload in `allowed_workloads.get(phase, set())` before `continue`.
2. **Sequential exception widening** — Fixed. Changed from `except RuntimeError` to `except Exception`, matching the parallel path.
3. **Per-pair warn in `_handle_slot_failure`** — Fixed. Each pair now gets a `warn()` line so users can see which specific workloads failed.

Acceptance criteria are met: per-workload `[OK]`/`[WARN]` lines print in real time, output is interleaved across slots, summary counts are accurate, and sequential path behavior is unchanged.

### Non-blocking observations (for future hardening, not this PR)

- `_handle_slot_failure` could theoretically double-count pairs if an exception propagates mid-extraction after some callbacks already fired. In practice this can't happen because all kubectl calls use `check=False`. A `if pair_key not in collected_pairs` guard would make it defensive.
- The "no progress data" fallback path and scoped (`--only`/`--workload`) path still catch only `RuntimeError`. These are pre-existing and out of scope for this PR, but could be widened for consistency in a follow-up.
- `_handle_slot_failure` docstring describes its precondition ("callback never fired") but not its behavior (emit per-pair warnings + mark as failed). Minor clarity improvement.

PR #219, opened 2026-05-22
feat: support multi-value --workload, --package, and --only flags

### LINKED ISSUE #212: Support multi-value --workload, --package, and --only across all deploy subcommands
## Problem

Only `collect --package` supports multiple values today (via `nargs="+"`). All other scoping flags (`--workload`, `--package` on other subcommands, `--only`) accept a single value. Users frequently want to scope operations to a subset of workloads or packages without running the command multiple times.

Example failure:
```
% python pipeline/deploy.py status --workload chatbot_mid chatbot_high
deploy.py: error: unrecognized arguments: chatbot_high
```

## Proposed Change

Support multi-value input for `--workload`, `--package`, and `--only` across all five subcommands that accept them:
- `collect`
- `status`
- `run`
- `reset`
- `wipe`

### Syntax

Accept **both** space-separated and comma-separated values:
```bash
--workload chatbot_mid,chatbot_high      # comma-separated
--workload chatbot_mid chatbot_high      # space-separated
--workload chatbot_mid,chatbot_high load # mixed
```

All produce the same flat list: `["chatbot_mid", "chatbot_high", "load"]`.

### Implementation approach

1. Add a shared helper to flatten comma-separated tokens:
   ```python
   def _parse_list(value):
       if isinstance(value, list):
           return [v for item in value for v in item.split(",")]
       return value.split(",")
   ```

2. Change argparse definitions for `--workload`, `--package`, and `--only` to `nargs="*"` (or `nargs="+"`) on all five subcommands.

3. Normalize early in each `_cmd_*` handler via `_parse_list`, then pass a list through existing filtering logic.

### Semantics

- **Union filter**: multiple values mean "match any of these" — same scheduling/execution logic, just a broader filter.
- No sequencing or ordering semantics — `--workload a,b` is equivalent to "all pairs whose workload is a OR b".

## Affected commands

| Subcommand | `--only` | `--workload` | `--package` |
|------------|:--------:|:------------:|:-----------:|
| `collect`  | multi    | multi        | already multi (normalize to also accept commas) |
| `status`   | multi    | multi        | multi       |
| `run`      | multi    | multi        | multi       |
| `reset`    | multi    | multi        | multi       |
| `wipe`     | multi    | multi        | multi       |

## Acceptance Criteria

- All three flags accept comma-separated, space-separated, or mixed input on all five subcommands
- `collect --package` retains backward compat (already supports space-separated)
- Existing single-value usage continues to work unchanged
- Invalid values still produce clear error messages listing valid options

kalantar:
## Summary

- Add `_parse_list` helper that normalizes comma-separated, space-separated, or mixed CLI input into flat lists
- Change `--only`, `--workload`, and `--package` to `nargs="+"` on all 5 subcommands (collect, status, run, reset, wipe)
- Update `_apply_run_filters` to iterate over multi-value `--only` (union of resolved keys) and use `in` for workload/package membership
- Update `_report_filter_mismatch` to format lists with comma-join
- Normalize `args.package` via `_parse_list` in `_cmd_collect`'s phase-filtering paths (both scoped and unscoped)

Closes #212

## Usage examples

```bash
# All equivalent:
deploy.py status --workload chatbot_mid chatbot_high
deploy.py status --workload chatbot_mid,chatbot_high
deploy.py status --workload chatbot_mid,chatbot_high load

# Multi-value --only (union):
deploy.py collect --only smoke-baseline smoke-treatment

# Multi-value --package with commas:
deploy.py run --package baseline,treatment
```

## Reviewer attention

- **`_parse_list` returns None (not empty list)** when input is None or produces no items. This preserves the existing "no filter given" semantics throughout `_apply_run_filters` and `_resolve_scope`.
- **`collect --package` already had `nargs="+"`** — the only change there is adding comma normalization via `_parse_list`, so `--package baseline,treatment` now works in addition to `--package baseline treatment`.
- **`--status` remains single-value** — it wasn't part of the issue scope and filtering by multiple statuses simultaneously is a different semantic (would need its own design decision).

## Test plan

- [x] Full suite: 795 tests pass (existing single-value tests exercise backward compat)
- [x] Existing `--workload` and `--only` scoped collect tests pass unchanged (single-value usage)
- [x] Existing `--package` multi-value tests pass (already `nargs="+"`)

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -195,7 +195,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--package NAME…` | Collect only these packages (comma or space-separated, phase-level filter) |
 | `--skip-logs` | Skip vLLM and EPP log files, collect only traces |
 
-When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). These pair-level flags compose with `--package` as AND: `--workload X --package baseline` pulls workload X from the baseline phase only. Requires progress data to resolve pairs.
+When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). Multiple values within a flag use OR (union): `--workload X Y` matches pairs for workload X or Y. Different flags compose as AND: `--workload X Y --package baseline` pulls workloads X and Y from the baseline phase only. Requires progress data to resolve pairs.
 
 **`deploy.py stop`** — deletes the `sim2real-orchestrator` Kubernetes Job (with cascading pod deletion) in the primary namespace. Only meaningful when the orchestrator runs as an in-cluster Job. Pair state is left as-is. If no remote orchestrator Job exists, prints a message and returns. Use `reset` separately to clear failed/stalled pair state.
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -152,9 +152,9 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | Flag | Default | Description |
 |------|---------|-------------|
 | `--remote` | — | Submit orchestrator as in-cluster Job instead of running locally |
-| `--only PAIR` | — | Scope execution to one specific pair key (`wl-` prefix optional) |
-| `--workload NAME` | — | Scope execution to pairs matching this workload |
-| `--package NAME` | — | Scope execution to pairs matching this package |
+| `--only PAIR…` | — | Scope execution to specific pair keys (comma or space-separated, `wl-` prefix optional) |
+| `--workload NAME…` | — | Scope execution to pairs matching these workloads (comma or space-separated) |
+| `--package NAME…` | — | Scope execution to pairs matching these packages (comma or space-separated) |
 | `--status STATE` | — | Scope execution to pairs with this status (e.g. `failed`, `timed-out`) |
 | `--skip-teardown` | — | Skip the Tekton teardown task, leaving namespace resources intact for debugging |
 | `--preserve-pipelineruns` | — | Do not delete PipelineRun objects after completion (keeps TaskRun logs for debugging) |
@@ -181,18 +181,18 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 
 | Flag | Description |
 |------|-------------|
-| `--only PAIR` | Scope to one pair key (`wl-` prefix optional) |
-| `--workload NAME` | Filter by workload name |
-| `--package NAME` | Filter by package name |
+| `--only PAIR…` | Scope to specific pair keys (comma or space-separated, `wl-` prefix optional) |
+| `--workload NAME…` | Filter by workload names (comma or space-separated) |
+| `--package NAME…` | Filter by package names (comma or space-separated) |
 | `--status STATE` | Filter by status (e.g. `running`, `done`, `failed`) |
 
 **`deploy.py collect`** — extracts results from the cluster PVC and writes to `workspace/runs/<run>/results/{phase}/<workload>/`.
 
 | Flag | Description |
 |------|-------------|
-| `--only PAIR` | Scope to one pair key — narrows both workload and package (`wl-` prefix optional; takes precedence over `--workload`) |
-| `--workload NAME` | Scope to pairs matching this workload |
-| `--package NAME…` | Collect only these packages (phase-level filter) |
+| `--only PAIR…` | Scope to specific pair keys — narrows both workload and package (comma or space-separated, `wl-` prefix optional; takes precedence over `--workload`) |
+| `--workload NAME…` | Scope to pairs matching these workloads (comma or space-separated) |
+| `--package NAME…` | Collect only these packages (comma or space-separated, phase-level filter) |
 | `--skip-logs` | Skip vLLM and EPP log files, collect only traces |
 
 When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). These pair-level flags compose with `--package` as AND: `--workload X --package baseline` pulls workload X from the baseline phase only. Requires progress data to resolve pairs.
@@ -203,9 +203,9 @@ When `--only` or `--workload` is given, only matching workload subdirectories ar
 
 | Flag | Description |
 |------|-------------|
-| `--only PAIR` | Scope reset to one specific pair key (`wl-` prefix optional) |
-| `--workload NAME` | Scope reset to pairs matching this workload |
-| `--package NAME` | Scope reset to pairs matching this package |
+| `--only PAIR…` | Scope reset to specific pair keys (comma or space-separated, `wl-` prefix optional) |
+| `--workload NAME…` | Scope reset to pairs matching these workloads (comma or space-separated) |
+| `--package NAME…` | Scope reset to pairs matching these packages (comma or space-separated) |
 | `--status STATE` | Scope reset to pairs with this status |
 | `--preserve-done-status` | Keep done pairs' status unchanged (cluster cleanup only) |
 | `--dry-run` | Print what would be reset without acting |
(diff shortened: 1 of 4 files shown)


kalantar:

The implementation is clean and correctly fulfills all acceptance criteria for issue #212 from a functional standpoint. The `_parse_list` helper, argparse changes, and filter logic updates are well-structured. However, three issues should be addressed before merge:

### 1. Unresolved `--only` keys are silently skipped (high severity)

In `_apply_run_filters`, the new multi-value `--only` loop silently drops keys that don't resolve in progress. Previously, a single unresolved key returned an empty set which triggered `_report_filter_mismatch` and `sys.exit(1)`. Now with `--only goodkey,typokey`, only `goodkey` is returned — the typo is never surfaced.

This is particularly concerning for destructive operations like `reset` or `wipe` where a user expects all specified keys to be affected.

**Suggested fix:** After the loop, warn (or error) for any unresolved keys:
```python
if unresolved:
    warn(f"--only: no match for {unresolved} — ignoring")
```

### 2. No tests for the new multi-value behavior

The PR adds no new tests. Existing tests exercise backward compatibility (single-value as strings), but none test:
- Multi-value `--only` union semantics
- Multi-value `--workload`/`--package` membership filtering
- Comma normalization (`"a,b"` → `["a", "b"]`)
- Mixed comma+space input (`["a,b", "c"]` → `["a", "b", "c"]`)
- `_parse_list` edge cases (None, empty list, whitespace-only)

At minimum: unit tests for `_parse_list` and multi-value tests for `_apply_run_filters`.

### 3. `pipeline/README.md` not updated

Per CLAUDE.md policy: "After any change to pipeline/ that affects CLI flags, update pipeline/README.md to match." The README still documents all three flags as single-value across all five subcommand tables.

### Minor items

- **Double `_parse_list` call** (lines 904-905): The `elif _parse_list(args.package):` / `pkg_filter = _parse_list(args.package)` pattern calls the function twice. The second collect path (line 990) already uses the correct assign-then-test pattern.
- **Redundant `.strip()`** in the `--only` loop: `_parse_list` already strips values.

kalantar:
Addressed all review items:

**Unresolved --only keys**: The multi-value `--only` loop now tracks unresolved keys and emits `[WARN] --only: no match for ['typokey']` instead of silently dropping them. The resolved keys still form the result set, so partial matches work while typos are surfaced.

**Tests**: Added 14 tests covering `_parse_list` (None, empty, single string, comma-separated, list input, mixed, whitespace, whitespace-only) and multi-value `_apply_run_filters` (multi-workload, multi-package, multi-only, prefix resolution, partial unresolved, comma-in-list).

**README**: Updated all 5 subcommand flag tables (run, status, collect, reset, wipe) to document multi-value syntax with `NAME…` / `PAIR…` and "(comma or space-separated)".

**Double `_parse_list` call**: Fixed with walrus operator in the scoped collect path.

**Redundant `.strip()`**: Removed from the `--only` loop since `_parse_list` already strips.

809 tests pass, lint clean.

kalantar:

All acceptance criteria from issue #212 are met. The second commit fully addresses the previous review feedback (warning on unresolved keys, 14 new tests, README updates, walrus operator fix). The implementation is clean, focused, and backward-compatible.

### What's good

- `_parse_list` correctly normalizes all input shapes (None, string, list, comma-in-list, whitespace)
- Union semantics properly implemented via `in` membership for workload/package and loop-with-union for `--only`
- 14 new tests cover `_parse_list` edge cases, multi-value filtering, prefix resolution, and the unresolved-key warning
- README flag tables updated across all 5 subcommands
- Existing 795 tests continue to pass (backward compat verified)

### Optional follow-up suggestions (non-blocking)

1. **Stricter handling for destructive ops**: When `--only` partially resolves, the function warns and proceeds. For `reset`/`wipe`, consider making unresolved keys a hard error in a follow-up — a warning to stderr is easily missed in scripts.

2. **Degenerate input edge case**: `--workload ",,,,"` normalizes to None (no filter) but triggers the "No pairs matched" error path with an empty diagnostic message. Unlikely in practice but could be improved by reporting raw values in `_report_filter_mismatch`.

3. **Docstring precision**: `_apply_run_filters` could note `--only` union semantics and the warn-on-unresolved behavior to help future maintainers.

None of these block the merge — the PR delivers what issue #212 specified.

kalantar:
## NEEDS CHANGES

Overall this is a clean, well-tested implementation that meets the issue #212 requirements. One production bug needs fixing before merge, plus a few smaller items.

### Blocking

**`_collect_run_flags` breaks multi-value forwarding to `--remote` jobs** — Line 2207 does `str(val)` on args that are now lists. For `args.workload = ["wl-smoke", "wl-heavy"]`, this forwards `--workload ['wl-smoke', 'wl-heavy']` (Python list repr) to the in-cluster Job CLI, which will fail to parse. Fix: extend with the list elements directly, e.g.:

```python
if isinstance(val, list):
    flags.extend([f"--{name}"] + val)
else:
    flags.extend([f"--{name}", str(val)])
```

Add a test verifying serialized output for list-valued args.

### Worth considering (non-blocking)

1. **Partial `--only` resolution silently narrows scope** — When some `--only` keys are typos, the operation proceeds on fewer pairs than requested with only a `warn()`. For destructive commands (`reset`, `wipe`), this could leave stale state the user doesn't notice. Consider aborting by default when any keys are unresolved, or at minimum printing a post-operation summary.

2. **`_parse_list` docstring says "comma/space/mixed"** but the function only splits on commas. Space separation is handled by argparse `nargs="+"`. Suggest: `"""Flatten a CLI flag value (possibly a list from nargs='+') by splitting each item on commas."""`

3. **Missing tests**: (a) multi-value `--only` with ALL keys unresolved through `_resolve_scope` abort path, (b) `_report_filter_mismatch` formatting with multi-value inputs, (c) `_collect_run_flags` with list-valued args.

4. **README composability note** could clarify that within a flag values use OR (union) while between flags the logic is AND.

### What looks good

- `_parse_list` returning `None` (not empty list) correctly preserves "no filter given" semantics
- `nargs="+"` is the right choice over `nargs="*"`
- 14 new tests cover core behavior well
- Backward compatibility confirmed by existing single-value tests passing unchanged
- `--status` exclusion is appropriate and well-documented

kalantar:
Addressed all review items in 16946e2:

**Fixed `_collect_run_flags` for list-valued args** — was using `str(val)` which serialized Python list repr (`"['a', 'b']"`) instead of proper CLI arguments. Now extends with individual elements for list values, preserving `str()` for scalars like `--status`.

**Unresolved `--only` keys now abort** — previously just warned and silently narrowed scope. Now prints the unresolved keys, lists all valid pair keys, and exits with code 1. This prevents partial operations on destructive commands like `reset` and `wipe` where a typo could leave stale state unnoticed.

**Fixed `_parse_list` docstring** — was claiming "comma/space/mixed" but the function only splits on commas. Space separation comes from argparse `nargs="+"`. Updated to accurately describe what the function does.

**Added tests** — `_collect_run_flags` list serialization (2 tests), all-keys-unresolved abort through both `_apply_run_filters` and `_resolve_scope`, `_report_filter_mismatch` multi-value formatting. Updated existing tests that expected empty-set returns for unresolved keys to expect `SystemExit(1)`.

**README composability note** — clarified that multiple values within a flag use OR (union) while different flags compose as AND.

All 814 tests pass, lint clean.

kalantar:
## Verdict: READY TO MERGE

All acceptance criteria from #212 are met. The implementation is correct, well-tested (14+ new tests), and backward-compatible. CI passes.

### Key Strengths

- `_parse_list` helper cleanly handles all input variants (None, string, list, mixed commas)
- Abort-on-unresolved for `--only` is the right safety choice for destructive ops (reset/wipe)
- `_collect_run_flags` serialization fix prevents broken `--remote` jobs with multi-value flags
- README composability docs (OR-within-flag, AND-between-flags) are clear and accurate

1. **`_resolve_scope` fragility** — `filters_given` uses raw `getattr` to detect whether flags were provided, while `_apply_run_filters` normalizes through `_parse_list`. Today this works (whitespace-only input triggers `_parse_list` → None → empty set → abort), but the semantic gap between "was a flag syntactically present" vs "does the flag have meaningful content" is fragile. Consider aligning `filters_given` to use `_parse_list` too.

2. **`_apply_run_filters` docstring** — now incomplete since `--only` can abort via `sys.exit(1)` rather than always returning a set. Worth a one-line addition documenting the abort path.

3. **`_collect_run_flags` comma forwarding** — values like `"smoke,load"` are forwarded unsplit to the remote Job, relying on the remote running the same `_parse_list`-aware version. Pre-normalizing before serialization would be more robust against version skew.

4. **Test gap: AND-between-flags with multi-values** — no test exercises `--workload X Y --package baseline` returning only the intersection. Low risk since the implementation is straightforward successive filtering, but would close the coverage loop on the documented composability semantics.

PR #221, opened 2026-05-26
feat: validate --workload and --package filter values before filtering

### LINKED ISSUE #220: Validate filter values before filtering: warn/error on unrecognized --workload and --package values
## Problem

When a user passes `--workload` or `--package` values that don't correspond to any pair in progress, the system silently ignores the invalid values. The principle should be: if the user uses a filter (`--only`, `--workload`, `--package`) and a value has no valid matches, report an error — don't silently fall back to a broader scope.

PR #219 added a warning for unresolved `--only` keys, but `--workload` and `--package` still lack equivalent validation.

### Example

```
deploy.py --run r1 status --package baseline1x
```

This errors because nothing matched — but the error message dumps the entire universe of valid keys, workloads, packages, and statuses. The user was clearly trying to restrict scope; the diagnostic should be proportional.

```
deploy.py --run r1 reset --package baseline1x treatment
```

Here `baseline1x` isn't a valid package, but `treatment` is. The command silently proceeds on just the treatment pairs with no indication that `baseline1x` was unrecognized. For destructive operations (`reset`, `wipe`), this is dangerous.

## Proposed Change

1. **Validate filter values against known values in progress** — before filtering, check that each `--workload` value exists in the set of workload values in progress, and each `--package` value exists in the set of package values. Warn (or error for destructive ops) on unrecognized values.

2. **Targeted diagnostics** — when a filter doesn't match, the error message should highlight which specific values were unrecognized and show only the relevant dimension's valid values, not dump the full universe.

## Affected Commands

All five subcommands that accept filter flags: `collect`, `status`, `run`, `reset`, `wipe`.

## Context

Surfaced during review of #219. The underlying issue predates that PR but becomes more visible now that multi-value filters are supported — users are more likely to pass multiple values and hit partial matches.

kalantar:
Closes #220

## Summary

- Pre-validates `--workload` and `--package` values against the set of known values in progress before applying filters
- If any value is unrecognized, prints the specific bad values plus valid values for that dimension and aborts (exit 1)
- Catches typos and partial mismatches immediately rather than silently narrowing scope or dumping the full universe of valid keys

## What changed

`_apply_run_filters` in `pipeline/deploy.py` now validates `--workload` and `--package` values before filtering, matching the existing `--only` validation pattern. The error message is dimension-specific (shows only valid workloads or packages, not the full pair key list).

## Reviewer attention

- The `collect` subcommand has its own separate `--package` validation (phase-level filter at lines 904-908) that bypasses `_resolve_scope` — it remains unchanged and unaffected.
- `_report_filter_mismatch` (the "No pairs matched" path) still serves a purpose: it fires when individually valid values combine to match zero pairs (e.g., valid workload + valid package with no pairs in common). The new validation only catches values that don't exist at all.
- Three existing tests were updated to expect the new targeted error message ("unrecognized") instead of the old broad "No pairs matched" message.

## Test plan

- [x] 4 new tests: invalid/partial-invalid for both `--workload` and `--package`
- [x] All 20 `_apply_run_filters` tests pass
- [x] Full suite: 751 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1388,13 +1388,31 @@ def _apply_run_filters(progress: dict, args) -> set:
     if not any([workload, package, status_filter]):
         return set()
 
-    candidates = {k for k in progress.keys() if _is_pair_key(k)}
+    pair_entries = {k: v for k, v in progress.items() if _is_pair_key(k)}
+
     if workload:
-        candidates = {k for k in candidates if progress[k].get("workload") in workload}
+        valid_workloads = {v.get("workload", "") for v in pair_entries.values()} - {""}
+        unknown = set(workload) - valid_workloads
+        if unknown:
+            err(f"--workload: unrecognized values {sorted(unknown)}")
+            print(f"  Valid --workload values: {', '.join(sorted(valid_workloads))}", file=sys.stderr)
+            sys.exit(1)
+
     if package:
-        candidates = {k for k in candidates if progress[k].get("package") in package}
+        valid_packages = {v.get("package", "") for v in pair_entries.values()} - {""}
+        unknown = set(package) - valid_packages
+        if unknown:
+            err(f"--package: unrecognized values {sorted(unknown)}")
+            print(f"  Valid --package values: {', '.join(sorted(valid_packages))}", file=sys.stderr)
+            sys.exit(1)
+
+    candidates = set(pair_entries.keys())
+    if workload:
+        candidates = {k for k in candidates if pair_entries[k].get("workload") in workload}
+    if package:
+        candidates = {k for k in candidates if pair_entries[k].get("package") in package}
     if status_filter:
-        candidates = {k for k in candidates if progress[k].get("status") == status_filter}
+        candidates = {k for k in candidates if pair_entries[k].get("status") == status_filter}
     return candidates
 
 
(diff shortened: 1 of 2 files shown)


kalantar:
## READY TO MERGE

Clean, well-scoped PR that satisfies issue #220's requirements. The implementation validates `--workload` and `--package` values against known values in progress before filtering, with targeted per-dimension error messages. All five affected subcommands are covered through the `_apply_run_filters` chokepoint.

**Key positives:**
- Validation placement in `_apply_run_filters` covers all subcommands without duplication
- Error messages are actionable: show the bad values and only the relevant dimension's valid set
- `_report_filter_mismatch` correctly preserved for the valid-but-empty-intersection case
- Tests are behavioral (exit code + message content), not coupled to implementation
- Consistent with existing `--only` validation pattern

**One observation (non-blocking, follow-up candidate):**
If progress entries exist but all have empty-string workload/package fields (e.g., malformed PipelineRun YAML missing `workloadName`), `valid_workloads` computes to an empty set and every user-supplied value is rejected with `Valid --workload values: ` (empty list). This is unlikely in normal operation since upstream guards prevent calling filters on empty/malformed progress, but a guard like `if not valid_workloads: warn(...)` would make the function self-defensive against corrupted data.

**Design choice noted:** The issue suggested warn-for-non-destructive/error-for-destructive. The PR always errors, which is strictly safer — proceeding on a typo is never useful regardless of command.

PR #236, opened 2026-05-28
fix: _cancel_and_delete_pipelinerun reports failure to caller

### LINKED ISSUE #79: deploy: _cancel_and_delete_pipelinerun should report failure to caller
## Context

Identified during review of #77. Pre-existing design issue affecting all call sites.

## Problem

`_cancel_and_delete_pipelinerun` uses `check=False` on all kubectl calls and never reports failure to its caller. If kubectl is unreachable or RBAC denies the delete, the PipelineRun continues running on the cluster consuming GPUs while the orchestrator believes the slot is freed and dispatches new work to it.

## Affected call sites

- Timeout handler (deploy.py poll loop)
- `_cleanup_pair` (cleanup subcommand + `--force`)
- `_handle_pending_pods` (early reclaim, added in #77)

## Proposed fix

Make `_cancel_and_delete_pipelinerun` return a boolean success indicator. Callers that free the slot should check the return value — if cancel/delete failed, leave the slot marked as busy and warn.

kalantar:
## Summary

- `_cancel_and_delete_pipelinerun` now returns `bool` — `True` when deletion succeeded (or PR didn't exist), `False` when `kubectl delete` fails
- All four callers check the return value and leave the slot marked busy on failure, preventing new work from being dispatched to a namespace whose PipelineRun is still consuming GPUs
- Added warning messages on failure so operators see exactly what happened

Closes #79

## Call sites updated

| Call site | Behavior on failure |
|-----------|-------------------|
| `_handle_pending_pods` (non-recoverable) | Returns `False`, entry stays `running` |
| `_handle_pending_pods` (recoverable timeout) | Returns `False`, entry stays `running` |
| `_reset_pair` (running pair) | Returns `False`, state NOT reset |
| Timeout handler (poll loop) | `continue`s loop, slot stays busy |

## Reviewer attention

- The `_reset_pair` path already had correct failure handling for non-running pairs (checking `kubectl delete` returncode directly). The running-pair path was the inconsistency — it now uses the same pattern via the return value.
- The "PR doesn't exist" case returns `True` (success) since there's nothing to clean up — this preserves backward compatibility for the reset subcommand which may encounter already-deleted PipelineRuns.

## Test plan

- [x] `python3 -m pytest pipeline/ -v` — 744 tests pass
- [x] 6 new tests covering the failure paths across all call sites

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -253,15 +253,19 @@ def _cmd_build(run_dir: Path, namespace: str, skip_build: bool) -> str:
 
 # ── PipelineRun helpers ──────────────────────────────────────────────────────
 
-def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> None:
+def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> bool:
     """If a PipelineRun with the given name exists, cancel it, wait for it to
-    finish cancelling, then delete it so a fresh one can be submitted."""
+    finish cancelling, then delete it so a fresh one can be submitted.
+
+    Returns True if the PipelineRun was successfully deleted (or didn't exist).
+    Returns False if the delete failed — caller should NOT free the slot.
+    """
     exists = run(
         ["kubectl", "get", "pipelinerun", pr_name, "-n", namespace],
         check=False, capture=True,
     )
     if exists.returncode != 0:
-        return  # doesn't exist, nothing to cancel
+        return True  # doesn't exist, nothing to cancel
 
     status = _check_pipelinerun_status(pr_name, namespace)
     info(f"Existing PipelineRun {pr_name!r} found (status: {status}); cancelling …")
@@ -282,11 +286,17 @@ def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> None:
         else:
             warn(f"PipelineRun {pr_name!r} did not cancel within 120 s; deleting anyway")
 
-    run(
+    result = run(
         ["kubectl", "delete", "pipelinerun", pr_name, "-n", namespace,
          "--ignore-not-found"],
         check=False, capture=True,
     )
+    if result.returncode != 0:
+        detail = result.stderr.strip() if result.stderr else ""
+        warn(f"Failed to delete PipelineRun {pr_name!r} in {namespace}"
+             + (f": {detail}" if detail else ""))
+        return False
+    return True
 
 
 def _delete_pipelinerun(pr_name: str, namespace: str) -> None:
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -499,7 +499,9 @@ def _handle_pending_pods(*, pr_name: str, namespace: str, entry: dict,
 
     if category == "non_recoverable":
         warn(f"[{entry.get('workload', '?')}] non-recoverable pending: {detail}")
-        _cancel_and_delete_pipelinerun(pr_name, namespace)
+        if not _cancel_and_delete_pipelinerun(pr_name, namespace):
+            warn(f"[{entry.get('workload', '?')}] cancel failed — slot NOT freed")
+            return False
         entry["status"] = "failed"
         entry["namespace"] = None
         entry["pending_since"] = None
@@ -523,7 +525,9 @@ def _handle_pending_pods(*, pr_name: str, namespace: str, entry: dict,
         return False
 
     warn(f"[{entry.get('workload', '?')}] pending {int(elapsed)}s > {pending_threshold}s threshold → reclaim")
-    _cancel_and_delete_pipelinerun(pr_name, namespace)
+    if not _cancel_and_delete_pipelinerun(pr_name, namespace):
+        warn(f"[{entry.get('workload', '?')}] cancel failed — slot NOT freed")
+        return False
     stalls = entry.get("pending_stalls", 0) + 1
     entry["pending_stalls"] = stalls
     entry["pending_since"] = None
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -258,33 +258,45 @@ def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> bool:
     finish cancelling, then delete it so a fresh one can be submitted.
 
     Returns True if the PipelineRun was successfully deleted (or didn't exist).
-    Returns False if the delete failed — caller should NOT free the slot.
+    Returns False if it could not be removed — caller should NOT free the slot.
+    The cancel-patch step is best-effort: if the patch fails the function still
+    attempts the delete, and the return value reflects whether delete succeeded.
     """
     exists = run(
         ["kubectl", "get", "pipelinerun", pr_name, "-n", namespace],
         check=False, capture=True,
     )
     if exists.returncode != 0:
-        return True  # doesn't exist, nothing to cancel
+        stderr = exists.stderr.strip() if exists.stderr else ""
+        if "NotFound" in stderr or "not found" in stderr:
+            return True  # doesn't exist, nothing to cancel
+        warn(f"Cannot reach PipelineRun {pr_name!r} in {namespace}"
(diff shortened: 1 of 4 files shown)


kalantar:
## Review summary

**Verdict: NEEDS CHANGES**

The PR correctly addresses the literal contract change in #79 — `_cancel_and_delete_pipelinerun` now returns `bool`, and all four call sites that free a slot (the three listed in #79 plus the renamed `_reset_pair`) check the return value and leave the slot busy on failure. Tests cover the main bool-return paths. Good shape.

But the same class of silent failure that motivated #79 is still present inside the function, and it bites in exactly the scenario the issue called out ("kubectl is unreachable or RBAC denies").

### Blocking

1. **`kubectl get` non-NotFound failure returns True** (`pipeline/deploy.py:267-268`). Any non-zero exit — RBAC denial, API server unreachable, kubeconfig context wrong, network blip — is treated as "doesn't exist" and the function returns `True`. The slot gets freed and a fresh PR is dispatched on top of a possibly-still-running one. This is the exact pattern #79 was filed to eliminate, just relocated upstream of the fix. Distinguish NotFound (`--ignore-not-found` + check stdout, or grep stderr for `NotFound`/`not found`) from other errors and return `False` on the latter.

2. **Cancel-patch failure is swallowed** (`pipeline/deploy.py:274-279`). The `kubectl patch ... cancelled` call uses `check=False, capture=True` and the result is never inspected. If the patch fails (RBAC, webhook, transient API), the wait loop runs to its 120s timeout and falls through to "deleting anyway" — and if delete succeeds, the function returns `True` even though cancellation never propagated. At minimum capture and warn on the patch result; consider returning False if patch fails so the caller leaves the slot busy.

3. **Misleading warn messages at call sites** (`pipeline/deploy.py:502, 529, 1879`). The warns say `"cancel failed — slot NOT freed"`, but per the function's contract `False` means **delete** failed, not cancel. An operator scanning logs will hunt for the wrong root cause. Either change the wording (`"could not remove PipelineRun — slot NOT freed"`) or actually have the function return False on cancel failure too. Bonus: the call-site warns include only `entry.get('workload')` — adding `pr_name` and `namespace` to each line would let an operator triage from a single grep.

4. **Main poll-loop timeout handler has no test** (`pipeline/deploy.py:~1878`). The diff adds `if not _cancel_and_delete_pipelinerun(...): warn(...); continue` inside `_work_remaining`, but no test exercises this path. A regression that flipped `continue` to fall-through would silently re-pend a slot that still owns a pod, and CI would not catch it. Add a test that sets up an over-`timeout_hours` entry, stubs cancel `False`, runs one poll iteration, and asserts the entry is unchanged.

### Non-blocking improvements

- Docstring on `_cancel_and_delete_pipelinerun` (`pipeline/deploy.py:256-261`) doesn't mention that the cancel patch is best-effort or that a 120s cancel timeout still returns `True` if delete succeeds. Worth one extra sentence so the next maintainer doesn't have to re-derive it.
- Stub pattern `lambda pr, ns: cancelled.append(...) or True` (`test_deploy_run.py:1317, 1416`) reads as a typo until you parse it. A named lambda or helper would be clearer.
- The "delete fails" test (`test_deploy_cancel.py:40`) sets `stderr = "connection refused"` but doesn't assert the warn includes that detail — locking in the operator-visible message would directly validate #79's intent.

### What's good

- All three issue-listed call sites plus `_reset_pair` consistently check the return and leave entry state intact on failure. Verified via `grep`.
- Tests at `test_deploy_run.py:2265+` and `test_deploy_reset.py:696+` assert the load-bearing invariants (`status == "running"`, `namespace` preserved), not just return values.
- Existing test fakes were updated to return `True` rather than left stubbed as `None`, so the new bool contract is enforced everywhere the function is mocked.

kalantar:
Addressed all four blocking items from the review:

1. **kubectl get non-NotFound failures now return False.** The function checks stderr for "NotFound"/"not found" — any other non-zero exit (RBAC denial, network error) is treated as "cannot verify state" and returns False, leaving the slot busy. This closes the relocated silent-failure gap the review identified.

2. **Cancel-patch failure now warns.** If the patch fails, the function logs the stderr detail and skips the 120s wait loop, proceeding directly to delete. The return value still reflects whether delete succeeded (per the docstring contract).

3. **Warn messages reworded.** All three call sites now say `"could not remove PipelineRun {pr_name!r} in {namespace} — slot NOT freed"` — accurately describes the delete failure and includes enough context for operator triage from a single grep.

4. **Timeout handler extracted into testable `_handle_timeout` helper.** Three new tests cover: cancel-fails (entry unchanged, returns False), cancel-succeeds (requeues, returns True), and not-yet-expired (returns None, no action). The poll loop calls the helper and handles slot cleanup based on the tri-state return.

Also added 2 tests for the new kubectl-get and patch-failure behaviors in `test_deploy_cancel.py`. 749 tests pass, lint clean.

kalantar:
## Re-review of commit 83d50ec8

**Verdict: READY TO MERGE** (with minor follow-ups worth filing as separate issues)

All four blocking items from the prior review are verifiably addressed:

1. **`kubectl get` non-NotFound failure now returns False** (`deploy.py:262-273`). Stderr-grep distinguishes NotFound from other errors. Covered by `test_cancel_returns_false_when_get_fails_not_notfound`. The substring heuristic is fragile in theory (locale, message phrasing), but the unhappy-path direction is fail-safe — ambiguous output is treated as "still active" and the slot stays busy.
2. **Cancel-patch failure now warns and skips the 120s wait** (`deploy.py:278-296`). Function still proceeds to delete and returns based on delete result, which the docstring now states explicitly. Covered by `test_cancel_patch_failure_still_attempts_delete`.
3. **Caller warn messages now read `"could not remove PipelineRun {pr_name!r} in {namespace} — slot NOT freed"`** at all three handlers (`deploy.py:514, 540, 582`). Includes the PR name and namespace for single-grep triage.
4. **`_handle_timeout` extracted with three tests** (`test_deploy_run.py:2364, 2400, 2437`). Tri-state return wired correctly into `_work_remaining`: `True` → free slot + save, `False` → continue (slot stays busy), `None` → fall through. The `import datetime as _dt` removed from `_cmd_run` is safe — no remaining references in that function; both `_handle_timeout` and `_handle_pending_pods` import it locally.

### Suggested follow-ups (non-blocking)

These are quality-of-implementation gaps in the refactor, not bugs in the bool contract that #79 was about:

- **`_handle_timeout` silently returns None when the timestamp can't be retrieved or parsed** (`deploy.py:568-569, 573-574`). If `kubectl get ... creationTimestamp` fails or `fromisoformat` raises, the function returns None — same as "not yet timed out." A PR with a corrupted timestamp would never be timed out. Fail-open is conservative, but a `warn(...)` on these branches would make the inconsistency visible.
- **`_handle_timeout` docstring is incomplete** (`deploy.py:557-561`). Documents three return values but omits that None also covers "timestamp could not be retrieved or parsed." A one-line addition would prevent surprise.
- **Max-retries path in `_handle_timeout` is untested.** Both new requeue tests use `retries=0, max_retries=3`. Need a test where `retries >= max_retries` and the entry transitions to `"timed-out"` (`deploy.py:589-591`).
- **Poll-loop integration with `_handle_timeout` is untested.** The new dispatch logic (`if timeout_result is True: del slots_busy[ns]; ... elif timeout_result is False: continue`) at `deploy.py:1916-1934` is exactly the orchestration the refactor was meant to clean up. A future swap of `is True`/`is False` for truthy checks would not be caught.
- **No tests verify the warn messages are emitted.** Adding `caplog`/`capsys` assertions would lock in the operator-visible signal that #79 was specifically about.
- **120s wait-timeout-then-delete path returns True if delete succeeds.** Documented in the docstring's "best-effort cancel" sentence; consistent with Tekton finalizer behavior. No action required, but worth noting in the warn message.

The blocking items are resolved and the residual concerns are about edge behaviors and test coverage of the new refactor, not about the issue #79 contract itself. Recommend merging this and filing a follow-up for the `_handle_timeout` silent-None paths and integration test.

PR #235, opened 2026-05-28
feat: per-algorithm translation and treatment configs

kalantar:
## Summary

Closes #225

Enables multi-algorithm manifests to produce independently translated, independently built, and independently deployed algorithm packages. Each algorithm gets its own:
- Generated files subdirectory (`generated/{algo_name}/`) with full relative paths
- Config overlay (`{algo_name}_config.yaml`)
- Translation output (`{algo_name}_output.json`)
- Container image tag (`run_name-algo_name`)

Single-algorithm manifests continue to work unchanged via fallback lookup paths.

## Changes

- **`assemble.py`** — `_resolve_overlay` checks `generated/{name}/{name}_config.yaml` before flat layout
- **`source_toggle.py`** — `restore_treatment` accepts `algo_name` param; reads from per-algo subdirectory with full paths
- **`copy_generated.py`** — `algo_name` param writes to `generated/{algo_name}/` preserving full paths; also writes per-algo output JSON
- **`epp.py`** — `inject_epp_image` accepts `algo_name`; tag becomes `{run_name}-{algo_name}`
- **`prepare.py` Phase 3** — loops over algorithms, checks per-algo output existence, targets next untranslated algorithm in `skill_input.json`, writes index `translation_output.json` when all complete
- **`prepare.py` Phase 4** — `_verify_generated_dir` and `_validate_assembly` handle per-algorithm index format
- **`prepare.py` Phase 5** — summary lists all algorithms with their plugin types
- **`deploy.py`** — `ensure_images` loads per-algo outputs, applies per-algorithm source toggle for each image build
- **`run_manager.py`** — `switch_run` copies from per-algorithm subdirectories with full paths

## Reviewer attention

- **Backward compat**: All existing single-algorithm behavior preserved via fallback paths. The per-algorithm format is only written when `generated/{algo_name}/{algo_name}_output.json` files exist.
- **`translation_output.json` dual format**: Can now be either legacy single-output (`{"plugin_type": ...}`) or per-algorithm index (`{"per_algorithm": {...}}`). All consumers check for `"per_algorithm"` key first.
- **Deploy build loop**: Per-algorithm mode does baseline→apply→build→baseline for each algorithm image. This is O(N) builds but ensures clean state between algorithms.
- **Subsumes #190 and #192**: Per-algorithm subdirectories with full relative paths eliminate the basename collision class entirely.

## Test plan

- [x] All 782 existing + new tests pass
- [x] Lint passes (ruff --select F)
[...]

PR #238, opened 2026-05-28
fix: update sim2real-translate skill for per-algorithm layout

### LINKED ISSUE #237: sim2real-translate skill not updated for per-algorithm layout from PR #235
## Summary

PR #235 ("feat: per-algorithm translation and treatment configs", merged 2026-05-28) updated the **pipeline** to expect per-algorithm subdirectories (`generated/{algo_name}/{algo_name}_config.yaml`, `generated/{algo_name}/{algo_name}_output.json`) and added an `algo_name` parameter to `copy_generated.copy_generated()`. **However, the `sim2real-translate` skill itself was not updated** to use any of this. As a result, when the skill runs against a manifest with multiple algorithms it produces flat-layout artifacts that prepare.py can only consume via legacy fallback paths, and there is no signal to prepare.py that an individual algorithm has been completed.

Observed concretely on the `expceil` experiment (two algorithms: `exponentialceiling`, `constantcontrol`):
- The skill ran once with `current_algorithm=exponentialceiling` and produced `runs/r3/generated/{policy.go, policy_test.go, runner.go, treatment_config.yaml, baseline_config.yaml}` (flat).
- No `generated/exponentialceiling/exponentialceiling_output.json` was written, so `prepare.py` Phase 3 (`pipeline/prepare.py:349-369`) cannot detect that algorithm as translated.
- `constantcontrol` was never translated; the skill has no internal loop and the SKILL.md never instructs the operator to invoke it again per algorithm.

## Required changes

All paths below are inside `.claude/skills/sim2real-translate/`.

### 1. `SKILL.md` — variable loading

Currently (lines 158–159) only `RUN_NAME` and `SCENARIO` are loaded from `skill_input.json`. Add:

```bash
CURRENT_ALGORITHM=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['current_algorithm'])")
```

Halt with a clear message if `current_algorithm` is missing — this is required since PR #235 (`pipeline/prepare.py:397`).

### 2. `SKILL.md` — placeholder substitution table (lines 352–368)

Add `{ALGO_NAME}` → `$CURRENT_ALGORITHM` so the writer/reviewer prompts can be parameterised on the algorithm being translated.

### 3. `SKILL.md` — Step 6 `copy_generated` call (line 582)

Currently:
```python
copy_generated('$TARGET_REPO', '$RUN_DIR')
```
Should be:
```python
copy_generated('$TARGET_REPO', '$RUN_DIR', algo_name='$CURRENT_ALGORITHM')
```
Per `scripts/copy_generated.py:18-26`, this is what triggers per-algorithm subdirectory layout and per-algo `{algo_name}_output.json` emission. Without it, prepare.py Phase 3's per-algo existence check (`prepare.py:318`, `prepare.py:353`) always misses.

### 4. `SKILL.md` — Step 6 verification (lines 615–632)

Currently reads top-level `translation_output.json` and expects flat `generated/{basename}` files. After change 3, it must:
- Read `$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json`
- Verify `files_created + files_modified` exist as `generated/$CURRENT_ALGORITHM/{full/relative/path}` (full path preserved, not basename)

### 5. `SKILL.md` — Step 6 completion summary (lines 645–660)

Update the printed paths from:
```
$RUN_DIR/translation_output.json
$RUN_DIR/generated/
```
to per-algo paths, and add a hint that the operator must re-run `prepare.py` to pick up the next untranslated algorithm (the cross-algorithm loop is owned by prepare.py per PR #235).

### 6. `SKILL.md` — resumability check (lines 215–223)

The current check ("If `translate` phase is `done` and `translation_output.json` exists") needs to be aware of the per-algorithm format described in PR #235:
- prepare.py writes the top-level `translation_output.json` only when ALL algorithms are done (`prepare.py:429-440`)
- Per-algorithm completion lives in `generated/{algo_name}/{algo_name}_output.json`

For a single-algorithm skill invocation, the check should be: if `generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json` exists and its referenced files are present, halt as already complete.

### 7. `prompts/agent-writer.md` — Phase 3 treatment config path

Currently (lines 78, 97, 100, 118, 139, 141, 158, 218, 224, 232, 255, 264) the writer is instructed to write `{RUN_DIR}/generated/treatment_config.yaml`. Per PR #235 (`pipeline/lib/assemble.py:106-126`) the canonical path is now `generated/{algo_name}/{algo_name}_config.yaml`; flat `treatment_config.yaml` is only a *legacy* fallback. The writer prompt should:
- Add `{ALGO_NAME}` to its variable header
- Write the treatment overlay to `{RUN_DIR}/generated/{ALGO_NAME}/{ALGO_NAME}_config.yaml`
- Send `treatment-ready: …{ALGO_NAME}/{ALGO_NAME}_config.yaml` so the main session displays the right path
- Update the Phase 4.5 `translation_output.json` writer to instead write `{RUN_DIR}/generated/{ALGO_NAME}/{ALGO_NAME}_output.json` (the top-level index is owned by prepare.py)
- Update the Step 3 snapshot script to copy the per-algo config

### 8. `prompts/agent-writer.md` — Phase 2 baseline config path

Baseline overlay handling needs a design decision:

- `assemble.py:_resolve_overlay` (lines 106–126) treats `generated/baseline_config.yaml` as a *shared fallback* across algorithms.
- With multi-algorithm manifests, every skill invocation writes the same baseline.

Two acceptable resolutions, pick one and document:
- **(a) Share**: keep writing `generated/baseline_config.yaml` flat, but the writer must skip Phase 2 if it already exists (probably the right answer — baselines don't change between algorithms).
- **(b) Per-algo**: write `generated/{ALGO_NAME}/{ALGO_NAME}_baseline_config.yaml` and update assemble.py to look there first. Higher impact; only worth it if baselines can ever differ per algorithm.

### 9. `prompts/agent-reviewer.md` — paths it reads

Currently (lines 70–72) the reviewer reads `{RUN_DIR}/generated/treatment_config.yaml` and `{RUN_DIR}/translation_output.json`. After change 7 these paths shift to `{RUN_DIR}/generated/{ALGO_NAME}/{ALGO_NAME}_config.yaml` and `{RUN_DIR}/generated/{ALGO_NAME}/{ALGO_NAME}_output.json`. Criterion 4/5/6 references to `treatment_config.yaml` (lines 116, 123, 135, 150, 160, 165) need the same update.

### 10. Tests

`scripts/copy_generated.py` already has `tests/test_copy_generated.py` covering the `algo_name` mode (added in PR #235). The skill-level changes above need at minimum:

- A unit test on the substitution layer that confirms `{ALGO_NAME}` lands in the writer/reviewer prompts.
- An end-to-end fixture that runs the skill against a two-algorithm manifest and asserts: per-algo subdirectories, per-algo `{algo_name}_output.json`, and that re-running the skill twice in succession (once per `current_algorithm`) produces both algorithms before prepare.py writes the index.

## Out of scope

- Updating `prepare.py` — already done in PR #235.
- Updating `copy_generated.py` — already done in PR #235.
- Updating `assemble.py`, `source_toggle.py`, `epp.py`, `deploy.py`, `run_manager.py` — already done in PR #235.

## Why this slipped through PR #235

PR #235's test plan box for "Manual: run pipeline with expceil experiment (two algorithms) to verify end-to-end" was unchecked at merge. The skill's own SKILL.md / prompt files were not in the PR's `files` list — only `scripts/copy_generated.py` and its test file under the skill directory.

## Reference

- Merged PR: https://github.com/inference-sim/sim2real/pull/235
- Closes related: https://github.com/inference-sim/sim2real/issues/225
- Concrete failure observed in: `kalantar-msb/expceil` workspace, run `r3` (only `exponentialceiling` translated, flat layout, `constantcontrol` untouched).

kalantar:
## Summary

- SKILL.md: loads `current_algorithm` from `skill_input.json`, adds `{ALGO_NAME}` placeholder, passes `algo_name='$CURRENT_ALGORITHM'` to `copy_generated()`, verifies output at per-algo subdirectory
- Writer prompt: writes treatment config to `generated/{ALGO_NAME}/{ALGO_NAME}_config.yaml` and output JSON to `generated/{ALGO_NAME}/{ALGO_NAME}_output.json`; skips Phase 2 if `baseline_config.yaml` already exists
- Reviewer prompt: reads from per-algo paths instead of flat `treatment_config.yaml`/`translation_output.json`
- New test: `test_skill_substitution.py` validates placeholder presence in all three files

## Design decision

Baseline config remains **shared/flat** (`generated/baseline_config.yaml`) — it doesn't vary between algorithms, so the writer skips Phase 2 if it already exists from a prior algorithm's skill invocation. This matches option (a) from issue #237 item 8.

## Reviewer attention

- The `copy_generated` script already supports the `algo_name` parameter (added in PR #235). This PR only wires it through the skill orchestration.
- `prepare.py` owns the cross-algorithm loop — it invokes the skill once per algorithm via the `current_algorithm` field. The skill itself does NOT loop.
- The `.state.json` `translate` phase entry now includes an `algorithm` field identifying which algorithm was translated in that invocation.

## Test plan

- [x] `ruff check pipeline/ .claude/skills/ --select F` — passes
- [x] `python3 -m pytest pipeline/ .claude/skills/sim2real-translate/tests/ -v` — 798 tests pass
- [ ] Manual: run pipeline with multi-algorithm experiment (expceil) to verify end-to-end

Closes #237

### DIFF
--- a/.claude/skills/sim2real-translate/SKILL.md
+++ b/.claude/skills/sim2real-translate/SKILL.md
@@ -139,7 +139,8 @@ import json, sys
 si = json.load(open('$RUN_DIR/skill_input.json'))
 required = ['run_name', 'run_dir', 'scenario', 'context_path', 'manifest_path',
             'algorithm_source', 'baseline_sim_config',
-            'target', 'build_commands', 'config_kind', 'context']
+            'target', 'build_commands', 'config_kind', 'context',
+            'current_algorithm']
 missing = [f for f in required if f not in si]
 if missing:
     print(f'HALT: skill_input.json missing fields: {missing}')
@@ -163,6 +164,7 @@ ALGO_SOURCE=$EXPERIMENT_ROOT/$(python3 -c "import json; print(json.load(open('$R
 ALGO_CONFIG=$(python3 -c "import json; v=json.load(open('$RUN_DIR/skill_input.json')).get('algorithm_config'); print('$EXPERIMENT_ROOT/' + v if v else '')")
 TARGET_REPO=$EXPERIMENT_ROOT/$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['target']['repo'])")
 CONFIG_KIND=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['config_kind'])")
+CURRENT_ALGORITHM=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['current_algorithm'])")
 CONTEXT_TEXT=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json')).get('context', {}).get('text', ''))")
 BASELINE_SIM_CONFIG=$(python3 -c "import json; v=json.load(open('$RUN_DIR/skill_input.json'))['baseline_sim_config']; print('$EXPERIMENT_ROOT/' + v if v else '')")
 BASELINE_REAL_CONFIG=$(python3 -c "import json; v=json.load(open('$RUN_DIR/skill_input.json')).get('baseline_real_config'); print('$EXPERIMENT_ROOT/' + v if v else 'null')")
@@ -215,12 +217,13 @@ BASELINE_DONE=$(grep '^BASELINE_DONE=' /tmp/resume_state.txt | cut -d= -f2)
 TREATMENT_DONE=$(grep '^TREATMENT_DONE=' /tmp/resume_state.txt | cut -d= -f2)
 ```
 
-**If `translate` phase is `done` and `translation_output.json` exists:**
-- Read `translation_output.json` and verify all `files_created` + `files_modified` exist
-  in `$RUN_DIR/generated/`
+**If `translate` phase is `done` and per-algorithm output exists:**
+- Check `$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json`
+- Read it and verify all `files_created` + `files_modified` exist
+  in `$RUN_DIR/generated/$CURRENT_ALGORITHM/`
 - If complete: print the summary (see Step 6) and HALT with:
-  `Translation already complete. Re-run prepare.py to continue.`
-- If `generated/` is missing files: jump directly to Step 6 to re-copy
+  `Translation for $CURRENT_ALGORITHM already complete. Re-run prepare.py to continue.`
+- If `generated/$CURRENT_ALGORITHM/` is missing files: jump directly to Step 6 to re-copy
 
 **If `context_file_populated=true` in state AND `$CONTEXT_PATH` exists AND `REBUILD_CONTEXT=false`:**
 - Skip Step 1 entirely
@@ -363,6 +366,7 @@ corresponding shell variables:
 - `{CONFIG_KIND}` → `$CONFIG_KIND`
 - `{REVIEW_ROUNDS}` → `$REVIEW_ROUNDS`
 - `{SCENARIO}` → `$SCENARIO`
+- `{ALGO_NAME}` → `$CURRENT_ALGORITHM`
 - `{BUILD_COMMANDS}` → `$BUILD_COMMANDS`
 - `{CONTEXT_TEXT}` → `$CONTEXT_TEXT`
 - `{MAIN_SESSION_NAME}` → `"main-session"`
@@ -432,7 +436,7 @@ SendMessage("writer", "continue")
 
 **On "treatment-ready: ...":**
 
-Read `$RUN_DIR/generated/treatment_config.yaml` and print to user:
+Read `$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_config.yaml` and print to user:
 ```
 ━━━ Treatment Config (derived from baseline + algorithm) ━━━
 <file contents>
@@ -471,7 +475,7 @@ import json
 o = json.load(open('$RUN_DIR/translation_output.json'))
 print('Plugin files:', o.get('files_created', []))
 "
-python3 -c "print(open('$RUN_DIR/generated/treatment_config.yaml').read())"
+python3 -c "print(open('$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_config.yaml').read())"
 ```
 
 Print to user:
@@ -579,7 +583,7 @@ python3 -c "
 import sys
 sys.path.insert(0, '$REPO_ROOT/.claude/skills/sim2real-translate/scripts')
 from copy_generated import copy_generated
-copy_generated('$TARGET_REPO', '$RUN_DIR')
+copy_generated('$TARGET_REPO', '$RUN_DIR', algo_name='$CURRENT_ALGORITHM')
 " || exit 1
 ```
 
@@ -593,10 +597,13 @@ from datetime import datetime, timezone
 
 state_path = Path('$RUN_DIR') / '.state.json'
 state = json.loads(state_path.read_text())
-o = json.load(open('$RUN_DIR/translation_output.json'))
+algo = '$CURRENT_ALGORITHM'
+algo_out = Path('$RUN_DIR/generated') / algo / f'{algo}_output.json'
+o = json.load(open(algo_out))
 state.setdefault('phases', {})['translate'] = {
     'status': 'done',
     'timestamp': datetime.now(timezone.utc).isoformat(),
+    'algorithm': '$CURRENT_ALGORITHM',
     'files': o['files_created'],
     'review_rounds': $REVIEW_ROUNDS_DONE,
     'consensus': '$FINAL_CONSENSUS',
@@ -615,19 +622,24 @@ Verify outputs:
 python3 -c "
 import json, sys
 from pathlib import Path
-o = json.load(open('$RUN_DIR/translation_output.json'))
+algo = '$CURRENT_ALGORITHM'
+algo_dir = Path('$RUN_DIR/generated') / algo
+algo_out = algo_dir / f'{algo}_output.json'
+if not algo_out.exists():
+    print(f'ERROR: {algo_dir}/{algo}_output.json not found')
+    sys.exit(1)
+o = json.load(open(algo_out))
 required = ['plugin_type', 'files_created', 'files_modified', 'package',
             'register_file', 'test_commands', 'config_kind',
             'treatment_config_generated', 'description']
 missing = [f for f in required if f not in o]
 if missing:
-    print(f'ERROR: translation_output.json missing: {missing}')
+    print(f'ERROR: {algo}_output.json missing: {missing}')
     sys.exit(1)
-gen = Path('$RUN_DIR/generated')
 for f in o['files_created'] + o.get('files_modified', []):
-    dst = gen / Path(f).name
+    dst = algo_dir / f
     if not dst.exists():
-        print(f'ERROR: generated/ missing: {Path(f).name}')
+        print(f'ERROR: generated/{algo}/{f} not found')
         sys.exit(1)
 print(f'Verified: plugin_type={o[\"plugin_type\"]}, '
       f'{len(o[\"files_created\"])} created, '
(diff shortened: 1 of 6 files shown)


kalantar:
## NEEDS CHANGES

All 10 requirements from issue #237 are addressed in the diff, but two runtime-breaking bugs prevent the skill from completing successfully in the per-algorithm workflow.

---

### Blocking Issues

**1. SKILL.md "review-passed" handler still reads top-level `translation_output.json` (line ~475)**

The writer now produces only `generated/{algo}/{algo}_output.json` (per the updated Phase 4.5 in agent-writer.md), but the review-passed display block still does:
```python
o = json.load(open('$RUN_DIR/translation_output.json'))
```
This file does not exist at this point in the workflow (prepare.py creates the top-level index only after ALL algorithms are translated). The handler will crash with `FileNotFoundError`.

Fix: `json.load(open('$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json'))`

**2. `copy_generated.py` contract break — reads `translation_output.json` that no longer exists**

`copy_generated()` unconditionally reads `rd / "translation_output.json"` (lines 76-81) regardless of whether `algo_name` is provided. Since the writer no longer creates this file, Step 6 will crash when calling `copy_generated('$TARGET_REPO', '$RUN_DIR', algo_name='$CURRENT_ALGORITHM')`.

The PR body states "The `copy_generated` script already supports the `algo_name` parameter" — true for per-algo *output* writing, but the *input* dependency on `translation_output.json` was not removed. Per CLAUDE.md "Pipeline Stage Contracts": changing an upstream output requires updating every consumer.

Fix: When `algo_name` is set, `copy_generated` should read from `generated/{algo_name}/{algo_name}_output.json` instead of the top-level file.

---

### Secondary Issues

**3. Snapshot script silently skips missing config** — The old code unconditionally copied the treatment config; the new code uses `if algo_cfg.exists()`. At snapshot time, the config should always exist (it was just written in Phase 3). This guard masks failures and produces incomplete snapshots with no warning.

**4. Empty `current_algorithm` not validated** — The presence check passes for an empty string, which would cascade into broken paths (`generated//`, `_config.yaml`). Add a truthiness check after the required-fields validation.

**5. Stale prose at line ~579** — Still says "overwrite `translation_output.json`"; should reference per-algo output.

---

### Needs Verification

**State marking and the multi-algorithm loop** — Step 6 sets `state.phases.translate.status = 'done'` after a single algorithm. If prepare.py checks this state before its per-algorithm file scan, it would skip Phase 3 for subsequent algorithms. Verify that prepare.py's per-algo detection (lines 349-369) runs before the phase status check.

---

### Test Coverage

The substitution tests (5 tests) satisfy the "unit test on the substitution layer" requirement. The e2e two-algorithm fixture remains unchecked — this is the same gap that caused issue #237 in the first place. Consider adding a negative assertion that `translation_output.json` references do not appear in SKILL.md executable code blocks (the current tests only check prompt files for `treatment_config.yaml`).

kalantar:
## NEEDS CHANGES (updated review)

All 10 requirements from issue #237 are addressed in the diff, but three critical runtime bugs prevent the multi-algorithm workflow from functioning. This is the exact scenario issue #237 targets.

---

### Critical — must fix before merge

**1. SKILL.md line 475: "review-passed" handler reads nonexistent `translation_output.json`**

The writer now writes only `generated/{algo}/{algo}_output.json` (Phase 4.5), but the review-passed handler still does:
```python
o = json.load(open('$RUN_DIR/translation_output.json'))
```
This file does not exist at this point. Guaranteed `FileNotFoundError` on every run.

Fix: `json.load(open('$RUN_DIR/generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json'))`

**2. `copy_generated.py` lines 76-81: unconditionally reads `translation_output.json`**

```python
to_path = rd / "translation_output.json"
o = json.loads(to_path.read_text())  # FileNotFoundError
```

This runs regardless of `algo_name`. The writer no longer creates this file. Step 6 will crash.

The issue body says `copy_generated.py` is "out of scope" — but that was written assuming the writer still produces `translation_output.json`. It does not. The fix is small: wrap lines 76-81 in `if algo_name is None:` and have the per-algo branch read directly from `generated/{algo}/{algo}_output.json` for metadata fields.

**3. State marking blocks the multi-algorithm loop (CONFIRMED)**

SKILL.md Step 6 sets `state['phases']['translate']['status'] = 'done'` after translating ONE algorithm. On re-run, `prepare.py` line 291:
```python
if state.is_done("translate") and not args.force:
    info("[skip] Translation already complete")
    return
```
This fires BEFORE the per-algorithm file scan (line 349). The second algorithm never gets translated. No error is emitted — just `[skip] Translation already complete`.

Fix: The skill should NOT mark `translate` as "done." Phase completion is owned by `prepare.py` (via `_write_translation_index` after ALL algorithms complete, line 361). Either remove the `translate` phase status entirely or use a per-algorithm key (e.g., `translate_exponentialceiling`) that `is_done("translate")` doesn't match.

---

### Secondary issues

**4. Snapshot guard silently drops config:** `if algo_cfg.exists()` should fail loudly — at snapshot time, the config must exist. Silent skip produces incomplete snapshots with no warning.

**5. Empty `current_algorithm` not validated:** Presence check passes for `""`. Cascade: paths become `generated//`, `_config.yaml`. Add truthiness check after required-fields validation.

**6. Test gap:** Tests assert `treatment_config.yaml` is absent from prompt files, but do NOT assert `translation_output.json` is absent from SKILL.md executable blocks. This gap allowed issue #1 to pass CI.

---

### Summary

The three critical bugs form an interaction chain that breaks the per-algorithm workflow at multiple points:
- Bug #2 crashes at "review-passed" (FileNotFoundError)
- Bug #1 crashes at Step 6 copy (FileNotFoundError)  
- Bug #3 silently blocks the second algorithm even if bugs 1 and 2 are fixed

All three must be fixed in this PR for multi-algorithm experiments to work.

kalantar:
All six issues addressed in 18810d8:

**Runtime bugs fixed:**
- Review-passed handler now reads `generated/$CURRENT_ALGORITHM/${CURRENT_ALGORITHM}_output.json` instead of the non-existent `translation_output.json`
- `copy_generated.py` in per-algo mode reads/updates the per-algo output JSON directly — no longer depends on top-level `translation_output.json` existing
- State marking no longer sets `translate.status = done` after a single algorithm (which would cause prepare.py line 291 to skip Phase 3 for subsequent algorithms). Instead records completion in a `completed_algorithms` sub-dict that prepare.py doesn't check.

**Secondary fixes:**
- Snapshot copies algo config unconditionally (removed silent `if exists()` guard)
- Added truthiness validation for `current_algorithm` (empty string rejected with clear HALT message)
- Stale prose "overwrite translation_output.json" updated

**Test coverage:**
- Added negative assertion test: `test_skill_md_no_translation_output_in_code_blocks` — verifies no code block in SKILL.md references `translation_output.json`
- Updated per-algo `copy_generated` tests to match new contract (pre-create per-algo output JSON, since `copy_generated` now reads it instead of the top-level file)

All 799 tests pass, lint clean.

kalantar:
## READY TO MERGE

The second commit (18810d8) addresses all three critical issues from prior reviews. All 10 requirements from issue #237 are fully implemented. Tests pass (798 tests including 6 new).

### What was fixed in commit 2:
1. **SKILL.md line 475**: Now reads per-algo output (no more FileNotFoundError)
2. **copy_generated.py**: Per-algo mode reads/updates per-algo output JSON; flat mode left unchanged for backward compat
3. **State marking**: Uses `completed_algorithms` sub-dict — does not set `translate.status = 'done'`, so `prepare.py` line 291 won't block subsequent algorithms
4. **Snapshot**: Unconditional copy (no silent `if exists()`)
5. **Empty validation**: Truthiness check added for `current_algorithm`
6. **New test**: `test_skill_md_no_translation_output_in_code_blocks` prevents regression

### Non-blocking follow-ups (optional):

**Docstring update** — `copy_generated.py` line 19 still says "update translation_output.json" in the one-liner and "run_dir: containing translation_output.json" in Args. Both are stale for the per-algo path.

**Resumability gap** — The SKILL.md resume check fires on `translate.status == 'done'` (set only by prepare.py after ALL algorithms). It doesn't check `completed_algorithms[current_algo]`, so re-running the skill for an already-completed algorithm (without re-running prepare.py first) would redo the work. Low-risk since prepare.py controls skill_input.json and won't target a completed algorithm.

**Print statement** — Step 6 prints "State: translate done" but the code now only records per-algorithm completion, not phase-level done. Minor mismatch.

PR #241, opened 2026-05-29
fix: make per-pair gpu_cost derived state, not persisted

### LINKED ISSUE #240: deploy: make per-pair gpu_cost derived state, not persisted
## Problem

`deploy.py run` logs `requires N GPUs` per pair from a value persisted in the run's progress ConfigMap. That value is computed once when the pair is first registered and is never refreshed on subsequent invocations, so any condition that produced a wrong value on the first run is locked in for the life of the run.

Concrete case: an experiment with `decode.replicas: 4` in every pair's `scenarioContent` logged

```
[INFO]  wl-codecompletion-mid-constantcontrol requires 1 GPUs
[OK]    [wl-codecompletion-mid-constantcontrol] → kalantar-0 (...)
[INFO]  wl-codecompletion-mid-expceil requires 1 GPUs
[INFO]  wl-codecompletion-mid-exponentialceiling requires 1 GPUs
```

while four GPU-bearing decode pods started for each pair. Re-running `_derive_pair_gpu_costs` against the current files yields `4` for all three pairs; the persisted `1` came from the first registration, when one of the three fallback exits in `_derive_pair_gpu_costs` fired (`defaults` is `None` at deploy.py:1698, empty/unparseable `scenarioContent` at deploy.py:1712, or derivation error at deploy.py:1717 — the last is the only one that emits a `warn()`).

## Root cause

deploy.py:1831–1843 only sets `gpu_cost` inside the `if key not in progress` branch:

```python
for key, meta in discovered.items():
    if key not in progress:
        progress[key] = {
            ...
            "gpu_cost": pair_costs[key],
            ...
        }
```

There is no path that updates `gpu_cost` for an existing entry, and `progress` survives across invocations via `ConfigMapProgressStore`. `gpu_cost` is being treated as durable state, but it is a deterministic function of `(scenarioContent, defaults, --default-gpu-cost)`, all of which are visible at the start of every run. Persisting it adds no value and creates exactly this staleness class.

## Proposed fix

**Stop persisting `gpu_cost`.** Compute it once per `_cmd_run` invocation, hold it in a local `dict[str, int]`, and pass that map (or a closure over it) to every consumer:

- `_capacity_limited_pairs` (deploy.py:1724) — receives the cost map instead of reading `progress[k]["gpu_cost"]`.
- backoff scarcity / capacity signals (deploy.py:2008, 2009, 2031, 2044) — read from the cost map.
- dispatch log line (deploy.py:2071–2072) — read from the cost map.

`ConfigMapProgressStore` keeps only durable per-pair fields: `status`, `namespace`, `completed_namespace`, `retries`, `pending_stalls`, `pending_since`. Old ConfigMaps that still carry a `gpu_cost` key are harmless (no reader looks for it).

## Orthogonal cleanup (do in same PR)

1. **Surface provenance in the dispatch log.** Today `info(f"{pair_key} requires {pair_cost} GPUs")` (deploy.py:2072) gives no hint whether the value came from a parsed scenario or a fallback. Have `_derive_pair_gpu_costs` return per-pair `(cost, source)` where `source ∈ {"derived", "defaults-only", "fallback"}`, and include the source in the log:

   ```
   wl-foo-bar requires 4 GPUs (derived from scenarioContent)
   wl-foo-bar requires 1 GPUs (fallback default — scenarioContent missing)
   ```

   This would have made the original debugging session instant.

2. **Log every fallback exit, not just one.** `_derive_pair_gpu_costs` currently only emits `warn()` on path 3 (derivation error, deploy.py:1717–1719). Paths 1 (`defaults is None`, line 1698) and 2 (empty/unparseable `scenarioContent`, line 1712) are silent — the same silence that hid this bug. Each fallback should `warn()` once per pair.

## Dead code to eliminate

After the change:

- deploy.py:1840 — `"gpu_cost": pair_costs[key]` removed from progress entry creation.
- deploy.py:1729, 1737, 1742 — `default_gpu_cost` parameter and the `progress[k].get("gpu_cost", default_gpu_cost)` defaulting pattern in `_capacity_limited_pairs` go away (the cost map is fully populated for every pending key, so no defaulting is needed at lookup sites).
- deploy.py:2008, 2009, 2031, 2044, 2071 — same pattern: `.get("gpu_cost", fallback_cost)` becomes a direct lookup in the cost map.

`--default-gpu-cost` CLI flag and the `fallback_cost` value remain, because `_derive_pair_gpu_costs` itself still needs a fallback when scenario derivation genuinely fails.

## Test impact

- `tests/test_deploy_run.py:test_init_progress_stores_gpu_cost` (line 1025) and `test_init_progress_gpu_cost_uses_fallback` (line 1060) assert that `progress[key]["gpu_cost"]` is written. These become invalid and should be replaced by tests asserting the cost map is recomputed correctly across invocations — including the case where `scenarioContent` changes between runs (the regression that motivated this issue).
- Tests around `_capacity_limited_pairs` (test_deploy_run.py:1098+, 1127+) that pre-populate `gpu_cost` in progress fixtures should pass a cost map argument instead.

kalantar:
## Summary

- Stop persisting `gpu_cost` in the ConfigMap progress store — compute it fresh at the start of each `_cmd_run` invocation from `scenarioContent` + defaults
- `_derive_pair_gpu_costs` now returns `(cost, source)` tuples for provenance tracking
- `_capacity_limited_pairs` takes a `cost_map` dict instead of reading from `progress` entries
- Dispatch log now shows provenance: `"derived from scenarioContent"`, `"derived from defaults only"`, or `"fallback default"`
- All three fallback exits in `_derive_pair_gpu_costs` now emit `warn()` (previously two were silent)

Closes #240

## Reviewer attention

- **Existing ConfigMaps** that still carry a `gpu_cost` key are harmless — no reader accesses it after this change. No migration needed.
- **Test fixtures** in other test functions (backoff, pending stalls) still include `"gpu_cost": 1` in their progress dicts as dead keys. Left intentionally to minimize diff churn — they are unused by production code.
- The `--default-gpu-cost` CLI flag and `fallback_cost` value remain because `_derive_pair_gpu_costs` itself still needs a fallback when scenario derivation genuinely fails.

## Test plan

- [x] All 758 tests pass (`python -m pytest pipeline/ -v`)
- [x] No remaining `progress[k].get("gpu_cost"` or `entry.get("gpu_cost"` patterns in deploy.py
- [x] New provenance tests cover all four derivation paths (derived, defaults-only, fallback via None defaults, fallback via error)
- [x] `_capacity_limited_pairs` tests updated to use `cost_map` parameter

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1681,22 +1681,21 @@ def _derive_pair_gpu_costs(
     *,
     defaults: dict | None,
     fallback_cost: int,
-) -> dict[str, int]:
+) -> dict[str, tuple[int, str]]:
     """Compute GPU cost per pair from its scenarioContent.
 
-    If defaults is None, returns fallback_cost for all pairs immediately.
-
-    Otherwise, per pair:
-      1. Parse scenarioContent → gpu_cost_per_pair(scenario, defaults)
-      2. If scenarioContent missing/invalid → gpu_cost_per_pair({}, defaults)
-      3. If derivation returns an error → fallback_cost
+    Returns dict mapping pair key to (cost, source) where source is one of:
+    - "derived": cost parsed from scenarioContent
+    - "defaults-only": scenarioContent missing/invalid, derived from defaults
+    - "fallback": derivation failed or defaults unavailable, using fallback_cost
     """
     from pipeline.lib.capacity import gpu_cost_per_pair
 
-    costs = {}
+    costs: dict[str, tuple[int, str]] = {}
     for key, meta in discovered.items():
         if defaults is None:
-            costs[key] = fallback_cost
+            warn(f"{key}: no defaults available — using fallback cost ({fallback_cost})")
+            costs[key] = (fallback_cost, "fallback")
             continue
 
         scenario_content = meta.get("scenario_content")
@@ -1709,37 +1708,39 @@ def _derive_pair_gpu_costs(
 
         if resolved and isinstance(resolved, dict):
             result = gpu_cost_per_pair(resolved, defaults)
+            source = "derived"
         else:
+            if scenario_content:
+                warn(f"{key}: scenarioContent not parseable as dict — deriving cost from defaults only")
+            else:
+                warn(f"{key}: no scenarioContent — deriving cost from defaults only")
             result = gpu_cost_per_pair({}, defaults)
+            source = "defaults-only"
 
         if isinstance(result, int):
-            costs[key] = result
+            costs[key] = (result, source)
         else:
             warn(f"{key}: GPU cost derivation failed: {result} — using fallback ({fallback_cost})")
-            costs[key] = fallback_cost
+            costs[key] = (fallback_cost, "fallback")
 
     return costs
 
 
 def _capacity_limited_pairs(
     pending: list[str],
-    progress: dict,
     *,
     free_gpus: int,
-    default_gpu_cost: int,
+    cost_map: dict[str, int],
 ) -> list[str]:
     """Select pending pairs that fit within available GPU capacity.
 
     Sorts by gpu_cost ascending to maximize dispatch count.
     """
-    sorted_pending = sorted(
-        pending,
-        key=lambda k: progress[k].get("gpu_cost", default_gpu_cost),
-    )
+    sorted_pending = sorted(pending, key=lambda k: cost_map[k])
     result = []
     budget = free_gpus
     for pair in sorted_pending:
-        cost = progress[pair].get("gpu_cost", default_gpu_cost)
+        cost = cost_map[pair]
         if budget >= cost:
             budget -= cost
             result.append(pair)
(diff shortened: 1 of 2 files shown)


kalantar:
## Verdict: READY TO MERGE

PR fully addresses all requirements from issue #240: gpu_cost is no longer persisted, costs are derived fresh per invocation, provenance tuples surface the source in dispatch logs, and all fallback paths now emit `warn()`.

### Key findings

**Correctness:** The invariant that `pair_costs` covers all keys in `_pending_pairs()` holds because both derive from `discovered` (PipelineRun YAMLs on disk). Progress initialization only creates entries for keys in `discovered`, so no orphaned progress key can appear in `pending` without also being in `pair_costs`. The change from `.get("gpu_cost", fallback)` to direct `cost_map[k]` lookup is safe given this invariant and would surface invariant violations earlier (KeyError at cost lookup vs. silent dispatch failure later).

**Non-blocking suggestions:**

1. **Stale docstring terminology** — `_capacity_limited_pairs` docstring still says "Sorts by gpu_cost ascending." Should be "Sorts by cost ascending" to match the renamed `cost_map` parameter.

2. **Double-warning on YAML parse error** — When `scenario_content` is present but invalid YAML, the catch block logs "invalid YAML" and then the subsequent `if scenario_content:` branch logs "not parseable as dict." The second warning is misleading since the real problem is a parse error, not a type mismatch.

3. **O(n) identical warnings when `defaults is None`** — Every pair emits its own "no defaults available" warning. Since `defaults` is the same for all pairs, a single summary warning before the loop would be cleaner.

4. **Test gap: warn() calls not verified** — The new `warn()` calls are untested. A `capsys` assertion or mock on at least the "no defaults" and "derivation error" paths would catch regressions.

None of these are merge-blockers — the core logic is correct, the design is sound, and all 758 tests pass.

PR #243, opened 2026-05-29
fix: scoped collect dispatches per-slot, not single namespace

### LINKED ISSUE #242: deploy.py collect: scoped path (--only/--workload) ignores per-pair completed_namespace, extracts all phases from a single slot
## Summary

`deploy.py collect` correctly records each completed pair's slot in the run-scoped progress ConfigMap (`completed_namespace`), and the **unscoped** collect path uses that mapping to dispatch one extractor pod per slot. The **scoped** path (`--only` and/or `--workload`) does not — it picks the *first* matching pair's `completed_namespace` for a given workload and uses that single namespace for **all** phases of that workload. When a workload's packages were spread across multiple slots (the normal parallel-pool case), this collects only from one slot and silently produces empty/missing results from the others.

## Reproduction

Run with packages spread across slots (here three packages, one workload, three slots):

```
$ python $SIM2REAL/pipeline/deploy.py reset --workload codecompletion_mid
$ python $SIM2REAL/pipeline/deploy.py run --remote --workload codecompletion_mid
$ python $SIM2REAL/pipeline/deploy.py collect --workload codecompletion_mid

━━━ Step 1: Collecting Results ━━━
[WARN]  mtime probe failed (rc=1): find: /data/r4/expceil: No such file or directory
command terminated with exit code 1 — falling back to full copy
[WARN]  mtime probe failed (rc=1): find: /data/r4/exponentialceiling: No such file or directory
command terminated with exit code 1 — falling back to full copy
[OK]    constantcontrol/codecompletion_mid    (kalantar-0)
[OK]    expceil/codecompletion_mid    (kalantar-0)
[OK]    exponentialceiling/codecompletion_mid    (kalantar-0)
```

Verify against the live ConfigMap:

```
$ kubectl get configmap sim2real-progress-r4 -n kalantar-0 -o jsonpath='{.data.progress}' \
    | jq -r '.[] | select(.workload=="codecompletion_mid") | "\(.package) -> \(.completed_namespace)"'
constantcontrol     -> kalantar-0
expceil             -> kalantar-1
exponentialceiling  -> kalantar-2
```

The progress data is correct. `collect` reports every pair as `(kalantar-0)` and the `find: /data/r4/{expceil,exponentialceiling}: No such file or directory` warnings show the extractor was launched in the wrong slot — those PVCs hold no data for those packages. The "falling back to full copy" path then masks the miss and reports `[OK]`.

## Affected flags

The scoped branch is gated by `scoped = (--only is not None) or (--workload is not None)` at `pipeline/deploy.py:983-985`.

- `--only` — affected
- `--workload` — affected
- `--only`/`--workload` in combination with `--package` — affected
- `--package` alone — **not** affected (falls through to the unscoped branch, which dispatches per-slot correctly)

## Root cause

`pipeline/deploy.py:1049-1068` (scoped branch):

```python
for wl_name in scoped_workloads:
    wl_ns = next(
        (progress[k].get("completed_namespace")
         for k in collectible
         if isinstance(progress[k], dict)
         and progress[k].get("workload") == wl_name
         and progress[k].get("completed_namespace")),
        None,
    )
    ...
    errors = _extract_phases_from_pvc(
        phases_to_collect, run_name, wl_ns, run_dir,
        skip_logs=skip_logs, workload=wl_name)
```

The iteration is by `wl_name` only, and `wl_ns` is the first matching pair's namespace. Iteration through the dict is insertion order, so `wl_ns` becomes whichever slot the first package landed on. Then a single extractor pod is launched in that one namespace with `phases_to_collect = [<all phases>]`.

The unscoped branch at `pipeline/deploy.py:1138-1247` already does the right thing: it builds `ns_phase_map: dict[str, list[str]]` and `ns_pair_map: dict[str, set[(pkg, wl)]]` keyed by `completed_namespace`, then runs one extractor pod per slot (in parallel via `ThreadPoolExecutor` when `len(ns_items) > 1`). Each pod is invoked with `allowed_workloads` restricting which (pkg, wl) pairs it pulls. The scoped branch should follow the same shape, restricted to the in-scope pairs.

## Expected behavior

`collect --only` / `--workload` (with or without `--package`) should group `done` pairs in scope by `completed_namespace` and extract per-slot — same model the unscoped path uses — so each phase is pulled from the slot that actually ran it.

## Secondary concern

The fallback at the `find: /data/<run>/<phase>: No such file or directory` step that "falls back to full copy" still ends up logging `[OK]` for the pair. Even after the routing fix, a missing source directory in the chosen slot probably should not surface as success — it likely warrants a real error, since by then the routing layer asserts the data is there.

## Environment

- run dir: `workspace/runs/r4`
- workload: `codecompletion_mid`
- packages: `constantcontrol`, `expceil`, `exponentialceiling`
- slots: `kalantar-0`, `kalantar-1`, `kalantar-2`

kalantar:
## Summary

Closes #242

- Rewrote the scoped `collect --only/--workload` branch to group collectible pairs by `completed_namespace` and dispatch one extractor pod per slot (parallel when >1 slot), matching the unscoped path's per-slot model.
- Removed the broken per-workload loop that picked only the first matching pair's namespace for all phases.
- Added `test_collect_scoped_multi_slot_per_workload` reproducing the exact bug scenario (one workload, packages in 3 different slots).

## Reviewer attention

The scoped branch now uses `allowed_workloads` + `on_workload_done` callbacks (same as the unscoped branch) instead of the old `workload=` parameter. Eight existing tests were updated to assert on the new dispatch shape. The behavioral contract is unchanged — scoped collect still restricts extraction to in-scope pairs — but the extraction is now correctly routed to the slot where each pair actually ran.

## Test plan

- [x] New test `test_collect_scoped_multi_slot_per_workload` verifies 3-slot dispatch
- [x] All 59 collect tests pass
- [x] Full pipeline suite: 759 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1013,9 +1013,6 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         scoped_phases = sorted({
             progress[k].get("package", "") for k in collectible
         } - {""})
-        scoped_workloads = sorted({
-            progress[k].get("workload", "") for k in collectible
-        } - {""})
 
         if not scoped_phases:
             warn("No done phases for scoped pairs.")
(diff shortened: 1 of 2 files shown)


kalantar:
## Verdict: READY TO MERGE

PR correctly fixes the scoped collect bug from issue #242. The broken per-workload loop (which picked a single namespace for all phases) is replaced with per-slot grouping and dispatch — matching the unscoped path's model exactly. The new test directly reproduces the bug scenario (1 workload, 3 packages, 3 slots) and verifies 3 separate dispatches.

### Key findings

**Correctness:** Grouping by `completed_namespace` is correct. Each slot gets its own `_extract_phases_from_pvc` call with `allowed_workloads` restricting extraction to pairs that actually ran there. The parallel ThreadPoolExecutor path mirrors the unscoped branch.

**Thread safety:** The `_on_workload_done` callback mutates shared lists from thread pool workers. This is the same pre-existing pattern as the unscoped path (benign under CPython's GIL — worst case is a harmless duplicate in the phase-level lists).

### Non-blocking suggestions

1. **Missing-namespace pairs no longer tracked as failures** — The old scoped code added pairs with missing `completed_namespace` to `failed_pairs`; the new code only warns. This aligns with the unscoped path's behavior, but means the summary shows "Collected: N/N" without accounting for skipped pairs. A one-line summary like `"Skipped: M pairs (missing namespace)"` after the warnings would improve observability.

2. **Double-counting on partial-success-then-crash** — If `_on_workload_done` reports success for some pairs before the extractor pod crashes, `_handle_slot_failure` adds ALL pairs (including already-collected ones) to `failed_pairs`. The phase-level cleanup removes them from `collected` but `collected_pairs` retains them, producing summary arithmetic that doesn't add up. This is inherited from the unscoped path — not a regression, but an opportunity.

3. **No test for scoped parallel partial failure** — The existing `test_collect_scoped_runtime_error` exercises the sequential (single-slot) exception path. A multi-slot variant where one slot fails and another succeeds would cover the parallel `_handle_slot_failure` path.

4. **Helper functions lack docstrings** — The scoped path's `_on_workload_done` and `_handle_slot_failure` have no docstrings, unlike their unscoped counterparts.

None of these are merge-blockers. The core fix is correct, well-tested (759 tests pass), and structurally aligned with the existing unscoped dispatch model.

PR #245, opened 2026-05-29
fix: deploy.py run — three regressions from PR #241

### LINKED ISSUE #244: deploy.py run: three regressions from PR #241 (missing entry, upfront cost derivation, defaults unavailable in remote)
## Summary

PR #241 (issue #240 — make `gpu_cost` derived per-invocation) introduced three regressions in `deploy.py run`:

1. **Crash on dispatch** — `UnboundLocalError: cannot access local variable 'entry'`
2. **GPU cost derived for all pairs before scoping** — produces 24 spurious warnings when only 3 pairs are in scope
3. **`defaults.yaml` unavailable in remote orchestrator** — causes all pairs to fall back to cost=1

All three are visible in a single `deploy.py run --remote --workload` invocation.

## Reproduction

```
$ python $SIM2REAL/pipeline/deploy.py run --remote --workload codecompletion_mid
```

Orchestrator log shows:
- 24× `[WARN] wl-...: no defaults available — using fallback cost (1)` (issues 2+3)
- `Scope: 3/24 pairs` (correct)
- `UnboundLocalError: cannot access local variable 'entry'` on first successful dispatch (issue 1)

### 1. Missing `entry = progress[pair_key]` in dispatch loop

PR #241 replaced:
```python
entry = progress[pair_key]
pair_cost = entry.get("gpu_cost", fallback_cost)
```
with:
```python
pair_cost = pair_costs[pair_key]
```

but left lines 2177–2179 referencing `entry`:
```python
entry["status"] = "running"
entry["namespace"] = ns
entry["pending_since"] = None
```

Fix: restore `entry = progress[pair_key]` before the `kubectl apply` block.

### 2. Cost derivation runs on all discovered pairs before scoping

`_derive_pair_gpu_costs` at line 1885 operates on `discovered` (all pairs from `cluster/`). `_resolve_scope` runs later at line 1905. Costs are only consumed by `_capacity_limited_pairs` which operates on `pending` — already a subset of scope.

Fix: move `_derive_pair_gpu_costs` call after `_resolve_scope`, passing only in-scope pairs.

### 3. `defaults.yaml` not available in remote orchestrator container

`load_defaults(REPO_ROOT)` looks for `llm-d-benchmark/config/templates/values/defaults.yaml`. The Dockerfile only copies `pipeline/` — the submodule is never included in the image.

Fix: add a `--defaults-path` CLI arg to the `run` subcommand. `deploy.py run --remote` (the local side that builds the Job) reads `defaults.yaml` locally, bundles it into the existing ConfigMap, and injects `--defaults-path <mount-path>` into the Job args. The user doesn't see or set this flag — it's internal plumbing injected by the `--remote` codepath.

## Affected Paths

- `deploy.py run` (local) — issues 1 and 2
- `deploy.py run --remote` (orchestrator Job) — all three

## Priority

High — issue 1 is a crash that blocks all remote dispatches. Issues 2 and 3 are noise/correctness but compound into confusing output.

## Files

- `pipeline/deploy.py` — all three fixes
- `pipeline/lib/remote.py` — bundle defaults.yaml into ConfigMap (issue 3)
- `pipeline/lib/capacity.py` — accept explicit path in `load_defaults` (issue 3)

kalantar:
Closes #244

## Summary

Fixes three regressions introduced by PR #241 (make `gpu_cost` derived per-invocation):

- **Crash on dispatch** — `UnboundLocalError` on `entry` in the dispatch loop. The PR removed `entry = progress[pair_key]` but left three lines referencing `entry`. Restored the assignment.
- **Spurious cost warnings** — `_derive_pair_gpu_costs` ran on all 24 discovered pairs before `_resolve_scope` filtered to the 3 in-scope. Moved cost derivation after scoping so only in-scope pairs are processed.
- **`defaults.yaml` unavailable in remote** — The orchestrator container image only contains `pipeline/`; the `llm-d-benchmark` submodule isn't present. Added plumbing to read `defaults.yaml` locally, bundle it into the ConfigMap, and inject `--defaults-path` into the remote Job args.

## Changes

| File | What |
|------|------|
| `pipeline/deploy.py` | Restore `entry` assignment; reorder cost derivation after scoping; use `--defaults-path` override; read+inject defaults in `_cmd_run_remote` |
| `pipeline/lib/capacity.py` | Add optional `defaults_path` kwarg to `load_defaults` |
| `pipeline/lib/remote.py` | Accept `defaults_content` in ConfigMap builder; handle `defaults.yaml` key in item mapping |
| `pipeline/tests/test_deploy_run.py` | Tests for dispatch entry fix + scoped cost derivation |
| `pipeline/tests/test_capacity.py` | Tests for `load_defaults` explicit path |
| `pipeline/tests/test_remote.py` | Tests for defaults bundling in ConfigMap |

## Reviewer attention

- The `--defaults-path` CLI arg is hidden (`argparse.SUPPRESS`) — it's internal plumbing injected only by the `--remote` codepath. Users never see or set it.
- The `scoped_discovered` filtering in Task 2 means `pair_costs` only has entries for in-scope pairs. This is safe because `_capacity_limited_pairs` and the dispatch loop only operate on `pending` (a subset of scope). Verified by tracing all consumers.

## Test plan

- [x] `python -m pytest pipeline/ -v` — 766 tests pass
- [ ] Manual: `deploy.py run --remote --workload <wl>` on cluster confirms no crash, scoped warnings only, and correct GPU costs

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1882,12 +1882,6 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     if not discovered:
         err("No pairs found in cluster/. Run prepare.py first."); sys.exit(1)
 
-    pair_costs_with_prov = _derive_pair_gpu_costs(
-        discovered, defaults=defaults_result, fallback_cost=fallback_cost,
-    )
-    pair_costs = {k: v[0] for k, v in pair_costs_with_prov.items()}
-    pair_provenance = {k: v[1] for k, v in pair_costs_with_prov.items()}
-
     # Initialize new entries (first run or new pairs added)
     for key, meta in discovered.items():
         if key not in progress:
@@ -1906,6 +1900,13 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     total_pairs = sum(1 for k in progress if _is_pair_key(k))
     if len(_scope) < total_pairs:
         info(f"Scope: {len(_scope)}/{total_pairs} pairs")
+
+    scoped_discovered = {k: v for k, v in discovered.items() if k in _scope}
+    pair_costs_with_prov = _derive_pair_gpu_costs(
+        scoped_discovered, defaults=defaults_result, fallback_cost=fallback_cost,
+    )
+    pair_costs = {k: v[0] for k, v in pair_costs_with_prov.items()}
+    pair_provenance = {k: v[1] for k, v in pair_costs_with_prov.items()}
 
     if getattr(args, "force", False):
         n = _force_reset(progress, _scope, discovered, namespaces=namespaces)
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2174,6 +2174,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 warn(f"[{pair_key}] kubectl apply failed: {result.stderr.strip()}")
                 continue
 
+            entry = progress[pair_key]
             entry["status"] = "running"
             entry["namespace"] = ns
             entry["pending_since"] = None
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1835,7 +1835,11 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     # CLI --gpu-resource-type overrides auto-derivation when explicitly set
     gpu_resource_type = args.gpu_resource_type  # None means auto-derive
     fallback_cost = args.default_gpu_cost
-    defaults_result = load_defaults(REPO_ROOT)
+    defaults_path_override = getattr(args, "defaults_path", None)
+    if defaults_path_override:
+        defaults_result = load_defaults(REPO_ROOT, defaults_path=defaults_path_override)
+    else:
+        defaults_result = load_defaults(REPO_ROOT)
     if isinstance(defaults_result, str):
         warn(defaults_result)
         defaults_result = None
@@ -2540,10 +2544,15 @@ def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
     workspace_dir = EXPERIMENT_ROOT / "workspace"
     run_name = run_dir.name
 
+    # Read defaults.yaml locally — not available in-cluster
+    defaults_path = REPO_ROOT / "llm-d-benchmark" / "config" / "templates" / "values" / "defaults.yaml"
+    defaults_content = defaults_path.read_text() if defaults_path.exists() else None
+
     try:
         cm = build_run_inputs_configmap(
             run_dir=run_dir, workspace_dir=workspace_dir,
             namespace=namespace, run_name=run_name,
+            defaults_content=defaults_content,
         )
     except OSError as exc:
         err(f"{exc} — run setup.py and prepare.py first")
@@ -2560,6 +2569,9 @@ def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
         sys.exit(1)
 
     run_flags = _collect_run_flags(args)
+    if defaults_content:
+        run_flags.append("--defaults-path")
+        run_flags.append("/data/workspace/defaults.yaml")
     job = build_orchestrator_job(
         namespace=namespace, image=orchestrator_image,
         run_name=run_name, run_flags=run_flags,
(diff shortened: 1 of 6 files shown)


kalantar:
## NEEDS CHANGES

All three issue #244 requirements are correctly addressed (crash fix, scoped cost derivation, remote defaults plumbing). The core logic is sound — invariants hold, the `entry` placement is correct, and the mount path is consistent. Two actionable issues should be fixed before merge:

### 1. Unhandled `OSError` on `defaults_path.read_text()` in `_cmd_run_remote`

```python
defaults_content = defaults_path.read_text() if defaults_path.exists() else None
```

If the file exists but is unreadable (permissions, NFS timeout, broken symlink), this crashes with an unhandled traceback. The `except OSError` block 4 lines later only wraps `build_run_inputs_configmap`, not this read. Wrap it:

```python
defaults_content = None
if defaults_path.exists():
    try:
        defaults_content = defaults_path.read_text()
    except OSError as exc:
        warn(f"defaults.yaml read failed: {exc} — remote Job will run without GPU cost defaults")
```

### 2. Inconsistent truthiness guards for `defaults_content`

`build_run_inputs_configmap` checks `if defaults_content is not None:` (packs empty string into ConfigMap), but `_cmd_run_remote` checks `if defaults_content:` (skips flag injection for empty string). If `defaults.yaml` is an empty file, it gets mounted but `--defaults-path` is never passed. Use `is not None` consistently in both places, or filter to `None` early when content is empty.

### Minor (non-blocking)

- **Docstring**: `load_defaults` describes `repo_root` as "experiment repo root" — it's actually the sim2real framework root (always `Path(__file__).parent.parent`). Future maintainers could confuse it with `EXPERIMENT_ROOT`.
- **Silent fallback**: When `--defaults-path` points to a non-existent file (e.g., initContainer copy failed), `load_defaults` returns `None` with no warning. Consider logging that the explicit path was not found.
- **Test gap**: No integration test for `_cmd_run_remote` wiring (reads defaults, bundles into CM, injects flag). Individual layers are tested but the glue is not.

kalantar:
Pushed `f830ad5` addressing review feedback:

**Fixed:**
- Wrapped `defaults_path.read_text()` in try/except OSError — a permissions or NFS failure no longer crashes `_cmd_run_remote`; it warns and continues without defaults.
- Changed `if defaults_content:` to `if defaults_content is not None:` so an empty `defaults.yaml` still triggers `--defaults-path` injection (matching the `is not None` guard in `remote.py`'s ConfigMap packing).
- Added a warning when the explicit `--defaults-path` points to a missing file (e.g., initContainer copy failed silently), so the operator knows GPU cost derivation is falling back.

**Not addressed (non-blocking):**
- Docstring for `load_defaults` param description ("experiment repo root" → framework root) — cosmetic, will fix separately.
- Integration test for the full `_cmd_run_remote` wiring — the individual layers are tested but the glue connecting them is not. Out of scope for this bugfix PR.

kalantar:
## READY TO MERGE

All three issue #244 requirements are correctly implemented, and all prior review feedback has been addressed by the 4th commit (OSError handling, consistent `is not None` guards, warning on missing explicit path).

### Verified

- **Crash fix**: `entry = progress[pair_key]` correctly placed after the `kubectl apply` success guard. Test `test_dispatch_sets_entry_running` would crash with `UnboundLocalError` if the line were reverted.
- **Scoped cost derivation**: `_derive_pair_gpu_costs` now runs only on `scoped_discovered`. All consumers of `pair_costs` access via `pending` (a subset of scope) — no KeyError possible.
- **Remote defaults plumbing**: Read locally → pack into ConfigMap → initContainer copies to `/data/workspace/defaults.yaml` → `--defaults-path` injected into Job args. Mount path is consistent with `MOUNT_BASE + /workspace`.
- **Hardening**: `try/except OSError`, `is not None` guard, and warning on missing explicit path all work as intended. Warning messages are accurate and actionable.

### Non-blocking observations (not required for merge)

1. **`UnicodeDecodeError` uncaught**: `except OSError` doesn't catch `UnicodeDecodeError` (it's a `ValueError` subclass). Affects both `_cmd_run_remote` read and `load_defaults` in `capacity.py`. Extremely unlikely for a YAML file, and the pre-existing `load_defaults` had this same gap before this PR. Could add `(OSError, UnicodeDecodeError)` in a follow-up.
2. **Docstring inaccuracy**: `load_defaults` describes `repo_root` as "experiment repo root" — it's actually the sim2real framework root. Cosmetic.
3. **Test coverage**: No integration test for `_cmd_run_remote` wiring end-to-end. Individual layers are well-tested. Acceptable for a high-priority bug fix.

PR #246, opened 2026-05-29
feat: integrate pod health monitoring into orchestrator

### LINKED ISSUE #228: Integrate pod health monitoring into orchestrator for remote mode
## Problem

`monitor.py` runs as a local sidecar alongside `deploy.py run`, detecting and remediating unhealthy vLLM pods (OOMKilled, Evicted, CrashLoopBackOff). In remote mode (`deploy.py run --remote`), the orchestrator runs as an in-cluster Job but has no pod health monitoring — thermally-killed or OOM'd vLLM pods sit broken until the 4-hour PipelineRun timeout.

The orchestrator's existing `_handle_pending_pods` only checks Tekton task pods (via `tekton.dev/pipelineRun` label) in `Pending` phase. It never sees vLLM model-serving pods.

## Proposed solution

Add a pod health check pass to the orchestrator's main loop (per busy slot, each poll cycle). Target vLLM pods using label selector `llm-d.ai/role in (decode, prefill)`.

### Tier 1 — auto-remediate

- **Evicted** pods → `kubectl delete pod`, log finding and action
- **OOMKilled** (≤ N attempts) → `kubectl delete pod`, log finding and action
- Track remediation attempts per pod; escalate to tier 2 after threshold

### Tier 2 — log with suggestion

- Persistent OOM (> N delete attempts)
- ImagePullBackOff
- Startup probe timeout
- Scheduling failures (Pending without resolution)

Log the issue and a brief suggestion to stdout.

### Tier 3 — log issue only

- CrashLoopBackOff
- Failed / Unknown phase

Log the signal. No Anthropic API call (not appropriate for in-cluster Job).

### Escalation

When tier-1 remediation fails persistently (OOM escalates to tier 2) and `--skip-teardown` is **not** set: cancel the PipelineRun and reclaim the slot (same as existing timeout/pending reclaim behavior).

When `--skip-teardown` **is** set: log only, preserve everything for debugging.

### Output

All findings logged to orchestrator stdout (streamable via `kubectl logs -f job/sim2real-orchestrator`). No `health_report.md` file.

## Non-goals

- Anthropic API diagnosis (tier 3 in current monitor)
- Standalone `monitor.py` changes (can be deprecated or kept for local-only use)
- PipelineRun/TaskRun failure detection (orchestrator already handles this via `_check_pipelinerun_status`)

## Notes

- Namespace resolution logic from `deploy.py` will be shared naturally since the health check lives in the orchestrator (addresses #156)
- Replaces the broad scope of #34 with a focused, actionable implementation

kalantar:
Closes #228

## Summary

Integrates pod health detection and auto-remediation into the orchestrator's main loop. Unhealthy pods (OOM, Evicted, CrashLoop, image pull failures, scheduling failures) are now detected and handled each poll cycle, rather than waiting for the 4-hour PipelineRun timeout.

## What changed

- **`pipeline/lib/health.py`** — Added `get_all_pods(namespace)` which queries all pods in a namespace excluding Tekton task/pipeline pods (filtered by `tekton.dev/*` labels). This gives broad coverage: vLLM model servers, EPP scorer pods, gateway pods — anything that might crash.

- **`pipeline/deploy.py`** — Added `_check_pod_health(namespace, pair_key, tracker, skip_teardown)` which triages each non-healthy pod via the existing `triage_pod` library:
  - **Tier 1** (Evicted, OOMKilled ≤ threshold): auto-deletes the pod, tracks attempts
  - **Tier 2** (persistent OOM, ImagePullBackOff, scheduling failure, startup probe): logs message + suggestion, escalates
  - **Tier 3** (CrashLoopBackOff, Failed/Unknown): logs only (no Anthropic API call)
  
  Wired into the main loop's "Running/Started" slot-processing block, after `_handle_pending_pods` and `_handle_timeout`. On escalation (tier 2 or persistent tier-1 failure), cancels the PipelineRun and reclaims the slot.

- **`--skip-teardown` interaction**: When set, tier-2 findings log but do NOT escalate (preserves everything for debugging).

## Reviewer attention

- **Broad pod selection**: `get_all_pods` excludes only Tekton pods (`tekton.dev/pipelineRun`, `tekton.dev/taskRun`, `tekton.dev/pipelineTask` labels). Every other pod in the namespace is checked. This catches vLLM, EPP, gateway — anything.
- **Escalation policy**: Tier 2 always escalates (unless `--skip-teardown`). This is intentional — if the cluster can't pull the image or schedule the pod, sitting on the slot for 4 hours helps nobody. The user sees the suggestion in logs and can fix the root cause.
- **No API diagnosis**: Tier 3 was `action="api"` in the old monitor.py. Now it's log-only, per issue #228's non-goals.
- **`RemediationTracker` lifetime**: Scoped to the orchestrator run (initialized before the while loop). Resets when a pod returns to Running+Ready.

## Test plan

- [x] `python -m pytest pipeline/ -v` — 773 tests pass
- [x] Unit tests for `get_all_pods` (excludes Tekton, handles errors)
- [x] Unit tests for `_check_pod_health` (tier 1 deletes, tier 2 escalates, skip_teardown prevents escalation, healthy resets tracker)
- [x] Integration test: full orchestrator loop with health escalation → pair marked failed

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -22,11 +22,15 @@ import time
 import yaml
 
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 # Ensure repo root is on sys.path when run as a script (python pipeline/deploy.py)
 _REPO_ROOT = Path(__file__).resolve().parent.parent
 if str(_REPO_ROOT) not in sys.path:
     sys.path.insert(0, str(_REPO_ROOT))
+
+if TYPE_CHECKING:
+    from pipeline.lib.health import RemediationTracker
 
 
 # ── Repo layout ──────────────────────────────────────────────────────────────
@@ -1996,6 +2000,9 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
         info(f"All {len(_scope)} pairs in scope already done — nothing to dispatch (use --force to reset)")
         return
 
+    from pipeline.lib.health import RemediationTracker as _HealthTracker
+    _health_tracker = _HealthTracker()
+
     while _work_remaining() or slots_busy:
 
         # ── Process completed/failed slots ───────────────────────────────
@@ -2095,6 +2102,30 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     store.save(progress)
                 elif timeout_result is False:
                     continue
+
+                # Check pod health (non-Tekton pods)
+                try:
+                    escalate = _check_pod_health(
+                        namespace=ns, pair_key=pair_key,
+                        tracker=_health_tracker,
+                        skip_teardown=getattr(args, "skip_teardown", False),
+                    )
+                except Exception as exc:
+                    warn(f"[{pair_key}] health check failed: {exc}")
+                    escalate = False
+                if escalate:
+                    warn(f"[{pair_key}] pod health escalation → cancelling PipelineRun")
+                    if _cancel_and_delete_pipelinerun(pr_name, ns):
+                        entry["status"] = "failed"
+                        entry["namespace"] = None
+                        store.save(progress)
+                        del slots_busy[ns]
+                        _last_log_state.pop("capacity", None)
+                        _last_log_state.pop("dispatch", None)
+                        _last_log_state.pop("backoff_skip", None)
+                        _zero_dispatch_count = 0
+                    else:
+                        warn(f"[{pair_key}] could not cancel PipelineRun — slot remains busy")
 
         # ── Capacity probe ───────────────────────────────────────────────
         capacity = probe_free_gpus(gpu_resource_type=gpu_resource_type)
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -654,9 +654,9 @@ def _check_pod_health(*, namespace: str, pair_key: str,
                       skip_teardown: bool) -> bool:
     """Check non-Tekton pods in namespace for health issues.
 
-    Returns True if escalation is needed (persistent tier-1 failure or tier-2
-    finding with skip_teardown=False), meaning caller should cancel the
-    PipelineRun and reclaim the slot.
+    Returns True if escalation is needed (tier-1 pod deletion failure, or
+    tier-2 finding with skip_teardown=False), meaning caller should cancel
+    the PipelineRun and reclaim the slot.
     """
     from pipeline.lib.health import (
         get_all_pods, get_events, triage_pod, delete_pod,
@@ -2100,6 +2100,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     _last_log_state.pop("backoff_skip", None)
                     _zero_dispatch_count = 0
                     store.save(progress)
+                    continue
                 elif timeout_result is False:
                     continue
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -645,6 +645,54 @@ def _handle_timeout(*, pr_name: str, namespace: str, entry: dict,
     return True
 
 
+def _check_pod_health(*, namespace: str, pair_key: str,
+                      tracker: "RemediationTracker",
+                      skip_teardown: bool) -> bool:
+    """Check non-Tekton pods in namespace for health issues.
+
+    Returns True if escalation is needed (persistent tier-1 failure or tier-2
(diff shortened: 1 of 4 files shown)


kalantar:
## NEEDS CHANGES

The implementation correctly addresses all nine issue #228 requirements — tier behaviors, escalation policy, skip-teardown interaction, and local/remote parity are all sound. Two issues should be fixed before merge:

### 1. KeyError crash when health check fires after timeout frees the slot

In the main loop, when `timeout_result is True` the code does `del slots_busy[ns]` but does NOT `continue` — execution falls through to the health check block. If `_check_pod_health` returns True (escalation), it attempts `del slots_busy[ns]` again on the already-deleted key → `KeyError` crash.

Fix: add `continue` after the `timeout_result is True` block (matching the pattern used by the `reclaimed` block above it):

```python
if timeout_result is True:
    del slots_busy[ns]
    ...
    store.save(progress)
    continue  # ← add this
elif timeout_result is False:
    continue
```

### 2. `get_all_pods` conflates "kubectl failed" with "no pods exist"

`get_all_pods` returns `[]` on kubectl failure (rc != 0, empty stdout, JSON decode error) — identical to a genuinely empty namespace. The caller treats `[]` as "nothing to check, return False (healthy)." If kubectl is broken (expired creds, network partition, API server overload), health monitoring silently becomes a no-op for the rest of the run, with no error-level logging.

Fix: at minimum, log at `warn` level inside `get_all_pods` when kubectl fails, so operators have visibility into health-check degradation:

```python
if rc != 0 or not stdout.strip():
    warn(f"get_all_pods({namespace}): kubectl failed (rc={rc})")
    return []
```

Optionally: return `None` on probe failure (vs `[]` on genuinely empty) and have `_check_pod_health` log that the health check was skipped.

### Additional observations (non-blocking)

- The broad `except Exception` in the caller uses `warn()` without a traceback — consider matching the adjacent `_handle_pending_pods` pattern (use `err()` + `traceback.print_exc`) so structural failures are diagnosable.
- Docstring says "persistent tier-1 failure" triggers escalation — in reality, escalation from tier-1 is caused by a *failed pod deletion*, not recurrence. Persistent OOM is promoted to tier-2 by `triage_pod` itself. Suggest: "tier-1 pod deletion failure, or tier-2 finding with skip_teardown=False".
- Test gaps: tier-1 delete failure → escalation, exception-in-health-check → caught, and cancel-failure → slot stays busy are all untested error paths worth covering.
- `TriageResult` for tier 3 still carries `action="api"` — pre-existing and non-blocking, but worth a follow-up to align the data model.

kalantar:
Both issues fixed in 372f11b:

**KeyError after timeout frees the slot** — added `continue` after the `timeout_result is True` block so execution no longer falls through to `_check_pod_health`. Matches the pattern used by the `reclaimed` block above it.

**Silent kubectl failures in `get_all_pods`** — now logs at warn level when kubectl returns non-zero or invalid JSON, so operators can see when health monitoring has degraded.

Also tightened the docstring: "persistent tier-1 failure" → "tier-1 pod deletion failure" to match what actually triggers escalation.

The three non-blocking observations (exception handler pattern, test gaps, tier-3 action field) are noted — skipping issues for now as they're tracked here in the thread.

kalantar:
## READY TO MERGE

All nine issue #228 requirements are correctly implemented. Both prior review findings (KeyError after timeout, silent kubectl failure) have been fixed. No blocking issues remain.

**What's solid:**
- Tier 1/2/3 routing is correct — delete, escalate, log-only behaviors all match the spec
- `--skip-teardown` correctly suppresses escalation (preserves debugging state)
- `continue` after timeout prevents the KeyError race
- `get_all_pods` now logs on kubectl failure and JSON decode error
- Docstring accurately describes escalation semantics
- 7 tests cover the primary behavioral contract and integration
- Local/remote parity: same `_cmd_run` code path, no mode branching

**Minor follow-up suggestions (non-blocking):**
- Add `traceback.print_exc(file=sys.stderr)` to the `except Exception` block in the health check caller — matches the adjacent `_handle_pending_pods` pattern and would make remote debugging easier
- Consider adding failure-path tests in a follow-up: tier-1 delete failure → escalation, cancel failure → slot stays busy, exception → caught without crash
- The cancel-failure path produces repeated warnings each poll cycle (correct behavior, just noisy) — could be throttled in a later PR if it proves annoying in practice

PR #248, opened 2026-05-30
feat: randomize pair-to-slot assignment in dispatch

### LINKED ISSUE #247: Randomize pair-to-slot assignment in orchestrator dispatch
## Summary

Today the orchestrator (`deploy.py run`) assigns pending pairs to free slots in deterministic order — dict-insertion order filtered by capacity. This means the same pair always lands in the same slot position, which can introduce systematic bias if slot behavior varies (e.g., different node affinity, timing, or resource contention).

## Proposal

Shuffle the `dispatchable` list **after** the capacity-limited selection but **before** the `zip(free_slots, dispatchable)` assignment. This randomizes which slot a given pair runs in across iterations without affecting:

- Which pairs are selected for dispatch (capacity packing is preserved)
- How many pairs are dispatched per iteration (same greedy selection)

### Sketch

```python
import random

# After capacity selection, before slot assignment:
random.shuffle(dispatchable)

for ns, pair_key in zip(free_slots, dispatchable):
    ...
```

### Location

`pipeline/deploy.py`, around line 2209 (the `zip(free_slots, dispatchable)` loop in `_cmd_run`).

## Motivation

Randomizing the slot assignment reduces systematic bias from slot-correlated factors (node placement, namespace-level resource contention, scheduling order effects) across repeated runs.

kalantar:
Closes #247

## Summary

- Adds `random.shuffle(dispatchable)` before `zip(free_slots, dispatchable)` in the orchestrator dispatch loop (`pipeline/deploy.py:2211`)
- Randomizes which namespace slot a given pair runs in across iterations, reducing systematic bias from slot-correlated factors (node placement, resource contention, scheduling order)
- Preserves capacity-limited selection (which pairs, how many) — only the slot assignment is randomized

## Test plan

- [x] New test `test_dispatch_shuffles_dispatchable` verifies `random.shuffle` is called on the dispatchable list
- [x] Full test suite passes (774 tests)
- [x] Lint passes (`ruff check pipeline/ --select F`)

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -15,6 +15,7 @@ Subcommands:
 import argparse
 import json
 import os
+import random
 import shutil
 import subprocess
 import sys
@@ -2206,6 +2207,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                         _last_log_state["dispatch"] = _disp_state
         else:
             dispatchable = pending
+
+        random.shuffle(dispatchable)
 
         for ns, pair_key in zip(free_slots, dispatchable):
             hf_secret_name = setup_config.get("hf_secret_name", "hf-secret")
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -2847,3 +2847,27 @@ def test_health_escalation_cancels_pipelinerun(tmp_path, monkeypatch):
 
     assert "wl-a-baseline" in saved
     assert saved["wl-a-baseline"]["status"] == "failed"
+
+
+def test_dispatch_shuffles_dispatchable(monkeypatch):
+    """Verify that dispatchable list is shuffled before slot assignment (#247)."""
+    from pipeline.deploy import _capacity_limited_pairs
+
+    cost_map = {"a-baseline": 4, "b-baseline": 4, "c-baseline": 4}
+    pending = ["a-baseline", "b-baseline", "c-baseline"]
+
+    dispatchable = _capacity_limited_pairs(pending, free_gpus=24, cost_map=cost_map)
+    assert len(dispatchable) == 3
+
+    shuffle_calls = []
+
+    def tracking_shuffle(lst):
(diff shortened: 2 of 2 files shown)


kalantar:
## NEEDS CHANGES

The production code is correct — `random.shuffle(dispatchable)` is placed at exactly the right location (after capacity selection, before slot assignment), satisfies all issue #247 requirements, and has no error-handling or correctness concerns.

**One issue to fix before merge:**

The test `test_dispatch_shuffles_dispatchable` does not exercise the production code path. It monkeypatches `random.shuffle`, then calls `mod.random.shuffle(dispatchable)` directly from the test body — it never invokes the dispatch loop. If the `random.shuffle(dispatchable)` line were removed from `deploy.py`, this test would still pass.

Fix: restructure the test to invoke the actual dispatch path (with sufficient mocking to reach the shuffle call) and assert `shuffle_calls` is populated as a side effect of that execution. Alternatively, run a minimal `_cmd_run` invocation (similar to the existing `test_dispatch_sets_entry_running` pattern) and verify the tracking shuffle was called with the expected dispatchable list.

Everything else is clean — no silent failures, no comment inaccuracies, production logic is sound.

kalantar:
Addressed — the test was rewritten to exercise the actual `_cmd_run` dispatch path (same pattern as `test_dispatch_sets_entry_running`). It patches `random.shuffle` with a tracking wrapper, invokes `_cmd_run` with mocked dependencies, and asserts the shuffle was called on the dispatchable list as a side effect of the dispatch loop.

Verified it catches a regression: removing the `random.shuffle(dispatchable)` line causes the test to fail. All 774 tests pass.

kalantar:
## READY TO MERGE

Prior review's sole concern (test not exercising the production path) is now fixed. The restructured test invokes `_cmd_run` with the full mock setup and asserts `shuffle_calls` is populated as a side effect of the actual dispatch loop — removing the `random.shuffle` line from deploy.py would fail this test.

Production code was already clean: correct placement, no error-handling concerns, all issue #247 requirements met.

PR #252, opened 2026-05-30
fix: use ConfigMap progress for pre-flight filter validation in --remote

### LINKED ISSUE #251: deploy.py run --remote: pre-flight filter validation uses wrong data source, rejects --status
## Bug

`deploy.py run --remote --status failed` errors with "No pairs matched --status 'failed'" even when `deploy.py status --status failed` correctly shows matching pairs.

### Root cause

`_cmd_run_remote` (line 2620-2625) validates filter flags before building images using a **synthetic progress dict** built from `_load_pairs()` (disk YAML metadata). This dict contains only `workload`, `package`, and `pr_path` — it has no `status` field. When `_resolve_scope` applies the `--status failed` filter, no entries match because the status field is absent.

```python
# line 2620-2625 in _cmd_run_remote
# Validate filter flags before building images (fail fast)
cluster_dir = run_dir / "cluster"
discovered = _load_pairs(cluster_dir)
if discovered:
    progress = {k: v for k, v in discovered.items()}  # ← no status field
    _resolve_scope(progress, args)                     # ← --status always fails
```

All other subcommands correctly read from the ConfigMap via `ConfigMapProgressStore.load()`:
- `status` — line 395 ✓
- `collect` — line 1028 ✓
- `run` (local) — line 1929 ✓
- `reset` — line 2306 ✓
- `wipe` — line 2361 ✓

### Reproduction

```bash
# Shows 2 failed pairs:
python pipeline/deploy.py status --status failed

# Rejects the same filter:
python pipeline/deploy.py run --remote --status failed
# [ERROR] No pairs matched --status 'failed'.
```

### Fix

The pre-flight validation in `_cmd_run_remote` should read progress from the ConfigMap (same as every other subcommand), not from disk YAML metadata. Replace lines 2622-2625 with a ConfigMapProgressStore load:

```python
cluster_dir = run_dir / "cluster"
discovered = _load_pairs(cluster_dir)
if discovered:
    store = ConfigMapProgressStore(namespace, run_name=run_dir.name)
    progress = store.load()
    if progress:
        _resolve_scope(progress, args)
```

If the ConfigMap is unreachable (cluster down), skip the pre-flight validation rather than using the wrong data source — the in-cluster orchestrator will validate filters when it starts.

### Note on --force interaction

Separately from this bug, using `--force` with `--status` on the local `_cmd_run` path works correctly because `_resolve_scope` runs before `_force_reset` (line 1958 vs 1970). The `--remote` path doesn't reach that logic at all due to this bug.

kalantar:
Closes #251

## Summary

`deploy.py run --remote --status failed` was incorrectly rejected with "No pairs matched" because the pre-flight validation built a synthetic progress dict from `_load_pairs()` (disk YAML metadata with no `status` field). Now reads from `ConfigMapProgressStore.load()` — the same data source used by every other subcommand.

## What changed

- **`pipeline/deploy.py`** — Replaced the synthetic `progress = {k: v for k, v in discovered.items()}` with a `ConfigMapProgressStore.load()` call. Wrapped in try/except so unreachable ConfigMaps skip validation (the in-cluster orchestrator validates when it starts).
- **`pipeline/tests/test_deploy_remote.py`** — Three new tests:
  - `test_run_remote_status_filter_uses_configmap` — `--status failed` passes when ConfigMap has failed pairs
  - `test_run_remote_status_filter_rejects_unmatched` — still errors when no pairs match
  - `test_run_remote_skips_validation_when_configmap_unreachable` — graceful degradation

## Reviewer attention

- The `discovered` check (`if discovered:`) is preserved — if no cluster YAMLs exist, we still skip validation entirely (nothing to validate against).
- The try/except catches any Exception from ConfigMapProgressStore (kubectl failure, network partition, RBAC) and skips validation. This matches the issue's recommendation: "skip pre-flight rather than using wrong data."

## Test plan

- [x] `python -m pytest pipeline/ -v` — 779 tests pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2631,8 +2631,14 @@ def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
     cluster_dir = run_dir / "cluster"
     discovered = _load_pairs(cluster_dir)
     if discovered:
-        progress = {k: v for k, v in discovered.items()}
-        _resolve_scope(progress, args)
+        from pipeline.lib.progress import ConfigMapProgressStore
+        try:
+            store = ConfigMapProgressStore(namespace, run_name=run_dir.name)
+            progress = store.load()
+        except Exception:
+            progress = None
+        if progress:
+            _resolve_scope(progress, args)
 
     _cmd_build(run_dir, namespace=namespace, skip_build=args.skip_build)
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2632,10 +2632,11 @@ def _cmd_run_remote(args, run_dir: "Path", setup_config: dict) -> None:
     discovered = _load_pairs(cluster_dir)
     if discovered:
         from pipeline.lib.progress import ConfigMapProgressStore
+        store = ConfigMapProgressStore(namespace, run_name=run_dir.name)
         try:
-            store = ConfigMapProgressStore(namespace, run_name=run_dir.name)
             progress = store.load()
-        except Exception:
+        except (RuntimeError, OSError) as exc:
+            warn(f"ConfigMap unreachable — skipping pre-flight filter validation: {exc}")
             progress = None
         if progress:
             _resolve_scope(progress, args)
--- a/pipeline/tests/test_deploy_remote.py
+++ b/pipeline/tests/test_deploy_remote.py
@@ -532,3 +532,74 @@ def test_main_routes_run_local(tmp_path, monkeypatch):
             mod.main()
 
     assert len(local_calls) == 1
+
+
+# ── Pre-flight filter validation (#251) ─────────────────────────────────────
+
+
+def test_run_remote_status_filter_uses_configmap(monkeypatch, tmp_path):
+    """--status filter in --remote mode reads from ConfigMap, not disk YAML (#251)."""
+    from pipeline.lib.progress import ConfigMapProgressStore
+
+    run_dir = _setup_run_dir(tmp_path)
+    monkeypatch.setattr(mod, "EXPERIMENT_ROOT", tmp_path)
+    monkeypatch.setattr(mod, "_check_existing_job", lambda ns: None)
+    monkeypatch.setattr(mod, "_cmd_build", lambda *a, **kw: "skip")
+    monkeypatch.setattr(mod, "_wait_for_job_pod", lambda *a, **kw: None)
+
+    progress_with_status = {
+        "wl-a-baseline": {"workload": "wl-a", "package": "baseline", "status": "failed",
+                          "namespace": None, "retries": 0, "pending_stalls": 0, "pending_since": None},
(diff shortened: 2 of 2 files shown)


kalantar:
## NEEDS CHANGES

The fix is functionally correct — it reads from ConfigMapProgressStore instead of the synthetic disk-YAML dict, which resolves the bug. Tests cover the three main branches (match, no-match, unreachable). One issue to fix before merge:

### Narrow `except Exception` and add a warning

The bare `except Exception` with no logging violates the project's established pattern. Six lines below this code (line 2648), the same function demonstrates the correct approach:

```python
except OSError as exc:
    warn(f"defaults.yaml read failed: {exc} — remote Job will run without GPU cost defaults")
```

The current catch silently masks programming bugs (`ValueError` from a corrupt ConfigMap, `TypeError` from bad inputs, `ImportError` if the module breaks). Move the import outside the try block and narrow the catch:

```python
if discovered:
    from pipeline.lib.progress import ConfigMapProgressStore
    store = ConfigMapProgressStore(namespace, run_name=run_dir.name)
    try:
        progress = store.load()
    except (RuntimeError, OSError) as exc:
        warn(f"ConfigMap unreachable — skipping pre-flight filter validation: {exc}")
        progress = None
    if progress:
        _resolve_scope(progress, args)
```

This lets real bugs propagate (corrupt data, logic errors) while still gracefully skipping on connectivity failures. The `store.load()` raises `RuntimeError` for kubectl failures and `OSError` when kubectl is missing — those are the legitimate "unreachable" cases.

### Non-blocking suggestion

Consider testing the empty-dict path (`store.load()` returns `{}`). This happens for fresh runs where the ConfigMap exists but the orchestrator hasn't populated it. Currently treated as "skip validation" (since `{}` is falsy) — correct behavior, but worth documenting with a test.

kalantar:
Pushed `3e2e426` addressing the review:

Narrowed the exception catch from bare `except Exception` to `except (RuntimeError, OSError)` — these are the two "unreachable" cases from `store.load()` (kubectl failure, kubectl missing). `ValueError` (corrupt ConfigMap data) and other programming errors now propagate. Also moved the import outside the try block and added a `warn()` with the exception message so operators can see why validation was skipped.

The non-blocking suggestion (test empty-dict path) is already handled correctly — `{}` is falsy, so `if progress:` skips validation. Didn't add a test since the behavior is trivially correct from the existing code path.

kalantar:
## READY TO MERGE

Previous review finding addressed correctly:
- Import moved outside try block (module-level errors propagate)
- Exception narrowed to `(RuntimeError, OSError)` — catches kubectl failures and missing binary
- `ValueError` (corrupt ConfigMap data) now propagates as it should
- Warning logged so the user knows validation was skipped and why
- Follows the project's established pattern (matches `defaults.yaml` error handling 6 lines below)

Fix is correct, tests cover the three main branches, all acceptance criteria from #251 are met.

PR #257, opened 2026-05-30
feat: shadow GPU reservation ledger to prevent over-subscription

### LINKED ISSUE #255: deploy.py run: shadow GPU reservation ledger to prevent over-subscription
## Problem

The orchestrator's GPU capacity probe (`probe_free_gpus`) has ~120s lag between dispatch and actual GPU consumption. During this window, the probe reports stale free-GPU counts, causing over-subscription: pairs are dispatched to slots that don't actually have capacity, leading to scheduling failures and backoff cycles.

The existing `--dispatch-cooldown` (PR #250) gates per-batch, which either serializes all dispatches unnecessarily (with high TTL) or doesn't prevent over-subscription (with low TTL). A per-dispatch cooldown would work but is too conservative — with 24 pairs at 120s each, dispatch alone takes 48 minutes.

## Proposed solution: shadow reservation ledger

Maintain a list of `(gpu_cost, dispatch_time)` tuples representing recently-dispatched pairs whose GPU consumption hasn't yet been reflected in the probe. Use this to compute an *effective* free GPU count:

```python
# Shadow ledger: list of (gpu_cost, dispatch_time)
_shadow_reservations: list[tuple[int, float]] = []

# Before dispatch:
now = time.time()
_shadow_reservations = [(c, t) for c, t in _shadow_reservations if now - t < ttl]
shadow_reserved = sum(c for c, _ in _shadow_reservations)
effective_free = probed_free - shadow_reserved

# Dispatch gate (per pair):
if effective_free >= pair_cost:
    # dispatch
    _shadow_reservations.append((pair_cost, now))
    effective_free -= pair_cost
```

### Behavior

- With 65 probed free GPUs and pairs costing 4 each: dispatches up to 16 pairs rapidly (65/4=16), then stops.
- With 8 probed free GPUs: dispatches 2 pairs (8/4=2), then stops — without needing to wait.
- After TTL (120s), entries age out naturally. By then the probe reflects actual consumption.
- If a pair completes quickly (slot frees + probe updates), new dispatches resume immediately — no artificial delay.

### Advantages over cooldown

- Dispatches as fast as capacity allows (no serialization)
- Self-limits when resources are tight (no over-subscription)
- Adapts to heterogeneous GPU costs (a 1-GPU pair dispatches even when a 4-GPU pair can't)
- No need to guess the "right" cooldown duration

## Acceptance criteria

1. Replace `--dispatch-cooldown` with `--shadow-ttl` (default 120s) — the duration to retain shadow reservations
2. Each successful dispatch appends `(pair_cost, time.time())` to the ledger
3. Dispatch gate: `effective_free = probed_free - sum(shadow_reserved)` must be >= pair cost
4. Entries older than TTL are pruned each iteration
5. `--shadow-ttl 0` disables shadow tracking (current behavior: dispatch purely on probe reading)
6. Forwarded to remote orchestrator via `_collect_run_flags`

## Location

`pipeline/deploy.py`, the "Assign pending work to free slots" section in `_cmd_run`. The shadow ledger replaces the current cooldown gate (lines 2179-2182).

## Interaction with existing features

- **Capacity-limited selection**: `_capacity_limited_pairs` should use `effective_free` instead of `probed_free` as its budget
- **Backoff controller**: Orthogonal — backoff gates on scarcity (min cost > free), shadow gates on committed-but-unreflected consumption. Both can be active.
- **`--default-gpu-cost`**: Used when pair cost can't be derived — shadow entries use the same `pair_costs[key]` map

## Observation

From production logs: ~120s from `kubectl apply` to GPU consumption showing in probe. This includes PipelineRun admission, task scheduling, pod creation, image pull, and vLLM startup.

kalantar:
## Summary

Closes #255

- Replaces the blunt `--dispatch-cooldown` timer with a **shadow reservation ledger** (`pipeline/lib/shadow.py`) that tracks committed-but-not-yet-probed GPU consumption
- Each dispatch records `(gpu_cost, timestamp)`. Before capacity-limiting, the orchestrator computes `effective_free = probed_free - shadow_reserved`, preventing over-subscription during the ~120s probe lag window
- New CLI flag: `--shadow-ttl` (default 120s, 0 to disable). Forwarded to remote orchestrator via `_collect_run_flags`
- Old `--dispatch-cooldown` flag removed entirely

## Design notes

The `ShadowLedger` class is a self-contained module with no dependencies beyond `time`. It prunes expired entries lazily on each `reserved()` call. With `ttl=0`, `record()` is a no-op so there's zero overhead when disabled.

The probe's blind spot is structural: `probe_free_gpus` skips pods without `spec.nodeName` (line 72 of `capacity.py`), so any pod still Pending is invisible. The shadow ledger fills this gap without changing the probe itself.

## Test plan

- [x] Unit tests for `ShadowLedger` (6 tests in `test_shadow.py`)
- [x] Integration test: shadow ledger gates dispatch to effective capacity (`test_shadow_ledger_prevents_over_subscription`)
- [x] Integration test: `--shadow-ttl 0` disables gating (`test_shadow_ttl_zero_disables_gating`)
- [x] Remote flag forwarding tests (non-default forwarded, default omitted)
- [x] Full suite passes (788 tests)
- [x] No remaining references to `dispatch_cooldown`/`dispatch-cooldown`

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2015,8 +2015,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     from pipeline.lib.health import RemediationTracker as _HealthTracker
     _health_tracker = _HealthTracker()
 
-    dispatch_cooldown = args.dispatch_cooldown
-    _last_dispatch_time: float = 0.0
+    from pipeline.lib.shadow import ShadowLedger
+    shadow = ShadowLedger(ttl=args.shadow_ttl)
 
     while _work_remaining() or slots_busy:
 
@@ -2187,12 +2187,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 backoff.signal_capacity(free_gpus=free_gpus, max_cost=max_cost)
 
         # ── Assign pending work to free slots ────────────────────────────
-        _cooldown_elapsed = time.time() - _last_dispatch_time if _last_dispatch_time > 0 else float('inf')
-        _in_cooldown = dispatch_cooldown > 0 and _cooldown_elapsed < dispatch_cooldown
-        if _in_cooldown:
-            free_slots = []
-        else:
-            free_slots = [ns for ns in namespaces if ns not in slots_busy]
+        free_slots = [ns for ns in namespaces if ns not in slots_busy]
         pending = _pending_pairs()
         if free_gpus is not None and pending:
             min_cost = min(pair_costs[k] for k in pending)
@@ -2203,9 +2198,10 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     _last_log_state["backoff_skip"] = _skip_state
                 dispatchable = []
             else:
+                effective_free = shadow.effective_free(free_gpus)
                 dispatchable = _capacity_limited_pairs(
                     pending,
-                    free_gpus=free_gpus, cost_map=pair_costs,
+                    free_gpus=effective_free, cost_map=pair_costs,
                 )
                 if len(dispatchable) == 0 and pending:
                     smallest = min(pair_costs[k] for k in pending)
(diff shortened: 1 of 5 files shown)


kalantar:
## READY TO MERGE

Clean implementation that exactly matches issue #255's design. All six acceptance criteria are met:

1. `--dispatch-cooldown` replaced with `--shadow-ttl` (default 120s)
2. Each dispatch records `(pair_cost, time.time())` via `shadow.record(pair_cost)`
3. Capacity selection uses `shadow.effective_free(probed_free)` as budget
4. Entries pruned lazily on each `reserved()` call
5. `ttl=0` disables tracking (record is a no-op, reserved returns 0)
6. Forwarded via `_collect_run_flags` (non-default only)

The `ShadowLedger` class is well-designed: self-contained, no external dependencies, lazy pruning, and the `effective_free` method clamps to 0 (no negative budgets). The integration is minimal — two lines in the dispatch path (one to compute effective capacity, one to record after dispatch).

Test coverage is solid: 6 unit tests for the ledger (including TTL expiry with mocked time, zero-TTL disable, negative-free clamping) plus 2 integration tests exercising the full orchestrator loop (capacity gating + TTL-zero bypass). Remote forwarding tested in both directions (non-default forwarded, default omitted).

No issues found.

Issue #263, opened 2026-06-02
Locate workload tolerations in resolved scenario for capacity probe filter


kalantar:
## Problem

The capacity probe filter in #261 needs workload tolerations to evaluate taint-based node eligibility. Tolerations are configured per workload (per role), but the exact key path inside the resolved scenario is currently unconfirmed.

`nodeAffinity` is known to live at `model.helmValues.{decode,prefill}.extraConfig.affinity` (see the suggestion text in `pipeline/lib/health.py:192`). Tolerations are presumed to live in a parallel location under `extraConfig`, but this needs verification against an actual resolved scenario and the llm-d-benchmark Helm chart it feeds.

## Initial assumption (to unblock #261)

For the initial probe-filter implementation, treat workload tolerations as **empty**:
- Cordoned-node filter applies.
- Untolerated-taint filter treats every `NoSchedule`/`NoExecute` taint on a node as untolerated, eliminating that node.

This is conservative — it under-counts available GPUs when a workload genuinely tolerates a taint — but it is correct in the dispatch-safety direction (won't over-count).

## Resolution

- Identify the exact resolved-scenario key path(s) for tolerations (per role: decode, prefill).
- Wire that path through to `probe_free_gpus` so the filter compares each node's taints against the actual tolerations.
- Add tests covering: pod tolerates the taint (node eligible), pod doesn't (node excluded), per-role asymmetry (decode tolerates, prefill doesn't, or vice versa).
- Verify against a real resolved scenario (e.g., admission-control baseline) that the key path matches what the rendered PipelineRun YAML actually carries.

## Related

- #261 — primary capacity-probe over-counting fix (the conservative no-tolerations assumption is acceptable for that fix; this issue lifts the assumption)

Issue #262, opened 2026-06-02
GPU capacity probe: per-node fragmentation for multi-GPU pods


kalantar:
## Problem

`pipeline/lib/capacity.py` `probe_free_gpus` returns a global sum: `eligible_alloc - eligible_req` across the whole cluster. Today every workload pod requests 1 GPU, so any free slot on any eligible node satisfies the pod and the global sum is a sufficient proxy for dispatchable capacity.

A scenario that schedules multi-GPU pods (tensor parallelism > 1, or `accelerator.count > 1` per pod) breaks this assumption. The probe might report N free GPUs while no single node has the required contiguous count free — fragmentation. The orchestrator would dispatch and the pod would sit Pending.

## Out of scope here

This issue is the follow-up to the per-node refinement called out at the bottom of #261. The minimum useful change in #261 (cordoned + tainted node filter) does not cover this.

## Proposed direction

When pods request more than one GPU, the probe should bin-pack per node:

- For each eligible node (after the #261 filters): `node_free = node_allocatable - sum(requests of pods on this node)`
- The orchestrator gates a dispatch on whether **some** node has `node_free >= pod.gpus_per_pod`, per role.
- Aggregate "free GPUs" stays useful for logging but should be computed as the sum of per-node free, capped by what's actually placeable.

## Trigger

Re-open / prioritize this when any scenario in active use sets `parallelism.tensor > 1`, `parallelism.dataLocal > 1`, or `accelerator.count > 1` for any role.

## Related

- #261 — primary capacity-probe over-counting fix (this is item 4 from that issue, split off)

PR #264, opened 2026-06-02
feat: capacity probe filters cordoned, tainted, and wrong-product nodes

### LINKED ISSUE #261: GPU capacity probe over-counts: includes cordoned, tainted, and wrong-model GPUs
## Problem

`pipeline/lib/capacity.py` `probe_free_gpus` computes `free = total_allocatable − total_requested` by summing `nvidia.com/gpu` across **every** node in the cluster. This over-reports the GPUs actually available to the workload pods the orchestrator dispatches, leading to dispatch decisions where pods then sit Pending while the orchestrator's `Capacity:` line claims tens of free GPUs.

Observed today on `kalantar-0` orchestrator: log reported `37 free GPUs (83 allocatable − 46 requested)`, yet a decode replica pod stayed Pending with:

```
0/20 nodes are available: 3 node(s) had untolerated taint {app: llm-d-benchmark-harness},
3 node(s) had untolerated taint {node-role.kubernetes.io/master: },
5 node(s) were unschedulable, 9 Insufficient nvidia.com/gpu.
```

Of 20 cluster nodes, only 9 were even candidates after taints/cordons were applied — and on those 9, no H100 was free.

## Root causes

The probe ignores filters the scheduler applies:

1. **Cordoned nodes counted.** `spec.unschedulable: true` doesn't zero `.status.allocatable.nvidia.com/gpu`, so cordoned nodes contribute. Today there are 5 such nodes (`pokprod-b93r38s3`, `b93r39s0`, `b93r39s2`, `b93r43s0`, `b93r44s3`).
2. **Tainted nodes counted regardless of tolerations.** The 3 `pokprod001work{0,1,2}` nodes carry `app=llm-d-benchmark-harness:NoSchedule`. Workload pods don't tolerate it, but their GPUs are still summed.
3. **GPU model not filtered.** Pods set `nodeAffinity` requiring `nvidia.com/gpu.product=NVIDIA-H100-80GB-HBM3`. Any non-H100 GPU in the cluster is counted but unreachable.

## Proposed direction

Filter nodes the same way the scheduler would, before summing:

```
for node in nodes:
    if node.spec.unschedulable: skip
    if any taint with NoSchedule/NoExecute is not tolerated by the workload pod: skip
    if node's gpu.product label doesn't match the scenario's required product: skip
    eligible_alloc += allocatable on this node
    eligible_req   += GPU requests of pods scheduled on this node
free = eligible_alloc - eligible_req
```

Required GPU product is already known to the orchestrator via the resolved scenario (nodeAffinity at `model.helmValues.{decode,prefill}.extraConfig.affinity` per `pipeline/lib/health.py:192`). Workload tolerations are also presumed to live in the resolved scenario, but the exact key path is not yet confirmed — see follow-up issue #263.

The filter is **per-role**: decode and prefill can require different node products and tolerate different taints. The aggregate "free GPUs" reported by the probe is the union of role-eligible nodes; the dispatch decision should be made per role.

## Initial implementation: assume no tolerations

To unblock this fix without resolving #263 first, treat workload tolerations as **empty** in the initial implementation. Every `NoSchedule`/`NoExecute` taint on a node eliminates that node. This under-counts when a workload legitimately tolerates a taint (conservative — never over-dispatches), and matches today's observed behavior since the workload pods in the incident did not tolerate `app=llm-d-benchmark-harness`. Once #263 lands, the filter can read tolerations from the scenario and lift the assumption.

## Minimum useful change

Excluding cordoned nodes and nodes with any `NoSchedule`/`NoExecute` taint (under the no-tolerations assumption above) would have prevented today's over-report. GPU-product filtering is a follow-up refinement within this issue.

## Out of scope (split out)

- **Per-node fragmentation for multi-GPU pods** — issue #262. Today every pod requests 1 GPU, so the global sum is a sufficient proxy. Multi-GPU pods need per-node bin-packing.
- **Locating tolerations in the resolved scenario** — issue #263. The conservative no-tolerations assumption above lets this issue proceed independently.

## Cross-path parity

`probe_free_gpus` has a single caller (`pipeline/deploy.py:2144`). No parallel run/replay/observe paths in the sim2real pipeline.

## Scope

- `pipeline/lib/capacity.py` (`probe_free_gpus`) — accept per-role filter input (required GPU product label, tolerations list — initially `[]`); apply unschedulable/taint/product filter before summing.
- `pipeline/deploy.py` `_cmd_run` (around line 2144) — pass the per-role filter derived from the resolved scenario.
- `pipeline/tests/test_capacity.py` — cases for cordoned, untolerated-taint, wrong-product, per-role asymmetry.

kalantar:
Closes #261

## Summary

Stops `pipeline/lib/capacity.py` `probe_free_gpus` from over-counting GPUs the K8s scheduler can't actually use. Before this change, the probe summed `nvidia.com/gpu` allocatable across **every** node and subtracted requests; the orchestrator then dispatched workloads onto eligibility-rejected nodes and watched them sit Pending while the `Capacity:` log line claimed tens of free GPUs.

## What changed

The probe now optionally accepts a list of `NodeFilter` objects (one per role). When supplied, a node is excluded from both allocatable and requested sums if it:

- is **cordoned** (`spec.unschedulable: true`),
- carries a **`NoSchedule` or `NoExecute` taint** that no role's tolerations match (initial implementation: tolerations are always `()` per follow-up #263, so any blocking taint excludes the node), or
- has a **`nvidia.com/gpu.product` label** not in the set required by the scenario's `nodeAffinity` (extracted per role from `model.helmValues.{decode,prefill}.extraConfig.affinity` matchExpressions with operator `In`).

Eligibility is **union across roles**: a node counts if some role's filter accepts it. With no filters supplied, behavior is byte-identical to the prior implementation (legacy callers and tests unaffected).

`pipeline/deploy.py` `_cmd_run` builds the filter dict once at startup from the same `cluster/baseline.yaml` it already loads for GPU-resource-type derivation, logs a one-line eligibility summary per role, and passes the filters on every probe call.

## Architecture

- **`pipeline/lib/capacity.py`**
  - `NodeFilter` frozen dataclass: `required_gpu_products: frozenset[str]`, `tolerations: tuple[dict, ...]`
  - `node_is_eligible(node, filters)` — union semantics; empty filter list = unfiltered
  - `_toleration_matches_taint`, `_node_is_cordoned`, `_node_blocking_taints`, `_filter_admits_node` — pure helpers
  - `probe_free_gpus(..., *, node_filters=None)` — keyword-only kwarg; pods on excluded nodes are dropped from `total_requested` so the (free, alloc, requested) tuple stays consistent
  - `extract_node_filters(resolved_scenario)` — builds per-role dict from scenario nodeAffinity
- **`pipeline/deploy.py`** — extracts filters at startup, forwards to probe
- **`pipeline/README.md`** — short paragraph documenting the filter behavior

## Reviewer attention

- **Tolerations are deliberately empty for now** (see #263). This is conservative: it under-counts when a workload would legitimately tolerate a taint, but never over-dispatches. The incident in #261 had pods that did not tolerate `app=llm-d-benchmark-harness:NoSchedule`, so the conservative behavior matches reality.
- **Per-role union, not per-role gating.** Aggregate `free_gpus` reflects the union of role-eligible nodes; the dispatch logic in `_cmd_run` continues to gate on the aggregate. Per-role gating remains future work — pair_cost is a sum across roles and most current scenarios share the same affinity across decode and prefill.
- **No new behavior with empty filters.** `extract_node_filters({})` returns `{}` and `list({}.values()) or None` → `None`, preserving legacy behavior. All 140 `test_deploy_run.py` tests pass without modification.
- **Per-node fragmentation is not addressed here.** Today every workload pod requests 1 GPU, so a global sum is sufficient. Multi-GPU pods will need per-node bin-packing — tracked in #262.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 902 passed
  - 32 new tests across `TestNodeFilter` (17), `TestProbeFreeGpus` filter cases (6), `TestExtractNodeFilters` (9)
- [x] Backward compat: `probe_free_gpus()` with no kwargs still sums every node (`test_no_filter_preserves_legacy_behavior`)
- [x] Pending pods (no `spec.nodeName`) still skipped under filter (`test_pod_with_no_nodename_still_skipped_under_filter`)
- [x] Pod requests on excluded nodes dropped from `total_requested` (`test_pod_request_on_filtered_node_excluded`)
- [x] Per-role union eligibility tested both ways (some role accepts, no role accepts)
- [x] PreferNoSchedule taint correctly ignored

## Follow-ups

- #262 — per-node fragmentation for multi-GPU pods
- #263 — locate workload tolerations in resolved scenario, lift the empty-tolerations assumption


kalantar:
## Verdict: NEEDS CHANGES

The core design is sound and the in-scope #261 root causes are addressed and tested (cordoned, tainted-no-toleration, GPU product). Backward-compat is preserved byte-identically, the (free, alloc, requested) tuple stays internally consistent under filtering, and the deferred-tolerations posture is provably conservative.

The blocker is that the error-handling envelope around the input does the opposite of what the filter does. The filter rejects when in doubt; the input parsing accepts when in doubt. Three places fail silently in the same direction — over-count → over-dispatch → Pending pods, which is exactly the bug this PR exists to fix.

### Please fix before merge

1. **`pipeline/deploy.py:1932-1939` swallows `yaml.YAMLError` with no warning.** A malformed `baseline.yaml` makes `extract_node_filters` return `{}`, which collapses to `node_filters=None`, which reverts the probe to legacy unfiltered behavior. The same file is already parsed 14 lines above with a `warn()` on failure — reuse that `resolved` instead of re-reading, or at minimum warn on the second parse too.

2. **`extract_node_filters` silently degrades to permissive on malformed affinity.** Any structural slip — `nodeSelectorTerms` wrong type, `operator: "in"` instead of `"In"`, key typo — produces `frozenset()` for the product set, and `_filter_admits_node` skips the product gate. The startup log prints `"Eligibility filter [decode]: no product constraint"` whether the user wrote no affinity or a broken one. When `helmValues.{role}.extraConfig.affinity` is present but no `In` expression on `nvidia.com/gpu.product` is extracted, warn the user — that's the only way to distinguish intent from typo.

3. **Unknown taint effects in `_node_blocking_taints` are silently dropped.** A taint with `effect: "Noschedule"` (case typo) or any future K8s effect is not blocking — node passes through the filter. Filter to the known pair, but warn once per probe cycle for any unrecognized non-empty effect so cluster-side typos surface.

### Worth fixing while you're in there

- **Mixed-eligibility pod accounting has no test.** `test_pod_request_on_filtered_node_excluded` asserts `(0,0,0)` because both nodes are excluded. The more diagnostic case — one included, one excluded, pods on each — would catch a regression that double-counts excluded-node pods.
- **`node_filters=None` vs `[]` vs `[NodeFilter()]` are three near-synonyms with subtly different meanings.** None and [] both skip filtering; a default-constructed NodeFilter still applies cordon/taint screening. Worth a sentence in the `probe_free_gpus` docstring; the current "When node_filters is provided" leaves `[]` ambiguous.
- **`_toleration_matches_taint` treats `effect: ""` as wildcard match.** Correct K8s semantics, but currently dead code (extract_node_filters always passes `tolerations=()`). When #263 lifts that, a mis-specified toleration silently under-counts blockers. Worth a `__post_init__` check on toleration dict shape, or a deferred TODO.

### Doc nits (not blocking)

- `NodeFilter.tolerations` docstring says "list of K8s toleration dicts" — stored type after `__post_init__` is `tuple[dict, ...]`.
- `extract_node_filters` docstring should mention it also returns `{}` when no scenario entry is present, not only when no helmValues.
- `extract_node_filters` reads only `scenario[0]` (consistent with `derive_gpu_resource_type` and `gpu_cost_per_pair`, but worth a one-line note for parity).

### What's good

- All four issue #261 root causes accounted for: 1–3 implemented, 4 deferred to #262 with clear rationale.
- 32 new tests covering legacy parity, cordoned, NoSchedule/NoExecute non-tolerated, PreferNoSchedule ignored, wrong-product, pending-pod-still-skipped, per-role union both ways, NotIn/Exists ignored.
- Tuple consistency: pods on filtered-out nodes drop from `total_requested` (`capacity.py:148-159`).
- Conservative-not-broken posture verified — empty tolerations cause under-count, never over-count.

kalantar:
Pushed `5ff5c2e` addressing the review.

The three blockers are fixed:
- The second YAML parse in `_cmd_run` is gone — the new block now reuses `resolved` from the existing parse, so a malformed `baseline.yaml` warns once and doesn't silently revert the probe to legacy unfiltered mode.
- `_extract_required_gpu_products` warns when a `nvidia.com/gpu.product` matchExpression is present but the operator is not `In` (catches typos like lowercase `in`). Plain absence of a product expression stays silent — the filter is for surfacing intent vs. typo, not nagging on legitimately unconstrained affinity.
- `probe_free_gpus` now scans node taints for unrecognized non-empty effects and emits one warn per probe cycle listing them. `Noschedule` and other typos surface immediately instead of slipping through.

Also addressed in the same commit: clarified the `probe_free_gpus` docstring on `None` vs `[]` vs `[NodeFilter()]`, fixed the `NodeFilter.tolerations` docstring (`list` → `tuple`), noted that `extract_node_filters` reads only `scenario[0]`, and added the mixed-eligibility pod test you flagged plus warn-coverage tests for the two new warn paths.

The `_toleration_matches_taint` empty-effect concern is filed as #265 to land alongside #263. Adding shape validation to `__post_init__` now would harden a code path that doesn't exist yet and compounds the coercion already noted as over-engineered.

Local CI parity: lint clean, 906 tests pass.

kalantar:
## Verdict: READY TO MERGE

The three blockers from the prior review are genuinely fixed: the redundant YAML parse in `_cmd_run` is gone and the existing `resolved` is reused, the operator-typo warn fires only when intent vs typo can be distinguished (and a companion test asserts no spurious warn on legitimate absence), and the unrecognized-taint-effect warn is aggregated to one message per probe cycle so it doesn't spam. The added mixed-eligibility test is solid — its `(5, 8, 3)` assertion catches both the "double-count excluded pod" regression and the "include excluded node entirely" regression, and the warn-path tests check message content rather than just structural coverage.

Doc fixes are accurate: the `probe_free_gpus` docstring correctly distinguishes None/[]/[NodeFilter()], the `tolerations` field type matches the stored tuple, and the `scenario[0]` parity note is in place.

Deferring the toleration shape-validation to follow-up #265 is the right call — adding `__post_init__` validation now would harden a code path that doesn't exist until #263 lifts the hardcoded `tolerations=()`.

Two small polish items if you want to fold them in before merge, otherwise fine as follow-ups:

- The operator-typo warn in `_extract_required_gpu_products` names the bad operator but not the role. With both decode and prefill running through extraction, a typo in only one role leaves the operator guessing which one. The caller has the role string already; passing it through would make the message immediately actionable.
- The unrecognized-taint-effect warn aggregates effects but not the offending node names. On a large cluster, `unrecognized taint effects ['Noschedule']` doesn't tell the operator where to look.

Neither is a correctness issue — just message quality. Both are one-line changes if you want them now.

Pre-existing test gaps still open (not introduced or worsened by this PR, worth tracking separately if they matter): multiple `nodeSelectorTerms` OR-semantics, `Equal` toleration with key match + value mismatch, multiple terms with disjoint product sets. These weren't called out in the prior review as blockers and don't need to gate this merge.

Approve.

PR #267, opened 2026-06-02
fix: shuffle pending pairs before capacity gate

### LINKED ISSUE #266: Dispatch order is biased toward insertion-order prefix; shuffle should run before the capacity gate
## Observation

In runs of `deploy.py run`, pairs in the first half of the manifest are consistently dispatched in the first wave, with second-half pairs only entering the schedule after first-half pairs start completing. The bias is visible across runs even though `random.shuffle` is called every cycle.

## Cause

In `pipeline/deploy.py:_cmd_run` the dispatch cycle does:

```python
pending = _pending_pairs()                       # insertion order
dispatchable = _capacity_limited_pairs(pending,  # sorts by cost asc, greedy fill
                                       free_gpus=effective_free,
                                       cost_map=pair_costs)
random.shuffle(dispatchable)                     # shuffle AFTER the budget filter
for ns, pair_key in zip(free_slots, dispatchable):
    ...
```

`_capacity_limited_pairs` (`pipeline/deploy.py:1848-1866`) uses `sorted(pending, key=cost)` — Python's `sorted` is stable. With every pair currently costing 4 GPUs, the sort is effectively a no-op and `dispatchable` keeps `pending`'s insertion order. The greedy fill then takes the first N entries that fit the budget.

Concretely with 24 pairs at 4 GPUs each and ~49 free GPUs: `dispatchable` is exactly the **first 12 pairs in insertion order**. The shuffle on line 2237 only reorders within those 12 — second-half pairs were never candidates this cycle. Second-half pairs only become eligible once first-half pairs finish and free budget for the next cycle.

The shuffle is doing only one of two things it could do: it's randomizing *which slot* a chosen pair lands on, but it's not randomizing *which pairs get chosen* in the first place.

## Relationship to #247 / PR #248

PR #248 (issue #247) added the post-gate `random.shuffle(dispatchable)` and was scoped explicitly to "randomize pair-to-slot assignment" — i.e., randomize *which slot* a chosen pair lands on. That contract is met and is preserved by the fix below.

This issue addresses a **different bias not covered by #248**: pair-selection bias (which pairs get picked for the first wave). The fix is additive — slot-assignment randomness from #248 still works after the move, because a stable sort on a shuffled input preserves the shuffled order for equal-cost groups.

## Proposed fix

Move the shuffle to run on `pending` *before* the capacity gate, and **before** the surrounding `if free_gpus is not None and pending:` branch so the no-filter `else: dispatchable = pending` path also gets the shuffled list:

```python
pending = _pending_pairs()
random.shuffle(pending)                          # randomize the full set
if free_gpus is not None and pending:
    if not backoff.should_dispatch(...):
        dispatchable = []
    else:
        dispatchable = _capacity_limited_pairs(pending, ...)
        # … existing capacity-limited / slot-limited logging
else:
    dispatchable = pending                       # already shuffled — no further work needed
for ns, pair_key in zip(free_slots, dispatchable):
    ...
```

This works because `sorted` is stable: with all-equal costs the shuffle order is preserved through the gate, and the greedy budget fill picks a random subset of size N from the full pool. With non-equal costs the sort still orders smallest-cost-first across cost groups, but randomizes within each group — budget packing is unchanged.

The post-gate `random.shuffle(dispatchable)` at line 2237 can be deleted; slot-assignment randomness is preserved because `dispatchable` is already in shuffled order when `zip(free_slots, dispatchable)` runs.

## Impacts

**Positive**
- First-wave pairs are an unbiased sample of the full pending list rather than the manifest's first 12 entries.
- Long-tail pairs no longer accidentally cluster at the start or end of the run based on their alphabetical/insertion position.
- Slot-assignment randomness preserved (#248's contract upheld).
- No change to budget packing for cost-asymmetric workloads.

**Negative / to be aware of**
- **Makespan may regress slightly in expectation.** Today the slowest pairs (`blindspot-under-*` at ~38 min) tend to land in the first wave because their alphabetical position puts them in the first 12. Random ordering distributes slow pairs uniformly; the slowest pair finishing last sets total wall time, so pushing slow pairs late hurts makespan. If makespan is the priority, the right answer is length-aware scheduling (LPT — Longest Processing Time first), not this fix. This change buys fairness, not optimal scheduling.
- **Run-to-run timing variance increases.** Today's runs are quasi-deterministic in dispatch order; after this change they vary fully. Two runs of the same 24 pairs may differ in total wall time by tens of minutes purely from luck of the draw on which slow pair lands last.
- **Determinism not addressed here.** `random` is unseeded today and would remain unseeded. Fixing both fairness and reproducibility together would require seeding the global RNG (e.g. `random.seed(args.seed or hash(run_dir.name))` near the top of `_cmd_run` and exposing `--seed`). Worth considering as a follow-up if debug-reproducibility matters.

**Unaffected**
- `_pending_pairs`, `_capacity_limited_pairs`, the shadow ledger, the backoff controller, and the slot-readiness check are all untouched.
- When `len(pending) ≤ len(free_slots)` (late in a run) every pending pair dispatches this cycle — same behavior as today.

## Cross-path parity

`_cmd_run` is sim2real-pipeline code with a single dispatch loop and no parallel run/replay/observe paths. No cross-command parity concerns.

## Scope

- `pipeline/deploy.py` `_cmd_run` dispatch cycle — move `random.shuffle` from after the capacity gate (line 2237) to immediately after `pending = _pending_pairs()` (line 2202), before the `if free_gpus is not None and pending:` block; delete the post-gate shuffle.
- Test: extract a small helper (e.g. `_select_dispatchable(pending, free_gpus, cost_map)` that does shuffle-then-`_capacity_limited_pairs`) so the unit test has an unambiguous target. Construct a 24-pair pending list with all-equal cost and a budget that fits 12, run the helper many times, and assert that over many trials the chosen pairs are roughly uniformly drawn from the full 24 (chi-squared or "every pair appears in ≥ K% of trials" works) — not just the first 12.

kalantar:
Closes #266

## Summary

Today the orchestrator's dispatch cycle calls `random.shuffle(dispatchable)` *after* `_capacity_limited_pairs` has already taken the first N entries of `pending` (in insertion order, because Python's `sorted` is stable and every pair currently costs the same). The shuffle therefore only randomizes which slot a chosen pair lands on; it does not randomize which pairs get chosen. The first wave is always the first 12 pairs in manifest order; second-half pairs only enter after first-half pairs complete.

This PR moves the shuffle upstream of the capacity gate via a new helper `_select_dispatchable(pending, *, free_gpus, cost_map)` that shuffles a copy of pending and then delegates to `_capacity_limited_pairs`. The greedy budget fill now picks a uniform random subset under all-equal costs; smallest-cost-first packing is preserved across cost groups under heterogeneous costs.

## What changed

- **`pipeline/deploy.py`**
  - New `_select_dispatchable(pending, *, free_gpus, cost_map)` adjacent to `_capacity_limited_pairs` — shuffles a `list(pending)` copy, then capacity-gates.
  - `_cmd_run` dispatch block: replace the inline `_capacity_limited_pairs(...)` call with `_select_dispatchable(...)`. The no-filter `else: dispatchable = pending` branch now takes a shuffled copy too. The trailing post-gate `random.shuffle(dispatchable)` is deleted (the shuffle is now upstream).
- **`pipeline/tests/test_deploy_run.py`**
  - New `TestSelectDispatchable` (7 tests): empty-pending, all-fits, no-input-mutation, uniform 50% per-pair appearance over 2000 trials, first-half-not-overrepresented over 1000 trials (the regression test for #266), smallest-first packing preserved under heterogeneous costs, intra-cost-group randomization.
  - Existing `test_dispatch_shuffles_dispatchable` continues to pass unchanged.

## Reviewer attention

- **The fairness invariant relies on Python's `sorted` being stable.** With all-equal costs (today's reality), the stable sort preserves shuffled input order through the gate, so the greedy fill picks a uniform random subset. With heterogeneous costs, sort still orders by cost ascending across groups while randomization applies within each group — `_capacity_limited_pairs`'s smallest-first packing is preserved.
- **The `else: dispatchable = pending` path now copies before shuffling.** Pre-fix, the post-gate `random.shuffle(dispatchable)` mutated `pending` itself when `free_gpus is None` or `pending` was empty. `_pending_pairs()` returns a fresh list each call (see `deploy.py:2031-2033`), so no caller relied on order — but `dispatchable = list(pending); random.shuffle(dispatchable)` removes that footgun.
- **Slot-assignment randomness from #248 is preserved.** `dispatchable` is in shuffled order when `zip(free_slots, dispatchable)` runs, just sourced from `_select_dispatchable` rather than the post-gate shuffle.
- **Issue #266 explicitly defers two related concerns.** Makespan-aware scheduling (LPT) is out of scope — this change buys fairness, not optimal scheduling, and may slightly regress wall time in expectation when slow pairs land late by luck. Run-to-run timing variance also increases. RNG seeding for reproducibility is also deferred.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 913 passed
- [x] Existing `test_dispatch_shuffles_dispatchable` passes unchanged (slot-assignment randomness contract from #248 preserved)
- [x] New `test_first_half_not_overrepresented_under_equal_costs` is the direct regression test for #266
- [x] `_capacity_limited_pairs` has no production callers remaining outside `_select_dispatchable`; existing tests of the lower-level function still pass

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -1866,6 +1866,28 @@ def _capacity_limited_pairs(
     return result
 
 
+def _select_dispatchable(
+    pending: list[str],
+    *,
+    free_gpus: int,
+    cost_map: dict[str, int],
+) -> list[str]:
+    """Shuffle pending then capacity-gate.
+
+    Shuffling before the gate (rather than after) makes the chosen subset an
+    unbiased random sample of the full pending list. With all-equal costs,
+    `_capacity_limited_pairs`'s stable sort preserves shuffled order so the
+    greedy fill picks a uniform random subset. With heterogeneous costs,
+    smallest-cost-first packing is preserved across cost groups while
+    randomization applies within each group.
+
+    Does not mutate `pending` — operates on a shuffled copy.
+    """
+    shuffled = list(pending)
+    random.shuffle(shuffled)
+    return _capacity_limited_pairs(shuffled, free_gpus=free_gpus, cost_map=cost_map)
+
+
 def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
     """Orchestrate parallel pool execution across namespace slots."""
     import tempfile as _tmp
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2232,7 +2232,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 dispatchable = []
             else:
                 effective_free = shadow.effective_free(free_gpus)
-                dispatchable = _capacity_limited_pairs(
+                dispatchable = _select_dispatchable(
                     pending,
                     free_gpus=effective_free, cost_map=pair_costs,
                 )
@@ -2254,9 +2254,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                         info(f"Dispatching {len(free_slots)}/{len(pending)} pending pairs (slot-limited)")
                         _last_log_state["dispatch"] = _disp_state
         else:
-            dispatchable = pending
-
-        random.shuffle(dispatchable)
+            dispatchable = list(pending)
+            random.shuffle(dispatchable)
 
         for ns, pair_key in zip(free_slots, dispatchable):
             hf_secret_name = setup_config.get("hf_secret_name", "hf-secret")
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -2853,6 +2853,117 @@ def test_health_escalation_cancels_pipelinerun(tmp_path, monkeypatch):
     assert saved["wl-a-baseline"]["status"] == "failed"
 
 
+class TestSelectDispatchable:
+    """Tests for _select_dispatchable: shuffle then capacity-gate."""
+
+    def test_returns_empty_when_pending_empty(self):
+        from pipeline.deploy import _select_dispatchable
+        result = _select_dispatchable([], free_gpus=8, cost_map={})
+        assert result == []
+
+    def test_returns_all_when_budget_fits_all(self):
+        from pipeline.deploy import _select_dispatchable
+        pending = [f"p{i}" for i in range(5)]
+        cost_map = {k: 1 for k in pending}
+        result = _select_dispatchable(pending, free_gpus=10, cost_map=cost_map)
+        assert sorted(result) == sorted(pending)
+
+    def test_does_not_mutate_input_pending_list(self):
+        from pipeline.deploy import _select_dispatchable
+        pending = ["a", "b", "c", "d"]
+        original = list(pending)
+        cost_map = {k: 1 for k in pending}
+        _select_dispatchable(pending, free_gpus=2, cost_map=cost_map)
+        assert pending == original, "input list must not be mutated"
+
+    def test_picks_uniform_random_subset_under_equal_costs(self):
+        """The fairness criterion from issue #266.
+
+        With 24 pairs at 4 GPUs each and budget=48 (fits 12), each pair should
+        appear in roughly 12/24 = 50% of trials. Run many trials and check the
+        per-pair appearance rate is within tolerance of 0.5.
+        """
+        from pipeline.deploy import _select_dispatchable
+        import random as _random
+        pending = [f"p{i:02d}" for i in range(24)]
+        cost_map = {k: 4 for k in pending}
+        trials = 2000
+        counts = {k: 0 for k in pending}
+        rng = _random.Random(42)  # deterministic for the test
+        for _ in range(trials):
+            with patch("pipeline.deploy.random", rng):
+                picked = _select_dispatchable(pending, free_gpus=48, cost_map=cost_map)
+            assert len(picked) == 12
+            for k in picked:
+                counts[k] += 1
+        # Each pair should appear in roughly 50% ± 5% of trials.
+        for k, c in counts.items():
+            rate = c / trials
+            assert 0.40 <= rate <= 0.60, (
+                f"{k} appeared {c}/{trials} = {rate:.2%}, expected ~50%"
+            )
+
+    def test_first_half_not_overrepresented_under_equal_costs(self):
+        """Direct test of the regression #266 fixes: first-half pairs are not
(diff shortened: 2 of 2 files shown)


kalantar:
## Verdict: READY TO MERGE

Faithful, minimal implementation of the fix proposed in #266. The core invariant works: with all-equal costs the stable sort preserves shuffled input through `_capacity_limited_pairs`, giving a uniform random subset; with heterogeneous costs the smallest-first packing across cost groups is preserved while randomization happens within each group. Slot-assignment randomness from #248 is preserved transitively because `dispatchable` is in shuffled order when `zip(free_slots, dispatchable)` runs. No new silent failures.

One thing worth correcting in the PR description: the `else: dispatchable = pending` branch was already shuffled before this PR (the unconditional post-gate `random.shuffle(dispatchable)` ran on it). The actual semantic change there is moving from in-place mutation of `pending` to operating on a copy — a latent-bug cleanup, not a new shuffle.

Two test-quality items worth folding in if convenient, neither blocking:

- `test_intra_cost_group_randomization` (#7 in the PR description) uses uniform cost-1 pairs with budget=2, which is just a smaller rerun of the uniform test rather than actually isolating within-group randomization from across-group ordering. A clearer test would mix tiers — e.g. 4 cost-1 + 4 cost-2 with budget large enough to admit all cost-1 plus 2 cost-2 — and assert that the cost-2 admits are random among the cost-2 pool while cost-1 always wins the budget contest.
- `test_first_half_not_overrepresented_under_equal_costs` uses a ±5% band on the aggregate first-half ratio. That catches the pre-fix bug (ratio=1.0, ~10σ off) but would let a real 55/45 bias slip through. With 1000 trials the binomial std is ~1.1%, so a tighter ±2-3% would still be flake-safe and detect subtler regressions.

Edge cases not covered today (low priority): `cost_map` missing a key, duplicate keys in `pending`, `free_gpus=0`. Worth contracting via type/raise or a one-line test, but the current production caller doesn't exercise any of them.

Doc nit: the test docstring at `test_deploy_run.py:2907` reads "regression #266 fixes" — minor — consider "regression that #266 fixes". The new `_select_dispatchable` docstring is good (explains the why and the sort-stability dependency).

Approve.

Issue #265, opened 2026-06-02
Validate toleration shape when #263 lifts the empty-tolerations assumption


kalantar:
## Problem

`pipeline/lib/capacity.py:_toleration_matches_taint` follows K8s semantics: an empty toleration `effect` matches any taint effect, and operator `Exists` with an empty `key` matches any taint key. These wildcard semantics are correct, but they make a mis-specified toleration dict (e.g. \`{\"key\": \"x\"}\` with no \`effect\`) silently match every blocking taint — under-counting blockers and over-counting eligible nodes.

This is dead code today: #261 ships with `extract_node_filters` always returning `tolerations=()`. The risk only materializes when #263 populates real tolerations from the resolved scenario.

## Resolution

When #263 wires actual toleration extraction, harden the data path. Options:

- Validate toleration dict shape at `NodeFilter` construction (require `effect` to be one of `\"\", \"NoSchedule\", \"NoExecute\", \"PreferNoSchedule\"`; require `operator` to be `\"Equal\"` or `\"Exists\"`; warn on anything else). Note: this should land alongside #263's extraction code, not before, so the validation lives at the boundary where shape variance actually originates.
- Or: warn at extraction time when a toleration has an unexpected operator/effect.

## Out of scope here

Adding validation to `NodeFilter.__post_init__` now would harden against a code path that does not yet exist and compounds the over-engineered coercion already in `__post_init__`. The right place is when extraction is implemented.

## Related

- #261 — primary capacity-probe filter (ships with empty tolerations)
- #263 — locate workload tolerations in resolved scenario

PR #269, opened 2026-06-02
fix: capacity probe reads acceleratorType + always screens cordon/taint

### LINKED ISSUE #268: Capacity probe filter bypasses cordon/taint screening when no GPU product is extracted; extractor reads wrong schema path
## Problem

PR #264 added cordoned-node, untolerated-taint, and wrong-product filtering to `pipeline/lib/capacity.py:probe_free_gpus`. In practice the filter never engages on real scenarios, leaving the probe in legacy unfiltered mode and the orchestrator over-dispatching as if #264 had never landed.

Two distinct defects combine to produce this. Either alone is enough to disable the filter; together they make it impossible for the filter to do its job.

## Defect 1: extractor reads a path that real scenarios don't use

`extract_node_filters` in `pipeline/lib/capacity.py:294-313` reads from:

```yaml
scenario:
- model:
    helmValues:
      decode:
        extraConfig:
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                - matchExpressions:
                  - key: nvidia.com/gpu.product
                    operator: In
                    values: [...]
```

That path was inferred from the human-facing suggestion text in `pipeline/lib/health.py:193`. It does not reflect the standard scenario schema. Real scenario bundles express the GPU product constraint as:

```yaml
scenario:
- name: expceil
  decode:
    acceleratorType:
      labelKey: nvidia.com/gpu.product
      labelValue: NVIDIA-H100-80GB-HBM3
```

This is the form produced by user-authored scenarios in experiment repos (e.g. `expceil/baselines/expceil.yaml`) and is what flows into `scenarioContent` on the rendered PipelineRun. The Helm chart translates `acceleratorType.labelKey`/`labelValue` into the actual K8s `nodeAffinity` on the rendered Deployment — but the orchestrator never sees the rendered form, only the scenario.

Result: `extract_node_filters` returns `{}` for every real scenario.

## Defect 2: empty filter dict bypasses cordon/taint screening

In `_cmd_run` (`pipeline/deploy.py`), the filter is forwarded as:

```python
node_filters = extract_node_filters(resolved)  # {} for real scenarios
...
capacity = probe_free_gpus(
    gpu_resource_type=gpu_resource_type,
    node_filters=list(node_filters.values()) or None,  # → None
)
```

In `probe_free_gpus`:

```python
filters = node_filters or []
...
for node in nodes.get("items", []):
    if filters and not node_is_eligible(node, filters):
        continue
```

When `filters` is empty, the `if filters and ...` short-circuits to False and every node passes — **including cordoned nodes and nodes with non-tolerated NoSchedule/NoExecute taints.**

This is the more serious defect. Cordon and taint are cluster facts: a cordoned node never accepts new pods regardless of what GPU product the workload wants. They should be screened *always*, independent of whether a product constraint can be extracted from the scenario. #264 conflated three independent filter rules into one all-or-nothing gate.

## Evidence

Live orchestrator run (`sim2real-orchestrator-ntcdp` in namespace `kalantar-0`, image `ghcr.io/inference-sim/sim2real/orchestrator@sha256:4187586f...03e26`, started 2026-06-02T15:27:27Z — confirmed post-#264, post-#267):

```
15:27:34 [INFO]  Capacity: 46 free GPUs (83 allocatable − 37 requested)
15:27:34 [INFO]  Dispatching 11/24 pending pairs (capacity-limited: 46 free GPUs)
```

`allocatable = 83` is the unfiltered cluster total. Direct cluster snapshot at the same time: 5 cordoned nodes contributing 35 GPUs, 6 eligible H100 nodes contributing 48 GPUs. With cordon+taint screening alone, allocatable should have been 48; with the H100 product filter on top, the same 48 since all eligible nodes are H100. The probe behaved as if #264 had not landed.

No `Eligibility filter [decode]: ...` line appeared at orchestrator startup — confirming `extract_node_filters` returned `{}`.

## Proposed fix

**Part 1 — always apply cordon/taint screening.** In `_cmd_run`, when `extract_node_filters` returns an empty dict, still pass at least a default `NodeFilter()` so cordon and taint screening fire on cluster facts. Smallest change: replace

```python
node_filters=list(node_filters.values()) or None
```

with

```python
node_filters=list(node_filters.values()) or [NodeFilter()]
```

A `NodeFilter()` with `required_gpu_products=frozenset()` and `tolerations=()` already short-circuits product screening (per `_filter_admits_node`'s `if filt.required_gpu_products` guard) but applies cordon and taint screening unconditionally. No change to `probe_free_gpus` or `node_is_eligible`.

**Part 2 — read `acceleratorType` from the actual scenario schema.** Extend `extract_node_filters` to read `scenario[0].{decode,prefill}.acceleratorType.{labelKey, labelValue}` for each known role. When `labelKey == "nvidia.com/gpu.product"`, use `{labelValue}` as the single-element `required_gpu_products` set. Keep the existing `model.helmValues.{role}.extraConfig.affinity` path as a fallback for users who override affinity directly. Same role-keyed dict result shape — no change to `probe_free_gpus` or `_cmd_run`.

## Tests

In `pipeline/tests/test_capacity.py`:

- New `TestExtractNodeFilters::test_extracts_from_acceleratorType_schema` — fixture shaped exactly like `expceil/baselines/expceil.yaml` with `decode.acceleratorType.labelKey` / `labelValue`. Assert filter has the expected single-product set.
- New `test_acceleratorType_with_non_gpu_product_label_ignored` — `labelKey: topology.kubernetes.io/zone` should not produce a product constraint.
- New `test_per_role_acceleratorType` — different `acceleratorType` on decode and prefill produces per-role filters with different product sets.
- Existing `model.helmValues...affinity` tests stay green (fallback path preserved).

In `pipeline/tests/test_deploy_run.py` (or extend existing):

- Test that with `extract_node_filters` returning `{}`, the call to `probe_free_gpus` from `_cmd_run` passes a non-None `node_filters` argument so cordon/taint screening is on. Mock both functions and assert the kwarg.

## Cross-path parity

`probe_free_gpus` has a single caller (`pipeline/deploy.py:_cmd_run`). No parallel run/replay/observe paths.

## Out of scope

- Workload tolerations remain `()` per the conservative assumption from #261; lifting that is tracked in #263.
- Per-node fragmentation for multi-GPU pods stays in #262.
- `_toleration_matches_taint` empty-effect wildcard validation stays in #265.

## Related

- #261 — original capacity-probe over-counting issue this PR was meant to fix
- #264 — the merged PR introducing the defects
- #263 — toleration extraction (deferred)

kalantar:
Closes #268

## Summary

Fixes two defects from PR #264 that combined to disable the capacity-probe filter on every real scenario.

## Defects

**Defect 1 — wrong schema path.** `extract_node_filters` read `scenario[0].model.helmValues.{role}.extraConfig.affinity.nodeAffinity...`, but real scenarios place the GPU constraint at `scenario[0].{role}.acceleratorType.{labelKey, labelValue}`. Filter returned `{}` for every real scenario.

**Defect 2 — empty filter dict bypassed cordon/taint screening.** When `extract_node_filters` returned `{}`, `_cmd_run` forwarded `node_filters=None`. In `probe_free_gpus`, `if filters and not node_is_eligible(...)` short-circuited, so every node passed — including cordoned and tainted ones.

Live evidence at orchestrator startup confirmed the probe behaved as if #264 had not landed: `allocatable=83` was the unfiltered cluster total despite 5 cordoned nodes contributing 35 GPUs.

## What changed

- `extract_node_filters` now reads `scenario[0].{role}.acceleratorType.{labelKey, labelValue}` — the canonical schema. The previous `model.helmValues.{role}.extraConfig.affinity` path matched no real scenario and is removed entirely (along with its helper and tests).
- `_cmd_run` now passes `[NodeFilter()]` instead of `None` when the extractor produces no per-role constraints, so cordon and taint screening always apply.
- A warn fires when `acceleratorType.labelKey == "nvidia.com/gpu.product"` but `labelValue` is missing/empty, and when `acceleratorType` is present but non-dict (string/list/null typo). Both name the role and the canonical example value.
- A warn fires when a role key is present in `scenario[0]` but its value is non-dict (e.g. `decode: null`).
- A new INFO line at startup states when cordon/taint-only screening is in effect (no per-role product constraint extracted) — separate from the existing per-role product log.

## Tests

- `TestExtractNodeFilters` exercises the canonical schema, non-GPU labelKey ignored, per-role asymmetry, missing-`labelValue` warns with example, non-GPU labelKey is silent, non-dict `acceleratorType` warns, and tolerations-empty invariant.
- `TestCmdRunForwardsNodeFilters` in `test_deploy_run.py` drives `_cmd_run` end-to-end, captures `probe_free_gpus` kwargs, and asserts both branches: empty extractor → `[NodeFilter()]`; populated role → role filter with the right product set. Also asserts the new INFO line fires.
- The pre-existing `test_filter_excludes_cordoned_nodes` already exercises `probe_free_gpus(node_filters=[NodeFilter()])` against a cordoned node — that's the regression assertion for defect 2.

## Reviewer attention

- The empty-filter case now applies cordon/taint screening rather than silently reverting to legacy unfiltered behavior. Conservative direction: under-counts when a workload would legitimately tolerate a taint, never over-counts.

## Out of scope

- Workload tolerations remain `()` per the conservative assumption from #261 (lifting that is tracked in #263).
- `acceleratorType.labelValues` (plural) takes precedence over the singular `labelValue` per defaults.yaml — not yet read; tracked in #270.
- Per-node fragmentation for multi-GPU pods stays in #262.
- `_toleration_matches_taint` empty-effect wildcard validation stays in #265.


kalantar:
## Review: NEEDS CHANGES

Both defects from issue #268 are correctly addressed by the canonical `acceleratorType` path, but the PR keeps the wrong-schema fallback alive instead of deleting it. Reviewers flagged three required changes.

### Required

1. **Delete the `helmValues.{role}.extraConfig.affinity` fallback in `extract_node_filters` (`pipeline/lib/capacity.py:345-347`).** The defaults that underlie scenario resolution (`llm-d-benchmark/config/templates/values/defaults.yaml`) place GPU constraints at `<role>.acceleratorType.{labelKey, labelValue}` and pod-affinity in a top-level `affinity:` block — neither matches `model.helmValues.<role>.extraConfig.affinity`. That path was inferred from stale operator-suggestion text in `health.py:193` (originally referencing a deleted `env_defaults.yaml`) and PR #264 picked it up. Keeping it as a fallback perpetuates the wrong-schema reading the canonical path was added to fix; it doesn't help any real user. Drop the fallback, the helper, and its tests, and remove the docstring/PR-description claims about helmValues precedence. This also dissolves the precedence-semantics concern in my earlier review.

2. **`TestNodeFiltersForwarding` is a tautology.** `pipeline/tests/test_deploy_run.py:3201-3221` evaluates the literal expression `list({}.values()) or [NodeFilter()]` in the test file — it never imports or calls `_cmd_run`. If a future change removes `or [NodeFilter()]` from `deploy.py:2181`, the test still passes. This is exactly the regression defect 2 represents. Mock `probe_free_gpus` and assert `_cmd_run` calls it with `node_filters=[NodeFilter()]` when the extractor returns `{}`. The class docstring is also misleading on what it actually verifies.

3. **Stale implementation plan committed.** `docs/superpowers/plans/2026-06-02-capacity-filter-fix.md` Task 4 prescribes a `test_default_filter_excludes_cordoned_node` test that was deliberately skipped (the PR relies on the pre-existing `test_filter_excludes_cordoned_nodes`). Either drop the plan from the repo or annotate it as historical with the deviation noted.

### Smaller observations (non-blocking)

- `_extract_required_gpu_products_from_accelerator_type` silently returns `frozenset()` when `acceleratorType` is a non-dict (string, list, null) — operator typos look identical to "not set." Consider warning the same way the empty-`labelValue` case does.
- The empty-`labelValue` warn at `capacity.py:311-313` doesn't tell the operator how to fix it. Adding the canonical example (e.g. `NVIDIA-H100-80GB-HBM3`) would match the discipline of the existing affinity-operator warn.
- No test covers a scenario with neither `acceleratorType` present on a role nor any role keys in `scenario[0]` — the trigger for the new INFO line.

### Strengths

- The `probe_free_gpus` `node_filters semantics:` doc block (`capacity.py:101-115`) explicitly documents the new `[NodeFilter()]` contract — the call site behavior matches.
- `_GPU_PRODUCT_LABEL` constant is reused.
- `test_acceleratorType_takes_precedence_over_helmValues_affinity` becomes obsolete once item 1 is applied (and that's the right outcome — the precedence question only existed because the fallback existed).
- `pipeline/README.md` is updated to reflect the new schema path; it should also be updated to drop any helmValues fallback wording when item 1 lands.

kalantar:
Pushed `eaa48a5` addressing the review.

The three required items are done:

- The `helmValues.{role}.extraConfig.affinity` fallback is gone. `_extract_required_gpu_products` and its tests are removed; `extract_node_filters` reads only `scenario[0].{role}.acceleratorType`. README and docstring no longer mention helmValues precedence. The precedence test became vacuous and is deleted with the rest.
- `TestNodeFiltersForwarding` is replaced by `TestCmdRunForwardsNodeFilters` in `test_deploy_run.py`. It drives `_cmd_run` end-to-end through the existing dispatch test pattern, captures `probe_free_gpus` kwargs, and asserts the empty-extractor path forwards `[NodeFilter()]` while a real `acceleratorType` path forwards the role filter. The third case in that class also covers the new INFO-line trigger.
- The stale plan doc is deleted from the repo.

The smaller items are also addressed in the same commit:

- `_extract_required_gpu_products_from_accelerator_type` now warns when `acceleratorType` is present but non-dict (string/list/null typo), naming the role and the actual type.
- The empty-`labelValue` warn now includes a canonical example value (`NVIDIA-H100-80GB-HBM3`) so the operator can see the expected shape.
- The new INFO-line trigger has a dedicated test (`test_no_per_role_constraint_logs_info_line`) plus the path is exercised by the empty-extractor integration test.

One observation surfaced while validating the helmValues drop: `defaults.yaml:768-774` documents an `acceleratorType.labelValues:` plural form that takes precedence over the singular `labelValue`. The current extractor doesn't read it, so a scenario using the plural form would silently lose its product constraint. Filed separately as #270 — out of scope for this PR.

Local CI: 839 tests pass, ruff `--select F` clean.

kalantar:
## Re-review (commit eaa48a5): NEEDS CHANGES (light)

All three required items from the prior review landed correctly:

- helmValues fallback and `_extract_required_gpu_products` helper are deleted from `capacity.py`; their tests are gone from `test_capacity.py`. README and docstrings no longer mention a fallback. Grep confirms no stragglers.
- `TestCmdRunForwardsNodeFilters` (`test_deploy_run.py`) drives real `_cmd_run` end-to-end via `_setup_dispatch_run`, monkeypatches `probe_free_gpus` at the right import seam, and captures kwargs. Both branches asserted: empty extractor → `[NodeFilter()]`; populated role → `frozenset({"NVIDIA-H100-80GB-HBM3"})`. INFO line emission is also asserted. No tautology.
- Stale plan doc deleted.

Bonus items also addressed: non-dict `acceleratorType` warn added, empty-`labelValue` warn now includes the canonical example, README updated.

### One blocker before merge

**PR description is stale.** The "What changed" section still says `extract_node_filters` "falls back to `model.helmValues.{role}.extraConfig.affinity`," "Reviewer attention" still claims `acceleratorType` wins precedence over helmValues, and "Tests" still lists a "fallback to helmValues affinity" test. None of that describes the merged code, and the merge-commit message will inherit it. Edit the PR body to drop fallback/precedence claims and the obsolete test bullet.

### Non-blocking — defer or follow-up

- A misspelled `acceleratorType` key (e.g. `acceleratoreType:`) is silently treated as "not set" — no warn. Same for a misspelled `labelKey` (the `labelKey != "nvidia.com/gpu.product"` branch is silent by design but covers typos too). Worth a follow-up issue.
- Non-dict role values (e.g. `decode: null` or `decode: "h100"`) are skipped silently with no warn — the new non-dict warn covers `acceleratorType` but not the role itself.
- `test_role_with_acceleratorType_forwards_role_filter` could also assert the existing "Eligibility filter [decode]" log line for symmetry with the empty-extractor INFO assertion. One-line addition.
- Minor smell: a few tests call `capsys.readouterr()` twice in a single expression — the second read is empty. Still passes; cleaner as a single read.

Once the PR description is updated, this is good to merge.

kalantar:
Pushed `904df47` plus the PR description edit.

The blocker is resolved — PR description no longer mentions a helmValues fallback, the precedence claim, or the obsolete fallback test. It now describes the merged code accurately and lists `labelValues` plural as out-of-scope (tracked in #270).

The non-blocking items are also folded in:

- A non-dict role value (e.g. `decode: null`, `decode: "h100"`) now warns with the role name and the actual type, instead of silently being skipped. Covered by `test_warns_on_non_dict_role_entry`.
- `test_role_with_acceleratorType_forwards_role_filter` now also asserts the `"Eligibility filter [decode]"` log line is emitted, for symmetry with the empty-extractor INFO-line assertion.
- `capsys.readouterr().out + capsys.readouterr().err` instances replaced with a single read into a captured object — the second call was reading an already-cleared buffer.

Misspelled top-level keys (e.g. `acceleratoreType:` with an extra 'e') is still a silent path — same INFO line emission whether the key was typo'd or legitimately absent. Filed as #271 for a fuzzy-match suggestion pass; out of scope for this PR.

Local CI: 840 tests pass, ruff clean.

PR #279, opened 2026-06-03
fix: merge extraObjects by Kubernetes identity, not list position (#278)

### LINKED ISSUE #278: deep_merge corrupts `extraObjects` list — positional fold by index instead of merge by Kubernetes identity
## Summary

Hand-edited entries added to `extraObjects:` in a scenario file (`baselines/<name>.yaml` or treatment YAML) get **structurally corrupted** when assembled against a generator-produced overlay (`generated/<name>_config.yaml`). The current `deep_merge` falls through to a positional, index-by-index list merge that folds dissimilar Kubernetes objects together, producing manifest entries with fields from two different kinds spliced into one document.

This is a deal-breaker for adding any hand-authored Kubernetes resource to a scenario when the overlay already defines `extraObjects`. It applies equally to baseline, control, and treatment packages — they all flow through the same `deep_merge(scenario, overlay)` step in `assemble_packages`.

## Where

`pipeline/lib/values.py:21-67` — `_merge_lists` three-tier strategy:

```python
_LIST_KEY_CANDIDATES = ("name", "mountPath", "containerPort")

def _detect_list_key(base_list: list, overlay_list: list):
    ...
    for candidate in _LIST_KEY_CANDIDATES:
        if all(candidate in d for d in all_items):
            return candidate
    return None
```

Tier 2 (named-key merge) only checks **top level** of each dict for `name` / `mountPath` / `containerPort`. Kubernetes objects in `extraObjects:` carry `name` at `metadata.name` (nested), so `_detect_list_key` returns `None`. Tier 3 fires:

```python
# Tier 3: positional merge — surplus from either side preserved
for i in range(max(len(base_list), len(overlay_list))):
    if i < len(base_list) and i < len(overlay_list):
        result.append(deep_merge(base_list[i], overlay_list[i]))   # ← fold
    ...
```

`deep_merge(Role, InferenceObjective)` then mixes their keys at the top level — overlay wins on shared scalar keys, base-only keys (`rules`, `subjects`, `roleRef`) survive as stray siblings on the overlay's `kind: InferenceObjective` document.

## Repro

Scenario file (`baselines/quartic_baseline.yaml`) — hand-authored, two RBAC objects in `extraObjects`:

```yaml
extraObjects:
  - apiVersion: rbac.authorization.k8s.io/v1
    kind: Role
    metadata: { name: ${model.idLabel}-gaie-epp-llm-d }
    rules:
    - apiGroups: ["llm-d.ai"]
      resources: ["inferenceobjectives", "inferencemodelrewrites"]
      verbs: ["get", "watch", "list"]
  - apiVersion: rbac.authorization.k8s.io/v1
    kind: RoleBinding
    metadata: { name: ${model.idLabel}-gaie-epp-llm-d }
    subjects:
    - kind: ServiceAccount
      name: ${model.idLabel}-gaie-epp
    roleRef:
      apiGroup: rbac.authorization.k8s.io
      kind: Role
      name: ${model.idLabel}-gaie-epp-llm-d
```

Overlay (`generated/baseline_config.yaml`) — generator output, two InferenceObjectives:

```yaml
extraObjects:
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferenceObjective
  metadata: { name: critical }
  spec: { poolRef: { ... }, priority: 100 }
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferenceObjective
  metadata: { name: sheddable }
  spec: { poolRef: { ... }, priority: -50 }
```

Resolved (`workspace/runs/quartic-jc/cluster/baseline.yaml`):

```yaml
extraObjects:
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferenceObjective
  metadata: { name: critical }
  rules:                            # ← stray Role.rules on InferenceObjective document
  - apiGroups: [llm-d.ai]
    ...
  spec: { poolRef: { ... }, priority: 100 }
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferenceObjective
  metadata: { name: sheddable }
  roleRef:                          # ← stray RoleBinding.roleRef on InferenceObjective document
    apiGroup: rbac.authorization.k8s.io
    kind: Role
    name: ${model.idLabel}-gaie-epp-llm-d
  spec: { ... }
  subjects:                         # ← stray RoleBinding.subjects on InferenceObjective document
  - kind: ServiceAccount
    name: ${model.idLabel}-gaie-epp
```

The hand-authored Role and RoleBinding **disappear entirely** as standalone documents; their fields are smeared onto the overlay's two InferenceObjective entries.

## Why it matters

This makes hand-edits to `extraObjects` impossible whenever the generator writes anything to the same field — which is true for baseline, control, and treatment in any scenario where InferenceObjectives or other Kubernetes resources are produced. The user has no single safe place to inject one extra Kubernetes manifest.

It is also silent: the bad output is valid YAML and Helm doesn't reject it (kubernetes-style `extraObjects` is a free-form list); the only signal is the missing-RBAC behavior at runtime, which is hard to diagnose back to the merge.

## Fix options

In rough order of cost / safety:

### A. Special-case `extraObjects` to append (smallest)

In `deep_merge` (`values.py:70-84`), recognize the field name `extraObjects` and concatenate base + overlay rather than calling `_merge_lists`. The current `_merge_lists` tier system is appropriate for typed structured fields (containers, ports, env), not for free-form Kubernetes-object lists.

The cost is one branch in `deep_merge` and a list of "always-append" field names (just `extraObjects` for now; future Kubernetes-object lists like `extraVolumes` could be added).

### B. Extend `_detect_list_key` with a composite Kubernetes identity (cleanest)

For entries that look like Kubernetes manifests, key by `(apiVersion, kind, metadata.name)`. That naturally separates `Role/<name>` from `InferenceObjective/critical` even at the same index, so neither base nor overlay folds into the other. This is the correct merge semantics if a user wants to override a specific manifest from the overlay (rare, but at least possible).

Implementation sketch: if all entries in both lists have `apiVersion`, `kind`, and `metadata.name`, build the key tuple and use Tier 2 with that composite. Falls back to Tier 3 otherwise.

### C. Both A and B

Use B as the default, but expose a list of "always-append" field names (configurable) for cases where the user explicitly wants stacking without identity-based merge. A is a safer default for `extraObjects` until B is validated.

The cosmetic counterpart — the surprising-silent-corruption symptom — should also be addressed: when Tier 3 falls through for a list whose entries differ in `apiVersion` or `kind` between base and overlay, that's almost certainly a misuse. Either log a warning at assembly time or refuse to merge and report. (Optional, but worth doing alongside the fix to catch other latent occurrences.)

## Severity

Real-world impact in this run: hand-edited Role + RoleBinding for the EPP service account never reached the cluster, the EPP got 403 errors trying to list `inferenceobjectives.llm-d.ai`, and runs failed in a way that took several rounds of debugging to attribute to the merge — the obvious explanation (\"chart RBAC bug\") was correct on its own merits but masked the second bug below it.

This is the kind of bug that compounds: every future contributor who tries to add a Kubernetes manifest as a hand-edit to a scenario will hit it, and the failure mode (silent fold) is hostile to debugging.

kalantar:
Closes #278.

## Problem

`deep_merge` corrupted hand-authored Kubernetes manifests in `extraObjects:`. When a scenario file and a generator-produced overlay both defined `extraObjects`, the list merge fell through to a **positional, index-by-index** fold: `deep_merge(base[i], overlay[i])`. Because K8s manifests carry their name at `metadata.name` (nested) rather than a top-level `name`, the named-key detector returned `None` and the positional tier fired — splicing a `Role`'s `rules`/`subjects`/`roleRef` onto an overlay's `InferenceObjective` document and dropping the hand-authored objects entirely.

This was silent (valid YAML, Helm accepts free-form `extraObjects`) and affected baseline, control, and treatment alike, since all three flow through `deep_merge(scenario, overlay)` in `assemble_packages`. Real-world impact: hand-edited RBAC for the EPP service account never reached the cluster → 403s listing `inferenceobjectives.llm-d.ai` → failed runs.

## Fix (option B — merge by Kubernetes identity)

Added a new tier to `_merge_lists` in `pipeline/lib/values.py`. When **every** entry in both lists is a Kubernetes manifest (has `apiVersion`, `kind`, and `metadata.name`), the lists merge by the composite identity `(apiVersion, kind, metadata.name)`:

- **Distinct identities are all preserved** — a base `Role` and an overlay `InferenceObjective` never fold together.
- **Same-identity manifests deep-merge** — an overlay entry can still patch the matching base manifest (overlay wins on scalar conflicts).

The existing named-key and positional tiers were refactored to share one `_merge_by_keyfn` helper. `deep_merge` and `_LIST_KEY_CANDIDATES` are unchanged.

## Why this can't regress typed lists

The new tier is checked first but is gated on `apiVersion` + `kind`, which typed config lists (`containers`, `env`, ports, the top-level `scenario:` list) never carry — so those keep merging by `name`/`containerPort` exactly as before. The two gates are mutually exclusive in practice.

## Tests

- `pipeline/tests/test_values.py` — unit tests for the new tier: distinct-identity preservation, same-identity merge, base-first ordering, the `metadata.name`-missing fallback to positional, and a containers no-regression check; plus a `TestK8sIdentity` class for the helper.
- `pipeline/tests/test_assemble.py` — end-to-end regression reproducing the exact #278 scenario through `assemble_scenarios` (hand-authored Role + RoleBinding survive an overlay of two InferenceObjectives). Confirmed it **fails** against pre-fix `values.py` and passes after.

CI commands pass locally: `ruff check pipeline/ .claude/skills/ --select F` clean; `pytest` → 927 passed.

## Reviewer attention

- **New same-identity merge behavior**: if a scenario and overlay both define a manifest with the same `(apiVersion, kind, metadata.name)`, they now deep-merge into one object (intended per option B). Previously this case produced corruption.
- **Deferred (noted in commit, not done here)**: the issue's optional Tier-3 divergence warning. With this tier in place the reported corruption never reaches the positional tier; the only residual is a K8s-object list where an entry omits `metadata.name` (e.g. `generateName`, `List` kinds), which still folds positionally. `values.py` has no logging today, so a warning is a separate change. Option C (configurable always-append field list) was skipped as YAGNI.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -360,9 +360,11 @@ Where `baseline_bundle` is the experiment's `baseline.yaml`, `treatment_diffs` i
   through untouched (base first, then overlay), never folded. A duplicate identity
   within either list raises `ValueError`
 - Lists of dicts with a common top-level `name` field merge by name
-- Lists of dicts without a common key merge positionally (a positional fold of two
-  entries with conflicting `apiVersion`/`kind` markers — a malformed manifest — raises
-  `ValueError` rather than silently smearing them together)
+- Lists of dicts without a common key merge positionally. If the two entries at a
+  position carry `apiVersion`/`kind` markers that differ and are not both absent — e.g.
+  a malformed manifest missing one of them — the fold raises `ValueError` instead of
+  silently smearing them. (Two malformed manifests with identical markers still fold;
+  malformed manifests are out of scope — kubectl/Helm reject them.)
 - Lists of scalars are replaced entirely
 - Treatment overlay only needs the delta from baseline_resolved (shared config propagates automatically)
 
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -352,7 +352,11 @@ Where `baseline_bundle` is the experiment's `baseline.yaml`, `treatment_diffs` i
 ### Deep merge semantics
 
 - Dict keys merge recursively (overlay overrides base)
-- Lists of dicts with a common `name` field merge by name
+- Lists where every entry is a Kubernetes manifest (has `apiVersion`, `kind`, and
+  `metadata.name` — e.g. `extraObjects:`) merge by object identity
+  `(apiVersion, kind, metadata.name)`: distinct manifests are all preserved, and an
+  overlay entry sharing an identity patches the matching base manifest
+- Lists of dicts with a common top-level `name` field merge by name
 - Lists of dicts without a common key merge positionally
 - Lists of scalars are replaced entirely
 - Treatment overlay only needs the delta from baseline_resolved (shared config propagates automatically)
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -352,12 +352,17 @@ Where `baseline_bundle` is the experiment's `baseline.yaml`, `treatment_diffs` i
 ### Deep merge semantics
 
 - Dict keys merge recursively (overlay overrides base)
-- Lists where every entry is a Kubernetes manifest (has `apiVersion`, `kind`, and
-  `metadata.name` — e.g. `extraObjects:`) merge by object identity
-  `(apiVersion, kind, metadata.name)`: distinct manifests are all preserved, and an
-  overlay entry sharing an identity patches the matching base manifest
+- Lists where every entry is a Kubernetes manifest (a dict with `apiVersion` and
+  `kind` — e.g. `extraObjects:`) merge by object identity. Entries with
+  `metadata.name` merge by `(apiVersion, kind, metadata.name)` — distinct manifests
+  are preserved, and a same-identity overlay entry patches the matching base manifest.
+  Entries without `metadata.name` (e.g. `generateName`, `List` kinds) are carried
+  through untouched (base first, then overlay), never folded. A duplicate identity
+  within either list raises `ValueError`
 - Lists of dicts with a common top-level `name` field merge by name
-- Lists of dicts without a common key merge positionally
+- Lists of dicts without a common key merge positionally (a positional fold of two
+  entries with conflicting `apiVersion`/`kind` markers — a malformed manifest — raises
+  `ValueError` rather than silently smearing them together)
 - Lists of scalars are replaced entirely
 - Treatment overlay only needs the delta from baseline_resolved (shared config propagates automatically)
 
--- a/pipeline/lib/values.py
+++ b/pipeline/lib/values.py
@@ -63,12 +63,18 @@ def _is_k8s_manifest(item) -> bool:
 
 
 def _k8s_markers_conflict(a: dict, b: dict) -> bool:
-    """True if a and b carry Kubernetes identity markers (apiVersion/kind) that disagree.
+    """True when a and b's (apiVersion, kind) markers differ and are not both absent.
 
-    Used to refuse a positional fold (Tier 3) of two entries that look like distinct
-    Kubernetes manifests — e.g. a malformed manifest missing apiVersion or kind that
-    failed the Tier 2a all-manifest gate and would otherwise smear two unrelated
-    objects together. Returns False for plain dicts carrying neither marker.
+    Fires whenever the two (apiVersion, kind) tuples are unequal — including the
+    one-sided case where one entry carries a marker and the other carries none. Used by
+    the Tier 3 positional merge to refuse folding two entries that look like distinct
+    Kubernetes manifests (e.g. a malformed manifest missing apiVersion or kind that
+    escaped the Tier 2a gate) rather than silently smearing unrelated objects together.
+
+    Known limitation: two malformed manifests with *identical* markers (e.g. both
+    missing apiVersion, same kind) are not a conflict here and still fold positionally.
+    Malformed-manifest handling is out of scope (kubectl/Helm reject them); see #278 and
+    the known-limitation tests in test_values.py.
     """
     a_markers = (a.get("apiVersion"), a.get("kind"))
     b_markers = (b.get("apiVersion"), b.get("kind"))
@@ -130,9 +136,9 @@ def _merge_lists(base_list: list, overlay_list: list) -> list:
              through, never folded. Covers free-form `extraObjects:`.
     Tier 2b: all-dict lists with a common top-level key field → merge by key.
     Tier 3:  all-dict lists without a common key → positional (index-based) merge;
-             refuses (raises) to fold two entries whose Kubernetes identity markers
-             (apiVersion/kind) disagree, which signals a malformed manifest that
-             escaped Tier 2a.
+             refuses (raises ValueError) to fold two entries whose (apiVersion, kind)
+             markers differ and are not both absent — a malformed manifest that escaped
+             Tier 2a. Entries with identical or wholly-absent markers still fold.
 
     Empty overlay → returns [] (explicit clear). Empty base → returns copy of overlay.
     """
--- a/pipeline/lib/values.py
+++ b/pipeline/lib/values.py
@@ -18,12 +18,54 @@ def _detect_list_key(base_list: list, overlay_list: list):
     return None
 
 
-def _merge_lists(base_list: list, overlay_list: list) -> list:
-    """Merge two lists using three-tier strategy.
+def _k8s_identity(item):
+    """Return a Kubernetes identity tuple (apiVersion, kind, metadata.name), or None.
 
-    Tier 1: either list contains non-dict items → overlay replaces base.
-    Tier 2: all-dict lists with a common key field → merge by key.
-    Tier 3: all-dict lists without a common key → positional (index-based) merge.
+    Used to merge free-form `extraObjects:`-style manifest lists by object identity
+    rather than by list position. Returns None for anything that is not a dict
+    carrying all three of apiVersion, kind, and metadata.name.
+    """
+    if not isinstance(item, dict):
+        return None
+    meta = item.get("metadata")
+    if not isinstance(meta, dict) or "name" not in meta:
+        return None
+    if "apiVersion" not in item or "kind" not in item:
+        return None
+    return (item["apiVersion"], item["kind"], meta["name"])
+
+
+def _merge_by_keyfn(base_list: list, overlay_list: list, keyfn) -> list:
+    """Merge two all-dict lists by a key extractor.
+
+    Base entries are emitted first (deep-merged with the same-keyed overlay entry
+    when present); overlay-only entries are appended in order.
+    """
+    overlay_by_key = {keyfn(item): item for item in overlay_list}
+    result = []
+    seen_keys = set()
+    for bitem in base_list:
+        k = keyfn(bitem)
+        seen_keys.add(k)
+        if k in overlay_by_key:
+            result.append(deep_merge(bitem, overlay_by_key[k]))
+        else:
+            result.append(copy.deepcopy(bitem))
+    for oitem in overlay_list:
+        if keyfn(oitem) not in seen_keys:
+            result.append(copy.deepcopy(oitem))
+    return result
+
+
+def _merge_lists(base_list: list, overlay_list: list) -> list:
+    """Merge two lists using a tiered strategy.
+
+    Tier 1:  either list contains non-dict items → overlay replaces base.
+    Tier 2a: all entries are Kubernetes manifests → merge by (apiVersion, kind,
+             metadata.name) identity. Keeps distinct manifests distinct and lets an
+             overlay patch a same-identity manifest. Covers free-form `extraObjects:`.
+    Tier 2b: all-dict lists with a common top-level key field → merge by key.
+    Tier 3:  all-dict lists without a common key → positional (index-based) merge.
 
     Empty overlay → returns [] (explicit clear). Empty base → returns copy of overlay.
     """
@@ -37,23 +79,14 @@ def _merge_lists(base_list: list, overlay_list: list) -> list:
             and all(isinstance(x, dict) for x in overlay_list)):
         return copy.deepcopy(overlay_list)
 
-    # Tier 2: named-key merge
+    # Tier 2a: Kubernetes-object identity merge (apiVersion, kind, metadata.name)
+    if all(_k8s_identity(x) is not None for x in base_list + overlay_list):
+        return _merge_by_keyfn(base_list, overlay_list, _k8s_identity)
+
+    # Tier 2b: named-key merge
     key_field = _detect_list_key(base_list, overlay_list)
     if key_field is not None:
-        overlay_by_key = {item[key_field]: item for item in overlay_list}
-        result = []
-        seen_keys = set()
-        for bitem in base_list:
-            k = bitem[key_field]
-            seen_keys.add(k)
-            if k in overlay_by_key:
-                result.append(deep_merge(bitem, overlay_by_key[k]))
-            else:
-                result.append(copy.deepcopy(bitem))
-        for oitem in overlay_list:
-            if oitem[key_field] not in seen_keys:
-                result.append(copy.deepcopy(oitem))
-        return result
+        return _merge_by_keyfn(base_list, overlay_list, lambda d: d[key_field])
 
     # Tier 3: positional merge — surplus from either side preserved
     result = []
--- a/pipeline/lib/values.py
+++ b/pipeline/lib/values.py
@@ -2,7 +2,7 @@
 import copy
 
 
-# ── Three-tier list merge strategy ───────────────────────────────────────────
+# ── Tiered list merge strategy ────────────────────────────────────────────────
 
 _LIST_KEY_CANDIDATES = ("name", "mountPath", "containerPort")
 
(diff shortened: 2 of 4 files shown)


kalantar:

The core fix is sound. Tier 2a merges `extraObjects` by Kubernetes identity `(apiVersion, kind, metadata.name)` exactly as issue #278's preferred option B describes — distinct manifests are preserved, same-identity manifests deep-merge, and typed config lists (`containers`, `env`, ports) are provably untouched. The `_merge_by_keyfn` extraction is a faithful refactor of the old Tier 2 loop, and the end-to-end test genuinely reproduces #278 through `assemble_scenarios`. For the normal path — hand-authored RBAC plus a generator overlay of InferenceObjectives, all carrying full identity — the bug is fixed and well covered.

The reason for NEEDS CHANGES is that the fix's central guarantee ("extraObjects never positionally fold") quietly breaks on inputs the issue specifically warned about, and the PR defers the guard the issue asked for. These are the same class of silent corruption #278 was filed against, so they matter more here than they would elsewhere.

### Should fix before merge

1. **Partial / asymmetric list silently re-introduces #278.** The Tier 2a gate is `all(...)` over `base_list + overlay_list`. A single entry missing `metadata.name` — most plausibly a manifest using `metadata.generateName`, a `List` kind, or one malformed entry from the translate skill — disqualifies the *whole* list, which then falls through to the Tier 3 positional fold. Verified: a base of `[Role(name=epp), RoleBinding(generateName=epp-)]` against an overlay of two InferenceObjectives produced 2 entries instead of 4, with `Role.rules` smeared onto an InferenceObjective and both RBAC objects gone — byte-for-byte the original corruption. The PR's own `test_k8s_identity_not_triggered_without_metadata_name` documents this fall-through as intended, but only exercises the benign single-entry case where the fold happens to be lossless. Recommended: partition entries per-item (merge those with identity by identity, append the rest), or at minimum raise/warn when a list is K8s-shaped but an entry lacks a complete identity. This is the divergence warning the issue called out as optional — given the above, it's load-bearing, not cosmetic.

2. **Duplicate identity silently loses data.** In `_merge_by_keyfn`, `overlay_by_key = {keyfn(item): item ...}` drops all but the last overlay entry sharing an identity; duplicate base entries both merge against the same overlay entry and get clobbered to an identical result. Duplicate `(apiVersion, kind, name)` is an invalid manifest set, so raising a `ValueError` naming the colliding identity is the right call rather than silently merging. (Note the overlay-side drop is also latent in the pre-existing name-merge tier, so this isn't strictly a regression — but identity-merge is a good place to make it loud.)

### Quick wins (same PR)

3. **Stale comments left by the change:**
   - `values.py:5` — `# ── Three-tier list merge strategy ──` is now four-tier; the rewritten `_merge_lists` docstring already says "a tiered strategy." Update the banner to match.
   - `values.py` `deep_merge` docstring — "Lists of dicts are merged by named key or positional index" omits the new identity tier (which runs first). Add it.
4. **`TestK8sIdentity` covers missing `kind`, missing `metadata.name`, non-dict metadata, non-dict item — but not missing `apiVersion`**, which is its own branch. One-line addition.

### Worth adding

5. **Treatment arm isn't tested end-to-end.** The only `assemble_scenarios` test uses the baseline arm with `treatment_path=None`. The treatment path is a double merge (`deep_merge(deep_merge(baseline_resolved, treatment_diffs), treatment_overlay)`), so extraObjects merge twice — a distinct path. Add a test asserting hand-authored RBAC survives the treatment double-merge.

The deferral of the divergence warning is reasonable in isolation, but combined with item 1 it leaves the headline guarantee false for any list containing one nameless manifest. Closing items 1 and 2 (or, at the team's discretion, item 1 as a guard now and a tracked follow-up for the rest) would move this to ready.

kalantar:
Thanks for the review — all five points are addressed in 19b2051.

The big one was the partial/asymmetric list. You're right that gating Tier 2a on every entry having a full identity meant a single `generateName` (or `List` kind) manifest disqualified the whole list and dropped it back into the positional fold — the exact corruption #278 is about. I reproduced it (2 entries instead of 4, `Role.rules` smeared onto an InferenceObjective) before fixing.

The fix changes the trigger and the merge:
- Tier 2a now fires when every entry is *manifest-shaped* (has `apiVersion` and `kind`), regardless of `metadata.name`. Typed lists like `containers`/`env`/ports never carry those fields, so they're still untouched.
- Within that tier, entries are partitioned per-item: manifests with a full identity merge by `(apiVersion, kind, metadata.name)`; manifests without `metadata.name` are carried through as their own objects — base first, then overlay — and never folded. So the headline "extraObjects never positionally fold" now actually holds even for mixed lists.

For duplicate identity, the merge now raises `ValueError` naming the colliding identity rather than silently keeping only the last entry. I left the pre-existing name-merge tier alone, since as you noted that drop is latent there and not a regression.

Also done: refreshed the stale "three-tier" banner and the `deep_merge` docstring to mention the identity tier; added the missing `apiVersion` case to the `_k8s_identity` tests; and added a treatment-arm end-to-end test that runs hand-authored RBAC through the double merge (baseline scenario + baseline overlay, then treatment overlay) and confirms all four objects survive intact.

Full CI green locally: ruff clean, 932 tests pass.

kalantar:
## Round-2 review: NEEDS CHANGES (close — strong progress)

Commit `19b2051bf` resolves everything from the previous review:

- **Partial/nameless lists no longer fold.** The gate broadened to `_is_k8s_manifest` (apiVersion + kind) and `_merge_k8s_objects` now carries nameless manifests (`generateName`, `List` kinds) through untouched instead of positionally folding them. Verified.
- **Duplicate identity is now loud**, raising `ValueError` for both base and overlay lists. Confirmed it propagates uncaught through `deep_merge` → `assemble.py` → `prepare.py` (the `except AssemblyError` there doesn't swallow it), and correctly does *not* fire on a legitimate base+overlay same-identity patch.
- Stale comments fixed, missing-`apiVersion` unit test added, treatment double-merge test added. The new tests assert behavior (object counts + no field smearing), not just structure.

The `_merge_k8s_objects` implementation was traced exhaustively and is correct — no dropped or double-emitted entries, ordering is base-then-overlay.

Two items remain before merge.

### 1. Residual #278 fold for a partial manifest (missing `apiVersion` or `kind`)

`_is_k8s_manifest` requires **both** `apiVersion` and `kind`. If a single entry has one but not the other, the `all(...)` gate fails for the whole list, it skips the safe identity merge, and drops to the Tier 3 positional fold — the exact #278 corruption. Reproduced:

```
base    = [Role(metadata.name=epp, rules=[...]) WITHOUT apiVersion]
overlay = [InferenceObjective(name=critical, spec=...)]
=> [{kind: InferenceObjective, rules: [...], spec: ...}]   # Role.rules smeared onto the objective
```

This is narrower than the round-1 case — it needs a *malformed* manifest (every valid manifest has `apiVersion`), not a valid one — so the normal path is genuinely fixed. But it's the same silent, debugging-hostile failure mode the issue is about, and it's the one residual the issue's own optional "divergence guard" recommendation was meant to catch. Preferred fix closes both at once: treat a dict carrying `apiVersion` *or* `kind` as manifest-shaped, route it to `_merge_k8s_objects`, and raise `ValueError` there if an entry has one but not both (a partial manifest is an authoring error — make it loud, not folded). Alternatively, if you'd rather accept this as a known limitation, pin it with a test + comment and file a follow-up — but silently folding it back into #278 shouldn't be the default.

### 2. README entry-condition is inaccurate (`pipeline/README.md:355-358`)

The bullet says the tier applies to lists where every entry "has `apiVersion`, `kind`, **and** `metadata.name`." The actual gate requires only `apiVersion` + `kind`; `metadata.name` is the *keying* field used inside the merge, not the entry condition — entries without it are carried through. As written, the README implies a `generateName` manifest list wouldn't qualify for this tier, which is exactly the fold this PR fixes. It also omits the nameless carry-through and the duplicate-identity `ValueError`. Suggested rewrite:

> - Lists where every entry is a Kubernetes manifest (a dict with `apiVersion` and `kind` — e.g. `extraObjects:`) merge by object identity. Entries with `metadata.name` merge by `(apiVersion, kind, metadata.name)` — distinct manifests preserved, a same-identity overlay entry patches the matching base manifest. Entries without `metadata.name` are carried through untouched (base first, then overlay), never folded. A duplicate identity within either list raises `ValueError`.

### Worth adding (not blocking)

- A combined test: one list mixing a same-identity base+overlay patch **and** a nameless carry-through, asserting it merges without raising — pins the new `ValueError` against a future regression on the legitimate-patch path.
- Overlay-side nameless carry-through (current nameless tests put `generateName` in the base).
- An end-to-end `assemble_scenarios` test confirming the duplicate-identity `ValueError` reaches the operator.

Item 1 is the only real call to make; item 2 is a one-paragraph doc fix. Address those (or downgrade item 1 to a tracked limitation with a pinning test) and this is ready.

kalantar:
Both blocking items are fixed in a034d44, plus the suggested extra tests.

On item 1 (the residual fold for a malformed manifest): I went a different route than the OR-gate you suggested, because broadening `_is_k8s_manifest` to `apiVersion` *or* `kind` has a side effect — RoleBinding `subjects` entries are `kind: ServiceAccount` with no `apiVersion`, so an all-`subjects` sublist would pass the broadened gate, get routed into the identity merge, and raise `ValueError` on every legitimate same-identity RoleBinding patch (the path you confirmed working in round 2).

Instead I kept the gate at `apiVersion` AND `kind` and added the divergence guard the issue recommended: the malformed manifest fails the Tier 2a gate, falls to the Tier 3 positional merge, and Tier 3 now refuses to fold two entries whose `apiVersion`/`kind` markers disagree — it raises instead of smearing. `subjects` merge by `name` (Tier 2b) and never reach Tier 3, so the patch path is untouched. I added a test for that exact case (`subjects` merge without raising) so it can't regress. Verified your repro now raises rather than producing the smeared object.

Item 2: corrected the README. The entry condition is `apiVersion` + `kind` (not `+ metadata.name`); `metadata.name` is the keying field used inside the merge. The bullets now also describe the nameless carry-through and the duplicate-identity / divergence `ValueError`s.

Also added the three tests you flagged as worth having: a combined same-identity-patch-plus-nameless-sibling case that asserts it merges without raising (pins the legitimate path against the new `ValueError`), an overlay-side nameless carry-through (the earlier nameless tests only put `generateName` in the base), and an end-to-end `assemble_scenarios` test confirming the duplicate-identity error actually surfaces to the operator.

ruff clean, 937 tests pass.

kalantar:
## Round-3 review: NEEDS DISCUSSION

First, the good news: the core fix is solid and the test suite is green (50/50). Every realistic case is handled correctly — valid manifests merge by identity, `generateName`/nameless manifests carry through, duplicate identities raise, and the new commit `a034d4475` correctly closes the round-2 residual (a manifest missing one of `apiVersion`/`kind` now raises instead of folding). The README rewrite is accurate. For the bug as filed and any well-formed input, this is done.

The remaining items are all about **malformed manifests** (entries missing `apiVersion` or `kind`), and they surface a design tension worth settling before merge rather than a simple fix.

### The tension

The Tier 3 guard refuses to fold when two entries' `(apiVersion, kind)` markers *differ*. That's necessary but not sufficient, and it cuts both ways — both reproduced by direct execution:

**It still folds silently when markers *match* (same class as #278).** Two manifests both omitting `apiVersion`, same `kind`, different names:
```
base    = [{kind: Role, metadata:{name: alpha}, rules: [...]}]
overlay = [{kind: Role, metadata:{name: beta},  rules: [...]}]
=> single Role {name: beta}  — "alpha" silently lost, no error
```
Both fail the Tier 2a gate (no `apiVersion`), `metadata.name` is nested so Tier 2b can't key them, Tier 3 sees matching markers `(None,'Role')` and folds. Same silent-data-loss failure mode the PR exists to kill, just triggered by symmetric rather than asymmetric malformation.

**It over-fires on nameless marker-bearing sublists.** A `kind`-bearing list without a top-level `name` (e.g. RBAC `subjects` of differing kinds nested inside two same-identity RoleBindings) reaches Tier 3 and raises spuriously — the guard treats a bare `kind` as a manifest marker at any recursion depth.

Both stem from one root: a depth-agnostic guard keyed on marker *divergence* can't tell a top-level manifest from a nested config entry, and can't see `metadata.name`.

### The decision

Pick the contract for malformed manifests; the current code is mid-way between two coherent designs:

1. **Carry-through (cleanest, IMO):** broaden the Tier 2a gate to "carries `apiVersion` *or* `kind`" and let `_merge_k8s_objects` carry through anything it can't fully key. Manifest-shaped lists then never reach Tier 3, which closes both the silent fold (both entries carried through, nothing lost) and the false-positive (the Tier 3 guard no longer runs on sublists). Trade-off: the round-2 case becomes carry-through instead of raise — arguably fine, since neither corrupts.
2. **Raise (current direction), made consistent:** also treat matching markers + differing `metadata.name` as a conflict, and scope the guard so a bare `kind` on a nameless nested entry doesn't trip it. More special-casing.
3. **Out of scope:** decide malformed manifests (missing `apiVersion`/`kind`) are the user's problem — kubectl/Helm will reject them anyway — accept current behavior, and pin it with tests (including a `xfail`/known-limitation test for the silent-fold case) so the boundary is explicit.

Any of the three is defensible. What I'd push back on is shipping the current in-between state silently, because the silent fold is the exact debugging-hostile failure #278 documents.

### Minor (do regardless)

The guard docstring and the README positional bullet understate when it raises — they say "two entries with conflicting markers"/"a malformed manifest," but it fires whenever the `(apiVersion, kind)` tuples differ and aren't both absent, including the one-sided case (one entry carries a marker, the other carries none). Reword to the actual predicate: raises unless both entries' `(apiVersion, kind)` are absent.

To be clear on scope: this is converging — there are no new bugs in the valid-input path, and I don't want to send you into an endless hardening spiral. It's one design call on malformed input plus a doc-wording fix.

kalantar:
Agreed the valid-manifest path is done, and I went with treating malformed manifests (missing `apiVersion` or `kind`) as out of scope — kubectl/Helm reject them anyway. So no behavior change this round; I accepted the current behavior and made the boundary explicit instead of leaving it silent in the code.

I reproduced both of your cases directly first: the symmetric matching-marker fold (two `Role`s both missing `apiVersion`, different names → folds to one, "alpha" lost) and the guard over-firing on a nameless `kind`-bearing nested list. Both now have xfail known-limitation tests that assert the *ideal* behavior and are marked xfail with the reason — so they document the gap and will flip to xpass (flagging us) if the behavior ever changes.

On the wording you flagged: the guard docstring, the `_merge_lists` Tier 3 line, and the README positional bullet all now state the real predicate — the fold raises whenever the two `(apiVersion, kind)` tuples differ and aren't both absent, including the one-sided case — and explicitly note that identical-marker malformed manifests still fold.

937 passed, 2 xfailed, ruff clean.

PR #280, opened 2026-06-04
fix: retain namespace on terminal pair failures so reset can clean helm releases (#277)

### LINKED ISSUE #1: chore: remove llm-d-inference-scheduler submodule from framework repo
## Background

`llm-d-inference-scheduler` is currently registered as a git submodule of the sim2real framework repo. As part of the experiment-repo split, this submodule belongs in each experiment repo (e.g. `admission-control`), not in the framework. The framework should be submodule-free for the scheduler target.

## Dependencies

Four files in the framework repo actively depend on the submodule directory being present at `<REPO_ROOT>/llm-d-inference-scheduler/`:

| File | How it uses the submodule |
|------|--------------------------|
| `pipeline/run.py` | Hardcodes `SUBMODULE_DIR = REPO_ROOT / "llm-d-inference-scheduler"` and passes it to `switch_run()` |
| `pipeline/lib/run_manager.py` | `switch_run()` resets the directory, deletes stale files, and copies generated scorer plugin files into it |
| `pipeline/prepare.py` | `_get_submodule_shas()` reads the submodule's git HEAD to record its SHA in context; `_phase_init()` validates `target.repo` exists |
| `pipeline/setup.py` | Uses `"llm-d-inference-scheduler"` as the default value for `--repo-name` |

Everything else (docs, config files, skills) is reference material — not load-bearing at runtime.

### Step 1 — Remove the git submodule

```bash
git submodule deinit -f llm-d-inference-scheduler
git rm llm-d-inference-scheduler
rm -rf .git/modules/llm-d-inference-scheduler
git commit -m "chore: remove llm-d-inference-scheduler submodule"
```

### Step 2 — Add `--experiment-root` to `run.py`

`run.py` currently hardcodes the three paths that must move to the experiment root:

```python
# Before
WORKSPACE_DIR = REPO_ROOT / "workspace"
SUBMODULE_DIR = REPO_ROOT / "llm-d-inference-scheduler"
SETUP_CONFIG  = WORKSPACE_DIR / "setup_config.json"
```

Add `--experiment-root PATH` (default: `Path.cwd()`) and derive these at runtime:

```python
# After (in main(), before dispatching subcommands)
experiment_root = Path(args.experiment_root).resolve() if args.experiment_root else Path.cwd()
workspace_dir   = experiment_root / "workspace"
submodule_dir   = experiment_root / "llm-d-inference-scheduler"
setup_config    = workspace_dir / "setup_config.json"
```

`run_manager.py` already takes `workspace_dir` and `submodule_dir` as parameters — no changes needed there.

### Step 3 — Update `prepare.py` `_get_submodule_shas()`

The function currently looks for both `inference-sim` and `llm-d-inference-scheduler` under `REPO_ROOT`. After the split, `inference-sim` stays in the framework, but `llm-d-inference-scheduler` lives in the experiment root.

```python
# Before
for name, path in [("inference-sim", "inference-sim"),
                   ("llm-d-inference-scheduler", "llm-d-inference-scheduler")]:
    sub = REPO_ROOT / path

# After
for name, path in [("inference-sim", "inference-sim"),
                   ("llm-d-inference-scheduler", "llm-d-inference-scheduler")]:
    sub = REPO_ROOT / path if name == "inference-sim" else (
        EXPERIMENT_ROOT / path if (EXPERIMENT_ROOT / path).exists()
        else REPO_ROOT / path   # backward compat fallback
    )
```

`EXPERIMENT_ROOT` is the module-level variable already introduced by the `--experiment-root` flag work in `prepare.py`.

### Step 4 — Update `setup.py` default repo name

```python
# Before
repo_default = defaults.get("repo_name", "llm-d-inference-scheduler")

# After — unchanged: the default string is still correct; the directory
# is now resolved relative to EXPERIMENT_ROOT, not REPO_ROOT.
```

No code change needed here, but confirm that wherever `repo_name` is used to build a path it is joined to `EXPERIMENT_ROOT`, not `REPO_ROOT`.

### Step 5 — Update CLAUDE.md submodules list

Remove `llm-d-inference-scheduler/` from the Submodules section. Add a note that experiment repos carry their own copy of this submodule.

### Step 6 — Verify

```bash
# No runtime reference to the submodule in framework pipeline code
grep -rn "llm-d-inference-scheduler" pipeline/ --include="*.py" \
  | grep -v "# " | grep -v "\"llm-d-inference-scheduler\""  # string literals for default/display are OK

# Tests pass
python -m pytest pipeline/ -v

# Submodule is gone
git submodule status  # should list only inference-sim and tektonc-data-collection
```

## Acceptance Criteria

- [ ] `git submodule status` no longer lists `llm-d-inference-scheduler`
- [ ] `python pipeline/run.py switch <name> --experiment-root <path>` works with the submodule in the experiment repo
- [ ] `python pipeline/prepare.py --experiment-root <path>` correctly records the scheduler submodule SHA from the experiment repo
- [ ] All pipeline tests pass
- [ ] CLAUDE.md submodules section updated

### LINKED ISSUE #2: Add install-blis task to install BLIS at pipeline runtime
## Summary

Create an \`install-blis\` Tekton task that builds the BLIS binary from the \`inference-sim\` submodule at pipeline runtime into the shared \`source\` workspace, eliminating the need for a pre-built \`blisImage\` container. Rename the existing \`run-workload-blis-observe\` task to \`run-workload-blis-observe-image\` (preserved for image-based use), and add a new \`run-workload-blis-observe\` that invokes the workspace-installed binary.

## Motivation

Currently, \`run-workload-blis-observe\` requires a pre-built container image (\`blisImage\`, sourced from \`observe.image\`). This image is constructed in \`prepare.py\` from \`git describe --tags\` on the \`inference-sim\` submodule — but whether that image actually exists in the registry is a separate concern, making the release hard to track. Building the binary directly from the pinned submodule commit at pipeline runtime:

- Eliminates the dependency on a pre-published BLIS container image
- Makes the exact version explicit and auditable (submodule commit → binary)
- Follows the \`install-inference-perf\` pattern already established for the \`source\` workspace

## Scope

- **\`tektonc-data-collection\` submodule** — new and renamed Tekton task YAML files under \`tekton/tasks/\`
- **\`inference-sim/sim2real\`** (this repo) — pipeline template, \`prepare.py\`, \`setup.py\`, \`deploy.py\`

### 1. Create \`install-blis\` task

Add \`tektonc-data-collection/tekton/tasks/install-blis.yaml\`:

- Workspace: \`source\` (consistent with \`install-inference-perf\`)
- Base image: \`golang:1.22\` (BLIS is a Go binary — \`pip install\` does not apply)
- Clone \`inference-sim\` at \`GIT_COMMIT\`, run \`go build -o blis .\`, write binary to \`$(workspaces.source.path)/blis/blis\`
- **Idempotency**: write commit SHA to \`$(workspaces.source.path)/blis/.commit\`; skip rebuild if marker exists, commit matches, and binary is executable
- Config via env vars: \`GIT_REPO_URL\`, \`GIT_COMMIT\` (consistent with \`install-inference-perf\`)

### 2. Rename and replace \`run-workload-blis-observe\`

- Rename \`run-workload-blis-observe.yaml\` → \`run-workload-blis-observe-image.yaml\` (task name: \`run-workload-blis-observe-image\`); no functional changes
- Add new \`run-workload-blis-observe.yaml\`: removes \`blisImage\` param; \`run-observe\` step uses \`debian:bookworm-slim\` and invokes \`$(workspaces.source.path)/blis/blis observe ...\`; adds \`source\` workspace

### 3. Update pipeline template

Update \`pipeline/templates/pipeline.yaml.j2\`:

- Add \`install-blis\` task with no \`runAfter\` (runs early, parallel to \`download-model\` / \`deploy-gateway\` / \`deploy-gaie\`)
- Add \`install-blis\` to \`run-workload\` \`runAfter\` list
- Remove \`blisImage\` param and \`{{ observe.image }}\` value; add \`git_commit\` param with value \`{{ observe.blis_commit }}\`
- Add \`source\` to pipeline workspaces list

### 4. Wire \`source\` workspace

- **\`setup.py\`**: add \`"source": {"persistentVolumeClaim": {"claimName": "source-pvc"}}\` to the \`setup_config.json\` workspaces block (the \`source-pvc\` PVC is already created in step 6; only the mapping is missing)
- **\`deploy.py\`**: add \`source-pvc\` to the pre-flight PVC bound check

### 5. Update \`prepare.py\`

- Replace \`git describe --tags\` + \`observe["image"]\` with \`git rev-parse HEAD\` + \`observe["blis_commit"]\`
- Remove \`observe.image\` injection; no change to \`config/env_defaults.yaml\` (the field was never there — it was injected at runtime)

## Resolved

- **BLIS is a Go binary** (\`main.go\` + \`go.mod\` in \`inference-sim/\`); \`pip install --target\` does not apply
- **Workspace**: \`source\` / \`source-pvc\` (consistent with \`install-inference-perf\`; \`source-pvc\` already created by \`setup.py\` step 6, just not wired)
- **Idempotency**: commit marker file \`$(workspaces.source.path)/blis/.commit\` — equivalent to pip's version check
- **\`blisImage\` fallback**: not needed; existing image-based task preserved as \`run-workload-blis-observe-image\`
- **Pipeline ordering**: \`install-blis\` is an early parallel task; \`run-workload\` gains it in \`runAfter\`

## References

- \`tektonc-data-collection/tekton/tasks/install-inference-perf.yaml\` — closest reference pattern (\`source\` workspace, env-var config, venv idempotency check)
- \`tektonc-data-collection/tekton/tasks/run-workload-blis-observe.yaml\` — task to be renamed
- \`pipeline/templates/pipeline.yaml.j2\` — pipeline template to be updated
- \`pipeline/prepare.py\` lines 437–448 — current \`observe.image\` injection to be replaced

---
*Migrated from [kalantar/sim2real#37](https://github.com/kalantar/sim2real/issues/37)*

### LINKED ISSUE #277: deploy.py reset/run --force leaves orphan helm releases when pair manifest has no namespace recorded
## Summary

`deploy.py reset` (and `run --force`, which shares the same code path) silently fails to clean up Helm releases whenever the pair entry in the progress ConfigMap has `namespace: null`. The pair status flips to `pending` and the operator gets `1 pair(s) reset`, but the actual cluster state — Helm releases, Deployments, Pods, InferencePool — is left in place. Reset's stated purpose is to bring cluster state back in line with the manifest; in this failure mode it does the opposite (manifest agrees nothing is running; cluster still has it).

This is the third in a series of "operator sees a misleading report" bugs in `deploy.py`, alongside #275 (build pod waits 30 min on Failed) and #276 (orchestrator pod failure prints `unknown reason` instead of pod logs). Common theme: the script exits cleanly while the underlying state is wrong, leaving the operator to discover the discrepancy via `helm list` / `kubectl get`.

## Repro

1. Start a fresh run: `python pipeline/deploy.py run --remote ...`.
2. Allow the orchestrator to claim a namespace and `helm install` the three releases (`infra-llmdbench`, `<model>-gaie`, `<model>-ms`).
3. Cause the run to fail mid-flight (any reason — EPP CrashLoop on RBAC works, as does an orchestrator argparse error if you have an invalid `--workload` value). The PipelineRun gets cleaned up; the pair entry's `namespace` field is cleared (or never written, depending on which step failed).
4. Run `python pipeline/deploy.py reset` (or `run --force`).

Output:
```
Resetting wl-balanced-20-baseline (status: failed, ns: —) — deleting PipelineRun, uninstalling helm releases
1 pair(s) reset
```

Reality:
```
$ helm list -n kalantar-0
infra-llmdbench                   ... llm-d-infra-v1.4.0          v0.4.0
qwen-qwe-079d55ee-wen3-14b-gaie   ... inferencepool-v1.5.0        v1.5.0
qwen-qwe-079d55ee-wen3-14b-ms    ... llm-d-modelservice-v0.4.12  v0.4.0
$ kubectl get pods -n kalantar-0
... 4 decode pods, 1 EPP in CrashLoop, 1 inference-gateway, 1 download Job
```

The pair manifest now claims `pending`, so subsequent `run --remote` re-installs over the existing releases (which produces a different class of failure), or the operator has to manually `helm uninstall` and clean up.

## Where

`pipeline/deploy.py:1535-1538` (the helm-uninstall step inside `_reset_pair`):

```python
# Discover and uninstall all Helm releases in the namespace
if ns:
    result = run([\"helm\", \"list\", \"-n\", ns, \"-q\"], check=False, capture=True)
    ...
```

The `if ns:` guard means: only enumerate when `entry[\"namespace\"]` is set. If it is null, the helm step is a no-op. The orphan-check fallback (`_uninstall_orphaned_helm` on `entry[\"completed_namespace\"]` at `:1519-1522`) only fires for `is_done == True`, so failed pairs without a recorded namespace fall through to a state-only reset.

The earlier action banner at `:1455-1462` does NOT consult `ns` and prints \"deleting PipelineRun, uninstalling helm releases\" anyway, so the operator gets a confident-sounding line that doesn't match what happened.

`run --force` reaches the same code via `_force_reset` at `:1563-1580`, which just iterates `_reset_pair`. So this bug applies to `--force` too — they are not separate code paths.

## Why the manifest goes out of sync

The pair entry's `namespace` field is written by the orchestrator only on terminal pair resolution, not at namespace-claim / helm-install time. Combine that with reset clearing `entry[\"namespace\"] = None` (`:1474, :1525`), and you reach a state where the cluster has installed releases but the manifest no longer knows which namespace they're in. Once that happens, no current code path will find them again.

## Suggested fixes

Either (or both):

1. **Write `entry[\"namespace\"]` at namespace-claim / pre-install time**, not only on success. Then any subsequent reset has a target to enumerate, regardless of how the run failed. This is the structural fix.

2. **Best-effort orphan sweep when `entry[\"namespace\"]` is null and status is non-pending.** Enumerate the configured namespace slots (`namespaces` arg threaded through `_cmd_reset`) and `helm list` each, looking for releases whose labels include the pair key (or matching by chart name + release naming convention). Uninstall matches. This covers historical state where #1 wasn't yet in effect.

The action-banner cosmetic bug at `:1455-1462` is a separate, small correctness fix — gate the \"uninstalling helm releases\" wording on `ns` actually being set.

## Severity

This breaks the fundamental contract of `reset`. Operators can't rely on it after partial failures, which is exactly when they need it to work. Today, the safe workaround is to manually `helm list -n <ns>` and `helm uninstall` after every non-trivial failure.

## Related

- #275 — `build-epp.sh` waits 30 min on Failed pods (also a misleading-state bug, in the build script)
- #276 — orchestrator pod failure prints `unknown reason` instead of pod logs (also a misleading-state bug)

The shared theme across these three is worth calling out separately if there's appetite: failures in pipeline scripts should propagate enough information for an operator to act, and success messages must reflect what actually happened.

kalantar:
Closes #277

## Problem

`deploy.py reset` (and `run --force`, same code path) silently fails to clean up Helm releases whenever a failed pair's progress-ConfigMap entry has `namespace: null`. The pair flips to `pending` and the operator sees `1 pair(s) reset`, but the Helm releases / Deployments / Pods / InferencePool are left running — the opposite of reset's contract. The operator then has to manually `helm list` / `helm uninstall`.

## Root cause (corrected from the issue)

The issue's suggested fix #1 ("write `namespace` at claim time") is **already implemented** — the dispatch loop sets `entry["namespace"] = ns` the moment a pair goes `running` (`deploy.py:2322`). The real defect is the **failure paths discarding it**: the live `_cmd_run` orchestrator cleared `entry["namespace"] = None` on every terminal outcome. `_reconcile_on_resume` was already fixed to *retain* namespace on failure (`:1780`, "Retain namespace so reset/cleanup can find the resources"), but that fix was never propagated to the live loop.

## Fix

Retain `namespace` on **all terminal pair states** so the existing `_reset_pair` `if ns:` Helm cleanup can find and uninstall the releases. Five paths cleared it; all are now consistent:

| Path | Terminal status | Before | After |
|------|----------------|--------|-------|
| `_cmd_run` hard failure (`:2086`) | `failed` | cleared | **retained** |
| `_cmd_run` pod-health escalation (`:2168`) | `failed` | cleared | **retained** |
| `_handle_pending_pods` non-recoverable (`:571`) | `failed` | cleared | **retained** |
| `_handle_pending_pods` max stalls (`:601`) | `stalled` | cleared | **retained** |
| `_handle_timeout` max retries (`:644`) | `timed-out` | cleared | **retained** |

The two genuinely-recoverable branches that re-dispatch to `pending` still release the namespace (the slot is freed and the pair reinstalls on the next dispatch):
- recoverable stall under threshold → `pending`
- timeout with retries remaining → `pending`

## Scope notes for reviewers

- **Beyond the issue's two named paths.** The issue named only the two inline `_cmd_run` sites. Deep inspection found three sibling terminal paths in the `_handle_pending_pods` / `_handle_timeout` helpers with the identical orphan bug; including them keeps the fix consistent (reset is meant to clean *any* non-pending pair). Confirmed with the maintainer before extending.
- **Out of scope (deliberately):** the issue's fix #2 (best-effort orphan sweep for already-broken null-namespace entries) and the cosmetic action-banner wording. This PR is the structural fix only. Pre-existing entries already written with `namespace: null` still need a one-time manual `helm uninstall`.
- **No false busy slots:** the startup `slots_busy` map keys only off `status == "running"` (`:2030`), so retaining namespace on terminal pairs does not occupy a slot or affect dispatch.

## Tests

- New: `test_cmd_run_hard_failure_retains_namespace`, `test_cmd_run_pod_health_escalation_retains_namespace`, `test_handle_timeout_max_retries_retains_namespace`.
- Updated to assert retention: `test_early_reclaim_non_recoverable_fails_immediately`, `test_early_reclaim_stalled_at_max_pending_stalls`.
- Preserved (re-dispatch clears): `test_early_reclaim_recoverable_threshold_exceeded`, `test_handle_timeout_cancel_succeeds_requeues`.

864 passed, 2 xfailed; `ruff check pipeline/ .claude/skills/ --select F` clean.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -523,9 +523,11 @@ def _handle_pending_pods(*, pr_name: str, namespace: str, entry: dict,
 
     Side effects on *entry* (caller must persist):
       On True (non-recoverable):
-        - status: "failed", namespace: None, pending_since: None
+        - status: "failed", pending_since: None
+        - namespace retained so reset/cleanup can find releases (issue #277)
       On True (recoverable threshold exceeded):
-        - status: "pending" or "stalled", namespace: None
+        - status: "pending" (re-dispatch — namespace cleared) or
+          "stalled" (terminal — namespace retained, issue #277)
         - pending_stalls: incremented, pending_since: None
       On False (no action):
         - pending_since: set on first recoverable detection, cleared when
@@ -1461,8 +1463,12 @@ def _reset_pair(key: str, entry: dict, discovered: dict, *,
             action = "state-only reset"
         elif is_done:
             action = "deleting PipelineRun, checking orphaned releases"
-        else:
+        elif ns:
             action = "deleting PipelineRun, uninstalling helm releases"
+        else:
+            # ns is null but a PipelineRun is known — helm cleanup needs the
+            # namespace, so it will be skipped (issue #277). Don't claim it.
+            action = "deleting PipelineRun (namespace unknown — skipping helm cleanup)"
         info(f"Resetting {key} (status: {status}, ns: {slot}) — {action}")
 
     # No namespace and no pr_name — just reset state
@@ -1474,6 +1480,12 @@ def _reset_pair(key: str, entry: dict, discovered: dict, *,
                     info(f"[DRY-RUN] {key}: would check for orphaned helm releases in {completed_ns}")
                 else:
                     _uninstall_orphaned_helm(key, completed_ns)
+        elif not dry_run:
+            # Terminal pair (callers only reset non-pending pairs) with no
+            # namespace recorded — helm cleanup is skipped. Warn rather than
+            # silently report success (issue #277).
+            warn(f"{key}: namespace unknown — skipped helm cleanup; if releases "
+                 f"were installed, remove them manually (helm list/uninstall across slots)")
         if not dry_run and not (is_done and preserve_done_status):
             entry["status"] = "pending"
             entry["retries"] = 0
@@ -1554,6 +1566,12 @@ def _reset_pair(key: str, entry: dict, discovered: dict, *,
             if helm_failed:
                 warn(f"{key}: some releases failed to uninstall — state NOT reset")
                 return False
+    else:
+        # PipelineRun was known but no namespace recorded — helm needs the
+        # namespace, so cleanup is skipped. Warn rather than silently report
+        # success (issue #277).
+        warn(f"{key}: namespace unknown — skipped helm cleanup; if releases "
+             f"were installed, remove them manually (helm list/uninstall across slots)")
 
     # Reset state
     entry["status"] = "pending"
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -568,7 +568,7 @@ def _handle_pending_pods(*, pr_name: str, namespace: str, entry: dict,
             warn(f"[{entry.get('workload', '?')}] could not remove PipelineRun {pr_name!r} in {namespace} — slot NOT freed")
             return False
         entry["status"] = "failed"
-        entry["namespace"] = None
+        # Retain namespace so reset/cleanup can find the helm releases (issue #277).
         entry["pending_since"] = None
         return True
 
@@ -596,12 +596,14 @@ def _handle_pending_pods(*, pr_name: str, namespace: str, entry: dict,
     stalls = entry.get("pending_stalls", 0) + 1
     entry["pending_stalls"] = stalls
     entry["pending_since"] = None
-    entry["namespace"] = None
     if stalls >= max_pending_stalls:
         entry["status"] = "stalled"
+        # Terminal: retain namespace so reset/cleanup can find releases (issue #277).
         warn(f"[{entry.get('workload', '?')}] reached max pending stalls ({max_pending_stalls}) → stalled")
     else:
         entry["status"] = "pending"
+        # Slot freed for re-dispatch — release the namespace.
+        entry["namespace"] = None
     return True
 
 
@@ -639,10 +641,12 @@ def _handle_timeout(*, pr_name: str, namespace: str, entry: dict,
              f"(attempt {retries + 1}/{max_retries})")
         entry["status"] = "pending"
         entry["retries"] = retries + 1
+        # Slot freed for re-dispatch — release the namespace.
+        entry["namespace"] = None
     else:
         warn(f"[{entry.get('workload', '?')}] timed out, max retries → timed-out")
         entry["status"] = "timed-out"
-    entry["namespace"] = None
+        # Terminal: retain namespace so reset/cleanup can find releases (issue #277).
     entry["pending_since"] = None
     return True
 
@@ -2083,7 +2087,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                             "PipelineRunStoppingTimeout"):
                 warn(f"[{pair_key}] hard failure ({status}) → failed")
                 entry["status"] = "failed"
-                entry["namespace"] = None
+                # Retain namespace so reset/cleanup can find the helm releases
+                # (issue #277). Mirrors _reconcile_on_resume's failure handling.
                 store.save(progress)
                 del slots_busy[ns]
                 _last_log_state.pop("capacity", None)
@@ -2165,7 +2170,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                     warn(f"[{pair_key}] pod health escalation → cancelling PipelineRun")
                     if _cancel_and_delete_pipelinerun(pr_name, ns):
                         entry["status"] = "failed"
-                        entry["namespace"] = None
+                        # Retain namespace so reset/cleanup can find the helm
+                        # releases (issue #277).
                         store.save(progress)
                         del slots_busy[ns]
                         _last_log_state.pop("capacity", None)
--- a/pipeline/tests/test_deploy_reset.py
+++ b/pipeline/tests/test_deploy_reset.py
@@ -114,6 +114,77 @@ def test_reset_pair_none_namespace_resets_state(monkeypatch):
     assert entry["retries"] == 0
 
 
+def test_timed_out_entry_flows_to_reset_helm_uninstall(monkeypatch):
+    """End-to-end (issue #277): a pair driven to timed-out by _handle_timeout
+    retains its namespace, and feeding that same entry to _reset_pair actually
+    uninstalls the helm releases. Reverting retain-on-failure breaks this — the
+    consumer would see namespace=None and skip helm cleanup."""
+    import datetime as _dt
+    import pipeline.deploy as mod
+
+    old_ts = (_dt.datetime.now(_dt.timezone.utc) - _dt.timedelta(hours=5)).strftime(
+        "%Y-%m-%dT%H:%M:%SZ")
+
+    entry = {"workload": "wl-heavy", "package": "baseline", "status": "running",
+             "namespace": "sim2real-0", "retries": 3, "pending_since": None}
+
+    calls = []
+
+    def fake_run(cmd, *, check=True, capture=False, cwd=None):
+        calls.append(cmd)
+        class _R:
+            returncode = 0
+            stdout = (old_ts if "creationTimestamp" in " ".join(cmd)
+                      else ("release-a\n" if cmd[:2] == ["helm", "list"] else ""))
+        return _R()
+
+    monkeypatch.setattr(mod, "run", fake_run)
+    monkeypatch.setattr(mod, "_cancel_and_delete_pipelinerun", lambda pr, ns: True)
+
+    # Producer: timeout at max retries → timed-out, namespace retained.
+    assert mod._handle_timeout(
+        pr_name="baseline-heavy-run1", namespace="sim2real-0",
+        entry=entry, timeout_hours=4.0, max_retries=3) is True
+    assert entry["status"] == "timed-out"
+    assert entry["namespace"] == "sim2real-0"
+
+    # Consumer: reset uses the retained namespace to actually uninstall helm.
+    calls.clear()
+    assert mod._reset_pair("wl-heavy-baseline", entry, _DISCOVERED) is True
+    helm_uninstalls = [c for c in calls if c[:2] == ["helm", "uninstall"]]
+    assert len(helm_uninstalls) == 1
+    assert "release-a" in helm_uninstalls[0]
+    assert entry["status"] == "pending"
+
+
+def test_reset_pair_null_namespace_skips_helm_and_warns(monkeypatch, capsys):
+    """Negative companion (issue #277): a terminal pair with no namespace cannot
+    run helm cleanup — _reset_pair must skip helm AND warn the operator rather
+    than silently reporting success."""
+    import pipeline.deploy as mod
+
+    entry = {"workload": "wl-load", "package": "baseline", "status": "timed-out",
+             "namespace": None, "retries": 1}
+
+    calls = []
+
+    def fake_run(cmd, *, check=True, capture=False, cwd=None):
+        calls.append(cmd)
+        class _R:
+            returncode = 0
+            stdout = ""
+        return _R()
+
+    monkeypatch.setattr(mod, "run", fake_run)
+
+    result = mod._reset_pair("wl-load-baseline", entry, _DISCOVERED)
+    assert result is True
+    assert [c for c in calls if c[:2] == ["helm", "uninstall"]] == []
+    assert [c for c in calls if c[:2] == ["helm", "list"]] == []
+    out = capsys.readouterr().out.lower()
+    assert "helm" in out and "manual" in out
+
+
 def test_reset_pair_kubectl_delete_failure_does_not_reset(monkeypatch):
     """When kubectl delete pipelinerun fails, state is NOT reset."""
     import pipeline.deploy as mod
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -1434,7 +1434,8 @@ def test_early_reclaim_non_recoverable_fails_immediately(monkeypatch):
 
     assert reclaimed is True
     assert entry["status"] == "failed"
-    assert entry["namespace"] is None
+    # Namespace retained so reset/cleanup can find the helm releases (issue #277)
+    assert entry["namespace"] == "sim2real-0"
     assert entry["pending_stalls"] == 0
     assert cancelled == [("baseline-smoke-run1", "sim2real-0")]
 
@@ -1486,6 +1487,8 @@ def test_early_reclaim_stalled_at_max_pending_stalls(monkeypatch):
     assert reclaimed is True
     assert entry["status"] == "stalled"
     assert entry["pending_stalls"] == 10
+    # Terminal stall retains namespace so reset/cleanup can find releases (issue #277)
+    assert entry["namespace"] == "sim2real-0"
 
 
 def test_early_reclaim_kubectl_failure_returns_false(monkeypatch, capsys):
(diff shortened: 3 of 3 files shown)


kalantar:
## Review: PR #280 vs issue #277 — NEEDS CHANGES (core fix is sound; finish-the-job items remain)

The structural fix is correct and well-reasoned. All terminal failure paths (failed / stalled / timed-out) now retain `entry["namespace"]`; the two genuinely-recoverable re-dispatch→pending branches still release it; the done paths correctly clear `namespace` only after setting `completed_namespace`. Verified there are no false busy slots (`slots_busy` keys off `status == "running"`), the live loop is now consistent with `_reconcile_on_resume`, and the "Mirrors _reconcile_on_resume" comment is genuinely accurate. The "cleared mid-flight" repro is fully closed. Five inline edits across three functions read clearly; a shared helper would reduce clarity, so keeping them inline is the right call.

Three things to address before merge, then a couple of nice-to-haves.

**1. Stale docstring (must fix — it's in the function you changed).** `_handle_pending_pods`'s "Side effects" block (deploy.py ~:524-529) still says the non-recoverable→failed and stalled outcomes set `namespace: None`. That's exactly the behavior this PR reversed. Update it to "namespace retained on terminal (failed/stalled); cleared only on recoverable →pending".

**2. The reset action-banner still lies for the null-namespace case (should fix, small).** This is the same misleading-success failure mode #277 exists to kill. For an entry that's terminal with `namespace: null` but a known `pr_name` (the issue's "never written" sub-case, and any entry persisted before this PR), `_reset_pair` still prints "deleting PipelineRun, uninstalling helm releases" and `reset` reports "N pair(s) reset" while the `if ns:` guard skips all helm cleanup — nothing warns the operator. The PR reasonably defers the full orphan sweep (fix #2), but at minimum the banner shouldn't assert an action it then skips: gate the "uninstalling helm releases" wording on `ns`, and `warn` that helm cleanup was skipped and manual `helm uninstall` is needed. Leaving the structural fix without this means the tool keeps reporting false success for the exact entries it can't clean.

**3. The headline contract isn't actually tested end-to-end (should add).** The new tests assert only the intermediate state (the entry dict keeps a namespace). No test wires the producer (a terminal-failure entry with retained namespace) to the consumer (`_reset_pair` actually running `helm uninstall`). The nearest existing test hardcodes the namespace and never references #277 — it would still pass if the fix were reverted. Add a producer→`_reset_pair` test (drive `_handle_timeout` to timed-out, feed that entry to `_reset_pair`, assert one `helm uninstall`) plus a negative companion (null namespace → no helm calls) so the real regression is visible.

Nice-to-haves (not blocking): a regression test that a retained-namespace terminal pair is not treated as busy/dispatchable on the next `_cmd_run`; an explicit `(issue #277)`-tagged release-on-requeue test for legibility; and a one-line note on the over-monkeypatched pod-health harness test clarifying it pins the wiring, not the health logic.

Scope note: deferring the best-effort orphan sweep (fix #2) to a follow-up is fine and worth filing — but item 2 above (don't print a cleanup you skipped) should land with this PR, since it's the issue's own "success messages must reflect reality" principle.

kalantar:
Addressed the three blocking items in 6ff4fae.

1. Stale docstring — `_handle_pending_pods`'s Side effects block still said `namespace: None` for the non-recoverable→failed and stalled outcomes, which is exactly what this PR reversed. Updated it to spell out: namespace retained on terminal (failed/stalled), cleared only on the recoverable→pending re-dispatch.

2. Reset banner no longer lies for the null-namespace case. When a terminal pair has no namespace recorded, `_reset_pair` can't run helm cleanup (it's guarded by `if ns:`). The banner now says "deleting PipelineRun (namespace unknown — skipping helm cleanup)" instead of claiming an uninstall, and a warning tells the operator helm cleanup was skipped and to remove releases manually. Covered both skip paths — the no-pr_name early return and the pr_name-known main path. I kept the state reset returning success (the entry is still reset); the change is purely that we no longer report a cleanup we didn't do.

3. Added the end-to-end test that was missing. The prior tests only checked the entry dict kept a namespace, so they'd pass even if the retain fix were reverted. New positive test drives `_handle_timeout` to timed-out and feeds that exact entry into `_reset_pair`, asserting a real `helm uninstall` happens. New negative companion confirms a null-namespace terminal pair skips helm entirely and emits the warning.

Left out, as discussed: the best-effort orphan sweep (the issue's fix #2) for entries already persisted with a null namespace — still a manual `helm uninstall` for those. Not filing a separate issue for it per your call. Also skipped the non-blocking nice-to-haves (dispatchability regression test, the #277-tagged requeue test, the harness-test comment).

866 passed, 2 xfailed; ruff F-codes clean.

kalantar:
## Re-review (round 2) — READY TO MERGE

All three round-1 items are resolved in `6ff4fae`, verified against the actual branch (181 tests pass in the touched files, ruff F-codes clean):

- **Stale docstring** — `_handle_pending_pods`'s "Side effects" block now accurately states namespace is retained on terminal (failed/stalled) and cleared only on the recoverable→pending re-dispatch. Matches the code in all three branches.
- **Banner no longer lies** — the "uninstalling helm releases" wording is now gated on `ns`; the null-namespace case prints "deleting PipelineRun (namespace unknown — skipping helm cleanup)" and both skip paths (the no-`pr_name` early return and the `pr_name`-known main path) emit a `warn` telling the operator cleanup was skipped and to remove releases manually. Traced exhaustively: no terminal-failure path now returns success with helm skipped and no warning. The dry-run paths stay correctly silent, and the done-pair path is correct by design (Tekton's finally task owns that teardown).
- **End-to-end test** — `test_timed_out_entry_flows_to_reset_helm_uninstall` drives the real `_handle_timeout` to timed-out and feeds that same entry to the real `_reset_pair`, asserting an actual `helm uninstall`. Mutation-checked: reverting retain-on-failure makes it fail. The negative companion confirms null-namespace → zero helm calls plus the warning. Good belt-and-suspenders.

The retain/release classification and the "Mirrors _reconcile_on_resume" claim from round 1 remain correct and unchanged.

Two optional, non-blocking follow-ups (not required for merge):
- The reset summary still counts warn-only (skipped-cleanup) pairs in "N pair(s) reset". The per-pair `[WARN]` lines precede it, so it's honest at the line level; if you want the summary line itself to reflect it, append "(M needed manual helm cleanup)" mirroring how errors are surfaced.
- A regression test that a retained-namespace terminal pair isn't treated as a busy slot / re-dispatched on the next `_cmd_run`. The behavior is already correct (`slots_busy` keys off `status == "running"`, verified by inspection in both rounds) — this would just guard the specific side effect this PR introduces.

Deferring the best-effort orphan sweep (the issue's fix #2) for pre-existing null-namespace entries remains reasonable; those still need a one-time manual `helm uninstall`, and the tool now says so out loud.

Issue #193, opened 2026-05-20
deploy build: add unit tests for happy-path and build-failure scenarios


kalantar:
## Context

PR #189 added `_cmd_build` but the unit tests only cover skip/error paths. The core value — staleness detection triggering a build, and build failure not saving a hash — has no unit coverage beyond the `ensure_image` library functions.

## Tests needed

1. `_cmd_build` returns "current" when source hash matches (needs monkeypatching `image_needs_build`)
2. `_cmd_build` returns "built" after successful build (monkeypatch subprocess + `image_needs_build`)
3. Build failure (non-zero exit from build-epp.sh) does not save hash to metadata

## Affected Files
- `pipeline/tests/test_deploy_build.py`

PR #281, opened 2026-06-04
fix: build-epp.sh fails fast on Failed build pod (#275)

### LINKED ISSUE #275: build-epp.sh waits 30 min on Failed build pod (kubectl wait checks only Succeeded)
## Summary

`pipeline/scripts/build-epp.sh` polls for the build pod's `Succeeded` phase only. When the pod transitions to `Failed`, the script keeps waiting until the 30-minute `--timeout` elapses, even though the outcome is decided.

## Where

`pipeline/scripts/build-epp.sh:179-180`:

```bash
if kubectl wait --for=jsonpath='{.status.phase}'=Succeeded \
  "pod/${BUILD_POD}" -n "${NAMESPACE}" --timeout=1800s 2>/dev/null; then
```

`kubectl wait` does not have an OR semantics across `--for` matchers, and `Failed` is not on the watch list, so a pod that exits with `phase=Failed` (or `reason=Error`) is treated identically to a still-running build for the full 1800 s timeout.

## Repro

Trigger any base-image pull failure inside the build pod (Docker Hub anonymous 429 is an easy way). The pod terminates within ~1 s with `phase=Failed` and the buildkit container in `Terminated{exitCode:1, reason:Error}`. `deploy.py run --remote` then sits at \"Building image (this may take several minutes)…\" for ~30 min before reporting failure.

## Expected

The script should detect terminal failure within seconds and surface logs immediately so the operator can react.

## Suggested fix

Replace the single `kubectl wait` with a short polling loop that breaks on either `Succeeded` or `Failed`:

```bash
deadline=$(( $(date +%s) + 1800 ))
while :; do
  phase=$(kubectl get pod \"\${BUILD_POD}\" -n \"\${NAMESPACE}\" -o jsonpath='{.status.phase}' 2>/dev/null || echo \"\")
  case \"\$phase\" in
    Succeeded) ok \"Image built and pushed: \${FULL_IMAGE}\"; break ;;
    Failed)    err \"Build failed. Logs:\";
               kubectl logs \"\${BUILD_POD}\" -n \"\${NAMESPACE}\" --tail=80 2>/dev/null || true
               exit 1 ;;
  esac
  if (( \$(date +%s) > deadline )); then
    err \"Timed out waiting for build pod (phase=\$phase). Last logs:\"
    kubectl logs \"\${BUILD_POD}\" -n \"\${NAMESPACE}\" --tail=80 2>/dev/null || true
    exit 1
  fi
  sleep 3
done
```

Equivalent improvement: keep the `kubectl wait` for `Succeeded` but run a second `kubectl wait --for=jsonpath='{.status.phase}'=Failed` in parallel and exit when either completes. Whichever pattern is preferred for the codebase.

## Notes

This was hit during `deploy.py run --remote` on a freshly built submodule update, when buildkit pulled `golang:1.25.11` anonymously and got Docker Hub 429-rate-limited. The 429 itself is a credentials issue separate from this bug — but the wait-loop behavior turned a 1-second build failure into a 30-minute hang, which is the part worth fixing here.

kalantar:
## Summary

`pipeline/scripts/build-epp.sh` waited the full 30-minute timeout whenever the build pod failed. `kubectl wait --for=jsonpath='{.status.phase}'=Succeeded` has no OR semantics across matchers, so a pod that exits with `phase=Failed` looked identical to a still-running build. A 1-second base-image pull failure (e.g. Docker Hub 429) turned into a ~30-minute hang at "Building image (this may take several minutes)…" in `deploy.py run --remote`.

## Change

Replaced the single `kubectl wait` (Step 5) with a 3-second polling loop that:
- breaks immediately on `Succeeded`,
- on `Failed`, surfaces the last 80 log lines and `exit 1` right away,
- keeps the 1800s deadline as a backstop for genuinely-stuck (`Pending`/`Unknown`) pods, also dumping logs on timeout.

## Acceptance criteria (from #275)

- [x] `phase=Failed` detected within seconds, not after 1800s
- [x] Pod logs surfaced on failure
- [x] Timeout backstop preserved for stuck pods
- [x] Success path unchanged — falls through to Step 6 (cleanup) and Step 7 (metadata)

## Reviewer attention

- **`set -euo pipefail` safety:** the only `(( ))` arithmetic is inside an `if` (exit status consumed, so a false condition won't abort the script), and `phase=$(kubectl get ... || echo "")` guards the `get` against `pipefail`.
- **Failed pod is left in place** on failure (we `exit 1` before Step 6 cleanup) — same behavior as before, intentional so logs remain inspectable.
- **No automated test:** there is no bash test harness in the repo; build-failure unit tests are tracked separately in #193. Verified with `bash -n` and the full `pytest pipeline/` suite (866 passed, 2 xfailed; no Python regression).

## Scope

The Docker Hub 429 / registry-credentials problem that triggered this is a separate concern and is **not** addressed here — only the wait-loop hang.

Closes #275

### DIFF
--- a/pipeline/scripts/build-epp.sh
+++ b/pipeline/scripts/build-epp.sh
@@ -175,15 +175,34 @@ spec:
 BUILDPOD
 
 # ── Step 5: Wait for build ────────────────────────────────────────
+# Poll for a terminal phase rather than `kubectl wait --for=...=Succeeded`,
+# which has no OR semantics across matchers: a pod that exits with
+# phase=Failed would otherwise be indistinguishable from a still-running
+# build and block for the full timeout. (#275)
 info "Building image (this may take several minutes)..."
-if kubectl wait --for=jsonpath='{.status.phase}'=Succeeded \
-  "pod/${BUILD_POD}" -n "${NAMESPACE}" --timeout=1800s 2>/dev/null; then
-  ok "Image built and pushed: ${FULL_IMAGE}"
-else
-  err "Build failed. Logs:"
-  kubectl logs "${BUILD_POD}" -n "${NAMESPACE}" --tail=50 2>/dev/null || true
-  exit 1
-fi
+BUILD_TIMEOUT=1800
+deadline=$(( $(date +%s) + BUILD_TIMEOUT ))
+while :; do
+  phase=$(kubectl get pod "${BUILD_POD}" -n "${NAMESPACE}" \
+    -o jsonpath='{.status.phase}' 2>/dev/null || echo "")
+  case "${phase}" in
+    Succeeded)
+      ok "Image built and pushed: ${FULL_IMAGE}"
+      break
+      ;;
+    Failed)
+      err "Build failed (phase=Failed). Logs:"
+      kubectl logs "${BUILD_POD}" -n "${NAMESPACE}" --tail=80 2>/dev/null || true
+      exit 1
+      ;;
+  esac
+  if (( $(date +%s) > deadline )); then
+    err "Timed out after ${BUILD_TIMEOUT}s waiting for build pod (phase=${phase:-unknown}). Last logs:"
+    kubectl logs "${BUILD_POD}" -n "${NAMESPACE}" --tail=80 2>/dev/null || true
+    exit 1
+  fi
+  sleep 3
+done
 
 # ── Step 6: Clean up build pod ─────────────────────────────────────
 kubectl delete pod "${BUILD_POD}" -n "${NAMESPACE}" --ignore-not-found --force --grace-period=0 2>/dev/null || true


kalantar:

The change correctly replaces the `kubectl wait --for=...=Succeeded` (which has no OR semantics and so can't distinguish `phase=Failed` from a still-running build) with a 3s polling loop that fast-exits on `Succeeded` or `Failed` and keeps the 1800s deadline as a backstop. All four acceptance criteria from #275 map directly to visible code branches, and the bash is safe under `set -euo pipefail` (the `(( ))` is inside an `if`, and `phase=$(... || echo "")` guards against `pipefail`).

### Findings

**No blocking issues.** Correctness, the comment accuracy, and the requirement coverage all check out. The "why" comment is well done — it documents exactly why a future maintainer shouldn't collapse the loop back into `kubectl wait`.

A note on one reviewer concern worth recording: an automated check raised that a Docker Hub 429 could surface as `phase=Pending` (ImagePullBackOff) and thus wait the full 1800s. That does not apply to the failure this PR targets — the issue's repro shows the 429 happens *inside* the buildkit container during the base-image pull, the container exits 1, and the pod reaches `phase=Failed` within ~1s. The new `Failed` branch catches exactly that. (A pod-level ImagePullBackOff of the buildkit image itself would still wait for the timeout, but that is a distinct, pre-existing edge, not the bug being fixed.)

### Optional, non-blocking follow-ups (separate issues, not for this PR)

1. `phase=$(kubectl get ... 2>/dev/null || echo "")` collapses every `kubectl get` failure — API unreachable, expired creds, RBAC denial — into the same empty string as "still pending," so a persistent connectivity/auth failure is masked for the full 1800s. The timeout backstop bounds the damage, but the final `phase=unknown` message gives the operator no hint of the real cause. Capturing stderr and surfacing it on timeout would help.
2. On the most useful failure case, `kubectl logs` may be empty (no container ran); a fallback to `kubectl describe pod` would carry the actual reason. Minor.
3. The repo has no shell linting in CI (`.github/workflows/test.yml` runs ruff + pytest only). A `shellcheck pipeline/scripts/*.sh` (or `bash -n`) gate would be cheap and catch future shell regressions. The "no automated test" claim here is consistent with this being a Python-test-only repo; #193 tracks the Python-side build coverage.

None of these block merging this fix.

PR #283, opened 2026-06-04
fix(deploy): surface orchestrator pod logs on Failed phase

### LINKED ISSUE #276: deploy.py reports "Orchestrator pod failed: unknown reason" instead of pod logs
## Summary

When the orchestrator Job's pod exits non-zero (e.g., argparse rejects an invalid flag value), `deploy.py run --remote` prints `Orchestrator pod failed: unknown reason` and exits without surfacing the pod logs. The user has to manually run `kubectl logs <pod> -c orchestrator` to see the actual error message — which the orchestrator emits cleanly.

## Where

`pipeline/deploy.py:2627-2630`:

```python
if phase == \"Failed\":
    msg = pod.get(\"status\", {}).get(\"message\", \"unknown reason\")
    err(f\"Orchestrator pod failed: {msg}\")
    sys.exit(1)
```

The code reads `pod.status.message`. That field is populated by the kubelet only for higher-level pod failures (evicted, preempted, OOMKilled at pod level). When a container simply exits non-zero, `pod.status.message` is empty and the per-container detail lives in `pod.status.containerStatuses[].state.terminated.{exitCode, reason}` — and the actual diagnostic is in the container's stdout/stderr. Neither is consulted.

## Repro

Run `deploy.py run --remote` with an invalid orchestrator flag — e.g., a `--workload` value that does not match the expected set:

```
[ERROR] Orchestrator pod failed: unknown reason
```

Manually fetched logs of the same pod show what `deploy.py` should have printed:

```
[ERROR] --workload: unrecognized values ['balanced-20']
  Valid --workload values: balanced_20, balanced_30, balanced_40, balanced_50
```

## Expected

When the pod transitions to `Failed`, `deploy.py` should fetch and print the orchestrator container logs (or at least the tail) before exiting. Bonus: include the container's terminated `exitCode` / `reason` from `containerStatuses` in the header line.

## Suggested fix

Replace the message-only branch with one that also surfaces the logs:

```python
if phase == \"Failed\":
    msg = pod.get(\"status\", {}).get(\"message\", \"\")
    err(f\"Orchestrator pod failed{': ' + msg if msg else ''}\")
    pod_name = pod.get(\"metadata\", {}).get(\"name\", \"\")
    if pod_name:
        logs = run([\"kubectl\", \"logs\", pod_name, \"-n\", namespace,
                    \"-c\", \"orchestrator\", \"--tail=80\"],
                   check=False, capture=True)
        if logs.stdout:
            err(\"Orchestrator pod logs:\\n\" + logs.stdout)
    sys.exit(1)
```

## Related

Same general theme as #275 (build-pod failures hidden by an inadequate wait/report path) — pipeline scripts should surface the actual failure cause from pod logs / container terminated state instead of relying on top-level `pod.status.message`.

kalantar:
Closes #276

## Problem

When the orchestrator Job's pod exits non-zero (e.g. argparse rejects an invalid `--workload` value), `deploy.py run --remote` printed:

```
[ERROR] Orchestrator pod failed: unknown reason
```

and exited without the actual error, forcing the user to manually run `kubectl logs <pod> -c orchestrator`.

The `Failed`-phase branch in `_wait_for_job_pod` read only `pod.status.message`. That field is populated by the kubelet only for **pod-level** failures (evicted, preempted, OOMKilled). When a container simply exits non-zero, `pod.status.message` is empty — the detail lives in `containerStatuses[].state.terminated.{exitCode,reason}` and the real diagnostic is in the container's stdout/stderr. Neither was consulted.

## Change

Extracted the failure-reporting into `_report_failed_pod(pod, namespace)`, which:

1. Uses `pod.status.message` when present (preserves prior behavior for OOMKilled/evicted).
2. Otherwise falls back to the container terminated `exitCode`/`reason` from `containerStatuses` / `initContainerStatuses` (the issue's "Bonus").
3. Tails the orchestrator container logs (`kubectl logs ... -c orchestrator --tail=80`) and prints them before exit.

Exit behavior is unchanged (`sys.exit(1)` on `Failed`).

## Reviewer attention

- **Init-container edge case:** if the failure is in an init container, the orchestrator container never started, so the `-c orchestrator` log fetch returns empty. Handled gracefully (logs only printed when `logs.stdout` is non-empty), and the container terminated detail still surfaces the init-container exit in the header.
- Log fetch uses `run(..., check=False, capture=True)` so a failed `kubectl logs` (e.g. container not found) never masks the original failure.
- Same theme as #275 (build-pod failures hidden by an inadequate report path).

## Tests

- Updated `test_wait_for_pod_failed_phase_exits` (its `fake_run` now also serves the `kubectl logs` call).
- Added `test_wait_for_pod_failed_surfaces_logs` — container exits 2 with empty `pod.status.message`; asserts both the terminated detail (`orchestrator exited 2 (Error)`) and the log text are surfaced.
- Added `test_report_failed_pod_no_logs_no_crash` — no logs available; reporting still prints the pod-level message without crashing.

Full suite: `840 passed, 2 xfailed`. Lint: `ruff check pipeline/ --select F` clean.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2543,6 +2543,45 @@ def _check_existing_job(namespace: str) -> "str | None":
     return "completed"
 
 
+def _report_failed_pod(pod: dict, namespace: str) -> None:
+    """Print the cause of an orchestrator pod failure before the caller exits.
+
+    `pod.status.message` is only populated for pod-level failures (evicted,
+    preempted, OOMKilled at the pod level). When a container simply exits
+    non-zero that field is empty, so fall back to the container's terminated
+    `exitCode`/`reason` and tail the orchestrator logs, where the real
+    diagnostic lives. See issue #276.
+    """
+    status = pod.get("status", {})
+    msg = status.get("message", "")
+    details = []
+    all_statuses = (status.get("initContainerStatuses", [])
+                    + status.get("containerStatuses", []))
+    for cs in all_statuses:
+        term = cs.get("state", {}).get("terminated", {})
+        if term and term.get("exitCode", 0) != 0:
+            cname = cs.get("name", "?")
+            reason = term.get("reason", "")
+            code = term.get("exitCode", "")
+            details.append(f"{cname} exited {code}"
+                           + (f" ({reason})" if reason else ""))
+
+    header = "Orchestrator pod failed"
+    if msg:
+        header += f": {msg}"
+    elif details:
+        header += ": " + "; ".join(details)
+    err(header)
+
+    pod_name = pod.get("metadata", {}).get("name", "")
+    if pod_name:
+        logs = run(["kubectl", "logs", pod_name, "-n", namespace,
+                    "-c", "orchestrator", "--tail=80"],
+                   check=False, capture=True)
+        if logs.stdout:
+            err("Orchestrator pod logs:\n" + logs.stdout)
+
+
 def _wait_for_job_pod(namespace: str, *, timeout: int = 120, poll: int = 5) -> None:
     """Poll until the orchestrator pod reaches Running or Succeeded.
 
@@ -2585,8 +2624,7 @@ def _wait_for_job_pod(namespace: str, *, timeout: int = 120, poll: int = 5) -> N
                 if phase in ("Running", "Succeeded"):
                     return
                 if phase == "Failed":
-                    msg = pod.get("status", {}).get("message", "unknown reason")
-                    err(f"Orchestrator pod failed: {msg}")
+                    _report_failed_pod(pod, namespace)
                     sys.exit(1)
                 all_statuses = (pod.get("status", {}).get("initContainerStatuses", [])
                                 + pod.get("status", {}).get("containerStatuses", []))
--- a/pipeline/tests/test_deploy_remote.py
+++ b/pipeline/tests/test_deploy_remote.py
@@ -269,6 +269,7 @@ def test_wait_for_pod_failed_surfaces_logs(monkeypatch, capsys):
         }],
     })
     log_text = "[ERROR] --workload: unrecognized values ['balanced-20']"
+    log_cmds = []
 
     def fake_run(cmd, *, check=True, capture=False, cwd=None):
         class _R:
@@ -276,6 +277,7 @@ def test_wait_for_pod_failed_surfaces_logs(monkeypatch, capsys):
             stderr = ""
         r = _R()
         if "logs" in cmd:
+            log_cmds.append(cmd)
             r.stdout = log_text
         else:
             r.stdout = pod_json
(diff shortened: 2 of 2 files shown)


kalantar:
## Review: approve with nits

The change does what issue #276 asks: when the orchestrator pod reaches `Failed`, `deploy.py` now surfaces the per-container terminated `exitCode`/`reason` and tails the orchestrator container logs instead of printing the unhelpful "unknown reason". It implements the issue's bonus (terminated detail in the header), checks both `initContainerStatuses` and `containerStatuses`, and preserves prior behavior for pod-level failures (OOMKilled/evicted via `pod.status.message`). Refactoring into `_report_failed_pod()` is a clean, testable improvement over the inline suggestion in the issue. Exit semantics are unchanged. Tests pass locally.

No blocking issues. A few nits to consider:

**Nits (non-blocking):**

1. **Headline init-container edge case is not test-covered.** The PR rightly calls out the init-container path as the key thing to verify (orchestrator container never started, log fetch returns empty, terminated detail must still surface in the header). But no test exercises a populated `initContainerStatuses` with a terminated non-zero state. `test_report_failed_pod_no_logs_no_crash` uses empty `containerStatuses` plus a pod-level `message`, so the `initContainerStatuses` concatenation branch and the "init exit surfaces in header" claim are untested. Adding one small test for that case would lock in the stated behavior.

2. **A failed `kubectl logs` (non-zero return, empty stdout) is silently dropped.** With `check=False` and the `if logs.stdout:` guard, a transient kubectl/network failure during log retrieval produces no diagnostic about the log fetch itself. This is the correct trade-off (never mask the original failure), and for the init-container case the header still carries the terminated detail — but consider a one-line hint (e.g. when returncode != 0 and stdout is empty) so a log-fetch failure isn't indistinguishable from "no logs". Optional.

3. **Test does not assert the kubectl args.** `fake_run` only checks `"logs" in cmd`; it does not verify `-c orchestrator` or `--tail=80` are passed. Minor — asserting the flags would guard against silent arg drift.

4. **Bare-header fallback.** If a pod is `Failed` with no non-zero container exit and an empty `pod.status.message`, the header is just "Orchestrator pod failed" with no detail. Graceful and acceptable; noting for completeness.

`err()` correctly routes to stderr, consistent with the existing `[ERROR]` convention. Same theme as #275 — good direction.

pipeline/deploy.py
```diff
@@ -2543,6 +2543,45 @@ def _check_existing_job(namespace: str) -> "str | None":
...
+        header += ": " + "; ".join(details)
+    err(header)
+
+    pod_name = pod.get("metadata", {}).get("name", "")
+    if pod_name:
+        logs = run(["kubectl", "logs", pod_name, "-n", namespace,
+                    "-c", "orchestrator", "--tail=80"],
+                   check=False, capture=True)
+        if logs.stdout:
+            err("Orchestrator pod logs:\n" + logs.stdout)
```
kalantar, line 2582: Optional: when `kubectl logs` fails it returns empty stdout and the diagnostic is on stderr, which is dropped here. In the worst case (orchestrator exits non-zero, `pod.status.message` empty, no terminated detail captured, and the logs fetch itself fails) the user is left with just a bare `Orchestrator pod failed` header and no detail. Rare, but consider printing `logs.stderr` (or a short hint) when stdout is empty so this path is never completely silent.

pipeline/tests/test_deploy_remote.py
```diff
@@ -253,6 +253,104 @@ class _R:
...
+        mod._wait_for_job_pod("ns", timeout=10, poll=1)
+    assert exc_info.value.code == 1
+    out = capsys.readouterr()
+    combined = out.out + out.err
+    # Container terminated detail appears in the header.
+    assert "orchestrator exited 2 (Error)" in combined
+    # The actual diagnostic from the pod logs is surfaced.
+    assert log_text in combined
+    # Logs are fetched for the orchestrator container with a bounded tail.
+    assert log_cmds, "expected a kubectl logs call"
```
kalantar, line 297: Optional: the test asserts the `kubectl logs` args (`-c orchestrator`, `--tail`) but not that the call passes `check=False`. Since `fake_run` never raises, a regression to `check=True` wouldn't be caught here. A small assertion that the captured call used `check=False` would lock in the "never mask the original failure" guarantee the PR description calls out.

kalantar (commented):
Overall this is a solid, well-scoped fix that matches the issue's intent and the suggested approach. It surfaces the container terminated detail in the header and tails the orchestrator logs, with the init-container edge case handled gracefully. Tests are meaningful and cover the main paths. **Verdict: approve with a few minor nits (none blocking).**

See the two inline notes for optional polish. Neither blocks merge.

PR #286, opened 2026-06-04
fix(deploy): handle corrupt progress ConfigMap across subcommands (#140)

### LINKED ISSUE #140: Handle corrupt progress.json gracefully in deploy.py subcommands
## Context

`LocalProgressStore.load()` raises `ValueError` on corrupt `progress.json`. Several deploy.py subcommands (`_cmd_wipe`, `_cmd_reset`, `_cmd_status`) call `store.load()` without catching this, resulting in a raw traceback for the user.

## Requirements

- Catch `ValueError` from `store.load()` in all subcommand handlers
- Print a clear error message: which file is corrupt, and suggest re-running the prepare phase or manually fixing the file
- Exit with non-zero status

Surfaced during review of #138 (wipe subcommand). Not specific to wipe — same gap exists in all subcommands that load progress.

kalantar:
## Summary

- Three subcommands (`_cmd_run`, `_cmd_reset`, `_cmd_wipe`) used `store.load()` bare and produced a raw traceback when the progress ConfigMap was corrupt. `_cmd_run_remote` caught `(RuntimeError, OSError)` but missed `ValueError`. `_cmd_status` and `_cmd_collect` caught `ValueError` but warned and continued — silently downgrading corruption to "0 pairs".
- Centralized the policy in a single `_load_progress(store, *, allow_unreachable)` helper that exits non-zero with a clear, actionable message on `ValueError` (corrupt data isn't recoverable by retry), and either re-raises or warns on `RuntimeError` depending on caller tolerance.
- Replaced all six `store.load()` call sites with the helper. Added 5 unit tests for the helper and updated one pre-existing collect test that asserted the old lenient behavior.

Closes #140

## Reviewer attention

- **Stale issue text:** issue #140 references `LocalProgressStore.load()` and `progress.json`, but the store moved to `ConfigMapProgressStore` since the issue was filed. The `ValueError` contract on corrupt data (`pipeline/lib/progress.py:67`) is unchanged, so the issue's intent applies directly.
- **Single helper, per user preference:** all six call sites now route through `_load_progress`. The helper takes `allow_unreachable` to keep the soft-fail behavior on transient `RuntimeError` for the three subcommands that intentionally degrade gracefully (`_cmd_status`, `_cmd_collect`, `_cmd_run_remote` pre-flight). `ValueError` always exits regardless of the flag — corruption is not transient.
- **Behavior change for `_cmd_status` and `_cmd_collect`:** previously they warned and showed "0 pairs" / continued without progress on corrupt data. They now exit non-zero, matching the issue's "exit with non-zero status" requirement. This is a small but intentional UX shift.
- **OSError no longer caught explicitly in `_cmd_run_remote`:** `ConfigMapProgressStore.load` already converts `OSError` to `RuntimeError` (`progress.py:52-53`), so the prior `(RuntimeError, OSError)` catch was over-broad. The new path catches `RuntimeError` only.
- **Test update:** `test_collect_corrupt_configmap_raises` was renamed to `test_collect_corrupt_configmap_exits` and asserts the new exit-on-corruption contract; phase extraction must not run on corrupt data.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 929 passed, 2 xfailed (was 924)
- [x] New `TestLoadProgressHelper` covers: success passthrough, ValueError exit + message, RuntimeError propagate, RuntimeError swallow with `allow_unreachable=True`, ValueError exit even when `allow_unreachable=True`
- [ ] Live smoke: corrupt the progress ConfigMap (`kubectl edit configmap sim2real-progress-<run> -n <ns>` and break the JSON), then run each of `deploy.py status`, `deploy.py collect`, `deploy.py run`, `deploy.py reset`, `deploy.py wipe`, `deploy.py run --remote` and confirm a clean error message + non-zero exit instead of a traceback

### DIFF
--- a/docs/superpowers/plans/2026-06-04-issue-140-progress-corrupt-handling.md
+++ b/docs/superpowers/plans/2026-06-04-issue-140-progress-corrupt-handling.md
@@ -1,0 +1,330 @@
+# Issue #140: Handle corrupt progress data gracefully in deploy.py Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop deploy.py subcommands from crashing with a raw traceback when the progress ConfigMap contains invalid JSON. Print a clear error and exit non-zero per issue #140.
+
+**Architecture:** Add a single `_load_progress(store, *, allow_unreachable: bool = False)` helper in `pipeline/deploy.py`. It catches `ValueError` (the corrupt-data signal raised by `ConfigMapProgressStore.load`) and exits with a user-actionable error message; `RuntimeError` (kubectl-unreachable) propagates by default but is caught and warned-only when `allow_unreachable=True`. Replace all six `store.load()` call sites with the helper. This consolidates the error policy in one place per the user's "minimize duplication" preference and gives consistent UX for the corrupt-progress case across `_cmd_run`, `_cmd_reset`, `_cmd_wipe`, `_cmd_status`, `_cmd_collect`, and `_cmd_run_remote`.
+
+**Tech Stack:** Python (deploy.py, pytest).
+
+**Closes:** #140
+
+---
+
+## Inventory of `store.load()` Call Sites
+
+| File:Line | Subcommand | Current handling | Target handling |
+|-----------|-----------|------------------|-----------------|
+| `deploy.py:418` | `_cmd_status` | `except (ValueError, RuntimeError)` → warn + `progress = {}` | `_load_progress(store, allow_unreachable=True)` (ValueError now exits with clear message; RuntimeError still falls through) |
+| `deploy.py:1049` | `_cmd_collect` | `except (ValueError, RuntimeError)` → warn + `progress = None` | `_load_progress(store, allow_unreachable=True) or None` |
+| `deploy.py:2009` | `_cmd_run` | bare — raw traceback on corrupt | `_load_progress(store)` |
+| `deploy.py:2346` | `_cmd_reset` | bare — raw traceback on corrupt | `_load_progress(store)` |
+| `deploy.py:2401` | `_cmd_wipe` | bare — raw traceback on corrupt | `_load_progress(store)` |
+| `deploy.py:2704` | `_cmd_run_remote` | `except (RuntimeError, OSError)` → warn + `progress = None` (misses ValueError) | `_load_progress(store, allow_unreachable=True)` |
+
+**Behavior change for `_cmd_status` and `_cmd_collect`:** Today, corrupt data is warn+continue (status shows "0 pairs", collect runs without progress). Per issue: corrupt data must exit non-zero. The helper makes both subcommands exit on corruption — a small but intentional UX shift, justified because "0 pairs" after corruption is misleading and silently hides data loss.
+
+**Behavior change for `_cmd_run_remote`:** Currently does not catch `ValueError`, so corruption produces a raw traceback during pre-flight filter validation. Helper unifies this with the rest.
+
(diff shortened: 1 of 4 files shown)


kalantar:

The PR cleanly addresses issue #140's three stated requirements (catch `ValueError` from `store.load()` everywhere, print a clear actionable error, exit non-zero). All six `store.load()` call sites in `pipeline/deploy.py` are now routed through the single `_load_progress` helper; 929 tests pass; ruff is clean. The behavior tightening for `_cmd_status` and `_cmd_collect` (corrupt → exit, instead of the previous warn-and-show-zero-pairs) is a real improvement and is properly called out in the PR description.

There is one finding that warrants a deliberate decision before merge rather than a unilateral fix:

### Discussion item: `_cmd_collect`'s unscoped path silently degrades on cluster-unreachable

In `_cmd_collect`, the new line is `progress = _load_progress(store, allow_unreachable=True) or None`. On `RuntimeError` (kubectl unreachable) the helper warns and returns `{}`, so `progress is None` here. With `--only`/`--workload`, the existing `if scoped and not progress` check exits with a clear message. **Without** scoping, the function proceeds into `_discover_phases(cluster_dir)` and collects from the PVC for every phase it finds locally — emitting only a stderr warning, not an error.

A user collecting after a transient kubectl outage would get a successful return code and a partial result set, with no signal that the cluster's authoritative progress data was never consulted. The `_cmd_status` and `_cmd_run_remote` paths have the same conflation between "cluster unreachable" and "no progress yet" because both cases collapse to `{}`.

This pattern was **pre-existing** — the prior code also caught `RuntimeError` and set `progress = None`. The PR didn't introduce the silent fallback; it preserved it while fixing the `ValueError` half. The question is scope:

- **Fix here**: Add a sentinel (e.g. raise a typed `ProgressUnavailable` from the helper, or return `(progress, was_unreachable)`) so callers can distinguish "no data yet" from "cluster unreachable" and exit on the latter where appropriate. This grows the PR.
- **Defer**: Merge as-is and file a follow-up issue for the unreachable-vs-empty conflation. This PR's stated requirement (corrupt-data ValueError) is fully met.

### Non-blocking polish (file in a follow-up regardless)

- `pipeline/deploy.py:135-136` — recovery hint contains literal `<name>` and `<namespace>` placeholders; the helper has the store in scope and could substitute the actual `store.configmap_name`. The propagated `ValueError` text from `progress.py:67-69` does include both, so the user can find them, but copy-pasting the suggested `kubectl edit configmap <name> -n <namespace>` literally fails.
- `pipeline/deploy.py:2030` — the `# Load or initialize progress` comment now sits above just the load; the initialization is on the lines that follow. Either trim to `# Load progress` or move it down to span the whole block.
- The OSError catch was removed from `_cmd_run_remote` because `ConfigMapProgressStore.load` converts `OSError` → `RuntimeError` at `progress.py:52-53`. Safe today, but the invariant is cross-file. Consider documenting it in `progress.py`'s class docstring so a future contributor doesn't leak an `OSError` from `load()`.
- Coverage gap: only `_cmd_collect` has an integration-style corrupt-ConfigMap test. `_cmd_status` would be the highest-value addition since it's the most user-facing and also changed semantics.

### What's good

- Single helper consolidates policy across six sites with clean separation between corrupt-data (always fatal) and unreachable (caller-policy via `allow_unreachable`).
- Helper unit tests are behavior-focused (exit code, message content, return value) rather than implementation-coupled. Five tests cover all branches plus the cross-product.
- The `test_collect_corrupt_configmap_raises` rename to `_exits` and the new `assert collected_phases == []` strengthens the test — it now asserts that phase extraction is **blocked** by corrupt data, where the prior version asserted the opposite (extraction proceeded).
- All three issue requirements verified in code: `ValueError` caught everywhere, message names the corrupt artifact via the propagated exception text and adds recovery guidance, `sys.exit(1)`.

PR #288, opened 2026-06-04
fix(deploy): raise RuntimeError when kubectl run fails creating extractor pod

### LINKED ISSUE #204: deploy collect: handle CalledProcessError from kubectl run in extractor
In `_extract_phases_from_pvc`, the `kubectl run` command (line 639) uses `capture=True` with the default `check=True`. If it fails, a `CalledProcessError` propagates unhandled — the caller's `except RuntimeError` doesn't catch it, so the user sees a raw traceback instead of a clean error message.

Should use the same `check=False` + explicit raise pattern as the adjacent `kubectl wait` call:
\`\`\`python
result = run(["kubectl", "run", ...], capture=True, check=False)
if result.returncode != 0:
    raise RuntimeError(f"kubectl run failed: {result.stderr.strip()}")
\`\`\`

This was pre-existing (not introduced by #202) but surfaced during review.

kalantar:
## Summary

In `_extract_phases_from_pvc`, the `kubectl run` that creates the extractor pod used the `run()` helper's default `check=True`, so a non-zero exit raised `subprocess.CalledProcessError`. The unscoped collect fallback at `pipeline/deploy.py:1408` catches only `RuntimeError`, so the error escaped as a raw traceback to the user.

Apply the same `check=False` + explicit `raise RuntimeError(...)` pattern already used three lines below for the adjacent `kubectl wait` call, and clean up any partial pod on failure (mirrors the wait-failure cleanup).

## Test plan

- [x] New unit test `test_extract_phases_kubectl_run_failure_raises_runtime_error` mocks a `kubectl run` failure and asserts `RuntimeError` is raised with the stderr message embedded
- [x] Full pipeline test suite: `python -m pytest pipeline/ -v` → 854 passed, 2 xfailed
- [x] Lint: `ruff check pipeline/ .claude/skills/ --select F` → clean

## Reviewer attention

- The new test exercises `_extract_phases_from_pvc` directly (matching the pattern already used by the `test_collect_*_clears_stale_files` tests below it), monkeypatching `subprocess.run` to fail only the create call.
- Pre-existing scoped collect callers (`_extract_one_slot` at lines 1205, 1236, 1366, 1397) catch the broader `Exception`, so they would have caught a `CalledProcessError` already — the user-visible regression was specifically on the unscoped fallback path. The fix unifies the two paths on `RuntimeError`.

Closes #204

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -866,8 +866,17 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
             "restartPolicy": "Never",
         }
     })
-    run(["kubectl", "run", pod_name, "--image=alpine:3.19", "--restart=Never",
-         "--overrides", overrides, "-n", namespace], capture=True)
+    create_result = run(
+        ["kubectl", "run", pod_name, "--image=alpine:3.19", "--restart=Never",
+         "--overrides", overrides, "-n", namespace],
+        check=False, capture=True,
+    )
+    if create_result.returncode != 0:
+        run(["kubectl", "delete", "pod", pod_name, "-n", namespace,
+             "--ignore-not-found", "--force", "--grace-period=0"],
+            check=False, capture=True)
+        raise RuntimeError(
+            f"Extractor pod {pod_name} create failed: {create_result.stderr.strip()}")
 
     result = run(
         ["kubectl", "wait", f"pod/{pod_name}", "--for=condition=Ready",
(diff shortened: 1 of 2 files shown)


kalantar:

The fix correctly converts kubectl run failures into RuntimeError matching the adjacent kubectl wait pattern, with appropriate cleanup and a regression test; remaining nits are stylistic.

### Important
- **Test does not assert that the cleanup `kubectl delete` is issued on create failure** (`pipeline/tests/test_deploy_collect.py:1888`) — deploy.py:874-877 adds a cleanup delete before raising, but the test (lines 1888-1899) only mocks subprocess.run and asserts the RuntimeError; it never records calls or asserts a delete was issued. If the cleanup line were removed or moved after `raise`, the test would still pass and a leaked pod could block the next collect attempt — exactly the production pain this PR is hardening against.

### Recommendations
- **mock_run dispatch mixes list-membership (`"run" in cmd`) with substring (`"--image=alpine:3.19" in cmd_str`)** (`pipeline/tests/test_deploy_collect.py:1890`) — Verified at line 1890: the two checks use different semantics on the same line. It works today because cmd is always a list with `run` as a token, but other mock_run helpers in this file consistently use cmd_str for both sides. Aligning the style would prevent future readers from misreading the intent.

PR #289, opened 2026-06-04
fix(deploy): distinguish 'cluster unreachable' from 'no progress yet'

### LINKED ISSUE #287: deploy: distinguish 'cluster unreachable' from 'no progress yet' in subcommands that tolerate kubectl flakiness
## Context

Surfaced during review of #286 (closes #140). PR #286 introduced `_load_progress(store, *, allow_unreachable)` and centralized progress-load error handling. When `allow_unreachable=True`, a `RuntimeError` from `store.load()` (kubectl unreachable) collapses to `{}` — same as the legitimate "ConfigMap NotFound" path in `pipeline/lib/progress.py:55-56` and the empty-data path at `:62-63`.

This means `_cmd_status`, `_cmd_collect`, and `_cmd_run_remote` cannot distinguish:

1. Cluster unreachable (kubectl failed)
2. ConfigMap doesn't exist yet (no run has produced progress)
3. ConfigMap exists but is empty
4. ConfigMap was deleted

## Why this is a silent failure

Concrete user-impact paths:

- **`_cmd_collect` unscoped path** (`pipeline/deploy.py:1074, :1241+`) — On unreachable, `progress is None`, the function falls through to `_discover_phases(cluster_dir)` and proceeds to collect from the PVC for every phase it finds locally. The user gets a successful exit code and a partial result set with only a stderr warning. They have no way to know the cluster's authoritative progress data was never consulted.
- **`_cmd_status`** — Prints `"0 pairs (no progress data)"` on cluster outage. Indistinguishable from a fresh run. Exits 0; a wrapper script polling for health gets a successful return on transient kubectl failure.
- **`_cmd_run_remote` pre-flight** — Filter validation is silently skipped on unreachable. The user might assume the failure aborted the run when in fact it proceeded.

This pattern is **pre-existing** — the pre-#286 code also caught `RuntimeError` and set `progress = None`, then proceeded. #286 fixed the `ValueError` half (corrupt → exit) but preserved the `RuntimeError` half. The conflation is now centralized in the helper, which is the right place to fix it.

## Affected files

- `pipeline/deploy.py` — `_load_progress` helper at lines 117-142; the three subcommands that pass `allow_unreachable=True` at lines 447 (`_cmd_status`), 1074 (`_cmd_collect`), 2725 (`_cmd_run_remote`).

## Proposed approach (sketch)

Make the helper distinguish unreachable from no-data by raising a dedicated typed exception or returning a tagged result, e.g.:

```python
class ProgressUnavailable(RuntimeError):
    """Progress store is reachable-failure (transient). Caller decides whether to proceed."""

def _load_progress(store) -> dict:
    try:
        return store.load()
    except ValueError as exc:
        err(f"Corrupt progress data: {exc}")
        err("Re-run prepare.py, or fix the ConfigMap manually.")
        sys.exit(1)
    except RuntimeError as exc:
        raise ProgressUnavailable(str(exc)) from exc
```

Then each caller catches `ProgressUnavailable` explicitly and decides what to do:

- `_cmd_collect` — should refuse to fall back to filesystem-discovered phases when progress is unreachable; exit with a clear message saying the cluster is unreachable and asking the user to retry.
- `_cmd_status` — exit non-zero with a clear "cluster unreachable" message rather than printing `0 pairs`.
- `_cmd_run_remote` pre-flight — keep the current "skip validation" behavior, but with a focused warning that names exactly what was skipped (the pre-#286 message was specifically `ConfigMap unreachable — skipping pre-flight filter validation` and lost specificity in the consolidation).

Alternative: return `(progress, was_unreachable)` from the helper. Slightly less clean but avoids a new exception type.

## Acceptance criteria

- A kubectl-unreachable scenario in any of the three subcommands does not silently fall through to "no progress" semantics.
- `_cmd_status` exits non-zero on cluster-unreachable.
- `_cmd_collect` does not collect from filesystem-discovered phases when the cluster is unreachable.
- `_cmd_run_remote` pre-flight skip is logged with a focused message naming what was skipped.
- ConfigMap-genuinely-not-found (a fresh run with no progress recorded) continues to behave as today — empty result, no error.

## Out of scope

The corrupt-data path (`ValueError`) — already addressed in #286.

kalantar:
## Summary

- `_load_progress(store, allow_unreachable=True)` previously collapsed `RuntimeError` (kubectl unreachable) to `{}` — indistinguishable from the legitimate `ConfigMap NotFound` / empty-data path. This made three subcommands silently degrade on cluster outage.
- Introduce a `ProgressUnavailable(RuntimeError)` sentinel that the helper raises only when `allow_unreachable=True`, so callers can pick their own policy.
- `_cmd_status` exits non-zero with a clear "cluster unreachable" message instead of printing `0 pairs`.
- `_cmd_collect` exits non-zero and refuses to fall through to filesystem-discovered phases when the cluster is unreachable.
- `_cmd_run_remote` pre-flight restores the previously-specific warning `ConfigMap unreachable — skipping pre-flight filter validation: <detail>` and keeps the soft-fail behavior.
- `ConfigMap genuinely empty / NotFound` (fresh run with no progress recorded) still returns `{}` — unchanged.

Closes #287.

## Reviewer attention

- **Helper contract change:** `_load_progress(store, allow_unreachable=True)` no longer returns `{}` on `RuntimeError` — it now raises `ProgressUnavailable`. This is intentional: the whole point of the issue is that the previous return value was lossy. `ProgressUnavailable` subclasses `RuntimeError`, so any pre-existing `except RuntimeError:` branches still catch it; the only callers using `allow_unreachable=True` are the three explicitly updated here.
- **`_cmd_status` UX shift:** previously printed `0 pairs (no progress data)` and exited 0 on cluster outage; now exits 1 with `Cluster unreachable — cannot read progress ConfigMap: ...`. Wrapper scripts polling for health were getting a false-OK; this is the intended fix.
- **`_cmd_collect` UX shift:** previously fell through to `_discover_phases(cluster_dir)` and proceeded with filesystem-discovered phases on cluster outage, succeeding with a partial result set and only a stderr warning. Now exits 1 with a message explaining the refusal. Genuinely-empty progress (a fresh run that has not produced any pair data yet) still falls through unchanged — only the unreachable case is refused.
- **`_cmd_run_remote` regression-recovery:** PR #286 inadvertently lost the specific `ConfigMap unreachable — skipping pre-flight filter validation` message in the consolidation. This PR restores it verbatim while keeping the helper as the single source of truth.
- **No change to the corrupt-data path:** `ValueError` continues to call `sys.exit(1)` with the same recovery message, regardless of `allow_unreachable`. Issue #287 is explicitly out of scope for the corrupt path (already addressed in #286).

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 933 passed, 2 xfailed (was 929)
- [x] New: `TestLoadProgressHelper.test_raises_progress_unavailable_when_allow_unreachable` — replaces the old swallow-and-return-`{}` assertion
- [x] New: `TestLoadProgressHelper.test_progress_unavailable_subclasses_runtime_error` — guards the catch-RuntimeError compatibility property
- [x] New: `test_status_unreachable_configmap_exits` — asserts non-zero exit + "unreachable" message and absence of `0 pairs`
- [x] New: `test_collect_unreachable_configmap_exits` — asserts non-zero exit, no extractor invocation, "unreachable" in output, even with a `cluster/` file present that `_discover_phases` would otherwise pick up
- [x] Updated: `test_run_remote_skips_validation_when_configmap_unreachable` — now also asserts the focused pre-flight skip message
- [ ] Live smoke: stop kubectl access and run each of `deploy.py status`, `deploy.py collect`, `deploy.py run --remote` against a setup config; confirm clean errors / warnings (manual, not in CI)

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -114,6 +114,16 @@ def _load_setup_config() -> dict:
 
 # ── Progress store loading ───────────────────────────────────────────────────
 
+class ProgressUnavailable(RuntimeError):
+    """Progress store is reachable-failure (transient kubectl error).
+
+    Distinct from the empty-result case (ConfigMap NotFound or empty data),
+    which ``ConfigMapProgressStore.load`` returns as ``{}``. Only raised by
+    ``_load_progress`` when ``allow_unreachable=True`` so callers can decide
+    whether to abort or degrade. See issue #287.
+    """
+
+
 def _load_progress(store, *, allow_unreachable: bool = False) -> dict:
     """Load progress data with consistent corrupt/unreachable handling.
 
@@ -125,8 +135,10 @@ def _load_progress(store, *, allow_unreachable: bool = False) -> dict:
     guidance, then ``sys.exit(1)``. Corrupt data is never recoverable by
     retry, so this applies regardless of ``allow_unreachable``.
 
-    On ``RuntimeError`` (e.g. kubectl cannot reach the cluster): re-raise by
-    default, or warn and return ``{}`` when ``allow_unreachable=True``.
+    On ``RuntimeError`` (e.g. kubectl cannot reach the cluster): re-raise the
+    original ``RuntimeError`` by default, or raise ``ProgressUnavailable`` when
+    ``allow_unreachable=True`` so callers can distinguish unreachable from
+    legitimate empty progress data (issue #287).
     """
     try:
         return store.load()
@@ -137,8 +149,7 @@ def _load_progress(store, *, allow_unreachable: bool = False) -> dict:
         sys.exit(1)
     except RuntimeError as exc:
         if allow_unreachable:
-            warn(f"Failed to load progress: {exc}")
-            return {}
+            raise ProgressUnavailable(str(exc)) from exc
         raise
 
 
@@ -444,7 +455,12 @@ def _cmd_status(args, run_dir: Path,
         err("No namespace configured. Run setup.py first.")
         sys.exit(1)
     store = ConfigMapProgressStore(primary_ns, run_name=run_dir.name)
-    progress = _load_progress(store, allow_unreachable=True)
+    try:
+        progress = _load_progress(store, allow_unreachable=True)
+    except ProgressUnavailable as exc:
+        err(f"Cluster unreachable — cannot read progress ConfigMap: {exc}")
+        err("Retry once kubectl can reach the cluster.")
+        sys.exit(1)
 
     if not progress:
         suffix = " (no progress data)"
@@ -1080,7 +1096,14 @@ def _cmd_collect(args, run_dir: Path, setup_config: dict):
         err("No namespace configured. Run setup.py first.")
         sys.exit(1)
     store = ConfigMapProgressStore(primary_ns, run_name=run_dir.name)
-    progress = _load_progress(store, allow_unreachable=True) or None
+    try:
+        progress = _load_progress(store, allow_unreachable=True) or None
+    except ProgressUnavailable as exc:
+        err(f"Cluster unreachable — cannot read progress ConfigMap: {exc}")
+        err("Refusing to collect from filesystem-discovered phases without "
+            "authoritative progress data. Retry once kubectl can reach the "
+            "cluster.")
+        sys.exit(1)
 
     # ── Pair-level scoping (--only / --workload) ──────────────────────────
     scope_only = getattr(args, "only", None)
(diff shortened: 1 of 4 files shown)


kalantar:
## Review verdict: READY TO READ
�Ysq�cleanly separates unreachable from empty-progress at the helper boundary, all three callers handle the new exception explicitly, and tests cover the regression in each subcommand.

### Recommendations
- **ProgressUnavailable docstring opens with the awkward phrase 'reachable-failure'** (`pipeline/deploy.py:118`) — deploy.py:118 reads 'Progress store is reachable-failure (transient kubectl error)' — the hyphenated compound parses as 'reachable but failed', which is the inverse of what the class actually represents. A small wording fix (e.g. 'Raised when the progress store cannot be reached') would prevent a future reader misinterpreting the contract.

PR #291, opened 2026-06-04
feat(deploy): add -s/--silent to status to print summary line only

### LINKED ISSUE #290: Support -s for deploy.py status
Currently, deploy.py status returns:

```
━━━ sim2real-deploy ━━━


PAIR                      STATUS       SLOT           RETRIES
-------------------------------------------------------------
wl-balanced-20-baseline   done         —              0
wl-balanced-20-control    done         —              0
wl-balanced-20-quartic    done         —              0
wl-balanced-30-baseline   done         —              0
wl-balanced-30-control    done         —              0
wl-balanced-30-quartic    done         —              0
wl-balanced-40-baseline   running      kalantar-1     0
wl-balanced-40-control    done         —              0
wl-balanced-40-quartic    running      kalantar-0     0
wl-balanced-50-baseline   done         —              0
wl-balanced-50-control    running      kalantar-2     0
wl-balanced-50-quartic    running      kalantar-3     0

  12 pairs: 8 done  4 running
```
The `-s` or `--silent` option would return just the summary line: 

```
  12 pairs: 8 done  4 running
````

kalantar:
## Summary

- Add `-s` / `--silent` flag to `deploy.py status`. When set, suppresses the per-pair table (header, separator, rows, surrounding blank lines) and prints only the trailing summary line — e.g. `12 pairs: 8 done  4 running`.
- The flag composes with `--only` / `--workload` / `--package` / `--status`; the summary reflects the filtered subset.
- The empty-progress path still prints the single `0 pairs (no progress data)` line (with optional `— filters ignored` suffix), unchanged.
- The error path (cluster-unreachable from #287, no-namespace-configured) is unchanged: still exits non-zero with the same messages, regardless of `-s`.

Closes #290.

## Reviewer attention

- **Short-flag namespace check:** `-s` was unused in the status subparser. `--status` (existing scoping flag) shares the leading letter, but argparse handles short vs long flags as distinct namespaces — `-s` is unambiguous, and a hypothetical `--s` would be ambiguous between `--silent` and `--status`. The standard `add_argument("-s", "--silent", ...)` pattern is used; users still write the long flag in full.
- **Backward compatibility:** `_cmd_status` reads the flag via `getattr(args, "silent", False)`, so any caller that constructs a hand-rolled `args` object without the attribute (existing tests, the `live` mode harness) still works without modification. Existing tests added the attribute only to silent-mode tests.
- **Counts loop duplication:** the count-tally is computed in the silent branch and within the per-row print loop in the verbose branch. I considered hoisting it to a shared pre-pass, but that would force two passes over `pairs` in the verbose case (currently one); given pair counts are small, the readability cost wasn't worth it.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 938 passed, 2 xfailed (was 933)
- [x] New: `test_status_silent_prints_summary_only` — asserts summary present, no per-pair rows, no `PAIR`/`STATUS` header
- [x] New: `test_status_silent_composes_with_filter` — `--workload wl-smoke` + `-s` reports `2 pairs`, no rows
- [x] New: `test_status_silent_empty_progress` — empty progress + `-s` still prints `0 pairs` line, no header
- [x] New: `test_status_silent_unreachable_still_exits` — cluster-unreachable + `-s` exits non-zero with the unreachable message (error path unchanged)
- [x] New: `test_status_silent_short_and_long_flag_parse` — `-s` and `--silent` both produce `args.silent=True`; default is `False`
- [ ] Live smoke: `python pipeline/deploy.py status -s` against a live cluster (manual, not in CI)

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -189,6 +189,7 @@ python pipeline/deploy.py pairs   [flags]   # list available pair keys, workload
 | `--workload NAME…` | Filter by workload names (comma or space-separated) |
 | `--package NAME…` | Filter by package names (comma or space-separated) |
 | `--status STATE` | Filter by status (e.g. `running`, `done`, `failed`) |
+| `-s`, `--silent` | Suppress the per-pair table and banner; print only the summary line (machine-readable) |
 
 **`deploy.py collect`** — extracts results from the cluster PVC and writes to `workspace/runs/<run>/results/{phase}/<workload>/`. Repeated collects are incremental: each workload's remote `trace_data.csv` mtime is probed and skipped if the local copy is already up to date. If the mtime probe fails (e.g., pod not running), collection falls back to a full copy — this is the expected degradation path.
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -2952,9 +2952,12 @@ Examples:
 def main():
     parser = build_parser()
     args = parser.parse_args()
-    machine_readable = (args.command == "pairs" and
-                         any(getattr(args, f, False)
-                             for f in ("keys_only", "workloads_only", "packages_only")))
+    machine_readable = (
+        (args.command == "pairs" and
+         any(getattr(args, f, False)
+             for f in ("keys_only", "workloads_only", "packages_only"))) or
+        (args.command == "status" and getattr(args, "silent", False))
+    )
     if not machine_readable:
         print(_c("36", "\n━━━ sim2real-deploy ━━━\n"))
 
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -476,35 +476,45 @@ def _cmd_status(args, run_dir: Path,
         return
 
     pairs = {k: progress[k] for k in _resolve_scope(progress, args)}
+    silent = getattr(args, "silent", False)
 
     if not pairs:
         print("  0 pairs")
-        print()
+        if not silent:
+            print()
         return
 
-    pair_w = max(len(k) for k in pairs) + 2
-    col_status = 12
-    col_slot = 14
-    col_retries = 7
-
-    header = (f"{'PAIR':<{pair_w}} {'STATUS':<{col_status}} {'SLOT':<{col_slot}} {'RETRIES':<{col_retries}}")
-    print()
-    print(header)
-    print("-" * len(header))
-
     counts: dict[str, int] = {}
-    for key, entry in sorted(pairs.items()):
-        status = entry.get("status", "unknown")
-        slot = entry.get("namespace") or "—"
-        retries = entry.get("retries", 0)
-        counts[status] = counts.get(status, 0) + 1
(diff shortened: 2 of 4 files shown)


kalantar:

Behavior is correct and well-tested; banner-suppression is a reasonable follow-up but not a blocker.

### Important
- **`-s` does not suppress the `━━━ sim2real-deploy ━━━` banner, so output isn't purely the summary line the issue describes** (`pipeline/deploy.py:2958`) — deploy.py:2955-2959 only treats `pairs --keys-only/--workloads-only/--packages-only` as machine-readable; `status -s` still prints the cyan banner before the summary. Issue #290's example output is just the summary line, so scripting use cases (e.g. `var=$(deploy.py status -s)`) will pick up the banner unless `main()` also short-circuits when `args.command == "status"` and `args.silent`.

kalantar:

Implementation is correct and well-tested, but the project-mandated README update for the new flag is missing and one early-return path leaks verbose text under -s.

### Must fix
- **README status flag table omits the new -s/--silent flag** (`pipeline/README.md:191`) — pipeline/README.md lines 186-191 list every other status flag but stop at --status STATE; -s/--silent is absent. CLAUDE.md states 'After any change to pipeline/ that affects CLI flags... update pipeline/README.md to match' — this PR adds a CLI flag, so the README update is a project-rule requirement, not optional polish.

### Important
- **Empty-progress early-return prints '(no progress data)' even under -s** (`pipeline/deploy.py:465`) — deploy.py lines 465-476 run before `silent = getattr(args, 'silent', False)` is set on line 479, so `deploy.py status -s` against an empty ConfigMap emits '  0 pairs (no progress data)' rather than the clean '  N pairs: ...' surface the rest of the silent path provides. Scripts doing exact-match parsing on the silent output get inconsistent output between the empty and non-empty cases.

### Recommendations
- **_cmd_status docstring still says 'Print a snapshot table' but silent mode prints no table** (`pipeline/deploy.py:451`) — deploy.py:451 docstring is now stale — with --silent the function prints only the summary line. Worth a one-line touch-up while the file is open.
- **test_status_silent_empty_progress docstring attributes the output to --silent, but the early-return runs before silent is consulted** (`pipeline/tests/test_deploy_run.py:191`) — tests/test_deploy_run.py docstring says '-s prints just the 0 pairs (no progress data) line', but deploy.py:465-476 fires that line regardless of -s — the test is really exercising the not-progress branch. A future maintainer reading the docstring would form the wrong mental model.

kalantar:
Updated pipeline/README.md to document the new -s/--silent flag in the status flag table per CLAUDE.md's CLI-flag rule. The two non-blocking findings (banner suppression and the empty-progress early-return) are tracked separately — banner suppression was already addressed in this PR; empty-progress consistency under -s is left for a follow-up since it's a behavior tweak rather than a docs gap.

kalantar:

The change is small, correct, and well-covered by tests; the surfaced concerns are doc/test polish that don't justify blocking merge.

Issue #274, opened 2026-06-02
Drop backoff state machine; always poll at base interval, skip GPU probe when slots are full


kalantar:
Refiling #273 with a simpler scope after investigating #65, #82, and the actual gating signals in \`_cmd_run\`.

## Problem

The dispatch loop in \`pipeline/deploy.py\` \`_cmd_run\` runs an exponential-backoff state machine (\`pipeline/lib/backoff.py\`) that adds significant code complexity without a documented benefit. Investigation:

- #65 introduced backoff to \"avoid wasted API calls and log noise\" during sustained scarcity. No production incident cited; the cost it avoids (a \`kubectl get nodes\` every 30s) is trivially cheap on the clusters this orchestrator targets.
- The two trigger paths (GPU-probe scarcity at \`deploy.py:2207-2222\`, reclaim signal at \`deploy.py:2114\`) both extend poll intervals up to 600s during sustained scarcity. The recovery signal — \`free_gpus >= max_cost\` of pending pairs — is GPU-keyed and doesn't account for slot exhaustion.
- The most common scenario observed in real logs is \"slots are full because the orchestrator dispatched everything that fits.\" The backoff state machine reads this as external GPU pressure and stretches polling to 8+ minutes, missing slot recovery.
- The base 30s polling interval is itself sufficient as a rate limit against hot-loop reclaim cycles. No additional backoff layer is needed for that protection.

## Observed log

\`\`\`
17:35:59 [WARN]  Scarcity detected: 1 free GPUs, entering backoff (next poll: 60s)
17:37:03 [INFO]  Backoff level 2 — next poll in 120s
17:39:08 [INFO]  Backoff level 3 — next poll in 240s
17:43:13 [INFO]  Backoff level 4 — next poll in 480s
17:51:18 [INFO]  Backoff probe: 7 free GPUs available → resuming normal dispatch
17:51:18 [INFO]  Dispatching 1/19 pending pairs (capacity-limited: 7 free GPUs)
\`\`\`

Slots were full throughout this window. The 8-minute blackout between 17:43:13 and 17:51:18 delayed slot-recovery detection. The 17:51:18 \"resuming normal dispatch\" line fired on GPU numbers that were a downstream effect of the orchestrator's own dispatches completing — the recovery signal didn't actually mean external pressure had eased.

## Proposed fix

**Delete the backoff state machine entirely. Always sleep at \`--poll-interval\` (default 30s). Skip the GPU probe when no slots are free.**

### Code changes

[...]

PR #292, opened 2026-06-04
fix(deploy): unify capacity log; report effective_free in dispatch-denied warn (#272)

### LINKED ISSUE #272: Orchestrator dispatch loop: capacity logs are inconsistent and warn message is misleading
## Problem

The orchestrator dispatch loop in `pipeline/deploy.py` `_cmd_run` emits capacity information in two different formats from two different code paths within the same poll cycle. This makes the log read as inconsistent and obscures what's actually gating dispatch. A separate warn message reports the wrong number for the decision it just made.

> **Update (2026-06-04):** Defect 3 below was superseded by PR #274 (commit `f2e3556`), which wrapped the GPU probe + dispatch in `if free_slots:` and added an explicit "all N slots busy" log. The remaining work is Defects 1 and 2. Line numbers below have been refreshed to current HEAD (`d880def`).

## Observed log

```
17:33:40 [OK]    [wl-codecompletion-mid-exponentialceiling] → kalantar-3
17:33:40 [INFO]  Capacity: 7 effective free GPUs (23 probed − 16 reserved)
17:33:40 [INFO]  wl-codecompletion-under-exponentialceiling requires 4 GPUs
17:33:40 [OK]    [wl-codecompletion-under-exponentialceiling] → kalantar-4
17:34:15 [INFO]  Capacity: 23 free GPUs (48 allocatable − 25 requested)
17:34:15 [WARN]  Dispatching 0/19 pending pairs — smallest cost (4) exceeds free GPUs (23)
```

Two unrelated capacity formats appear 35 seconds apart. The 17:34:15 warn says "4 exceeds 23" — which is mathematically false. Reading the code, the dispatch decision was actually made on `effective_free` (probed minus shadow-ledger reservations), which was ≈0 because the shadow ledger still held the recent dispatches under its 120s TTL. The warn just interpolates the wrong variable.

## Defect 1 — two capacity formats from the same poll cycle

`pipeline/deploy.py:2267` (top of poll cycle, just after probe):

```python
info(f"Capacity: {free_gpus} free GPUs ({allocatable} allocatable − {requested} requested)")
```

`pipeline/deploy.py:2321` (inside dispatch loop, just before applying a PipelineRun):

```python
info(f"Capacity: {_effective} effective free GPUs ({free_gpus} probed − {_reserved} reserved)")
```

The first never mentions the shadow ledger; the second never mentions the cluster numbers. Whether you see one or the other depends on whether a dispatch happens this cycle. Reading a single log line, you cannot tell which view you're seeing without checking what came right after.

## Defect 2 — dispatch-denied warn reports the wrong number

`pipeline/deploy.py:2292`:

```python
warn(f"Dispatching 0/{len(pending)} pending pairs — smallest cost ({smallest}) exceeds free GPUs ({free_gpus})")
```

This warn fires inside the `if len(dispatchable) == 0 and pending:` branch at `deploy.py:2287-2293`, where `dispatchable` was just computed against `effective_free` (line 2282-2286). The denial reason is that `smallest > effective_free`, but the warn message reports `free_gpus` (the probed number). When the shadow ledger is the gating factor, the reported math doesn't add up.

## Defect 3 — superseded

Originally read as: "no log when all slots are busy." Resolved by PR #274 — `deploy.py:2257` now wraps the GPU probe + dispatch in `if free_slots:`, and `deploy.py:2378-2382` emits `"Dispatching 0/<pending> pending — all <total> slots busy"` when `free_slots` is empty. No code change required for this defect.

## Proposed fix

**Unify the capacity log to one format that always shows all four numbers.** Both emission points become:

```
Capacity: <effective> effective free GPUs (<probed> probed − <reserved> reserved; cluster: <allocatable> allocatable − <requested> requested)
```

When the ledger is empty the reservation term is `0` and the line still reads cleanly. The reader sees the same fields in the same order regardless of when the line was emitted.

**Update the dispatch-denied warn to use `effective_free` and name the reservation context:**

```
Dispatching 0/<pending> pending pairs — smallest cost (<min_cost>) exceeds <effective_free> effective free GPUs (<probed> probed − <reserved> reserved)
```

## Out of scope

- No change to the dispatch decision itself. `effective_free` is already the right gate at line 2285; only the log strings change.
- No change to backoff state machine.
- No change to `probe_free_gpus` or `extract_node_filters`.
- No change to the all-slots-busy log path (already correct after #274).

## Tests

- Capture orchestrator stdout in a fake-cluster integration test that drives `_cmd_run` through one cycle with shadow reservations present. Assert the cycle emits exactly one `"Capacity:"` format containing all four numbers, and that the dispatch-denied warn (when triggered) reports `effective_free` rather than `free_gpus`.

## Related

- #255 — introduced the shadow GPU reservation ledger
- #261 — original capacity-probe over-counting
- #268 / #269 — current capacity-probe filter fix
- #274 — superseded Defect 3 (skip GPU probe when slots full; explicit "all slots busy" log)

kalantar:
## Summary

Fixes two log-quality defects in the orchestrator dispatch loop in `pipeline/deploy.py`:

- **Defect 1** — The `Capacity:` log used two distinct format strings emitted from two different sites within the same poll cycle. Whichever you saw depended on whether a dispatch happened that cycle, and the formats did not share fields.
- **Defect 2** — The dispatch-denied `WARN` interpolated `free_gpus` (the probed number) when reporting the denial, but the gating decision had actually been made on `effective_free` (probed minus shadow-ledger reservations). When the shadow ledger was the gating factor, the printed math did not add up (e.g. `\"smallest cost (4) exceeds free GPUs (23)\"` reported "23" while the gate had actually evaluated against ≈0).

Defect 3 from the issue (no log when all slots busy) was already addressed by #274 and is not touched here.

Closes #272.

## What changed

**`pipeline/deploy.py`**

- New module-level helper `_format_capacity(effective, probed, reserved, allocatable, requested)` returns the unified line:

  ```
  Capacity: <effective> effective free GPUs (<probed> probed − <reserved> reserved; cluster: <allocatable> allocatable − <requested> requested)
  ```

  Both emission sites — the post-probe site and the pre-apply site inside the dispatch for-loop — now go through this helper, so a future format change must touch one place.

- Dispatch-denied warn rewritten to:

  ```
  Dispatching 0/<pending> pending pairs — smallest cost (<min>) exceeds <effective_free> effective free GPUs (<probed> probed − <reserved> reserved)
  ```

  Its dedup state tuple now keys on `(effective_free, reserved)` rather than `free_gpus`, so a draining shadow ledger correctly re-emits the warning when the gating answer changes.

**`pipeline/tests/test_log_dedup.py`**

- Mocks updated to the new format (parameter list now includes `effective`, `reserved`).
- Added `test_changed_reservation_logs_again` — same probe but changing reservation re-emits the capacity line.
- Added `test_zero_dispatch_includes_reservation_in_state` — regression for the dedup-keys-on-effective_free fix in Defect 2.
- Added `TestUnifiedCapacityFormat` — pins the exact output of `_format_capacity` and confirms the empty-ledger case still reads cleanly.

## Reviewer attention

- **Drift prevention.** The whole point of this fix is that two callers must produce the same string. I extracted a small helper rather than duplicating the format literal — please push back if the helper feels over-engineered, but the alternative is to leave two literals in sync by convention, which is exactly the failure mode #272 documents.
- **`max(0, probed - reserved)` vs `shadow.effective_free(probed)`.** The pre-apply site previously called `shadow.effective_free(free_gpus)` (which prunes the ledger again). I changed both sites to compute `_effective = max(0, free_gpus - _reserved)` after a single `shadow.reserved()` call, to avoid double-pruning. Behavior is identical given the same input.
- **`allocatable`/`requested` scoping at Site B.** Site B (inside the for-loop) reads `allocatable` and `requested` from the enclosing function scope. They are only set inside the `isinstance(capacity, tuple)` branch, but Site B is guarded by `if free_gpus is not None:` — `free_gpus` is also only set in that same branch, so the variables are guaranteed in scope when Site B fires. Probe-failure path is unchanged.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/` — 944 passed, 2 xfailed.
- [ ] Manual: run a real `deploy.py run` cycle and eyeball that the capacity line is identical at both emission points and that the denial warn shows effective_free.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -61,10 +61,13 @@ def _format_capacity(effective: int, probed: int, reserved: int,
                      allocatable: int, requested: int) -> str:
     """Return the unified orchestrator capacity log line (issue #272).
 
-    All five numbers (three ledger: effective/probed/reserved; two cluster:
-    allocatable/requested) appear in a fixed order so a single log line is
+    All five numbers appear in a fixed order so a single log line is
     self-contained — the reader does not need surrounding context to know
-    which view they are seeing.
+    which view they are seeing. The five numbers split across three layers:
+
+      - effective: derived (probed − reserved, clamped at 0)
+      - probed, allocatable, requested: cluster probe (`probe_free_gpus`)
+      - reserved: shadow ledger
     """
     return (
         f"Capacity: {effective} effective free GPUs "
@@ -2276,13 +2279,17 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 gpu_resource_type=gpu_resource_type,
                 node_filters=list(node_filters.values()) or [NodeFilter()],
             )
+            # Snapshot the shadow ledger once per cycle. Every log line and
+            # gating decision in this cycle uses the same _reserved value,
+            # so `free_gpus − _reserved == effective_free` holds in every
+            # printed line (issue #272).
             if isinstance(capacity, tuple):
                 free_gpus, allocatable, requested = capacity
                 _reserved = shadow.reserved()
-                _effective = max(0, free_gpus - _reserved)
-                _cap_state = (_effective, free_gpus, _reserved, allocatable, requested)
+                effective_free = max(0, free_gpus - _reserved)
+                _cap_state = (effective_free, free_gpus, _reserved, allocatable, requested)
                 if pending and _cap_state != _last_log_state.get("capacity"):
-                    info(_format_capacity(_effective, free_gpus, _reserved,
+                    info(_format_capacity(effective_free, free_gpus, _reserved,
                                           allocatable, requested))
                     _last_log_state["capacity"] = _cap_state
                 if _probe_fail_count > 0:
@@ -2291,6 +2298,8 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 _last_probe_error = ""
             else:
                 free_gpus = None
+                _reserved = 0
+                effective_free = 0
                 _probe_fail_count += 1
                 if capacity != _last_probe_error or _probe_fail_count % 10 == 0:
                     warn(f"Capacity probe failed: {capacity} — dispatching without GPU gating")
@@ -2298,14 +2307,12 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
 
             # ── Assign pending work to free slots ────────────────────────
             if free_gpus is not None and pending:
-                effective_free = shadow.effective_free(free_gpus)
                 dispatchable = _select_dispatchable(
                     pending,
                     free_gpus=effective_free, cost_map=pair_costs,
                 )
                 if len(dispatchable) == 0 and pending:
                     smallest = min(pair_costs[k] for k in pending)
-                    _reserved = shadow.reserved()
                     _disp_state = ("zero", len(pending), effective_free, _reserved, smallest)
                     _zero_dispatch_count += 1
                     if _disp_state != _last_log_state.get("dispatch") or _zero_dispatch_count % 10 == 0:
@@ -2340,9 +2347,7 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
                 source = pair_provenance[pair_key]
                 _source_labels = {"derived": "derived from scenarioContent", "defaults-only": "derived from defaults only", "fallback": "fallback default"}
                 if free_gpus is not None:
-                    _reserved = shadow.reserved()
-                    _effective = max(0, free_gpus - _reserved)
-                    info(_format_capacity(_effective, free_gpus, _reserved,
+                    info(_format_capacity(effective_free, free_gpus, _reserved,
                                           allocatable, requested))
                 info(f"{pair_key} requires {pair_cost} GPUs ({_source_labels[source]})")
                 pr_meta = discovered.get(pair_key, {})
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -61,9 +61,10 @@ def _format_capacity(effective: int, probed: int, reserved: int,
                      allocatable: int, requested: int) -> str:
     """Return the unified orchestrator capacity log line (issue #272).
 
-    All four cluster + ledger numbers appear in a fixed order so a single
-    log line is self-contained — the reader does not need surrounding
-    context to know which view they are seeing.
+    All five numbers (three ledger: effective/probed/reserved; two cluster:
+    allocatable/requested) appear in a fixed order so a single log line is
+    self-contained — the reader does not need surrounding context to know
+    which view they are seeing.
     """
     return (
         f"Capacity: {effective} effective free GPUs "
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -57,6 +57,21 @@ def _is_pair_key(key: str) -> bool:
     return not key.startswith("_")
 
 
+def _format_capacity(effective: int, probed: int, reserved: int,
+                     allocatable: int, requested: int) -> str:
+    """Return the unified orchestrator capacity log line (issue #272).
+
+    All four cluster + ledger numbers appear in a fixed order so a single
+    log line is self-contained — the reader does not need surrounding
+    context to know which view they are seeing.
+    """
+    return (
+        f"Capacity: {effective} effective free GPUs "
+        f"({probed} probed − {reserved} reserved; "
+        f"cluster: {allocatable} allocatable − {requested} requested)"
+    )
+
+
 def step(n, title: str) -> None:
     print("\n" + _c("36", f"━━━ Step {n}: {title} ━━━"))
 
@@ -2262,9 +2277,12 @@ def _cmd_run(args, run_dir: Path, setup_config: dict) -> None:
             )
             if isinstance(capacity, tuple):
                 free_gpus, allocatable, requested = capacity
-                _cap_state = (free_gpus, allocatable, requested)
+                _reserved = shadow.reserved()
+                _effective = max(0, free_gpus - _reserved)
+                _cap_state = (_effective, free_gpus, _reserved, allocatable, requested)
                 if pending and _cap_state != _last_log_state.get("capacity"):
-                    info(f"Capacity: {free_gpus} free GPUs ({allocatable} allocatable − {requested} requested)")
+                    info(_format_capacity(_effective, free_gpus, _reserved,
+                                          allocatable, requested))
                     _last_log_state["capacity"] = _cap_state
                 if _probe_fail_count > 0:
                     info(f"Capacity probe recovered after {_probe_fail_count} failure(s)")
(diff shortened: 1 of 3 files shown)


kalantar:

Core fixes are correct but the issue's stated integration-test acceptance criterion is unmet, and a docstring/test-name miscount documents the wrong invariant.

### Important
- **Integration test driving _cmd_run through one cycle with shadow reservations is missing despite being a stated acceptance criterion in the issue** (`pipeline/tests/test_log_dedup.py`) — Issue #272 Tests section explicitly requires: 'Capture orchestrator stdout in a fake-cluster integration test that drives _cmd_run through one cycle with shadow reservations present. Assert the cycle emits exactly one Capacity: format containing all four numbers, and that the dispatch-denied warn (when triggered) reports effective_free rather than free_gpus.' The PR ships unit tests of _format_capacity and copy-of-the-dedup state machine, but none drives _cmd_run end-to-end. A future refactor could call only one of the two production emission sites or pass wrong args without the unit tests catching it. test_deploy_run.py already has the pattern (monkeypatched probe_free_gpus, shadow ledger setup) for adding this.
- **Docstring on _format_capacity claims 'four numbers' but the function takes and emits five** (`pipeline/deploy.py:64`) — deploy.py:64 docstring says 'All four cluster + ledger numbers appear in a fixed order' but the signature is _format_capacity(effective, probed, reserved, allocatable, requested) — five distinct numbers, all five present in the output. The test test_log_dedup.py:271 test_format_includes_all_four_numbers_in_fixed_order also claims four while the assertion at 272-277 verifies five values (7, 23, 16, 48, 25). The whole point of the fix is that the unified format names every relevant quantity; mis-stating the count in the spec docstring and the named test invariant undermines that contract for future readers.

### Recommendations
- **Zero-dispatch warn test helpers omit the '(probed − reserved)' parenthetical, so the production warn string is not fully pinned by any test** (`pipeline/tests/test_log_dedup.py:155`) — Tests at test_log_dedup.py:155-159, 186, 209 emit only 'effective free GPUs' without the parenthetical, while the production warn at deploy.py:2311-2315 includes '({free_gpus} probed − {_reserved} reserved)'. TestUnifiedCapacityFormat pins the capacity info line but no test pins the warn line, so a future change could silently drop the parenthetical from the war

kalantar:
Addressed both important findings from the review.

Added an integration test in `test_deploy_run.py` that drives `_cmd_run` through one cycle with the shadow ledger pre-seeded to 20 reserved GPUs against a 23-probed/48-allocatable/25-requested probe. The test asserts the cycle emits exactly one `Capacity:` line containing all five numbers in the unified format, and that the dispatch-denied warn reports `effective_free=3` rather than the raw `free_gpus=23`. This pins both production emission sites (and the warn variable) against silent regression.

Fixed the "four numbers" miscount: `_format_capacity` actually emits five (three ledger + two cluster), so the docstring on `deploy.py:64`, the test method name `test_format_includes_all_four_numbers_in_fixed_order`, and the docstring on `test_zero_dispatch_includes_reservation_in_state` (which was also incomplete — the dedup tuple keys on both `effective_free` AND `reserved`, not just one) are corrected.

The single recommendation in the verdict (zero-dispatch warn parenthetical not fully pinned by unit-test mocks) is now covered by the new integration test, which asserts on `"23 probed"` and `"20 reserved"` substrings of the actual production warn line.

945 pytest passed (up from 944), ruff clean.

kalantar:

The unification fix is correct and well-tested, but the dispatch-denied warn still computes `effective_free` and `_reserved` from two separate shadow queries, which contradicts the PR's stated single-snapshot design and can produce log arithmetic that doesn't add up — exactly the user-confusion class this issue set out to eliminate.

### Important
- **Dispatch-denied warn uses two separate shadow queries; arithmetic in the printed line can be internally inconsistent** (`pipeline/deploy.py:2308`) — Line 2301 computes `effective_free = shadow.effective_free(free_gpus)` (which calls `reserved()` internally), then line 2308 calls `shadow.reserved()` a second time. The warn message at lines 2312-2316 prints both numbers as `{effective_free} effective free GPUs ({free_gpus} probed − {_reserved} reserved)`. With TTL=120s and per-cycle gaps of microseconds the practical risk is tiny, but the PR body explicitly states the design is to use `max(0, free_gpus - _reserved)` after a single `shadow.reserved()` call — and that pattern was applied to the post-probe site (2281-2283) and pre-apply site (2342-2344) but NOT here. Issue #272 is precisely about log lines whose math doesn't add up; leaving this site with an unsnapshotted recomputation reintroduces the same bug class the PR is fixing.

### Recommendations
- **_format_capacity docstring labels `probed` as a ledger number when it is a cluster probe** (`pipeline/deploy.py:64`) — Docstring says 'three ledger: effective/probed/reserved'. Verified at line 60-67: `probed` is the raw GPU count from `probe_free_gpus`, not a ledger value; only `reserved` comes from the shadow ledger and `effective` is derived. Worth a one-line correction since this PR's whole purpose is to make the cluster vs. ledger distinction explicit in the log.

kalantar:
Iteration 2 review came back NEEDS DISCUSSION on the unsnapshotted shadow query in the dispatch-denied warn. Going with the option that fixes all three sites.

Now snapshotting `shadow.reserved()` once at the top of the `if free_slots:` block and threading the snapshot through every consumer in the cycle: the post-probe Capacity log, the gating decision (replaces `shadow.effective_free()` with `max(0, free_gpus - _reserved)`), the dispatch-denied warn, and the pre-apply per-dispatch Capacity log. After this change every line in a cycle satisfies `probed − reserved == effective_free` exactly.

One behavior change worth flagging: the per-dispatch log inside the for-loop no longer refreshes after `shadow.record(pair_cost)` updates between dispatches in the same cycle. Each pair logged in the same cycle shows the cycle-top snapshot — what the orchestrator saw when it decided which pairs were dispatchable. The running-balance narrative in the issue's example log is gone; the line now answers "why was this pair dispatched" rather than "what's the live ledger state." Open to feedback if you want the running balance back as a separate line.

Also fixed the docstring recommendation: probed/allocatable/requested are cluster-probe values, not ledger values.

Extended the integration test to parse both the Capacity line and the dispatch-denied warn with regex and assert (a) the same probed/reserved/effective triple appears in both lines and (b) effective == max(0, probed − reserved). Pins the snapshot invariant against future regressions.

945 tests passed, ruff clean.

kalantar:

Both named defects are correctly fixed at the right altitude with regression tests; nothing in the diff blocks merge.

### Recommendations
- **The cap_limited partial-dispatch log at deploy.py:2328 still prints raw `free_gpus` and keys its dedup tuple on `free_gpus`, leaving the same smell this PR removes from the sibling zero-dispatch warn.** (`pipeline/deploy.py:2328`) — Verified at deploy.py:2326-2328: `_disp_state = ("cap_limited", len(dispatchable), len(pending), free_gpus)` and the message reads `capacity-limited: {free_gpus} free GPUs`. The actual gate at line 2312 is `effective_free`, so when the shadow ledger is non-zero the printed number does not match the decision criterion. The issue body lists only Defects 1 and 2 and excludes this branch from scope, so it is genuinely optional, but a reviewer who just unified the other two sites would naturally note the third.

PR #294, opened 2026-06-04
feat(sim2real-bootstrap): bare-flag prefix-caching input + symmetric output

kalantar:
## Summary

Expanded scope of #293 after design discussion. The bootstrap script now handles `enable_prefix_caching` symmetrically:

**Input (`config.md`) — accept either form:**

| Row label                        | Value column     | Meaning |
|----------------------------------|------------------|---------|
| `enable_prefix_caching`          | `true` / `false` | Caching ON / OFF (legacy keyed) |
| `--enable-prefix-caching` (alias)| `true` / `false` | Caching ON / OFF (legacy keyed) |
| `--enable-prefix-caching`        | empty            | Caching ON (bare flag) |
| `--no-enable-prefix-caching`     | empty            | Caching OFF (bare flag) |

Multiple agreeing rows are reconciled. Contradictory specifications (e.g. both bare flags present) error with the observed sources.

**Output (scenario YAML) — always a single bare flag:**

| Resolved state | Emitted flag |
|----------------|--------------|
| ON             | `--enable-prefix-caching` |
| OFF            | `--no-enable-prefix-caching` |
| Field absent   | Emit nothing (defer to vLLM's per-model default) |

This brings the script into alignment with vLLM's CLI conventions (see `vllm/engine/arg_utils.py:_set_default_chunked_prefill_and_prefix_caching_args`, which resolves an unset value to `model_config.is_prefix_caching_supported`).

Closes #293

## Test plan

- [x] `ruff check pipeline/ .claude/skills/ --select F` — clean
- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-translate/tests/ -v` — 940 passed
- [x] Inline behavioral check: legacy true/false, bare positive, bare negative, absent, and agreeing-duplicate cases all produce the expected output flag
- [x] Conflict path: contradictory bare flags exit 1 with a message naming both source rows

PR #296, opened 2026-06-05
fix(sim2real-bootstrap): default to --enable-prefix-caching when config.md is silent

### LINKED ISSUE #295: sim2real-bootstrap: default to --enable-prefix-caching when config.md is silent
## Background

PR #294 (closes #293) made `sim2real-bootstrap` symmetric over `enable_prefix_caching`:

- Input in `config.md` accepts legacy keyed rows, bare `--enable-prefix-caching`, or bare `--no-enable-prefix-caching`.
- Output in scenario YAML is a single bare flag.
- **When `config.md` is silent on the field, the bootstrap emits nothing**, deferring to vLLM's per-model default (`vllm/engine/arg_utils.py:_set_default_chunked_prefill_and_prefix_caching_args`, which resolves to `model_config.is_prefix_caching_supported`).

## Problem

The per-model default resolution is a recent vLLM change. The vLLM version we deploy against predates it, so an unset value does **not** resolve to "ON for supported models" — it falls back to the older static default, which is not what we want for our scenarios.

## Proposed change

When `config.md` mentions neither `enable-prefix-caching` nor `no-enable-prefix-caching`, `sim2real-bootstrap` should emit `--enable-prefix-caching` in the scenario YAML (instead of emitting nothing).

Updated absent-field row:

| Resolved state | Emitted flag |
|----------------|--------------|
| ON             | `--enable-prefix-caching` |
| OFF            | `--no-enable-prefix-caching` |
| Field absent   | `--enable-prefix-caching` (was: emit nothing) |

The two explicit forms (`enable-prefix-caching` ON/OFF and `no-enable-prefix-caching`) keep their PR #294 semantics.

## Notes

- Once we move to a vLLM version that ships the per-model default, we may want to revisit and restore the "emit nothing" behavior for absent fields.
- Tests and the `config.md` documentation in the bootstrap skill should be updated to reflect the new default.

Refs: #293, #294

kalantar:
## Summary

Closes #295.

The deployed vLLM version predates per-model default resolution for `enable_prefix_caching` (`vllm/engine/arg_utils.py:_set_default_chunked_prefill_and_prefix_caching_args`). PR #294 emitted nothing for an absent `enable_prefix_caching` row in `config.md`, which on the deployed version falls back to OFF rather than ON — not the intent for our scenarios.

This PR flips the absent-field default to `--enable-prefix-caching`. The two explicit forms (legacy keyed and bare flags, both ON and OFF) keep their PR #294 semantics.

| Resolved state | Emitted flag |
|----------------|--------------|
| ON             | `--enable-prefix-caching` (unchanged) |
| OFF            | `--no-enable-prefix-caching` (unchanged) |
| Field absent   | `--enable-prefix-caching` (was: emit nothing) |

## Reviewer attention

- Provenance: the synthesized flag carries the source string `"sim2real-bootstrap default (config.md silent; deployed vLLM requires explicit ON)"` so it's distinguishable in the YAML provenance comments from a flag that came from `config.md`.
- `extract_fields` was deliberately not changed — it still does not synthesize an `enable_prefix_caching` entry when the field is absent. Only `build_additional_flags`'s output differs. This keeps `extract_fields` purely a parser of input and confines the policy choice to the output stage.
- The "Notes" section of #295 flags that this default should be revisited if/when we move to a vLLM version that ships per-model default resolution. The code comment + SKILL.md text both reference #295 so the rationale is discoverable.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-bootstrap/tests/ .claude/skills/sim2real-translate/tests/` — 964 passed, 2 xfailed
- [x] Two existing absent-field tests flipped to assert the new default; all explicit-form, contradiction, and unparseable-boolean tests unchanged and passing

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -178,7 +178,7 @@ Generate llm-d-benchmark scenario YAML(s) for baseline arm(s).
    The script will:
    - Parse the vLLM configuration table from config.md
    - Apply MODEL_METADATA and HARDWARE_LABELS lookups
-   - Emit a single bare prefix-caching flag matching the user's intent: `--enable-prefix-caching` if explicitly enabled, `--no-enable-prefix-caching` if explicitly disabled, nothing if unspecified in config.md (defers to vLLM's per-model default). Input accepts either the legacy keyed form (`enable_prefix_caching | true|false`) or a bare flag row label (`--enable-prefix-caching` / `--no-enable-prefix-caching` with an empty value column); contradictory specifications error.
+   - Emit a single bare prefix-caching flag matching the user's intent: `--enable-prefix-caching` if explicitly enabled, `--no-enable-prefix-caching` if explicitly disabled, `--enable-prefix-caching` (with a `sim2real-bootstrap default` provenance source) if unspecified in config.md. The deployed vLLM version predates per-model default resolution, so silent → OFF rather than ON; defaulting to ON keeps caching enabled for our scenarios (see issue #295). Input accepts either the legacy keyed form (`enable_prefix_caching | true|false`) or a bare flag row label (`--enable-prefix-caching` / `--no-enable-prefix-caching` with an empty value column); contradictory specifications error.
    - Write YAML with inline provenance comments showing each value's source
    - Error on unknown models/hardware (update lookup tables in the script if needed)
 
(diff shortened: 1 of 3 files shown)


kalantar:

Tightly scoped behavioral flip from "emit nothing when silent" to "default to --enable-prefix-caching with bootstrap provenance," fully covered by updated unit tests.

PR #323, opened 2026-06-11
feat(deploy): redact Secret data in collected plans and resource manifests

### LINKED ISSUE #312: Redact sensitive object data when collecting plan YAMLs
## Problem

`deploy.py collect` pulls llm-d-benchmark plan YAMLs from the cluster PVC into `results/<phase>/plans/<flow>/`. Some scenarios render Kubernetes `Secret` objects into these files (e.g. HuggingFace tokens, image pull credentials, TLS material). The collected artifacts end up on developer laptops, in shared run directories, and potentially in commits — none of which should hold credentials.

The plan files are valuable for configuration analysis, so the goal is **redact the sensitive content while preserving the object structure**, not drop whole files.

## Proposed change

Add a YAML-aware redaction pass in `_extract_phase_plans` (deploy.py:1119–1175), after `kubectl cp` lands the files locally and before they're written to the run dir. The pass walks every YAML document in every plan file and stubs out sensitive field values for docs whose `kind` matches a denylist.

### Default behavior (`kind: Secret`)

For each `kind: Secret` document:

- Keep `apiVersion`, `kind`, `type`, `metadata` (name, namespace, labels, annotations, ownerReferences) — these describe the object's role and identity, useful for config analysis.
- Replace every value under `data:` and `stringData:` with the literal string `REDACTED` (preserving the keys).
- Prepend a comment marker to the file: `# REDACTED by sim2real collect: 1 Secret stubbed`.

Example before:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: hf-token
  namespace: jchen4
type: Opaque
data:
  token: aGZfYWJjMTIzZGVmNDU2...
```

After:

```yaml
# REDACTED by sim2real collect: 1 Secret stubbed
apiVersion: v1
kind: Secret
metadata:
  name: hf-token
  namespace: jchen4
type: Opaque
data:
  token: REDACTED
```

The Secret still exists in the file as a parseable object with its name, namespace, type, and the *names* of the data keys (often diagnostic — "this Secret carried a `token` and a `ca.crt`" is a useful fact). Only the values are gone.

### Multi-doc and non-Kubernetes files

- Multi-doc YAML (`---`-separated) is handled per-document. Non-Secret docs in a mixed file pass through unchanged.
- Helmfile values files and other plain-config YAMLs (no `kind:` field) are passed through unchanged in v1. Helm values can carry credentials too, but matching them requires key-name heuristics (separate problem — see "Out of scope").

### Configurable kinds (no filename hardcoding)

A small constant list at the top of the redactor:

```python
DEFAULT_REDACT_KINDS = {"Secret"}
```

Future kinds (e.g. `ConfigMap` if a specific scenario starts shipping tokens via configmap data) can be added by editing the constant. Filename-independent — survives any llmdbenchmark renaming.

Optional CLI flag: `--redact-kinds Secret,ConfigMap` to extend at the call site without code changes. Default unset → use `DEFAULT_REDACT_KINDS`.

## Acceptance criteria

- [ ] Every `kind: Secret` document in collected plan files has its `data`/`stringData` values replaced with `REDACTED`
- [ ] Metadata, type, and key names within `data`/`stringData` are preserved
- [ ] Multi-doc files: only the matching docs are stubbed; others pass through unchanged
- [ ] Files without `kind:` (helmfile values, config.yaml) are passed through unchanged
- [ ] A `# REDACTED by sim2real collect: ...` header is present on every file that had at least one doc redacted
- [ ] Tests cover: standalone Secret, multi-doc Secret+ConfigMap, Secret with `stringData` only, Secret with no data fields, files without `kind:`
- [ ] No regression on non-plan extraction paths (trace files, logs, calibration data)

## Out of scope (deferrable to follow-ups if real cases surface)

- **ConfigMap data redaction.** Some configmaps carry tokens. Easy to add later — same kind-denylist mechanism.
- **Container env-value redaction.** Pods/Deployments/Jobs sometimes have credentials inlined as `env[*].value` (instead of `valueFrom.secretKeyRef`). Walking container specs is more invasive than the doc-level filter; skip for v1.
- **Helmfile values key-name redaction.** Regex over keys like `(?i)token|secret|password|credential` in non-Kubernetes YAMLs. Harder to scope (false positives mean unredacted values, false negatives mean broken configs). Skip for v1.

If any of these surface as a real leak, file a follow-up issue — the redaction infrastructure from this issue extends naturally.

## Implementation notes

- Use `yaml.safe_load_all` / `yaml.safe_dump_all` to handle multi-doc files. Preserve key order with the round-trip dumper if available, or accept reformat.
- Walk happens after `kubectl cp` completes, before the file lands in `results/`. Operate on a temp directory and move into place atomically when done — avoid leaving half-redacted files on partial failure.
- Failure to parse a YAML file (e.g. it isn't actually YAML) → leave the file untouched, log a warning. Don't fail the collect.

kalantar:
Closes #312

## Summary

- Adds `pipeline/lib/redact.py`: a YAML-aware redactor that stubs `data`/`stringData` values in every doc whose `kind` is in `DEFAULT_REDACT_KINDS` (currently `{"Secret"}`). Exposes `redact_yaml_file(path)` and `redact_yaml_tree(root)`.
- Wires `redact_yaml_file` into `_extract_phase_plans` (per-yaml call after `kubectl cp` success) — covers the path called out in the issue.
- Wires `redact_yaml_tree` into `_extract_phases_from_pvc` after the resources directory cp (added by #315) — covers the resource manifest tree, which can also contain Secret docs.
- Multi-doc-safe, per-file atomic write, non-fatal on parse / IO failure.

## Reviewer attention

- **Scope expanded beyond the issue text.** Issue #312 was written before #315 shipped the `resources/` collection path. That new path can ship the same kinds of K8s Secret manifests, so this PR redacts both — leaving #315's path leaking would be worse than not having a redactor at all. Mentioned here so reviewers don't get surprised by the second wire-up.
- **Default scope is `{"Secret"}` only.** ConfigMap data redaction is explicitly out of scope per the issue (some configmaps carry tokens, but key-name heuristics are the right tool for that and would over-redact). Adding more kinds is a one-line constant edit.
- **No CLI flag in v1.** The issue lists `--redact-kinds` as optional. Skipped here to minimize surface area; add when a real call site needs it.
- **Idempotency:** a second pass on an already-redacted file changes nothing and does not rewrite (covered by `test_idempotent_on_already_redacted`).
- **Failure mode:** unreadable / unparseable files are logged and skipped — collect never fails because of redaction.
- **Atomic write:** sibling tmp file + `os.replace` ensures no half-redacted files on crash.

## Test plan

- [x] `python -m pytest pipeline/tests/test_redact.py pipeline/tests/test_deploy_collect_plans.py -v` — 19 passed
- [x] `python -m pytest pipeline/ -v` — 895 passed, 2 xfailed
- [x] Full CI command: `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-bootstrap/tests/ .claude/skills/sim2real-translate/tests/` — 990 passed

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -50,6 +50,7 @@ def _c(code: str, text: str) -> str:
 
 
 from pipeline.lib.log import info, ok, warn, err
+from pipeline.lib.redact import redact_yaml_file, redact_yaml_tree
 
 
 def _is_pair_key(key: str) -> bool:
@@ -1023,6 +1024,8 @@ def _extract_phases_from_pvc(phases: list[str], run_name: str, namespace: str,
                             check=False, capture=True)
                     if r.returncode != 0 and "no such file" not in r.stderr.lower():
                         wl_errors.append(f"{wl_name}/resources: {r.stderr.strip()}")
+                    else:
+                        redact_yaml_tree(res_dest)
                     if wl_errors:
                         phase_errors.extend(wl_errors)
                     if on_workload_done:
@@ -1243,6 +1246,7 @@ def _extract_phase_plans(pod_name: str, run_name: str, phase: str,
                 copy_errors.append(f"{yaml_name}: {cp_result.stderr.strip()}")
             else:
                 copied += 1
+                redact_yaml_file(local_path)
         if copy_errors:
             warn(f"[{phase}/plans/{flow}] copy errors: {'; '.join(copy_errors)}")
         if copied or skipped:
--- a/pipeline/lib/redact.py
+++ b/pipeline/lib/redact.py
@@ -1,0 +1,124 @@
+"""YAML redaction for collected plan files.
+
+Stubs out sensitive field values in Kubernetes objects whose `kind`
+matches a denylist, so collected plan YAMLs do not carry credentials
+into developer laptops or shared run dirs.
+
+Behavior:
+  - `data` and `stringData` values in matching docs are replaced with
+    the literal string ``REDACTED``. Key names are preserved.
+  - Multi-doc YAML files are processed per-document; non-matching docs
+    pass through unchanged.
+  - Files without any matching docs are not rewritten.
+  - Unreadable / unparseable files are left untouched (warning logged).
+  - Writes go through a sibling tmp file + atomic ``os.replace`` so a
+    process crash mid-write cannot leave a half-redacted file on disk.
+"""
+from __future__ import annotations
+
+from collections import Counter
+from pathlib import Path
+from typing import Iterable
+
+import yaml
+
+from pipeline.lib.log import warn
+
+REDACTED = "REDACTED"
+
+DEFAULT_REDACT_KINDS: frozenset[str] = frozenset({"Secret"})
+
+
+def _stub_data_fields(doc: dict) -> bool:
+    """Replace every value under data/stringData with REDACTED in-place.
+
+    Returns True if any value was actually changed. Already-redacted
+    values are left as-is and do not count as a change (so an
+    already-redacted file produces a 0-count pass and skips rewrite).
+    """
+    changed = False
+    for field in ("data", "stringData"):
+        section = doc.get(field)
+        if not isinstance(section, dict):
+            continue
+        for key, value in list(section.items()):
+            if value != REDACTED:
+                section[key] = REDACTED
+                changed = True
+    return changed
+
+
+def _format_header(counts: Counter) -> str:
+    parts = []
+    for kind, n in sorted(counts.items()):
+        suffix = "s" if n != 1 else ""
+        parts.append(f"{n} {kind}{suffix} stubbed")
+    return f"# REDACTED by sim2real collect: {', '.join(parts)}\n"
+
+
+def redact_yaml_file(path: Path, kinds: Iterable[str] | None = None) -> int:
+    """Redact data/stringData values for kind-matching docs in `path`.
+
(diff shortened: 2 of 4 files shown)


kalantar:
## Review verdict: READY TO READY TO MERGE

Solid, well-tested redaction with one minor coverage gap on the resources-tree wire-up that does not block merge.

### Recommendations
- **Consider adding an integration test for the redact_yaml_tree call on the resources/ directory in _extract_phases_from_pvc** (`pipeline/tests/test_deploy_collect.py`) — deploy.py:1028 adds a second wire-up site that is exercised only by unit tests of redact_yaml_tree itself; no integration test verifies the wiring inside _extract_phases_from_pvc. The existing end-to-end test in test_deploy_collect_plans.py covers only the plans path. A small test mirroring that one for the resources path would lock the wire-up in.
- **Tighten the else-branch around redact_yaml_tree so it only fires on copy success, not on the 'no such file' non-error case** (`pipeline/deploy.py:1025`) — Verified at deploy.py:1025-1028: the else fires both on success and on 'no such file in' failures. In practice this is harmless because shutil.rmtree at line 1001-1002 wipes res_dest before the iteration and the freshly-mkdir'd directory is empty when 'no such file' triggers. Restructuring to gate on r.returncode == 0 would make intent clearer and remove a needless tree walk.

Issue #335, opened 2026-06-16
pipeline: remove dead emission and tests for baseline_sim_config / baseline_real_config / baseline_real_notes


kalantar:
## Background

Follow-up to #333. After #333 lands, the prompt-side surface for these fields will be gone, and the v3 manifest schema no longer accepts `baselines[].sim` / `baselines[].real` (they were dropped in the manifest refactor — neither the bootstrap skill nor any real `transfer.yaml` produces them). The only producer of non-empty values today is a test fixture, validating an emission path that nothing consumes.

This issue tracks deleting the pipeline-side emission and its tests.

## Confirmed: no producer in production

- `pipeline/lib/manifest.py:63-73` — validator requires only `name` + `scenario` per baseline; `sim` and `real` are silently accepted but not part of the documented v3 shape.
- `.claude/skills/sim2real-bootstrap/SKILL.md:318-321` — manifest template emits only `name` + `scenario`.
- `../sim2real-config-exploration-validation/sim2real/transfer.yaml` — the only real `transfer.yaml` in the workspace; no `sim:` / `real:` blocks.
- `repository-wide grep` for non-test occurrences of `sim:` / `real:` under a `baselines:` parent — none.

The only non-empty value anywhere in the repo is the literal string `"Use EndpointPickerConfig"` at `pipeline/tests/test_prepare.py:697`, hand-authored to exercise the emission contract.

## Scope

Files to change:

### `pipeline/prepare.py`
- L177-184 — drop `bl.get("sim", {}).get("config")` and `bl.get("real", {}).get("config")` file-existence validation.
- L431-439 — drop `sim_config` / `real_config` / `real_notes` keys from the per-baseline `skill_input["baselines"][*]` entries.
- L453-455 — drop the legacy top-level `baseline_sim_config` / `baseline_real_config` / `baseline_real_notes` keys.
- Re-check the `skill_input.json` key allowlist (around L141 in older drafts; current line drift) for any references.

### `pipeline/tests/test_prepare.py`
- L657-680 — delete `test_skill_input_includes_baseline_sim_config`.
- L682-738 — delete `test_skill_input_includes_baseline_real_config_when_present`.
- L14-18 — strip `"sim": {...}` from `MINIMAL_MANIFEST`.

### `pipeline/lib/manifest.py`
- Optional: add an explicit rejection for `sim` / `real` keys at the baseline level so future manifests can't quietly re-introduce the dead fields. Lower priority — current schema simply ignores them.

[...]

Issue #332, opened 2026-06-16
translate: InferenceObjective example in pipeline/README.md is over-prescriptive; writer copies stale apiVersion


kalantar:
## Problem

The translate writer agent emits InferenceObjective YAML with `apiVersion: inference.networking.x-k8s.io/v1alpha2`. The deployed EPP watches the `llm-d.ai` API group, so the emitted CRs are silently ignored. Per-request EPP logs show `"No associated InferenceObjective found, using default"` for every request — both `critical` and `sheddable` collapse to default priority, which invalidates any class-shielding experiment.

## Root cause

`pipeline/README.md:393-409` ("InferenceObjective requirements") shows a YAML example that the writer copies verbatim. Its purpose is to communicate one structural rule:

> `spec.poolRef.name` must be `${model.idLabel}-gaie` so llm-d-benchmark substitutes it at render time.

That rule is correctly honored. But the surrounding example also pins `apiVersion: inference.networking.x-k8s.io/v1alpha2` (and omits `poolRef.group`). The writer treats the whole example block as gospel and copies the wrong apiVersion, overriding the project-supplied truth in the context document.

The same `${model.idLabel}-gaie` rule is also stated in `agent-writer.md:93`, so removing the section from `pipeline/README.md` doesn't lose the requirement — it consolidates authority in the writer prompt.

## Evidence

Project-side context document (correct, built from `config.md`):
```yaml
apiVersion: llm-d.ai/v1alpha2
kind: InferenceObjective
spec:
  poolRef:
    group: inference.networking.k8s.io
    name: ${model.idLabel}-gaie
```

`pipeline/README.md:397-407` (over-prescriptive example):
```yaml
extraObjects:
  - apiVersion: inference.networking.x-k8s.io/v1alpha2   # ← stale; not the rule
    kind: InferenceObjective
    spec:
      poolRef:
        name: ${model.idLabel}-gaie                      # ← the actual rule
```

Test fixtures pin the same wrong apiVersion and lock the bug in:
- `pipeline/tests/test_values.py` lines 62, 64, 85, 89, 129, 131
- `pipeline/tests/test_assemble.py` lines 119, 122, 180, 196, 241, 251, 254

Deployed cluster RBAC grants the EPP `apiGroups: [llm-d.ai]` for `inferenceobjectives`, confirming `llm-d.ai` is the correct group.

## Proposed fix

[...]

PR #336, opened 2026-06-16
fix(translate): remove dead {BASELINE_SIM_CONFIG} / {BASELINE_REAL_CONFIG} / {BASELINE_REAL_NOTES} references from agent prompts

### LINKED ISSUE #333: translate: remove dead {BASELINE_SIM_CONFIG} / {BASELINE_REAL_CONFIG} / {BASELINE_REAL_NOTES} references from agent prompts
## Problem

Three placeholder variables — `{BASELINE_SIM_CONFIG}`, `{BASELINE_REAL_CONFIG}`, `{BASELINE_REAL_NOTES}` — are referenced as primary inputs by the translate writer/expert/reviewer prompts, but in current usage they are always empty:

- The bootstrap skill never emits `baselines[].sim` or `baselines[].real` blocks (see `.claude/skills/sim2real-bootstrap/SKILL.md:318-320` — the manifest template only sets `name` and `scenario`).
- The v3 manifest validator (`pipeline/lib/manifest.py:67-69`) only requires `name` + `scenario` for baselines; sim/real are silently optional.
- prepare.py emits these fields with `null` / `""` when the manifest doesn't supply them.

Effect on the writer: it's told its primary inputs are these three variables plus the context document, but in practice the three are always empty, leaving only the context document. Worse, the prompt phrasing ("Use `{BASELINE_REAL_NOTES}`, `{BASELINE_SIM_CONFIG}`, and the context document …") suggests the variables carry authoritative content when they don't, biasing the writer toward looking past the context document for non-existent data. This contributed to the InferenceObjective translation bug tracked in #332.

agent-writer.md additionally has two redundant input lists — an upfront table at lines 24-31 and a per-phase "Read:" list at lines 79-84. The Phase 2 list adds nothing real (it duplicates the table) but does add the dead variables.

### A. `.claude/skills/sim2real-translate/prompts/agent-writer.md`

- L82: `2. \`{BASELINE_SIM_CONFIG}\` (if non-null) — the sim baseline policy`
- L83: `3. \`{BASELINE_REAL_CONFIG}\` (if not null) — a reference real config (for guidance, not literal copy)`
- L84: `4. \`{BASELINE_REAL_NOTES}\` — translation hints describing what the baseline should contain`
- L104: `- Use \`{BASELINE_REAL_NOTES}\`, \`{BASELINE_SIM_CONFIG}\`, and the context document to`
- L108: `- If \`{BASELINE_REAL_CONFIG}\` is null and \`{BASELINE_SIM_CONFIG}\` is null, derive content`

(Line numbers reflect HEAD after #332 merged; the original issue cited L98 / L102 but those references shifted.)

Plus consolidation: delete the entire Phase 2 "Read:" list at lines 79-84 (it duplicates the upfront table at 24-31). Replace with a one-liner: *"Use the inputs listed in the upfront 'Inputs — Read These Now' table."*

Update content rules at 103-109 to lean only on `{CONTEXT_PATH}` and the Expert.

The Phase 3 "Read:" list at 121-124 is fine — it references live, distinct, per-phase files. Leave alone.

### B. `.claude/skills/sim2real-translate/prompts/agent-expert.md`

- L34: `4. If \`{BASELINE_SIM_CONFIG}\` is non-empty: read it — the simulation baseline policy`
- L35: `5. If \`{BASELINE_REAL_CONFIG}\` is not null: read it — real config template`

### C. `.claude/skills/sim2real-translate/prompts/agent-reviewer.md`

- L31: `4. If \`{BASELINE_REAL_CONFIG}\` is non-null: read it — the real config structure for reference`

### D. `.claude/skills/sim2real-translate/SKILL.md`

- **L141**: `'baseline_sim_config'` entry in the `required = [...]` allowlist that HALTs the skill if the field is missing from `skill_input.json`. After this issue lands, the prompts no longer consume `BASELINE_SIM_CONFIG`, so the skill no longer needs to require it. Removing this entry here keeps issue #335 (which deletes the pipeline-side emission) self-contained — otherwise #335 would HALT the skill on every run unless landed simultaneously with another SKILL.md edit.
- L172-174: bash exports of `BASELINE_SIM_CONFIG`, `BASELINE_REAL_CONFIG`, `BASELINE_REAL_NOTES` from `skill_input.json`
- L364-366: substitution-table entries `{BASELINE_SIM_CONFIG}` / `{BASELINE_REAL_CONFIG}` / `{BASELINE_REAL_NOTES}` → shell variables

## Out of scope (tracked separately)

The pipeline-side emission of these fields (in `pipeline/prepare.py`) and its tests (in `pipeline/tests/test_prepare.py`) are tracked separately in #335. After this issue lands, removing the L141 allowlist entry in section D above means #335 can land cleanly without further coordination.

The `graphify-out/GRAPH_REPORT.md:533` mention is auto-generated and will refresh when the graph is rebuilt.

## Verification

After the changes, run the existing test suite to confirm nothing regresses, then do a fresh `prepare.py` Phase 4 dry run on a project that doesn't supply `sim`/`real` blocks (e.g., the soft-reflective experiment) and confirm:

1. The writer prompt no longer references the dead variables.
2. SKILL.md's bash export step doesn't try to set them.
3. SKILL.md's `required = [...]` allowlist no longer contains `'baseline_sim_config'`.
4. `skill_input.json` still contains the three fields (so pipeline-side support is intact pending #335); the prompt and the skill validator simply don't surface or require them.

## Related

- #332 — InferenceObjective translation issue (the user-visible bug this dead-reference clutter helped mask).
- #335 — pipeline-side cleanup (depends on this issue landing first).

kalantar:
## Summary

Closes #333.

Removes the three dead placeholder variables (`{BASELINE_SIM_CONFIG}`, `{BASELINE_REAL_CONFIG}`, `{BASELINE_REAL_NOTES}`) from the translate skill prompts. These variables were referenced as primary writer inputs but are always empty in production: the bootstrap skill never emits `baselines[].sim` / `baselines[].real`, the v3 manifest validator treats both as silently optional, and no real `transfer.yaml` in this workspace populates them. Their misleading phrasing biased the writer to look past the context document, contributing to #332.

## Changes

| File | Change |
|---|---|
| `.claude/skills/sim2real-translate/prompts/agent-writer.md` | Replaced duplicative Phase 2 `Read:` list with a one-liner pointing at the upfront "Inputs — Read These Now" table. Stripped placeholder mentions from the content rules so they lean on `{CONTEXT_PATH}` and the Expert only. |
| `.claude/skills/sim2real-translate/prompts/agent-expert.md` | Dropped the two conditional `BASELINE_SIM_CONFIG` / `BASELINE_REAL_CONFIG` read items. |
| `.claude/skills/sim2real-translate/prompts/agent-reviewer.md` | Dropped the conditional `BASELINE_REAL_CONFIG` read item. |
| `.claude/skills/sim2real-translate/SKILL.md` | Removed `'baseline_sim_config'` from the `required = [...]` allowlist, deleted the three bash exports, deleted the three substitution-table entries. |

## Reviewer attention

- **Why touch SKILL.md L141 (the allowlist)?** It was the asymmetric piece: only `baseline_sim_config` was required, not the other two. After this PR, the skill's prompt surface no longer consumes the value, so requiring it in the allowlist would just HALT future runs once #335 (pipeline-side cleanup) lands. Removing it here keeps both PRs self-contained.
- **Pipeline emission untouched.** `pipeline/prepare.py:177-184, 431-439, 453-455` still produce these fields in `skill_input.json`; `pipeline/tests/test_prepare.py:657-738` still pin them. That cleanup is tracked in #335 and now unblocked.

## Stale-reference sweep

After the diff, grepped the entire repo for `BASELINE_SIM_CONFIG`, `BASELINE_REAL_CONFIG`, `BASELINE_REAL_NOTES`, `baseline_sim_config`, `baseline_real_config`, `baseline_real_notes`, and the unprefixed `sim_config` / `real_config` / `real_notes` keys. The only remaining hits are in `pipeline/prepare.py` and `pipeline/tests/test_prepare.py` — both deferred to #335. No skill, prompt, or doc residue.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-{analyze,bootstrap,translate}/tests/ -q` — 1007 passed, 2 xfailed (pre-existing)
- [x] Acceptance criteria from #333: writer prompt no longer references the dead variables; SKILL.md bash exports gone; SKILL.md required-fields allowlist no longer requires `'baseline_sim_config'`; `skill_input.json` still contains the three fields (pipeline emission intact pending #335).

## Related

- Closes #333
- Unblocks #335
- Follow-up to #332

### DIFF
--- a/.claude/skills/sim2real-translate/SKILL.md
+++ b/.claude/skills/sim2real-translate/SKILL.md
@@ -138,7 +138,7 @@ python3 -c "
 import json, sys
 si = json.load(open('$RUN_DIR/skill_input.json'))
 required = ['run_name', 'run_dir', 'scenario', 'context_path', 'manifest_path',
-            'algorithm_source', 'baseline_sim_config',
+            'algorithm_source',
             'target', 'build_commands', 'config_kind', 'context',
             'current_algorithm']
 missing = [f for f in required if f not in si]
@@ -169,9 +169,6 @@ TARGET_REPO=$EXPERIMENT_ROOT/$(python3 -c "import json; print(json.load(open('$R
 CONFIG_KIND=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['config_kind'])")
 CURRENT_ALGORITHM=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json'))['current_algorithm'])")
 CONTEXT_TEXT=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json')).get('context', {}).get('text', ''))")
-BASELINE_SIM_CONFIG=$(python3 -c "import json; v=json.load(open('$RUN_DIR/skill_input.json'))['baseline_sim_config']; print('$EXPERIMENT_ROOT/' + v if v else '')")
-BASELINE_REAL_CONFIG=$(python3 -c "import json; v=json.load(open('$RUN_DIR/skill_input.json')).get('baseline_real_config'); print('$EXPERIMENT_ROOT/' + v if v else 'null')")
-BASELINE_REAL_NOTES=$(python3 -c "import json; print(json.load(open('$RUN_DIR/skill_input.json')).get('baseline_real_notes', ''))")
 ```
 
 Verify source files exist:
@@ -361,9 +358,6 @@ corresponding shell variables:
 - `{CONTEXT_PATH}` → `$CONTEXT_PATH`
 - `{ALGO_SOURCE}` → `$ALGO_SOURCE`
 - `{ALGO_CONFIG}` → `$ALGO_CONFIG`
-- `{BASELINE_SIM_CONFIG}` → `$BASELINE_SIM_CONFIG`
-- `{BASELINE_REAL_CONFIG}` → `$BASELINE_REAL_CONFIG`
-- `{BASELINE_REAL_NOTES}` → `$BASELINE_REAL_NOTES`
 - `{EXPERT_AGENT_NAME}` → `"expert"`
 - `{TARGET_REPO}` → `$TARGET_REPO`
 - `{CONFIG_KIND}` → `$CONFIG_KIND`
--- a/.claude/skills/sim2real-translate/prompts/agent-expert.md
+++ b/.claude/skills/sim2real-translate/prompts/agent-expert.md
@@ -31,9 +31,7 @@ Read the following to build your foundation:
 1. Read `{EXPERIMENT_ROOT}/transfer.yaml` — understand scenario, algorithm, baseline, context
 2. Read `{ALGO_SOURCE}` — the simulation algorithm being translated
 3. If `{ALGO_CONFIG}` is non-empty: read it — algorithm weights and thresholds (ground truth)
-4. If `{BASELINE_SIM_CONFIG}` is non-empty: read it — the simulation baseline policy
-5. If `{BASELINE_REAL_CONFIG}` is not null: read it — real config template
-6. Read all files in context.files (listed in transfer.yaml)
+4. Read all files in context.files (listed in transfer.yaml)
 
 Then do targeted exploration of the target repo:
 
--- a/.claude/skills/sim2real-translate/prompts/agent-reviewer.md
+++ b/.claude/skills/sim2real-translate/prompts/agent-reviewer.md
@@ -28,8 +28,7 @@ Read and hold in context:
    available plugin types and their type strings
 2. **Algorithm source** `{ALGO_SOURCE}` — the simulation Go file being translated
 3. If `{ALGO_CONFIG}` is non-empty: read it — weights and thresholds (ground truth)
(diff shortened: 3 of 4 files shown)


kalantar:
## Review verdict: READY TO DIOMERGE

Clean surgical deletion-only PR; the only remaining specialist concerns are pre-existing wording issues outside the diff and optional regression tests on a 14-line removal.

### Recommendations
- **Consider one consolidated regression test asserting the three placeholder names are absent from prompts and SKILL.md** (`.claude/skills/sim2real-translate/tests/test_skill_substitution.py`) — Verified via grep that test_skill_substitution.py contains no negative assertions for {BASELINE_SIM_CONFIG} / {BASELINE_REAL_CONFIG} / {BASELINE_REAL_NOTES} or the required-list entry. The existing file already has the pattern of asserting strings must (or must not) appear in SKILL.md, so a single small test would lock in the cleanup without bloat — the three separate tests the analyzer proposes are overkill for a deletion of this size.

PR #341, opened 2026-06-16
feat(sim2real-check): validate generated configs against config.md and assert InferenceObjective resolution

### LINKED ISSUE #339: sim2real-check: validate generated configs against config.md and assert InferenceObjective resolution in EPP logs
## Problem

`sim2real-check` currently has no static rule that compares the project's `config.md` against the generated EPP/InferenceObjective YAML, and no runtime rule that asserts InferenceObjective resolution in the EPP logs. As a result, the apiVersion-mismatch class of translation bug (see #332) only surfaces during trace-stage analysis — long after compute is burned. Even then, the failure mode is indirect: the runtime check at 5c.2 ("priority bands recognized in scheduling logs") catches a downstream symptom, not the load-time error that would point at the cause.

The skill already reads `config.md` for vLLM server params (Step 2a) and treats `extraObjects (InferenceObjective CRDs)` in Step 2c — but Step 2c is a **cross-variant identity** rule, not a correctness rule. All three variants can be wrong in the same way and 2c still passes.

## Proposed additions

Two minimal, evidence-grounded checks. Each is a single parser pass or a grep — no new dependencies.

### A. Step 2d (new) — EPP & InferenceObjective config parity vs `config.md`

Add a sub-section under existing **Step 2 (CONFIG PARITY)**.

If `<sim>/config.md` contains EPP config blocks (markdown sections matching `## llm-d EPP Configuration — Baseline|Treatment|Control`) and/or InferenceObjective entries (`## Priority Bands`, or any fenced YAML with `kind: InferenceObjective`), parse them as YAML and structurally diff against the generated artifact:

| `config.md` block | Generated artifact |
|---|---|
| `## llm-d EPP Configuration — Baseline` | `<real>/generated/baseline_config.yaml` → `inferenceExtension.pluginsCustomConfig.custom-plugins.yaml` (parse the inner YAML string) |
| `## llm-d EPP Configuration — Treatment` | `<real>/generated/<treatment>/<treatment>_config.yaml` deep-merged onto baseline |
| `## llm-d EPP Configuration — Control` | `<real>/generated/<control>/<control>_config.yaml` deep-merged onto baseline |
| InferenceObjective entries (`kind: InferenceObjective`) | `<real>/generated/baseline_config.yaml` → `extraObjects` filtered to `kind: InferenceObjective` |

Compare these fields (order-independent for lists):

- `apiVersion`, `kind`
- `featureGates` (set equality)
- `plugins` — list of `(type, name, parameters)` tuples
- `schedulingProfiles` — for each profile, the `(pluginRef, weight)` map
- `flowControl.usageLimitPolicyPluginRef`
- For each InferenceObjective: `metadata.name`, `spec.priority`, `spec.poolRef.name`, `spec.poolRef.group`

**Verdict**: PASS only if every field matches. Show a side-by-side table — `field path | config.md value | generated value | match`. Skip with NOTE if `config.md` has no EPP / InferenceObjective blocks.

This catches the apiVersion bug (#332) and any future drift in EPP plugin lists, weights, or feature gates without needing a deep policy audit.

### B. Step 5c.6 (new) — InferenceObjective resolution at runtime

Add a sub-section under existing **Step 5c (EPP Runtime Proof)**.

**Conditional**: run only if `<real>/generated/baseline_config.yaml` declares at least one `kind: InferenceObjective` in `extraObjects`. Otherwise SKIP with note "no objectives configured."

**Check 1 — Negative signal absent.** Grep all `epp_logs/*.log` for the literal string:

```
"No associated InferenceObjective found, using default"
```

This message is emitted at `pkg/epp/requestcontrol/director.go:219` (in the llm-d-router source) whenever `datastore.ObjectiveGet(reqCtx.ObjectiveKey)` returns nil — i.e., the InferenceObjective CR isn't loaded into the API group the EPP watches, or the request's `objectiveKey` doesn't match any loaded objective name.

- PASS: 0 occurrences across all per-pod EPP logs.
- FAIL: any occurrence. Emit: *"InferenceObjective wiring broken (apiVersion or poolRef mismatch) — see #332-class issues. Skipping downstream signal/policy analysis."*

**Check 2 — Positive signal present.** Grep `epp_logs/*.log` for log lines emitted at `director.go:275` carrying a `priority` field with non-default values. For each `spec.priority` value declared in the configured InferenceObjectives, expect at least one matching log line. If only `"priority":0` (the `defaultPriority` fallback at `director.go:222`) appears, the request flow never associated configured priorities with requests.

- PASS: every configured priority value appears at least once.
- FAIL: any configured priority value missing. (Less common than Check 1; usually means objectives loaded but request headers don't match objective names.)

**Output**: table per pod — `failure-string count | distinct priority values seen | verdict`.

## Why these two are the right scope

- **(A) is static** — runs before compute. Catches the `apiVersion`/`poolRef` mismatch class without deploying.
- **(B) is runtime but cheap** — a single grep on already-collected `epp_logs/`. Confirms the static fix actually wired through.
- Both are conditional on what the project actually configures: skip silently when no InferenceObjectives or EPP config blocks exist.
- Together they pin both ends of the wire (config-time YAML and request-time runtime), which is the gap that let #332 escape every existing rule.

## Verification on the soft-reflective run

Both checks would FAIL immediately on `workspace/runs/sr-mk/results.wrong.{1,2}` (formerly `results.{1,2,3}`):

- **(A)** `config.md` declares `apiVersion: llm-d.ai/v1alpha2` for InferenceObjective; the committed `generated/baseline_config.yaml` (commit `db84e0d`) has `apiVersion: inference.networking.x-k8s.io/v1alpha2`. Field mismatch.
- **(B) Check 1**: `grep -c "No associated InferenceObjective found"` on any per-pod EPP log returns thousands of hits.
- **(B) Check 2**: only `"priority":0` ever appears in scheduling logs; the configured `100` (critical) and `-50` (sheddable) values never resolve.

## Related

- #332 — translation bug that motivated these checks
- #333 — translate-skill cleanup

kalantar:
## Summary
- Adds **Step 2d** (CONFIG PARITY) — structurally diff generated EPP and InferenceObjective YAML against the `## llm-d EPP Configuration — Baseline|Treatment|Control` and `kind: InferenceObjective` blocks in `<sim>/config.md`. Compares `apiVersion`, `kind`, `featureGates`, `plugins`, `schedulingProfiles`, `flowControl.usageLimitPolicyPluginRef`, and per-objective `metadata.name` / `spec.priority` / `spec.poolRef.{name,group}`. Conditional — skips silently when `config.md` has no such blocks.
- Adds **Step 5c.6** (EPP Runtime Proof) — greps `epp_logs/*.log` for `"No associated InferenceObjective found, using default"` (emitted at `pkg/epp/requestcontrol/director.go:219`), and verifies every configured `spec.priority` appears at least once in scheduling logs. Conditional — skips when `baseline_config.yaml` has no `kind: InferenceObjective` in `extraObjects`.

Together these pin both ends of the InferenceObjective wire — config-time YAML and request-time runtime — closing the gap that let the apiVersion-mismatch class of translation bug (#332) reach trace-stage analysis. Existing Step 2c (identity check across variants) and Step 5c.2 (priority bands seen at runtime) both pass when all variants are wrong in the same way and only catch the downstream symptom; these new checks catch the cause.

## Reviewer attention
- The two sub-sections are independent — the static check (2d) runs before deploy, the runtime check (5c.6) runs after `epp_logs/` are collected. Each has its own SKIP condition, so projects that don't use InferenceObjective configs see no new noise.
- 2d compares against `config.md` parsed as YAML (markdown sections + fenced YAML); the implementer of the skill at run time is expected to handle malformed YAML by surfacing the parse error rather than silently passing.
- 5c.6 references `director.go:219`, `:222`, `:275` from the llm-d-router source — line numbers are documentation pointers for the skill operator, not runtime imports.

## Sweep
- Searched `**/*.md`, `docs/`, `.claude/skills/`, and `README*` for references to specific Step labels (`Step 2c`, `Step 5c.5`, etc.) that would collide with renumbering. No external references to internal step labels — only `README.md` mentions `/sim2real-check` at the command level.
- No code symbols renamed/removed; no schema changes; no test files for this skill (CI's `pytest` stanza covers `sim2real-analyze`/`-bootstrap`/`-translate`, not `sim2real-check`).
- `ruff check pipeline/ .claude/skills/ --select F` — passes.

Closes #339

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -349,21 +349,25 @@ Step 5c.2 (priority bands recognized) is a downstream symptom check — it sees 
 
 **Conditional**: run only if `<real>/generated/baseline_config.yaml` declares at least one `kind: InferenceObjective` in `extraObjects`. Otherwise SKIP with note "no objectives configured."
 
+**Verbosity prerequisite.** Both checks below depend on log lines that are gated on EPP verbosity. The llm-d-router defines `VERBOSE = 3` and `DEBUG = 4` in `pkg/common/observability/logging/const.go`; an EPP deployed at the default Info level (`V=0`) will emit neither. Before relying on the absence of the failure string in Check 1, confirm VERBOSE-level logging is active by greping `epp_logs/*.log` for any line containing `"objectiveKey"` (added at `director.go:275` and emitted at `V(logutil.DEBUG)` further down). If no such line appears, treat both checks as INAPPLICABLE and surface that the EPP must be redeployed with `-v=4` (or at minimum `-v=3` for Check 1 alone). Otherwise the absence of the failure string is indistinguishable from "logging too quiet to detect failure."
+
 **Check 1 — Negative signal absent.** Grep all `epp_logs/*.log` for the literal string:
 
 ```
 "No associated InferenceObjective found, using default"
 ```
 
-This message is emitted at `pkg/epp/requestcontrol/director.go:219` (in the llm-d-router source) whenever `datastore.ObjectiveGet(reqCtx.ObjectiveKey)` returns nil — i.e., the InferenceObjective CR isn't loaded into the API group the EPP watches, or the request's `objectiveKey` doesn't match any loaded objective name.
+This message is emitted at `pkg/epp/requestcontrol/director.go:219` (in the llm-d-router source) at `V(logutil.VERBOSE)` (requires `-v=3` or higher) whenever `datastore.ObjectiveGet(reqCtx.ObjectiveKey)` returns nil — i.e., the InferenceObjective CR isn't loaded into the API group the EPP watches, or the request's `objectiveKey` doesn't match any loaded objective name.
 
-- PASS: 0 occurrences across all per-pod EPP logs.
+- PASS: 0 occurrences across all per-pod EPP logs **and** the verbosity prerequisite above is met (some `"objectiveKey"` line is present).
 - FAIL: any occurrence. Emit: *"InferenceObjective wiring broken (apiVersion or poolRef mismatch) — see #332-class issues. Skipping downstream signal/policy analysis."*
+- INAPPLICABLE: 0 occurrences but no `"objectiveKey"` lines anywhere in the logs — verbosity too low to evaluate.
 
-**Check 2 — Positive signal present.** Grep `epp_logs/*.log` for log lines emitted at `director.go:275` carrying a `priority` field with non-default values. For each `spec.priority` value declared in the configured InferenceObjectives, expect at least one matching log line. If only `"priority":0` (the `defaultPriority` fallback at `director.go:222`) appears, the request flow never associated configured priorities with requests.
+**Check 2 — Positive signal present.** Grep `epp_logs/*.log` for the message `"LLM request assembled"`, emitted at `director.go:277` at `V(logutil.DEBUG)` (requires `-v=4` or higher). Each such line carries the `priority` field set from the resolved `infObjective.Spec.Priority` (attached via `logger.WithValues(...)` at `director.go:275`). For each `spec.priority` value declared in the configured InferenceObjectives, expect at least one matching log line. If only `"priority":0` (the `defaultPriority` fallback at `director.go:222`) appears, the request flow never associated configured priorities with requests.
 
 - PASS: every configured priority value appears at least once.
 - FAIL: any configured priority value missing. (Less common than Check 1; usually means objectives loaded but request headers don't match objective names.)
+- INAPPLICABLE: no `"LLM request assembled"` lines at all — EPP verbosity is below `-v=4`.
 
 **Show:** table per pod — `failure-string count | distinct priority values seen | verdict`.
 
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -178,6 +178,34 @@ The real bundle may contain multiple experiment variants (e.g., `baseline/`, `tr
 - The generated Go code for each variant's policy plugin — confirm control returns constant 1.0 and treatment implements the sim algorithm
 - Verdict: PASS if only flow control policy differs, FAIL if routing/scheduling/priority/feature-gates diverge
 
+**2d. EPP & InferenceObjective Config Parity vs `config.md`**
+
+Step 2c is a cross-variant **identity** check — all variants can be wrong in the same way and 2c still passes. This sub-section is the **correctness** check against the project-supplied truth in `<sim>/config.md`.
+
+**Conditional**: run only if `<sim>/config.md` contains EPP config blocks (markdown sections matching `## llm-d EPP Configuration — Baseline|Treatment|Control`) and/or InferenceObjective entries (`## Priority Bands`, or any fenced YAML with `kind: InferenceObjective`). Otherwise SKIP with note "no EPP / InferenceObjective blocks in config.md."
+
+**Source-of-truth → generated artifact mapping:**
+
+| `config.md` block | Generated artifact |
+|---|---|
+| `## llm-d EPP Configuration — Baseline` | `<real>/generated/baseline_config.yaml` → `inferenceExtension.pluginsCustomConfig.custom-plugins.yaml` (parse the inner YAML string) |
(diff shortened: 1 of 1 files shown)


kalantar:

The 5c.6 runtime check has two factual flaws in its log-source guidance that will mislead operators applying it; static-config 2d is solid.

### Important
- **Check 1's grep target requires V(logutil.VERBOSE) but the doc gives no verbosity prerequisite, so a default-verbosity EPP will silently PASS even when wiring is broken** (`.claude/skills/sim2real-check/SKILL.md:358`) — Confirmed at /Users/kalantar/projects/go.workspace/src/github.com/kalantar-msb/soft-reflective/llm-d-router/pkg/epp/requestcontrol/director.go:219 — the message is emitted at V(logutil.VERBOSE) and pkg/common/observability/logging/const.go defines VERBOSE = 3. An EPP deployed at default Info level produces zero occurrences whether the wiring is healthy or completely broken, so 'PASS: 0 occurrences' is indistinguishable from 'check inapplicable.' Since this check is the primary detection mechanism for the #332-class bug (per the PR body), the doc must state the -v=3+ prerequisite or add a positive sentinel grep that confirms VERBOSE logging is active before trusting an absence as a pass.
- **Check 2 attributes the priority log to director.go:275, but line 275 is a logger.WithValues() context-enrichment call and emits nothing — the actual emission is the V(logutil.DEBUG) Info("LLM request assembled") call at director.go:277** (`.claude/skills/sim2real-check/SKILL.md:363`) — Verified at the same director.go: line 275 is `logger = logger.WithValues("objectiveKey", ..., "priority", infObjective.Spec.Priority)` (no output) and line 277 is `logger.V(logutil.DEBUG).Info("LLM request assembled")` (the actual emission, at DEBUG=4). An operator searching log output 'emitted at director.go:275' will find nothing. The doc should reference the message string ("LLM request assembled") and note the -v=4+ requirement, or correct the line number to 277.

kalantar:
Addressed both blocking items from the review.

5c.6 Check 1 now states the `-v=3` prerequisite explicitly. Before trusting "0 occurrences" as a pass, the operator confirms VERBOSE-level logging is actually flowing by greping for any `"objectiveKey"` line. If no such line appears, the verdict is INAPPLICABLE (verbosity too low to detect failure) rather than a false PASS.

5c.6 Check 2 now references the actual emission — `"LLM request assembled"` at `director.go:277`, `V(logutil.DEBUG)`, requires `-v=4`. The line 275 mention is preserved as the `logger.WithValues(...)` call that attaches the priority field, since that's the structural reason the priority appears on the line 277 record. Same INAPPLICABLE verdict path when no such lines exist.

Confirmed both citations against `director.go` in the soft-reflective tree: line 219 is `log.FromContext(ctx).V(logutil.VERBOSE).Info("No associated InferenceObjective found, using default", ...)`; lines 275–277 are `logger = logger.WithValues(..., "priority", ...)` then `logger.V(logutil.DEBUG).Info("LLM request assembled")`. `VERBOSE = 3` and `DEBUG = 4` in `pkg/common/observability/logging/const.go`.

kalantar:

Documentation-only PR adds two well-scoped check sections to a skill prompt; the surfaced issues are precision nits in prose, not behavioral defects in the prompt's logic.

### Recommendations
- **Verbosity-prerequisite explanation in 5c.6 misattributes the lowest-verbosity emit of "objectiveKey"** (`.claude/skills/sim2real-check/SKILL.md`) — SKILL.md says "objectiveKey" is "added at director.go:275 and emitted at V(logutil.DEBUG) further down" — but director.go also emits "objectiveKey" at V(logutil.VERBOSE) in the "Request handled" line (line 462 in the current source). The gate happens to function at -v=3 because of that VERBOSE emit, but the prose suggests it requires -v=4. Tightening the explanation to mention both emission sites would make the "-v=3 sufficient for Check 1" caveat self-consistent.
- **INAPPLICABLE verdict introduced in 5c.6 is not represented in the Output Format summary table** (`.claude/skills/sim2real-check/SKILL.md:55`) — Line 55 defines the canonical verdict set as PASS/WARN/FAIL and the summary table at lines 351-359 has only those three columns. Adding INAPPLICABLE without indicating where it lands in the summary leaves the agent to improvise. Mapping INAPPLICABLE to WARN (or adding a column) would close the loop.

PR #334, opened 2026-06-16
fix(translate): remove over-prescriptive InferenceObjective example from pipeline/README.md

kalantar:
## Summary

Closes #332.

The translate writer agent was emitting InferenceObjectives with `apiVersion: inference.networking.x-k8s.io/v1alpha2`, which the deployed EPP (watching `llm-d.ai`) silently ignored. Every request fell back to default priority — class-shielding experiments measured nothing meaningful.

Root cause: `pipeline/README.md` carried an InferenceObjective YAML example whose only purpose was to communicate the `spec.poolRef.name: ${model.idLabel}-gaie` rule. The writer copied the entire block — including the wrong apiVersion and a missing `poolRef.group` — overriding the correct values from the project context document.

## Changes

- **Delete** the `### InferenceObjective requirements` section from `pipeline/README.md` (formerly lines 393–409). The poolRef rule survives in the writer prompt, so no rule is lost.
- **Rewrite** the InferenceObjective bullet in `.claude/skills/sim2real-translate/prompts/agent-writer.md` (line 93) to be explicit: only `spec.poolRef.name` is pinned by this format; `apiVersion`, `spec.priority`, `spec.poolRef.group`, etc. come from the project's context document, not from any example.
- **Add disclaimer comments** to `pipeline/tests/test_values.py` (TestMergeLists class) and `pipeline/tests/test_assemble.py` (module docstring): the apiVersion strings in those fixtures are illustrative; the tests verify deep-merge identity behavior, not the canonical apiVersion of any resource. Fixture strings left unchanged.

## Reviewer attention

- Verified `agent-writer.md:93` was the only place outside the deleted README section that stated the `${model.idLabel}-gaie` rule, so no requirement is lost by the deletion.
- Tests still pass with the literal `inference.networking.x-k8s.io/v1alpha2` strings in fixtures — the disclaimer is the only thing distinguishing "what this fixture demonstrates" from "what to copy into a real overlay."

## Stale-reference sweep

Grepped `**/*.md` and Python sources for:
- `InferenceObjective requirements` section title — no remaining cross-refs.
- `x-k8s.io/v1alpha2` in markdown — only in this PR's diff (now gone) and skill prompts (only in code-fenced output examples that don't pin a value).

## Test plan

- [x] `python -m pytest pipeline/ -v` — 912 passed, 2 xfailed (pre-existing)
- [x] `python -m pytest .claude/skills/sim2real-{analyze,bootstrap,translate}/tests/ -v` — 95 passed
[...]

PR #344, opened 2026-06-16
fix(translate): require verbatim field-level copy of config.md InferenceObjective fields

### LINKED ISSUE #343: translate: writer drops spec.poolRef.group from InferenceObjective (second instance of #332-class drop)
## Summary

The translation writer reads `config.md` from the context document and copies most InferenceObjective fields, but **silently drops `spec.poolRef.group`**. This is the same class of bug as #332 (which addressed `apiVersion`), at a different field. The post-#332 fix to `agent-writer.md` mentions `spec.poolRef.group` in prose but doesn't enforce field-level copying — so this drop slipped past the rewording.

Effect: identical to #332. Without `poolRef.group`, the EPP cannot match the InferenceObjective CR to the deployed `InferencePool`. `datastore.ObjectiveGet(reqCtx.ObjectiveKey)` returns nil for every request → priority falls back to default (0) → `AllOrderedPriorityLevels()` returns `[0]` → soft-reflective plugin's `n<=1` carve-out keeps it dormant → treatment, control and baseline are indistinguishable at the admission layer.

### Source of truth (`config.md`)

```yaml
# config.md:96-104
apiVersion: llm-d.ai/v1alpha2
kind: InferenceObjective
metadata:
  name: critical
spec:
  poolRef:
    group: inference.networking.k8s.io   # <-- present
    name: ${model.idLabel}-gaie
  priority: 100
```

Both InferenceObjective entries (critical and sheddable) carry `poolRef.group`.

### What the writer produced (commit `db84e0d`, before any hand-edits)

```yaml
# generated/baseline_config.yaml from db84e0d
extraObjects:
  - apiVersion: inference.networking.x-k8s.io/v1alpha2   # <-- #332 (now fixed)
    kind: InferenceObjective
    metadata:
      name: critical
    spec:
      poolRef:
        name: ${model.idLabel}-gaie                        # <-- group dropped
      priority: 100
```

The writer copied `name` from `poolRef` but dropped `group`. Same pattern on the `sheddable` entry. This was committed alongside the apiVersion bug from #332.

### What the configmap-as-deployed shows

`workspace/runs/sr-mk/run-inputs-configmap.yaml` (the cluster apply payload, mtime `Jun 12 18:03`) carries the writer's output verbatim:

```yaml
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferenceObjective
  metadata: {name: critical}
  spec:
    poolRef:
      name: ${model.idLabel}-gaie
    priority: 100
```

### Runtime impact (across three replicates, results.{1,2,3})

EPP logs at `pkg/epp/requestcontrol/director.go:219`:

```
"No associated InferenceObjective found, using default" objectiveKey="critical"
"No associated InferenceObjective found, using default" objectiveKey="sheddable"
```

| Replicate | Total failure-string lines (12 variant×workload combos) |
|---|---:|
| results.1 | ~96K |
| results.2 | ~96K |
| results.3 | ~96K |

The objectiveKey field shows the request *did* carry the right SLO label (`critical` / `sheddable`); the lookup failed inside `datastore.ObjectiveGet`. With both `apiVersion` wrong (#332) and `poolRef.group` missing (this issue), the EPP couldn't resolve either CR to the deployed `InferencePool`, regardless of which bug caused the lookup miss.

`FlowRegistry` initializes with `orderedPriorities:[0]` — only the default band exists at runtime; the configured `100` and `-50` never register. Consequence chain (verified in code, llm-d-router):

1. Datastore returns nil objective → `director.go:222-224` sets `defaultPriority = 0`.
2. Registry's `AllOrderedPriorityLevels()` returns `[0]`, so `priorities = [0]` (length 1).
3. Soft-reflective plugin `policy.go:79-81`: `if n == 1 { ceilings[0] = 1.0; return ceilings }` — the index-0 carve-out fires before the reflective formula.
4. Constant-ceiling plugin: also returns `1.0` always (by design).
5. Baseline (`flowControl: {}`): default policy, `1.0` always.
6. **All three variants behave identically at the admission layer.** The three replicates cannot distinguish soft-reflective from baseline.

### Why the post-#332 rewording didn't catch this

`agent-writer.md:88-94` (post-#332 state) reads:

```markdown
- InferenceObjectives go in `extraObjects`. The only field this format pins is:
  - `spec.poolRef.name: ${model.idLabel}-gaie` — llm-d-benchmark substitutes
    `${model.idLabel}` at render time (requires llm-d-benchmark >= PR #1103).

  All other InferenceObjective fields (`apiVersion`, `spec.priority`,
  `spec.poolRef.group`, etc.) come from the project's context document
  (typically `config.md`) — copy them from there, not from any example.
```

This wording mentions `spec.poolRef.group` correctly but uses ambient language ("come from the project's context document") rather than a verb that forces field-level copy. A writer paraphrasing the rule reasonably reads it as "use config.md as the source for these fields if you choose to include them" rather than "every field that appears in config.md's InferenceObjective MUST appear in the generated output." The drop is consistent with that softer reading.

## Minimal fix

Tighten `agent-writer.md:88-94` so the writer must reproduce the entire `spec.poolRef` block (and any other nested map under `spec`) from `config.md`, not just selected sub-fields.

**Proposed wording (replaces lines 88-94):**

```markdown
- InferenceObjectives go in `extraObjects`. The format pins one rule:
  - `spec.poolRef.name` MUST equal `${model.idLabel}-gaie` so llm-d-benchmark
    substitutes `${model.idLabel}` at render time (requires llm-d-benchmark
    >= PR #1103).

  Everything else MUST be copied verbatim from the project's context document
  (typically `config.md`) at the field level. In particular: copy the entire
  `spec.poolRef` block (including `group`, `name`, and any other keys), not
  just `name`. Same applies to `apiVersion`, `metadata`, `spec.priority`, and
  any other nested fields the project's `config.md` declares — if it's in
  config.md's InferenceObjective, it MUST appear in the generated output.
  Do not paraphrase or selectively include fields.
```

The change is one paragraph in `agent-writer.md`; no code changes. The static check filed in #339 (Step 2d) catches this class of drop at config-time once it lands.

## Verification

After applying the wording change:

1. Run a fresh translate against `soft-reflective`'s `transfer.yaml` and confirm the new `generated/baseline_config.yaml` carries `spec.poolRef.group: inference.networking.k8s.io` on both InferenceObjective entries.
2. Apply the regenerated configmap and run one workload at v=3.
3. Grep the EPP logs for `"No associated InferenceObjective found, using default"` — expected count: 0.

## Related

- #332 — first instance of the same translation drop pattern (`apiVersion` field). Closed without surfacing the second drop because the static check (#339) wasn't in place when #332 was being verified.
- #333 — dead `{BASELINE_*}` reference cleanup (independent).
- #339 — sim2real-check additions that would have caught this at Step 2d (config.md ↔ generated diff). Adopting #339 is the structural defense against future field-drops; tightening agent-writer.md is the immediate fix for the writer behavior.

## Note on deployment hygiene (out of scope, flag only)

Independent of the translation bug, the run that produced results.{1,2,3} (Jun 16 11:10) appears to have been deployed using the Jun 12 18:03 `run-inputs-configmap.yaml` — the configmap was never regenerated after the working-tree fixes were hand-applied. That's a deployment-pipeline issue (probably user-side workflow, not codebase). Worth a separate conversation about either auto-regenerating the configmap on deploy or refusing to deploy a configmap older than the source files it was assembled from. Not part of this fix.

kalantar:
Closes #343.

## Summary

The translation writer was paraphrasing the InferenceObjective rule rather than enforcing it. The post-#332 wording in `agent-writer.md:88-94` said *"All other InferenceObjective fields … come from the project's context document"* — ambient language a writer can reasonably read as "use config.md if you choose to include these fields," not "every field config.md declares MUST appear in the generated output."

That softer reading is consistent with the observed bug: the writer copied `spec.poolRef.name` from `config.md` but dropped `spec.poolRef.group`. The runtime impact (verified across three replicates of `workspace/runs/sr-mk/results.{1,2,3}`): every InferenceObjective lookup fails — `datastore.ObjectiveGet` returns nil because the EPP can't match the objective to the deployed `InferencePool` without a properly-grouped `poolRef` — and treatment/control/baseline collapse to identical admission behavior. ~96K `"No associated InferenceObjective found, using default"` lines per replicate.

This PR replaces the wording with a forcing rule: `spec.poolRef.name` MUST equal the template literal; everything else MUST be copied verbatim from the context document at the field level. The wording explicitly calls out the entire `spec.poolRef` block (including `group`, `name`, and any other keys), not just `name`. Same for `apiVersion`, `metadata`, `spec.priority`, and any other nested fields the project's `config.md` declares.

## What changed

`.claude/skills/sim2real-translate/prompts/agent-writer.md` lines 88-94 — one paragraph rewritten. No code changes, no test changes.

## Reviewer attention

- **Wording strength.** The new text uses "MUST" and "verbatim" repeatedly. That's deliberate — the prior bug class is the writer paraphrasing or selectively including fields, so the rule is written to forbid both behaviors. If the wording feels heavy, the alternative (a programmatic field-level diff guard) is tracked in #341/#339; this PR is the lightweight prompt-only fix.
- **Scope.** The rule applies to InferenceObjectives in `extraObjects`. It does not extend to other artifacts (e.g., the EPP plugin config), which haven't shown the same drop pattern.

## Sweep

Per the implement-issue Step 6 sweep:

- `pipeline/README.md` — no longer carries the over-prescriptive InferenceObjective example (#332's fix landed in #334). Generic structural mentions at lines 388/418/425 are unaffected.
- `.claude/skills/sim2real-translate/prompts/agent-expert.md` and `agent-reviewer.md` — no parallel paragraph requiring updates.
- `.claude/skills/sim2real-check/SKILL.md` — already references `spec.poolRef.group` as a field to validate in Step 2d (#341's static check). Consistent with this PR's direction.
- `pipeline/tests/test_values.py` / `test_assemble.py` — fixtures use illustrative `inference.networking.x-k8s.io/v1alpha2` strings, already noted as illustrative per #332's plan. They don't contradict the new rule (which governs translator output, not test fixtures).
- Phase-3 line at `agent-writer.md:138` ("Do NOT repeat `extraObjects` …") — unrelated; that's about deep-merge propagation, not field copying.

## Verification

Per the issue's recipe (manual, since this is a prompt change without automated guards):

1. Re-translate against `soft-reflective`'s `transfer.yaml` and confirm the new `generated/baseline_config.yaml` carries `spec.poolRef.group: inference.networking.k8s.io` on both InferenceObjective entries.
2. Apply the regenerated configmap and run one workload at v=3.
3. `grep -c "No associated InferenceObjective found" results.<n>/*/*/epp_logs/*.log` should be 0.

Local pre-push checks:
- `python -m pytest pipeline/ -v` — 920 passed, 2 xfailed.
- `ruff check pipeline/ --select F` — clean.
- `git -C <parent-repo> status` — only the worktree shows changes; no leakage.

## Related

- #332 — first instance of this drop pattern (`apiVersion` field), fixed in #334.
- #333 — dead `{BASELINE_*}` reference cleanup, fixed in #336 (independent).
- #339/#341 — sim2real-check additions that automate the equivalent field-level diff at config-time. This PR is the immediate prompt-only fix; #341 (now landed) is the structural defense for future drops.

### DIFF
--- a/.claude/skills/sim2real-translate/prompts/agent-writer.md
+++ b/.claude/skills/sim2real-translate/prompts/agent-writer.md
@@ -85,13 +85,18 @@ that will be deep-merged onto the experiment's `baseline.yaml` by `prepare.py`.
 - Top-level `scenario:` list with one dict
 - The `name` field MUST match the scenario name in the experiment's baseline file exactly
   (mismatched names cause llmdbenchmark to deploy multiple scenarios instead of merging)
-- InferenceObjectives go in `extraObjects`. The only field this format pins is:
-  - `spec.poolRef.name: ${model.idLabel}-gaie` — llm-d-benchmark substitutes
-    `${model.idLabel}` at render time (requires llm-d-benchmark >= PR #1103).
+- InferenceObjectives go in `extraObjects`. The format pins one rule:
+  - `spec.poolRef.name` MUST equal `${model.idLabel}-gaie` so llm-d-benchmark
+    substitutes `${model.idLabel}` at render time (requires llm-d-benchmark
+    >= PR #1103).
 
-  All other InferenceObjective fields (`apiVersion`, `spec.priority`,
-  `spec.poolRef.group`, etc.) come from the project's context document
-  (typically `config.md`) — copy them from there, not from any example.
+  Everything else MUST be copied verbatim from the project's context document
+  (typically `config.md`) at the field level. In particular: copy the entire
+  `spec.poolRef` block (including `group`, `name`, and any other keys), not
+  just `name`. Same applies to `apiVersion`, `metadata`, `spec.priority`, and
+  any other nested fields the project's `config.md` declares — if it's in
+  config.md's InferenceObjective, it MUST appear in the generated output.
+  Do not paraphrase or selectively include fields.
 - Plugin config in `inferenceExtension.pluginsCustomConfig` as a YAML-in-YAML string
 - Only include fields you are adding or overriding
 


kalantar:

A focused prompt-only tightening that directly addresses the soft-wording root cause; structural concerns the specialists raised are pre-existing and explicitly tracked elsewhere.

### Recommendations
- **Consider adding a static prompt-content test pinning the new MUST/verbatim language** (`.claude/skills/sim2real-translate/tests/test_skill_substitution.py`) — test_skill_substitution.py already contains six tests that pin critical prompt constraints (e.g., test_writer_prompt_uses_algo_name at line 17 forbids regression to flat treatment_config.yaml). Adding a parallel test asserting 'verbatim', 'group', and 'Do not paraphrase' are present in agent-writer.md would catch a future soft-reword regression at PR time, matching the pattern used after #332. The PR author may reasonably defer this since #341's post-translate static check already provides defense in depth.

PR #367, opened 2026-06-20
feat(deploy): show completed_namespace in status SLOT column for done pairs

### LINKED ISSUE #366: deploy.py status: SLOT column shows `—` for done pairs even though completed_namespace is recorded
## Problem

In `pipeline/deploy.py status`, pairs that previously completed and were reset back to `pending` show their stale `completed_namespace` in the SLOT column. Operators reading the table believe these pairs are running in / assigned to those namespaces; they are not.

PR #367 introduced a display-layer fallback (`namespace or completed_namespace or "—"`) intended to surface the run namespace for `done` pairs. That fallback exposed a latent data-shape problem: `completed_namespace` is set on every completion ([`pipeline/deploy.py:2040`](../blob/main/pipeline/deploy.py#L2040), [`2330`](../blob/main/pipeline/deploy.py#L2330)) but is **never cleared anywhere** in the file. After a reset, the field persists into the new `pending` state and leaks into the display.

## Live evidence (`kalantar-0` cluster, `sim2real-progress-sr-mk` ConfigMap)

| Pair | status | namespace | completed_namespace | SLOT shown after #367 |
|------|--------|-----------|---------------------|----------------------|
| wl-code-generation-16-softreflective | pending | None | kalantar-1 (stale) | **kalantar-1** ❌ |
| wl-interactive-chat-15-softreflective | pending | None | kalantar-1 (stale) | **kalantar-1** ❌ |
| wl-interactive-chat-5-softreflective | pending | None | kalantar-0 (stale) | **kalantar-0** ❌ |
| wl-reasoning-0-2-softreflective | pending | None | kalantar-2 (stale) | **kalantar-2** ❌ |

## Root cause

`completed_namespace` was treated half as state and half as history. The invariant should be:

> `completed_namespace is not None` ⟺ `status == \"done\"`

The reset paths violate this:

- [`pipeline/deploy.py:1828-1832`](../blob/main/pipeline/deploy.py#L1828-L1832) (`_reset_pair`) — clears `namespace`, retries, pending_stalls, pending_since. Does **not** clear `completed_namespace`.
- `_force_reset` at line 1836+ — same gap.

Both readers of `completed_namespace` (`collect` at [`deploy.py:1354`](../blob/main/pipeline/deploy.py#L1354), `_reset_pair`'s own orphan-helm cleanup at [`1728`](../blob/main/pipeline/deploy.py#L1728), `wipe` at [`1787`](../blob/main/pipeline/deploy.py#L1787)) gate on `is_done` before reading. None of them depend on a historical \"this pair completed sometime ago\" semantic. By operator workflow, results are collected before any reset is performed. So clearing on reset breaks no real path.

## Proposed fix

Tighten the reset paths so `completed_namespace` follows the same lifecycle as `namespace`:

- `_reset_pair`'s state-reset block (around lines 1828-1832): add `entry[\"completed_namespace\"] = None`.
- `_force_reset` reset block: same.

The display-layer fallback in #367 (`pipeline/deploy.py:526`) **stays as-is**:

```python
slot = entry.get(\"namespace\") or entry.get(\"completed_namespace\") or \"—\"
```

With the cleared-on-reset invariant, this fallback becomes correct for every state without any status gating:

| status | `namespace` | `completed_namespace` | SLOT |
|--------|-------------|-----------------------|------|
| pending | None | None (cleared on reset) | `—` ✓ |
| running | live | None | live ✓ |
| done | None (cleared on completion) | where it ran | where it ran ✓ |
| failed / stalled / timed-out | retained per #277 | None | retained `namespace` ✓ |
| pending after reset of a previously-done pair | None | None (cleared on reset) | `—` ✓ |

## Tests

- The existing `test_status_done_pair_shows_completed_namespace` continues to assert the done-row case.
- New regression test: place a pending entry with a non-None `completed_namespace` in `_PROGRESS` (mirroring the cluster-observed pre-fix shape) and assert the SLOT column for that row is `—`. This locks in the invariant from the display side regardless of how the data shape is enforced upstream.
- New unit test on `_reset_pair` (and/or `_force_reset`): given a done entry with `completed_namespace` set, after reset the entry has `status == \"pending\"` AND `completed_namespace is None`.

## Out of scope

- Renaming the column or adding a separate \"LAST\" column.
- Audit of other fields in the entry that might have similar half-state/half-history shape.

kalantar:
## Summary

`pipeline/deploy.py status` now shows where a done pair ran, in the SLOT column. Previously every done pair printed `—` even though the orchestrator records the run namespace under `entry["completed_namespace"]` (it has to clear `entry["namespace"]` so the slot can be reused — see [`deploy.py:2327-2332`](../blob/main/pipeline/deploy.py#L2327-L2332)).

One-line fix at [`pipeline/deploy.py:526`](../blob/main/pipeline/deploy.py#L526):

```python
slot = entry.get("namespace") or entry.get("completed_namespace") or "—"
```

Closes #366.

## Reviewer attention

- **Failed / stalled / timed-out pairs are unaffected.** Those paths intentionally retain `entry["namespace"]` (per #277, so reset/cleanup can find helm releases). The first lookup wins; behavior unchanged.
- **Test fixture realignment**: `pipeline/tests/test_deploy_run.py:11` previously had an unrealistic `done` entry that retained `namespace=sim2real-0`. Real done entries have `namespace=None` and `completed_namespace` set. The fixture was updated to mirror the real lifecycle, and a new test (`test_status_done_pair_shows_completed_namespace`) asserts the SLOT column displays the namespace rather than `—` for the done row. Existing tests that rely on `_PROGRESS` (status filtering, summary count, key presence) don't depend on the namespace value, so they're untouched.
- No public-API or schema changes; this is a display-only fix.

## Test plan

- [x] `python3 -m pytest pipeline/tests/test_deploy_run.py -v` — 164/164 pass, including the new assertion.
- [x] Stale-reference sweep over `docs/`, `pipeline/README.md`, `.claude/skills/`. `deploy.py status` is described only in generic prose; no SLOT-column example or contract docs to update.
- [ ] On a real cluster: complete a deploy, run `deploy.py status`, confirm done pairs show their actual namespace in the SLOT column rather than `—`.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -523,7 +523,7 @@ def _cmd_status(args, run_dir: Path,
 
         for key, entry in sorted(pairs.items()):
             status = entry.get("status", "unknown")
-            slot = entry.get("namespace") or "—"
+            slot = entry.get("namespace") or entry.get("completed_namespace") or "—"
             retries = entry.get("retries", 0)
             counts[status] = counts.get(status, 0) + 1
             print(f"{key:<{pair_w}} {status:<{col_status}} {slot:<{col_slot}} {retries}")
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -8,7 +8,7 @@ from pipeline.lib.progress import ConfigMapProgressStore
 
 
 _PROGRESS = {
-    "wl-smoke-baseline":   {"workload": "wl-smoke",  "package": "baseline",   "status": "done",      "namespace": "sim2real-0", "retries": 0},
+    "wl-smoke-baseline":   {"workload": "wl-smoke",  "package": "baseline",   "status": "done",      "namespace": None,         "completed_namespace": "sim2real-0", "retries": 0},
     "wl-smoke-treatment":  {"workload": "wl-smoke",  "package": "treatment",  "status": "running",   "namespace": "sim2real-1", "retries": 0},
     "wl-load-baseline":    {"workload": "wl-load",   "package": "baseline",   "status": "pending",   "namespace": None,         "retries": 0},
     "wl-load-treatment":   {"workload": "wl-load",   "package": "treatment",  "status": "timed-out", "namespace": "sim2real-2", "retries": 1},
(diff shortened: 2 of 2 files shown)


kalantar:

One-line display fix is correct, the fallback chain is sound, and the regression test pins the bug.

PR #373, opened 2026-06-21
feat(deploy): re-read namespace slot list mid-run for live add/remove (#372)

kalantar:
Closes #372.

## Summary

Adds `_refresh_namespaces`, called once per cycle in `_cmd_run` just before the `free_slots` computation ([pipeline/deploy.py:2479](pipeline/deploy.py#L2479) post-change). Operators can now edit `workspace/setup_config.json` while `deploy.py run` is in flight and vary the slot pool without restarting the orchestrator.

## Behavior

- **Add a namespace mid-run** → next cycle includes it in `free_slots` and dispatches to it (subject to existing `_check_slot_ready` — PVCs bound, HF secret present). Unprovisioned namespaces produce a `Slot ... not ready` warn and are skipped, not crashed.
- **Remove a namespace mid-run** → drops from `free_slots`; the status/completion loop iterates `slots_busy` directly, so a pair already running there keeps being polled and the slot is freed normally on success/failure/timeout/reclaim. **No cancellation** — pairs drain on their own.
- **`namespaces[0]` is pinned for the run's lifetime.** The run-scoped progress ConfigMap (`sim2real-progress-{run}`) is bound to the startup primary; mid-run primary changes are logged and ignored.
- **Parse / IO errors keep the prior list** (best-effort).
- **Quiet cycles emit nothing.** Only adds, removes, primary-mismatch, and parse-failure paths log.

## Touch-point validation

The issue body documents the full analysis. The only references to `namespaces` inside the dispatch loop body are line 2445 (the `free_slots` computation — needed live data) and lines 2581/2583 (cosmetic `len(namespaces)` log strings — self-correcting once the list is refreshed). Startup-only consumers (`_cmd_build`, `_configmap_namespace`/`store`, `_force_reset`, `_reconcile_on_resume`, final summary print) remain pinned to the startup snapshot — primary-pinning protects the ConfigMap binding.

## Tests (8 new)

In `pipeline/tests/test_deploy_run.py`:

- `test_refresh_namespaces_no_change` — equal lists return same list, no log
- `test_refresh_namespaces_add` — addition logs `Slot pool: +ns2`
- `test_refresh_namespaces_remove_drains_naturally` — removal logs `Slot pool: -ns2 (will drain on their own)`
- `test_refresh_namespaces_add_and_remove` — mixed in same cycle
- `test_refresh_namespaces_primary_change_ignored` — warns + returns current
- `test_refresh_namespaces_empty_keeps_current` — wiped/missing config: silent keep
[...]

PR #282, opened 2026-06-04
refactor(deploy): drop backoff state machine; poll at base interval, skip GPU probe when slots full (#274)

kalantar:
Closes #274.

## What changed

The orchestrator dispatch loop in `pipeline/deploy.py` `_cmd_run` ran an exponential-backoff state machine (`pipeline/lib/backoff.py`). It misread the common "slots full because we dispatched everything that fits" condition as external GPU pressure and stretched the poll interval up to 600s, delaying slot-recovery detection by minutes (see the observed log in #274). The base 30s poll is itself a sufficient rate limit against hot-loop reclaim cycles.

This PR deletes the backoff machinery and replaces it with slot-aware probe skipping:

- **Deleted** `pipeline/lib/backoff.py` and `pipeline/tests/test_backoff.py`.
- **`_cmd_run`**: removed all `BackoffController` references; the loop now always sleeps at `--poll-interval` (default 30s).
- **Slot gating**: the GPU capacity probe *and* the dispatch computation now run only when at least one slot is free (`if free_slots:`). When every slot is busy, only the PipelineRun status check at the top of the cycle runs, and the loop logs `Dispatching 0/<pending> pending — all <total> slots busy` (deduped per state transition). Slot recovery is therefore detected within one poll interval.
- **CLI**: removed the `--max-backoff` flag, its default, and its `--remote` forwarding entry.
- **Progress schema**: stopped writing the `_orchestrator` key and dropped its display in `deploy.py status`.
- **Docs**: updated `pipeline/README.md` (flag table + polling section) and the `CLAUDE.md` library table.

## Out of scope (unchanged)

Reclaim handling (only the backoff *signal* from reclaims is removed), the shadow GPU reservation ledger (#255), and the #272 observability fixes.

## Reviewer attention

- **Backward compatibility**: old progress ConfigMaps in live clusters may still carry a legacy `_orchestrator` key. `_is_pair_key` filters any `_`-prefixed key, so such files still render cleanly in `status` and are excluded from the pair set. `test_status_backward_compat_ignores_legacy_orchestrator` and the `_orchestrator`-exclusion tests cover this.
- **Loop restructure**: the probe + dispatch block moved inside `if free_slots:`; the capacity-probe-failure fallback (`free_gpus = None` → dispatch without gating) is preserved within that branch.
- **`_last_log_state` keys**: the old `backoff_skip`/`backoff_level` dedup keys are gone; a new `slots_busy` key dedups the busy log and is cleared at every slot-state-change site.

## Verification

[...]

PR #375, opened 2026-06-23
ci(orchestrator-image): make smoke test self-maintaining and surface failures

### LINKED ISSUE #374: ci(orchestrator-image): smoke test imports deleted `backoff` module — image push silently broken since 2026-06-04
## Problem

The `Build Orchestrator Image` workflow has been failing on every push to `main` since commit f2e3556 (PR #282, 2026-06-04). The Docker build itself succeeds; the failure is in the post-build `Smoke test` step:

```
ImportError: cannot import name 'backoff' from 'pipeline.lib' (/app/pipeline/lib/__init__.py)
```

`.github/workflows/orchestrator-image.yml:61` runs:

```python
from pipeline.lib import assemble, backoff, capacity, ..., values; print('OK')
```

PR #282 ("refactor(deploy): drop backoff state machine; poll at base interval, skip GPU probe when slots full") deleted `pipeline/lib/backoff.py` but did not update this hand-maintained import list.

## Why it kept the published image stale

Workflow step order in the `build-and-push` job: `Build (push: false) → Smoke test → Push`. When Smoke test fails, the explicit `Push` step never runs — so neither `:latest` nor the `:<sha>` tag is pushed to GHCR. The workflow status correctly reports failure on the Actions tab, but nothing surfaces that to operators or the README.

## Consequence

`:latest` and every `:<sha>` tag from the last ~19 days on `ghcr.io/inference-sim/sim2real/orchestrator` are from before 2026-06-04. Orchestrator pods started since then are running pre-PR-#282 code and are missing roughly 19 merged PRs of behavior, including but not limited to:

- #373 — `feat(deploy): re-read namespace slot list mid-run for live add/remove`
- #371 — `fix(setup): preserve deploy-owned run_metadata fields on re-run`
- #368 — `fix(deploy): clear completed_namespace on reset`
- #367 — `feat(deploy): show completed_namespace in status SLOT column for done pairs`
- #361 — `feat(tekton): pin generated PipelineRun SA to helm-installer`

Detected when a user manually added a namespace to `setup_config.json` mid-run expecting #373's live add/remove behavior. Nothing happened. The running pod's `/app/pipeline/deploy.py` has no `_refresh_namespaces`. The image itself was stale.

## Smallest fix (literally 3 chars)

Delete `backoff, ` from line 61 of `.github/workflows/orchestrator-image.yml`. Push step turns green and operators stop pulling pre-#282 images. Doesn't prevent recurrence — the next module rename/delete that misses this list will silently re-break the same way.

## Sturdier fix (this issue)

Three changes, in `.github/workflows/orchestrator-image.yml` and `README.md`. Each is independent; all three together form the recommendation.

### 1. Self-maintaining smoke test (mechanism — the import list drifts no more)

Replace the literal import list at line 61 with dynamic discovery:

```yaml
      - name: Smoke test
        run: |
          docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
          docker run --rm --entrypoint kubectl ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --client
          docker run --rm --entrypoint helm  ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --short
          docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "
          import pkgutil, importlib, pipeline.lib as p
          failures = []
          for m in pkgutil.iter_modules(p.__path__):
              try:
                  importlib.import_module(f'pipeline.lib.{m.name}')
              except Exception as e:
                  failures.append(f'{m.name}: {type(e).__name__}: {e}')
          if failures:
              raise SystemExit(chr(10).join(failures))
          print('OK: all pipeline.lib modules imported')
          "
```

- New module added → covered automatically.
- Module deleted → stops being checked automatically.
- Any module breaks at import time (syntax error, missing dep, circular import) → surfaces in CI; **all** failures reported, not just the first.
- `pipeline.lib.__init__` itself failing to load short-circuits at the `import pipeline.lib as p` line, which is a meaningful signal too.

### 2. README workflow badge (passive visibility)

Add to the top of `README.md`:

```markdown
[![Build Orchestrator Image](https://github.com/inference-sim/sim2real/actions/workflows/orchestrator-image.yml/badge.svg)](https://github.com/inference-sim/sim2real/actions/workflows/orchestrator-image.yml)
```

Anyone viewing the repo sees red. A 19-day silent failure couldn't survive this with anyone glancing at the README.

### 3. Auto-open issue on failure (active visibility, for cases nobody visits the README)

Append a step at the end of the `build-and-push` job:

```yaml
      - name: Notify on failure
        if: failure() && github.event_name == 'push'
        uses: actions/github-script@v7
        with:
          script: |
            const labels = ['ci-broken', 'orchestrator-image'];
            const { data: existing } = await github.rest.issues.listForRepo({
              owner: context.repo.owner, repo: context.repo.repo,
              state: 'open', labels: labels.join(','),
            });
            const body = `Run [${context.runId}](${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}) failed on \`${context.sha.slice(0,7)}\`.`;
            if (existing.length === 0) {
              await github.rest.issues.create({
                owner: context.repo.owner, repo: context.repo.repo,
                title: 'CI: Build Orchestrator Image is failing on main',
                body, labels,
              });
            } else {
              await github.rest.issues.createComment({
                owner: context.repo.owner, repo: context.repo.repo,
                issue_number: existing[0].number, body,
              });
            }
```

Guard `github.event_name == 'push'` keeps `workflow_dispatch` recovery attempts from piling on the same open issue.

Workflow needs `issues: write` permission alongside the existing `contents: read` and `packages: write`.

## Test plan

- [ ] On a feature branch with all three changes applied, deliberately break a module (e.g., add `1/0` at the top of `pipeline/lib/log.py`). Smoke test fails reporting `log: ZeroDivisionError: division by zero`.
- [ ] Revert the deliberate break. Smoke test passes; OK line lists every module.
- [ ] Add a new `pipeline/lib/example.py` containing only `pass`. Smoke test still passes (module appears implicitly in the iteration).
- [ ] After merging the fix to main, manually trigger via `workflow_dispatch`. Confirm `:latest` digest on GHCR updates and that `oc -n <ns> exec <orchestrator-pod> -- grep _refresh_namespaces /app/pipeline/deploy.py` (after a pod restart) returns hits.
- [ ] Confirm README badge renders green on the rendered README.
- [ ] Push a temporary commit that breaks a module on `main`-via-PR; confirm a new issue is opened with labels `ci-broken,orchestrator-image`. Push a follow-up still-broken commit; confirm it comments on the existing issue rather than opening a new one.

## Out of scope (file separately if needed)

- **`--remote` mode setup_config gap.** PR #373's "edit setup_config.json" guidance is workstation-only; the in-cluster orchestrator reads from a per-pod emptyDir populated once by an init container. The CM update path doesn't refresh the running pod. Surfaced in the same investigation; should be its own issue and fix.
- **Operator-side staleness signal.** `prepare.py` / `setup.py` could compare the orchestrator image's `:<sha>` tag against the local working tree HEAD and warn when they diverge by more than N commits. Would have caught this from the operator side independent of CI visibility.

kalantar:
Closes #374.

## Summary

The post-build smoke test in `.github/workflows/orchestrator-image.yml` hardcoded a literal list of `pipeline.lib` modules to import. When PR #282 deleted `pipeline/lib/backoff.py`, the list wasn't updated, so every push to `main` since 2026-06-04 has built the image, failed at the smoke test with `ImportError: cannot import name 'backoff' from 'pipeline.lib'`, and never reached the Push step. `:latest` on GHCR sat stale for ~19 days while merges piled up — caught when a user hit PR #373's missing `_refresh_namespaces` in a running pod.

## Changes

1. **Smoke test is now self-maintaining.** Replace the literal import list with `pkgutil.iter_modules(pipeline.lib.__path__)` + `importlib.import_module`. Modules added/removed under `pipeline/lib/` are picked up automatically. All import failures are collected and reported, not just the first.
2. **README workflow status badge.** Added at the top of `README.md` so a stale build is visible to anyone glancing at the repo. Passive layer.
3. **`Notify on failure` step.** Opens a labeled tracking issue (or comments on the existing one) when the workflow fails on a `push` event. Labels `ci-broken` and `orchestrator-image` are created idempotently inside the step (catches 422 = already exists). `workflow_dispatch` recovery reruns are intentionally excluded so they don't pile new comments onto the open issue. Active layer.

`permissions:` block gains `issues: write` to enable the new step (already had `contents: read`, `packages: write`).

## Reviewer attention

- **Python heredoc inside YAML.** The smoke-test `run:` step uses a YAML literal-block `|` with a `docker run -c "..."` whose argument spans multiple lines. YAML strips the common leading indent before passing the script to bash; bash preserves the embedded newlines through to `python -c`. The Python code uses `chr(10)` rather than `'\n'` to dodge any escape-sequence ambiguity. Verified by parsing the YAML with `yaml.safe_load` and by running the same Python snippet locally against the current `pipeline/lib/`.
- **Idempotent `createLabel`.** The `Notify on failure` step creates both labels via `github.rest.issues.createLabel`, swallowing only HTTP 422 (label exists) and re-throwing anything else. So the first failure on `main` provisions the labels; subsequent failures find them present and proceed. No out-of-band label management needed.
- **`failure() && github.event_name == 'push'` guard.** `failure()` is true if any prior step in the job failed (Build, Smoke test, or Push). `push` excludes `workflow_dispatch` so manual recoveries don't comment on the tracking issue. Implication: if a `workflow_dispatch` rerun fails, no notification fires; that's intentional — the operator triggering it knows it failed.
- **Org policy on workflow `issues: write`.** Repo's `default_workflow_permissions` is `read`, but per-workflow elevations work (we already use `packages: write`). No org rule visible in `gh api` blocks `issues: write`. If org policy turns out stricter, the symptom will be the `Notify on failure` step itself failing with a 403; the build/smoke/push pipeline is unaffected.

## Test plan (locally verified)

- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/orchestrator-image.yml'))"` — YAML parses.
- [x] Run the new smoke-test Python locally against current `pipeline/lib/` — all 17 modules import, exits 0, prints `OK: all pipeline.lib modules imported`.
- [x] Add `pipeline/lib/_failtest_374.py` containing `raise RuntimeError("simulated module-load failure")`; rerun the snippet — exits 1, output is `_failtest_374: RuntimeError: simulated module-load failure`. Removed.

## Test plan (post-merge, requires running CI)

- [ ] After merge, `:latest` digest on GHCR updates from the pre-2026-06-04 stale digest. Confirm by `oc -n <ns> exec <orchestrator-pod> -- grep _refresh_namespaces /app/pipeline/deploy.py` (after a pod restart) returning hits.
- [ ] README badge renders green on the rendered README.
- [ ] Push a temp commit that breaks a `pipeline/lib/*.py` module on a feature branch → merge → confirm a new issue opens with labels `ci-broken,orchestrator-image`. Push a follow-up still-broken commit → confirm it comments rather than re-creating.

## Out of scope (filed separately if/when needed)

- `--remote` mode `setup_config.json` gap (PR #373's edit-the-file guidance is workstation-only; the in-cluster orchestrator reads from a per-pod emptyDir populated once by an init container).
- Operator-side staleness check in `prepare.py`/`setup.py` (compare orchestrator image's `:<sha>` to local working-tree HEAD).

### DIFF
--- a/.github/workflows/orchestrator-image.yml
+++ b/.github/workflows/orchestrator-image.yml
@@ -20,6 +20,7 @@ jobs:
     permissions:
       contents: read
       packages: write
+      issues: write
 
     steps:
       - uses: actions/checkout@v4
@@ -58,7 +59,67 @@ jobs:
           docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} --help
           docker run --rm --entrypoint kubectl ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --client
           docker run --rm --entrypoint helm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} version --short
-          docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "from pipeline.lib import assemble, backoff, capacity, context_builder, epp, health, manifest, pod_pending, progress, run_manager, state_machine, tekton, values; print('OK')"
+          docker run --rm --entrypoint python ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.sha }} -c "
+          import pkgutil, importlib, pipeline.lib as p
+          failures = []
+          for m in pkgutil.iter_modules(p.__path__):
+              try:
+                  importlib.import_module(f'pipeline.lib.{m.name}')
+              except Exception as e:
+                  failures.append(f'{m.name}: {type(e).__name__}: {e}')
+          if failures:
+              raise SystemExit(chr(10).join(failures))
+          print('OK: all pipeline.lib modules imported')
+          "
 
       - name: Push
         run: docker push --all-tags ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - name: Notify on failure
+        if: failure() && github.event_name == 'push'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const labels = ['ci-broken', 'orchestrator-image'];
+            const colors = { 'ci-broken': 'd93f0b', 'orchestrator-image': 'c5def5' };
+
+            // Idempotent label creation: 422 = already exists, anything else propagates.
+            for (const name of labels) {
+              try {
+                await github.rest.issues.createLabel({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  name,
+                  color: colors[name],
+                });
+              } catch (e) {
+                if (e.status !== 422) throw e;
+              }
+            }
+
+            const { data: existing } = await github.rest.issues.listForRepo({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              labels: labels.join(','),
+            });
+
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            const body = `Run [${context.runId}](${runUrl}) failed on \`${context.sha.slice(0, 7)}\`.`;
+
+            if (existing.length === 0) {
+              await github.rest.issues.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                title: 'CI: Build Orchestrator Image is failing on main',
+                body,
+                labels,
+              });
+            } else {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: existing[0].number,
+                body,
+              });
+            }
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # sim2real
 
+[![Build Orchestrator Image](https://github.com/inference-sim/sim2real/actions/workflows/orchestrator-image.yml/badge.svg)](https://github.com/inference-sim/sim2real/actions/workflows/orchestrator-image.yml)
+
 Pipeline for transferring simulation-discovered routing algorithms from [inference-sim](inference-sim/) to production [llm-d](https://github.com/llm-d/llm-d).
 
 ## Install


kalantar:

CI fix is correct and well-scoped; specialist findings are either hypothetical edge cases or polish suggestions that don't warrant blocking merge.

Issue #51, opened 2026-05-07
deploy.py reset: warn when --status filter conflicts with done/pending exclusion


kalantar:
## Context

`deploy.py cleanup` excludes `done` and `pending` pairs from its actionable set. If a user runs `cleanup --status done` or `cleanup --status pending`, the filter selects those pairs but the exclusion immediately removes them all. The user sees "No pairs need cleanup (all are done or pending)" with no explanation that their `--status` filter conflicts with the hardcoded exclusion.

## Proposed Fix

When the intersection of filtered pairs and the exclusion set is non-empty, emit a targeted message:

```
[INFO]  --status 'done' pairs are excluded from cleanup (already terminal)
```

## Origin

Review feedback on PR #49 (item 6, classified as follow-up).

kalantar:
The `cleanup` subcommand was renamed to `reset` in #124. Updated title to match.

kalantar:
`cleanup` is no longer a valid subcommand.

PR #384, opened 2026-06-24
fix(setup): adopt clean RBAC pattern after roles.yaml split

### LINKED ISSUE #380: setup.py: probe-and-augment RBAC apply (always namespaced base, optional cluster augment)
## Goal

`pipeline/setup.py` Step 3 (`step_rbac`, `pipeline/setup.py:385`) should always apply the namespaced RBAC bundle and additionally apply the cluster-scoped bundle only when the user has permission to create `ClusterRole`. Today it unconditionally applies the monolithic `roles.yaml`, which fails on clusters where the operator doesn't have cluster-admin.

## Design

- Always: apply the namespaced base (SA + Role + RoleBinding) — works in any project the user owns.
- Conditional: probe `kubectl auth can-i create clusterrole`. If yes, apply the cluster-scoped augment (`ClusterRole` + `ClusterRoleBinding`) for cross-namespace functions (currently: `stream-gpu-stats` Phase 2/2.5/3 cross-ns exec into DCGM exporter pods in `nvidia-gpu-operator`).
- Same for the OpenShift SCC bindings at `pipeline/setup.py:407-414` — probe before applying or apply per-namespace where possible.

## Blocked on / depends on

This depends on inference-sim/tektonc-data-collection#48, which splits the existing monolithic `roles.yaml` into a clean namespaced base + a cluster-scoped augment. Until that lands, `roles.yaml` and `roles-ns.yaml` are independently curated alternates rather than additive layers.

## Workaround in tree today

`pipeline/setup.py` has been switched locally to apply `roles-ns.yaml` instead of `roles.yaml`, with an inline `MUST CHANGE` marker. This unblocks Step 3 on non-admin clusters at the cost of cross-ns DCGM Phase 2/3 GPU metrics. Revert once tektonc-data-collection#48 ships and the probe-and-augment design lands.

## Related

- inference-sim/tektonc-data-collection#48 — submodule split
- #379 — Step 2 cluster-scope assumptions (related symptom on non-admin OpenShift)

## Update — orchestrator behavior in degraded mode (2026-06-23)

When `sim2real-runner-cluster.yaml` cannot be applied (non-admin cluster), the orchestrator runs in **degraded mode**:

- `probe_free_gpus()` (`pipeline/lib/capacity.py:85`) fails with a Forbidden error string. `deploy.py:2511-2514` increments `_probe_fail_count` and emits a rate-limited `warn`: *"Capacity probe failed: {error} — dispatching without GPU gating"*.
- `deploy.py:2543-2545` falls through to ungated dispatch: shuffle all pending and assign to free slots regardless of cluster GPU availability.
- Consequence: PipelineRuns whose cumulative GPU cost exceeds free cluster capacity will be created; their pods queue `Pending`. The existing pending-pod handling (`pipeline/lib/pod_pending.py`) eventually classifies them, but the orchestrator can no longer proactively pace dispatch.

**No orchestrator code change is required for correctness** — the warn-and-dispatch path already exists. But two follow-up changes may be worth tracking under this issue (or split out):

1. **Loud setup-time warning** — done locally; setup.py prints a `DEGRADED MODE` warn at install time so the operator knows what to expect.
2. **Alternative gating signal** — for clusters where cluster-scoped reads are permanently off the table, add an out-of-band cap (e.g. `setup_config.json` `gpu_cap` honored when probe fails) so the orchestrator still paces dispatch instead of running fully open.

Minimum permissions audit (per discussion): the existing `sim2real-runner-cluster.yaml` (`nodes,pods` `get,list` cluster-scoped) is already the lean minimum for the gating design. Verbs are minimum; `nodes` is unavoidably cluster-scoped; `pods` cluster-wide is required for correct cross-tenant accounting. Reducing further requires changing the design (single-tenant assumption → namespace-scoped pod accounting; or out-of-band cap as above).

kalantar:
## Summary

Closes the workaround that PR #383 punted on. Replaces the `MUST CHANGE` block in `step_rbac` with the proper probe-and-augment design now that the YAML split lands in tektonc-data-collection.

`step_rbac` now applies five YAMLs as `(path, required)` tuples — three namespaced bases (always required) and two cluster augments (best-effort on `forbidden`):

| File | Status |
|------|--------|
| `tektonc-data-collection/tekton/roles-ns.yaml` | required |
| `tektonc-data-collection/tekton/roles-cluster.yaml` | best-effort |
| `tektonc-data-collection/tekton/rbac/sim2real-runner.yaml` | required |
| `pipeline/rbac/sim2real-runner-ns.yaml` | required |
| `pipeline/rbac/sim2real-runner-cluster.yaml` | best-effort |

Best-effort means: on a `forbidden` reply (operator lacks cluster-scoped RBAC create), skip the file with a `warn` and continue. Any other failure (network blip, malformed YAML, non-RBAC admission failure) still exits non-zero.

**Drops the imperative `oc adm policy add-scc-to-user` block** (`anyuid -z default`, `anyuid -z helm-installer`, `privileged -z helm-installer`). All three SCC bindings now live in `tektonc-data-collection/tekton/roles-ns.yaml` as RoleBindings to `system:openshift:scc:{restricted,anyuid,privileged}`, so the YAML bundle is the single source of truth — anyone applying the bundle (with or without `setup.py`) gets the same SCC posture.

**Bumps the tektonc-data-collection submodule pin** to pick up the split.

## Dependency

This PR depends on inference-sim/tektonc-data-collection#51 merging first. Marked Draft until then. Once that merges, I'll update the submodule pin to the merge SHA and mark this PR ready.

## Test plan

- [x] `ruff check pipeline/ .claude/skills/ --select F` — passed
- [x] `python -m pytest pipeline/ -v` — 988 passed, 2 xfailed (matches main)
- [ ] Re-run `python pipeline/setup.py --experiment-root <repo>` against a non-admin OpenShift cluster, confirm Step 3 passes with the warn for the cluster augments
- [ ] Re-run against a cluster-admin context, confirm both cluster augments apply cleanly

Closes #380

### DIFF
--- a/pipeline/setup.py
+++ b/pipeline/setup.py
@@ -369,24 +369,34 @@ def step_namespace(cfg: SetupConfig) -> None:
 # ── Step 3: RBAC ─────────────────────────────────────────────────────
 
 def step_rbac(cfg: SetupConfig) -> None:
-    """Step 3: Apply RBAC roles and OpenShift SCC policies."""
+    """Step 3: Apply RBAC bundles for the helm-installer and sim2real-runner SAs.
+
+    Each SA has a namespaced base (always required) and an optional cluster
+    augment (best-effort). On a kubectl Forbidden reply that specifically
+    targets cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding),
+    a best-effort file is skipped with a warn and the run continues. Any
+    other failure (network, malformed YAML, SCC/webhook/quota denial
+    unrelated to cluster-scoped RBAC) still aborts the run.
+
+    SCC RoleBindings (restricted, anyuid, privileged) ride along in the
+    namespaced bases — no imperative `oc adm policy add-scc-to-user` calls.
+
+    Degradations when a cluster augment is skipped:
+      - tektonc roles-cluster.yaml: llm-d-benchmark node auto-discovery is
+        unavailable; cross-ns DCGM exec for stream-gpu-stats Phase 2/2.5/3
+        is skipped. Admin-only stack installs are already gated by the
+        install task's `--non-admin` flag (set by llmdbenchmark-standup).
+      - sim2real-runner-cluster.yaml: orchestrator GPU capacity probe
+        (kubectl get nodes/pods) fails; the probe-failure branch of
+        deploy.py's dispatch loop falls through to ungated dispatch with
+        a rate-limited warn.
+    """
     step(3, 8, "RBAC")
     if cfg.no_cluster:
         ok("RBAC (skipped — --no-cluster)"); return
 
     primary_ns = cfg.namespaces[0]
     env = {**os.environ, "NAMESPACE": cfg.namespace, "PRIMARY_NAMESPACE": primary_ns}
-    # Apply RBAC as paired (namespaced base, optional cluster augment) bundles.
-    # Required files exit on failure. Best-effort files exit-gracefully on a
-    # `forbidden` reply (operator lacks cluster-scoped RBAC create) — the
-    # corresponding ClusterRole/ClusterRoleBinding is skipped and downstream
-    # consumers degrade as documented:
-    #   - tektonc roles-cluster.yaml absent → llm-d-benchmark auto-discovery
-    #     unavailable; cross-ns DCGM exec for stream-gpu-stats Phase 2/2.5/3
-    #     skipped (admin-only stack installs already gated by --non-admin).
-    #   - sim2real-runner-cluster.yaml absent → orchestrator GPU capacity
-    #     probe (kubectl get nodes/pods) fails; deploy.py falls through to
-    #     ungated dispatch with a rate-limited warn (deploy.py:2511-2545).
     yaml_paths = [
         (TEKTONC_DIR / "tekton" / "roles-ns.yaml",                          True),
         (TEKTONC_DIR / "tekton" / "roles-cluster.yaml",                     False),
@@ -407,17 +417,24 @@ def step_rbac(cfg: SetupConfig) -> None:
         if result.returncode == 0:
             continue
         combined = (result.stdout + result.stderr).lower()
-        if not required and "forbidden" in combined:
+        # Narrow predicate: only skip when the failure is a Forbidden on
+        # cluster-scoped RBAC resources (clusterrole/clusterrolebinding).
+        # Other Forbidden reasons (SCC admission, webhook denials, quota
+        # exhaustion) signal that the operator's intent isn't being met
+        # and the run should abort.
+        is_cluster_rbac_forbidden = (
+            not required
+            and "forbidden" in combined
+            and "clusterrole" in combined
+        )
+        if result.stderr: print(result.stderr, end="", file=sys.stderr)
+        if is_cluster_rbac_forbidden:
             warn(f"Skipped {yaml_path.name} — user lacks cluster-scoped RBAC. "
                  "Downstream features that depend on this augment will degrade; "
-                 "see step_rbac docstring for which.")
+                 "see step_rbac docstring for the catalog.")
             continue
-        if result.stderr: print(result.stderr, end="", file=sys.stderr)
         sys.exit(result.returncode)
 
-    # SCC bindings are now carried by tekton/roles-ns.yaml (helm-installer-
-    # restricted-scc, helm-installer-anyuid-scc, helm-installer-privileged-scc).
-    # No imperative `oc adm policy add-scc-to-user` calls needed.
     ok("RBAC roles applied")
 
 # ── Step 4: Secrets ──────────────────────────────────────────────────
--- a/pipeline/setup.py
+++ b/pipeline/setup.py
@@ -376,28 +376,48 @@ def step_rbac(cfg: SetupConfig) -> None:
 
     primary_ns = cfg.namespaces[0]
     env = {**os.environ, "NAMESPACE": cfg.namespace, "PRIMARY_NAMESPACE": primary_ns}
-    for yaml_path in [
-        TEKTONC_DIR / "tekton" / "roles.yaml",
-        TEKTONC_DIR / "tekton" / "rbac" / "sim2real-runner.yaml",
-        REPO_ROOT / "pipeline" / "rbac" / "sim2real-runner-cluster.yaml",
-        REPO_ROOT / "pipeline" / "rbac" / "sim2real-runner-ns.yaml",
-    ]:
+    # Apply RBAC as paired (namespaced base, optional cluster augment) bundles.
+    # Required files exit on failure. Best-effort files exit-gracefully on a
+    # `forbidden` reply (operator lacks cluster-scoped RBAC create) — the
+    # corresponding ClusterRole/ClusterRoleBinding is skipped and downstream
+    # consumers degrade as documented:
+    #   - tektonc roles-cluster.yaml absent → llm-d-benchmark auto-discovery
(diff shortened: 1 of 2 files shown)


kalantar:

The forbidden-tolerance path needs a narrower error predicate or a pre-flight probe, the stderr swallowing on the skip path hides the real server response, and step_rbac's branching has zero test coverage despite a clear pattern next door — plus a stale docstring that still claims SCC policies are applied imperatively.

### Important
- **'forbidden' substring match is too broad; non-RBAC denials on best-effort files would be misclassified as missing-RBAC and silently skipped** (`pipeline/setup.py:410`) — setup.py:410 matches `'forbidden' in (stdout+stderr).lower()` against the full kubectl error text. kubectl returns 'forbidden' for SCC admission failures, OPA/Gatekeeper denials, ResourceQuota exhaustion, and validating-webhook rejections — none of which mean the operator lacks cluster-scoped RBAC. Narrowing to e.g. `'clusterroles' in combined and 'forbidden' in combined` (or running `kubectl auth can-i create clusterrole` as a pre-flight per the issue design) would gate the skip on the actual condition the catalog claims.
- **stderr is not printed on the forbidden-skip branch, hiding the real API server message from the operator** (`pipeline/setup.py:410`) — setup.py:410-414 takes `continue` without printing result.stderr; the only stderr print is at line 415 on the fatal path. When the skip fires, the operator sees only the generic warn — the actual server response (specific resource, denying controller, webhook message) is gone, which makes later debugging much harder. Printing stderr to sys.stderr before `continue` would preserve the human-readable warn while keeping the raw response visible.
- **step_rbac has zero test coverage despite five new branches and a clear pattern in test_setup_pipeline.py** (`pipeline/setup.py:397`) — grep for `step_rbac` in pipeline/tests/ returns nothing; meanwhile test_setup_pipeline.py already mocks `run`, patches TEKTONC_DIR/REPO_ROOT, and uses tmp_path to test step_tekton. The forbidden-skip path, the required-and-non-forbidden fatal path, and the file-not-found exit are all untested. A future change that capitalizes 'Forbidden', flips a `required` flag, or widens the substring match would silently regress with nothing to catch it.
- **step_rbac docstring still says 'and OpenShift SCC policies' but the imperative oc-adm block was removed** (`pipeline/setup.py:372`) — setup.py:372 reads `"""Step 3: Apply RBAC roles and OpenShift SCC policies."""` while the body no longer issues `oc adm policy add-scc-to-user` calls (those bindings now live declaratively in tekton/roles-ns.yaml, verified by inspection). A maintainer reading the docstring will look in this function for SCC handling and find none. The runtime warn at line 413 also points readers at "step_rbac docstring" for the degradation catalog, but that catalog is in the inline comment block — both references should be reconciled.

### Recommendations
- **deploy.py line citation in the inline comment is stale (2511-2545); the actual probe-failure dispatch is around 2597-2656** (`pipeline/setup.py:389`) — Verified by `grep -n 'Capacity probe\|dispatchable = list' pipeline/deploy.py`: the warn is at 2625 and the ungated fallback at 2656. Hard-coded line numbers in cross-file comments rot quickly — prose alone (e.g. "see the probe-failure branch of the orchestrator dispatch loop") is more durable.
- **'gated by --non-admin' wording could mislead readers searching setup.py for that flag** (`pipeline/setup.py:386`) — `--non-admin` is a real flag, but it lives on the tektonc-data-collection llmdbenchmark-standup task, not setup.py (which only has --no-cluster). Qualifying the comment as e.g. "already gated by the install task's --non-admin flag" would save the next reader a fruitless grep.

kalantar:
Addressed the review. Pushed in 4e85a51.

The forbidden-skip predicate now requires both `forbidden` and `clusterrole` in the error text — that narrows it to the actual condition (operator can't create cluster-scoped RBAC) and lets SCC admission failures, webhook denials, and quota errors propagate normally. The skip branch also prints kubectl's stderr before the warn, so the API server's actual message stays visible for debugging.

Added `TestStepRbacApply` with six cases covering the five branches plus a happy path: all-succeed, required-failure-aborts, cluster-rbac-forbidden-skips-with-warn, cluster-augment-non-rbac-forbidden-aborts, skip-branch-emits-stderr, required-file-not-found-aborts. Suite is now 994/2xfailed (was 988).

Updated the docstring to drop "and OpenShift SCC policies" (the imperative `oc adm policy` block was removed in the original commit; SCC bindings are declarative in `tekton/roles-ns.yaml` now). Moved the degradation catalog into the docstring so the runtime warn's "see step_rbac docstring" pointer actually leads somewhere.

Replaced the stale `deploy.py:2511-2545` line citation with prose ("the probe-failure branch of deploy.py's dispatch loop"). Clarified that `--non-admin` is on the install task, not setup.py.

The per-lens code-reviewer finding about file-not-found semantics for `required=False` entries — where a missing optional file would still fatal-exit — is filed as #386 for follow-up. It only manifests under future submodule churn (both optional files exist today), so it didn't feel right to bundle into this PR.

kalantar:

CI is currently failing on this PR because four of the six new TestStepRbacApply tests assume the tektonc-data-collection submodule is checked out, which it is not under the workflow's submodules:false configuration.

### Must fix
- **New TestStepRbacApply tests fail in CI because they call step_rbac() against real submodule paths that aren't checked out under submodules:false** (`pipeline/tests/test_setup_pipeline.py:469`) — Confirmed by running pytest in a fresh clone with no submodules: test_all_succeed, test_required_failure_aborts, test_cluster_rbac_forbidden_skips_with_warn, and test_skip_branch_emits_stderr all raise SystemExit at setup.py:409 (file-not-found guard fires before any mock is reached). The PR's CI run on commit head currently shows the test job failing for this reason. test_cluster_augment_non_rbac_forbidden_aborts passes only because both its expected exit code and the file-not-found exit code are 1 — it passes for the wrong reason and is silent on the predicate it claims to test. Fix: monkeypatch TEKTONC_DIR and REPO_ROOT to a tmp_path with the five YAML stubs pre-created, following the test_redeploy_default_pipeline_missing_errors pattern already in this file.

### Recommendations
- **The sys.exit on a non-skippable kubectl failure prints no labeled [ERROR] line, only raw kubectl stderr** (`pipeline/setup.py:436`) — setup.py:436 is the only abort path in this file that fires without a preceding err() call (compare 409, 453, 508, 538, 689). The user sees kubectl's stderr — which names resources but not the YAML path being applied — and then the process exits. A one-line err(f"kubectl apply failed for {yaml_path.name} (exit {result.returncode})") before sys.exit would match the rest of the file and make the failing input obvious.

kalantar:
Addressed both items from the second review pass. Pushed in ee1a793.

The new step_rbac tests now stage stub YAMLs under `tmp_path` and monkeypatch `TEKTONC_DIR` and `REPO_ROOT` to that path, so they don't depend on the tektonc-data-collection submodule being checked out. The previous commit's tests broke under CI's `submodules: false` because the file-not-found guard fired before any mock reached kubectl — four of six tests SystemExit at setup.py:409. The refactor also caught a real test bug: `test_cluster_augment_non_rbac_forbidden_aborts` was passing for the wrong reason (file-not-found and the predicate-fail abort both return exit 1), so it was silent on what it claimed to test. It now also asserts the loop reached the second apply call before aborting.

The non-skippable apply abort now emits an `err()` line before `sys.exit` — `[ERROR] kubectl apply failed for <filename> (exit N) — RBAC setup aborted` — matching every other fatal path in setup.py (compare lines 409, 453, 508, 538, 689). `test_required_failure_aborts` asserts the file name appears in stderr to lock this in.

CI is running. Local: lint clean, full suite 994/2xfailed unchanged.

PR #389, opened 2026-06-25
fix(health): gate startup-probe escalation on event count

### LINKED ISSUE #387: health: startup probe failures escalate to PipelineRun cancellation on first Unhealthy event
## Symptom

vLLM model pods loading from a cold PVC routinely fail the startup probe a few times while weights load. The deploy orchestrator currently treats the first such failure as terminal and cancels the PipelineRun, even though the pod would have recovered on its own under the helm chart's `failureThreshold`.

## Root cause

`pipeline/lib/health.py:198-216` triggers Tier 2 on the *existence* of any matching `Unhealthy` event, not on its count:

```python
# Tier 2: Startup probe timeout
if pod.phase == \"Running\" and not pod.ready:
    startup_fail = next(
        (e for e in pod_events
         if e.reason == \"Unhealthy\" and \"startup probe\" in e.message.lower()),
        None,
    )
    if startup_fail:
        return TriageResult(tier=2, action=\"suggest\", ...)
```

`EventRecord.count` is parsed (`parse_events:107` — `count=item.get(\"count\", 1)`) but never consulted in this branch. So a k8s event with `count=1` flips Tier 2 immediately.

For comparison, OOMKilled deliberately tolerates retries before escalating (`health.py:115`, `:141-148`):

```python
_OOM_MAX_ATTEMPTS = 2  # tier-1 retries before escalating
...
if pod.reason == \"OOMKilled\":
    attempt = tracker.count(pod.name) + 1
    if attempt <= _OOM_MAX_ATTEMPTS:
        return TriageResult(tier=1, ...)
    return TriageResult(tier=2, ...)
```

There's no analogous budget for startup-probe failures.

## How Tier 2 cancels the PipelineRun

`pipeline/deploy.py:888-893`:

```python
elif result.tier == 2:
    warn(f\"[{pair_key}] {result.message}\")
    if result.suggestion:
        info(f\"  Suggestion: {result.suggestion}\")
    if not skip_teardown:
        needs_escalation = True
```

And `deploy.py:2572-2574`:

```python
if escalate:
    warn(f\"[{pair_key}] pod health escalation → cancelling PipelineRun\")
    if _cancel_and_delete_pipelinerun(pr_name, ns):
```

So `needs_escalation = True` flows straight to `_cancel_and_delete_pipelinerun`.

## Fix

Gate the Tier 2 escalation on a threshold rather than mere existence. Options:

1. **Event count threshold** — e.g. `startup_fail.count >= N`. Simple; matches the existing local patch (`>= 10`). Caveat: k8s coalesces \"similar\" events with a ~6 min window, so the count can be coarser than expected.
2. **Time-based threshold** — escalate only when the most recent `Unhealthy` event is older than N minutes (`last_timestamp` age). Decouples from k8s event-coalescing heuristics; better matches \"the pod has been failing for too long\" semantics.

Either way, the existing test pins the buggy behavior and needs updating:

`pipeline/tests/test_health.py:259-268`:

```python
def test_triage_startup_probe_unhealthy():
    pod = _make_pod(phase=\"Running\", ready=False)
    events = [_make_event(reason=\"Unhealthy\",
                          message=\"Startup probe failed: connection refused\")]
    result = triage_pod(pod, events, RemediationTracker())
    assert result is not None
    assert result.tier == 2
```

A single event → tier 2. Need to add an under-threshold case (returns `None`) and an over-threshold case (returns tier 2).

## Affected

- `pipeline/lib/health.py` — fix lives here
- `pipeline/tests/test_health.py` — update existing test, add threshold cases

kalantar:
Closes #387

## Summary

`triage_pod` was returning Tier 2 on the first Unhealthy event matching `startup probe`, which propagates through `_check_pod_health → needs_escalation → _cancel_and_delete_pipelinerun`. vLLM model pods loading from a cold PVC routinely produce a few early Unhealthy events while weights load, so the orchestrator was preempting the helm chart's `failureThreshold` and killing runs that would have recovered on their own.

## Change

Add a count-based threshold and gate the escalation on it:

```python
_STARTUP_PROBE_MIN_EVENT_COUNT = 10
...
if startup_fail and startup_fail.count >= _STARTUP_PROBE_MIN_EVENT_COUNT:
    return TriageResult(tier=2, ...)
```

The escalation message now includes the event count so it's visible in the orchestrator's stdout (`pod-name: startup probe failing (count=12)`).

## Why count-based vs. time-based

Issue #387 spelled out two candidates: count-based (`startup_fail.count >= N`) and time-based (`last_timestamp` age > N min). Going with count-based for two reasons:

1. **Working precedent.** The user's local patch used `count >= 10` and was confirmed to unblock the symptom. Time-based would be unproven here.
2. **Symmetry with the OOM gate.** `_OOM_MAX_ATTEMPTS = 2` is also a count-based threshold living in the same module; readers comparing the two paths get the same shape.

The caveat is documented at the constant: k8s coalesces "similar" events on a ~6-min window, so the count is coarser than wall-clock failures. `10` is generous in coalesced units.

## Tests

- `test_triage_startup_probe_unhealthy` (which pinned the buggy behavior) is replaced by two tests:
  - `test_triage_startup_probe_unhealthy_below_threshold_returns_none` — `count=1` event → `triage_pod` returns `None`.
  - `test_triage_startup_probe_unhealthy_at_threshold_escalates` — `count=_STARTUP_PROBE_MIN_EVENT_COUNT` event → returns Tier 2 with the same suggestion as before.
- `_make_event` helper extended with a `count=` kwarg (default 1, matching prior behavior).

Full suite: `1090 passed, 2 xfailed in 28.21s`. Lint: `ruff check pipeline/ .claude/skills/ --select F` clean.

## Reviewer attention

- The escalation message format changed from `pod-name: startup probe failing` to `pod-name: startup probe failing (count=N)`. No code consumes this message programmatically (`grep -r` clean), but worth a sanity check.
- I considered escalating only when both count and a time threshold are crossed (the time-based option from the issue). Decided against it for this PR — the working local patch is count-only and adding wall-clock awareness introduces a second tunable to reason about. If we ever want it, swap to `last_timestamp` age and delete the count constant.

## Sweep for stale references

- `grep` across `**/*.md` and `**/*.py` for `startup probe`, `triage_pod`, `_OOM_MAX_ATTEMPTS`, `startup_fail`: no external doc references; only `deploy.py` consumes `triage_pod`, and the consumer signature is unchanged.
- No CLAUDE.md or README updates needed.

### DIFF
--- a/pipeline/lib/health.py
+++ b/pipeline/lib/health.py
@@ -114,6 +114,14 @@ def parse_events(json_str: str) -> list[EventRecord]:
 
 _OOM_MAX_ATTEMPTS = 2  # tier-1 retries before escalating
 
+# Tolerate this many startup-probe Unhealthy events before escalating to
+# Tier 2. vLLM model pods loading from a cold PVC routinely fail the
+# startup probe a few times while weights load; the helm chart's
+# failureThreshold lets the pod recover on its own. k8s coalesces
+# similar events with a ~6-min window, so the count is coarser than
+# wall-clock failures — 10 is generous in coalesced units.
+_STARTUP_PROBE_MIN_EVENT_COUNT = 10
+
 
 def triage_pod(
     pod: PodState,
@@ -202,10 +210,11 @@ def triage_pod(
              if e.reason == "Unhealthy" and "startup probe" in e.message.lower()),
             None,
         )
-        if startup_fail:
+        if startup_fail and startup_fail.count >= _STARTUP_PROBE_MIN_EVENT_COUNT:
             return TriageResult(
                 tier=2, action="suggest", needs_logs=False,
-                message=f"{pod.name}: startup probe failing",
+                message=(f"{pod.name}: startup probe failing "
+                         f"(count={startup_fail.count})"),
                 suggestion=(
                     "Startup probe timing out before model finishes loading.\n"
                     "Increase failureThreshold in "
--- a/pipeline/tests/test_health.py
+++ b/pipeline/tests/test_health.py
@@ -155,9 +155,9 @@ def _make_pod(**kwargs):
     return PodState(**defaults)
 
 
-def _make_event(reason="", message="", involved_object="p"):
+def _make_event(reason="", message="", involved_object="p", count=1):
     from pipeline.lib.health import EventRecord
-    return EventRecord(reason=reason, message=message, count=1,
+    return EventRecord(reason=reason, message=message, count=count,
                        last_timestamp="", involved_object=involved_object)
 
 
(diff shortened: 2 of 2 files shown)


kalantar:

The fix is correct, focused, and well-tested for the regression it targets; remaining suggestions are polish that don't justify blocking.

### Recommendations
- **Add a count=_STARTUP_PROBE_MIN_EVENT_COUNT-1 boundary test to pin the >= semantics.** (`pipeline/tests/test_health.py:259`) — Verified at pipeline/tests/test_health.py:259-282: below-threshold uses count=1 and at-threshold uses count=10. A regression flipping >= to > would still pass both. A single extra case at count=9 closes the off-by-one gap cheaply. Three specialists raised this independently.
- **Consider widening the Tier 2 section comment and suggestion text from 'timeout' to 'failure'.** (`pipeline/lib/health.py:206`) — Verified at pipeline/lib/health.py:206 and 217: the match is `'startup probe' in e.message.lower()` (any startup probe failure including connection refused / HTTP / exec exits), but the section label says 'timeout' and the suggestion says 'timing out'. Minor wording mismatch a future maintainer could trip over.

PR #400, opened 2026-06-28
fix(deploy): apply --package as a pair-scope filter in collect

kalantar:
## Summary

`deploy.py collect` built a synthetic `_ScopeArgs` with `package=None` before calling `_resolve_scope`, dropping `--package` from pair scope and violating the cumulative-filter rule that every other `deploy.py` subcommand follows. `--workload X --package baseline` then resolved pair scope to "every pair of workload X," and the follow-on \"Scoped pair _ has status 'pending' — skipping\" warn fired on pairs of other packages the user never asked about.

This treats `--package` as a pair-scope filter on the same footing as every other caller. The synthetic `experiment` value is the documented carve-out: it doesn't name a pair package, so when the user passes `--package experiment` we pass `None` to `_resolve_scope` and let the existing phase-scope logic expand it to every package directory of the scoped pairs.

Closes #398

## Changes

- `pipeline/deploy.py` (`_cmd_collect`): hoist `scope_package = _parse_list(args.package)` above the synthetic `_ScopeArgs` and pass it through as the pair-scope `package` value unless it's the literal `["experiment"]`; reuse the same parsed value at the later phase-scope decision so `args.package` is parsed once.
- `pipeline/tests/test_deploy_collect.py`: three new tests covering `--workload + --package baseline` (no spurious warn on out-of-scope pairs), `--workload + --package experiment` (today's behavior preserved — warn fires on every non-done pair), and `--workload + --package baseline,softreflective`.
- `pipeline/README.md`: refresh the `collect` flag table and prose to describe `--package` as a pair-scope filter with the `experiment` carve-out.

## Reviewer attention

- The carve-out is a literal `scope_package == ["experiment"]` check. Mixing values like `--package experiment baseline` is rejected downstream by `_apply_run_filters` as an unknown package, which is arguably an improvement over the previous quiet acceptance — flagging it here in case reviewers prefer a more permissive `"experiment" in scope_package` check.
- The change is scoped to `_cmd_collect`. The other `_resolve_scope` callsites already pass `args` directly, so no parity work is needed.

## Sweep

Grepped `**/*.md`, `pipeline/`, `docs/`, `.claude/skills/`, and `README*` for `--package`, "phase-level filter", and adjacent terms. Only `pipeline/README.md` documented the old behavior; updated. `docs/superpowers/plans/2026-05-14-collect-scoping-flags.md` is a historical planning artifact and accurately describes the design at the time it was written — left as-is.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-{analyze,bootstrap,translate}/tests/` — 1096 passed
- [x] Three new regression tests in `test_deploy_collect.py`

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -202,10 +202,10 @@ Per phase, the resolved llm-d-benchmark plan YAMLs are also pulled into `workspa
 |------|-------------|
 | `--only PAIR…` | Scope to specific pair keys — narrows both workload and package (comma or space-separated, `wl-` prefix optional; takes precedence over `--workload`) |
 | `--workload NAME…` | Scope to pairs matching these workloads (comma or space-separated) |
-| `--package NAME…` | Collect only these packages (comma or space-separated, phase-level filter) |
+| `--package NAME…` | Scope to pairs matching these packages (comma or space-separated). Pass the synthetic value `experiment` to collect every package directory of the scoped pairs. |
 | `--skip-logs` | Skip vLLM and EPP log files, collect only traces |
 
-When `--only` or `--workload` is given, only matching workload subdirectories are pulled from the PVC (instead of entire phase directories). Multiple values within a flag use OR (union): `--workload X Y` matches pairs for workload X or Y. Different flags compose as AND: `--workload X Y --package baseline` pulls workloads X and Y from the baseline phase only. Requires progress data to resolve pairs.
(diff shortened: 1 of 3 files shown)


kalantar:

The fix correctly forwards --package to pair scope when combined with --only/--workload, and the carve-out for the synthetic "experiment" value is sound; remaining items are comment-clarity polish that does not block merge.

### Recommendations
- **Section header at deploy.py:1397 lists --package but the `scoped` sentinel at 1401 excludes it, creating a visible header/code mismatch** (`pipeline/deploy.py:1401`) — Read pipeline/deploy.py:1397-1405. The header `Pair-level scoping (--only / --workload / --package)` reads as if all three set `scoped=True`, but `scoped` is still `scope_only is not None or scope_workload is not None`. The behavior is intentional and matches the PR body's reasoning, but a one-line comment near 1401 noting why `scope_package` alone is not enough (pair-scope resolution needs --only or --workload to be meaningful) would prevent a future reader from assuming it's a bug.
- **README collect flag table could note that --status is intentionally not a collect filter** (`pipeline/README.md:201`) — Read pipeline/README.md:201-208. The status/reset tables list --status while collect does not; `_ScopeArgs` in `_cmd_collect` hardcodes `status = None`. The omission is correct but the asymmetry across three otherwise-symmetric tables invites a `what about --status?` question. A single sentence in the prose paragraph would close that.

PR #401, opened 2026-06-28
fix(remote): drop activeDeadlineSeconds from orchestrator Job

kalantar:
## Summary

Removes the hardcoded `activeDeadlineSeconds: 18000` (5h) from `build_orchestrator_job`. Keeps `backoffLimit: 0` and `restartPolicy: Never`.

## Why

The deadline only fires when the orchestrator has already failed to make progress on its own — and when it fires, it makes the failure worse: the pod is SIGTERMed with no handler, leaving `running` ConfigMap entries stranded until the next `deploy.py run` triggers `_reconcile_on_resume`. Multi-pair sweeps that exceed 5h hit this every time.

The orchestrator's own exit conditions (`_work_remaining()` plus per-PipelineRun `timeout_hours × max_retries` plus `_handle_pending_pods` plus `_check_pod_health`) already bound runtime. The deadline is redundant in the success case and harmful in the failure case.

A no-progress watchdog ("no entry changed state in N hours → exit") would be the right mitigation for an orchestrator that genuinely hangs, but that's a separate enhancement and shouldn't gate this removal. The two related concerns called out in #396 (orchestrator SIGTERM handler; `deploy.py status` reconciling against PR state) are also out of scope here.

## Changes

- `pipeline/lib/remote.py` — drop the `activeDeadlineSeconds` line.
- `pipeline/tests/test_remote.py` — replace `test_job_active_deadline` (which asserted `== 18000`) with `test_job_has_no_active_deadline`, a regression guard against reintroduction.
- `pipeline/tests/test_deploy_run.py` — update the `_fmt_duration` width-budget docstring; the upper bound is no longer the Job deadline but the per-PR timeout × retries × pair count.

## Stale-reference sweep

Grepped `activeDeadlineSeconds`, `18000`, `5h`, and `5 hours` across `pipeline/`, `.claude/skills/`, `docs/`, and READMEs. Only hits were unrelated (a `_fmt_duration` test input and section headers in `sim2real-check/SKILL.md`).

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-bootstrap/tests/ .claude/skills/sim2real-translate/tests/` — 1093 passed, 2 xfailed
- [ ] Smoke test on the next remote sweep that takes >5h (cannot exercise locally)

Closes #396

### DIFF
--- a/pipeline/lib/remote.py
+++ b/pipeline/lib/remote.py
@@ -86,7 +86,6 @@ def build_orchestrator_job(
         "metadata": {"name": JOB_NAME, "namespace": namespace},
         "spec": {
             "backoffLimit": 0,
-            "activeDeadlineSeconds": 18000,
             "template": {
                 "spec": {
                     "serviceAccountName": SERVICE_ACCOUNT,
--- a/pipeline/tests/test_deploy_run.py
+++ b/pipeline/tests/test_deploy_run.py
@@ -296,11 +296,11 @@ def test_fmt_duration_invalid_returns_dash(value):
 def test_fmt_duration_width_budget_holds():
     """Every formatted duration in the realistic range fits in 7 chars.
 
-    The realistic upper bound is set by `activeDeadlineSeconds: 18000` (5h)
-    on the orchestrator Job, multiplied by retries — at most a handful of
-    hours, never near a day. The 7-char column budget holds comfortably for
-    anything operators will actually see; values past ~999 days would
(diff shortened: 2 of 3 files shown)


kalantar:

The one-line removal of `activeDeadlineSeconds` is correct, self-contained, and well-tested; specialist nits about a test docstring's arithmetic and parallelism phrasing are too minor to block merge.

PR #338, opened 2026-06-16
feat(pipeline): expose blis observe tuning params on the sim2real Pipeline (closes #234)

kalantar:
## Summary

Closes #234.

This is **PR B** of a two-PR plan:
- **PR A** ([tektonc-data-collection#37](https://github.com/inference-sim/tektonc-data-collection/pull/37), merged at `acf4055`): added params to the `run-workload-blis-observe-binary` Task itself.
- **PR B** (this one): bumps the submodule pointer to PR A's merge SHA and threads matching Pipeline-level params so PipelineRuns can override without touching the Task YAML.

## Changes

**`pipeline/pipeline.yaml`** — adds five Pipeline-level params with defaults matching the Task's, and threads them through the `run-workload-blis-observe-binary` task invocation:

| Param | Default | Source flag | Notes |
|---|---|---|---|
| `maxConcurrency` | `"10000"` | `--max-concurrency` | |
| `timeout` | `"1800"` | `--timeout` | seconds |
| `warmupRequests` | `"50"` | `--warmup-requests` | |
| `prewarmDuration` | `"60s"` | `--prewarm-duration` | `0s` disables priming entirely |
| `extraArgs` | `""` | catch-all | appended verbatim to the `blis observe` command line |

**`tektonc-data-collection`** — submodule pointer bumped from `d33fa35` → `acf4055` (PR #37 merge commit).

## What this PR does NOT do

Per #234's Scope and the cross-reference to #337, **scenario-config-driven values are not threaded** in this PR. `pipeline/lib/tekton.py:make_pipelinerun_scenario()` is unchanged — generated PipelineRuns do not emit the new params, so Tekton uses the Pipeline-level defaults (which match the previously hardcoded Task values). That keeps behavior identical for existing callers.

The follow-up that wires these from `transfer.yaml` / scenario / workload through to PipelineRun param values is tracked in #337. Once #337 resolves the manifest-hierarchy question, the threading can land in a focused PR.

## Reviewer attention

- **Default parity.** The Pipeline defaults (`"10000"` / `"1800"` / `"50"` / `"60s"` / `""`) intentionally mirror the Task defaults from PR #37. Drifting them apart would silently change behavior — keep them in sync until #337 lands a manifest-driven source.
- **`extraArgs` shell behavior.** The Task script appends `$(params.extraArgs)` unquoted on its own continuation line. With the empty default, POSIX `\<newline>` line-joining produces trailing whitespace which the shell discards — verified with `bash -c` repro before pushing PR A.
[...]

PR #404, opened 2026-06-28
feat(manifest): add blis_observe section to transfer.yaml

### LINKED ISSUE #337: Decide where blis observe tuning params (timeout, warmupRequests, maxConcurrency) live in the manifest hierarchy
## Background

Follow-up to #234. Once the `run-workload-blis-observe-binary` Tekton Task is parameterized (`maxConcurrency`, `timeout`, `warmupRequests`), the PipelineRun construction in `pipeline/lib/tekton.py:make_pipelinerun_scenario()` (currently lines 62-78) needs to source values for those new params. This issue tracks the design decision of *where* in the manifest hierarchy those values are specified.

## The question

A PipelineRun is generated per `(scenario, workload)` pair. The new tuning params can plausibly live in any of three places:

1. **The workload file** (`workloads/*.yaml`) — per-workload override.
2. **`transfer.yaml`** — single set of values applied to every PipelineRun in this transfer.
3. **The scenario file** (`baseline.yaml` / `treatment.yaml`) — per-scenario override.

## Initial preference: `transfer.yaml`

Reasoning from the #234 comment thread:

- The workload file describes *what request stream to drive* (model, prompt distribution, request rate). Stuffing client-side tuning into it conflates "what we're measuring" with "how we're driving the measurement."
- The scenario file describes *what server-side configuration to deploy* (EPP plugins, helm values, images). Stuffing client-side driver tuning into it conflates "what we're testing" with "how we're testing it."
- `transfer.yaml` already plays the role of cross-cutting transfer-level configuration (workloads list, baselines, algorithms). Tuning knobs that apply uniformly to every PipelineRun in the transfer fit its scope without overloading it.

The tradeoff: `transfer.yaml`-only means no per-workload differentiation. If saturation tests need a longer timeout than golden-path tests, a single transfer can't express that. If/when that becomes a real constraint, escalate to per-workload overrides — but the workload-file location can be reconsidered then.

## Resolution criteria

A decision on this issue should produce:

1. The chosen location (one of the three above, or a hybrid).
2. The schema change required in `pipeline/lib/manifest.py` (likely a new optional top-level section in v3, e.g. `observe: { timeout, maxConcurrency, warmupRequests }`).
3. An example `transfer.yaml` snippet showing the new section.
4. Acceptance criteria for the schema change tests.

The actual param-plumbing in `pipeline/lib/tekton.py:make_pipelinerun_scenario()` and `pipeline/pipeline.yaml` is implementation work covered by #234 — but it cannot land cleanly without this design decision.

## Out of scope

- Adding the Task params themselves (#234).
- Threading values from manifest → PipelineRun once the location is decided (#234, after this issue resolves).

## Related

- #234 — the parent issue that introduces the Task params.
- The Tekton Task lives in `tektonc-data-collection/tekton/tasks/run-workload-blis-observe-binary.yaml` — that submodule may also need to coordinate, but the manifest-design decision is sim2real-side.

kalantar:
## Summary

Closes #337. Threads per-transfer overrides for the five `blis observe` tuning params into the PipelineRuns the orchestrator generates.

Today every PipelineRun uses the Pipeline-level defaults declared in `pipeline/pipeline.yaml` (`maxConcurrency: "10000"`, `timeout: "1800"`, `warmupRequests: "50"`, `prewarmDuration: "60s"`, `extraArgs: ""`) — wired up in PR #338. There has been no way for an experiment bundle to specify different values short of editing the framework's Pipeline definition. This adds:

```yaml
blis_observe:
  maxConcurrency: 5000
  timeout: 3600
  warmupRequests: 25
  prewarmDuration: 30s
  extraArgs: "--foo bar"
```

at the top level of `transfer.yaml`. All keys optional; omitted keys fall through to the Pipeline-level defaults. Section absent → today's behavior, byte-identical.

## Changes

- **`pipeline/lib/manifest.py`** — new optional `blis_observe:` validator at the end of `_validate_v3_fields`. Mirrors the `defaults:` validator's style (default value when absent, mapping type-check, unknown-key rejection, per-key scalar type-check). Booleans are explicitly rejected (otherwise `isinstance(True, int) == True` would let them through).
- **`pipeline/lib/tekton.py`** — `make_pipelinerun_scenario` gains an `observe: dict | None = None` kwarg. Emits one PipelineRun `params` entry per specified key, in a stable order (`_OBSERVE_PARAM_ORDER`), with `str()` coercion so YAML ints (`timeout: 3600`) become Tekton param strings (`"3600"`).
- **`pipeline/prepare.py`** — Phase 4 reads `manifest.get("blis_observe") or {}` once and passes it to every `make_pipelinerun_scenario` call in the `(packages × workloads)` loop.
- **`pipeline/tests/test_manifest.py`** — six new cases: absent, full, partial, non-mapping, unknown key, non-scalar value (parametrized over bool/list/dict/None).
- **`pipeline/tests/test_tekton.py`** — four new cases: absent kwarg, empty dict, partial dict, full dict.
- **`pipeline/README.md`** — adds the new section to the v3 schema block.

## Reviewer attention

- **Compat guarantee.** Existing manifests that don't have `blis_observe:` are unaffected. The validator defaults it to `{}`; `make_pipelinerun_scenario` with `observe={}` emits no new param entries; Tekton falls through to Pipeline defaults. Verified with `test_observe_absent_emits_no_observe_params` and `test_observe_empty_dict_emits_no_observe_params`.
- **No Pipeline.yaml change.** The Pipeline-level params declared in PR #338 stay as-is. They remain the defaults for any field a transfer doesn't specify.
- **Bool rejection.** `isinstance(True, int)` is `True` in Python; the validator explicitly checks `isinstance(v, bool)` to keep booleans out of the type-accepts-scalar branch. Tested via the parametrized `test_blis_observe_rejects_non_scalar_values`.
- **Tekton TaskRun ceiling.** The new `timeout:` field is passed to `blis observe` as its `--timeout` flag (workload duration). The outer TaskRun wall-clock ceiling is still hardcoded to `90m` in `pipeline/lib/tekton.py:11-14`. A `--timeout` value > the TaskRun ceiling minus setup/teardown overhead would be silently truncated by Tekton. Flagged as out-of-scope in the #337 decision comment; worth a separate validator check if it becomes a real issue.
- **README schema block.** The surrounding v3 schema block already has unrelated drift from PR #109 (target/config/epp_image vs the current `component:` section) — tracked in #402. This PR only adds the new section; it doesn't try to fix the surrounding drift.

## Stale-reference sweep

Grepped `blis_observe`, `make_pipelinerun_scenario`, and the five param names across `pipeline/`, `.claude/skills/`, `docs/`. Every hit is either new code I just wrote, the existing Pipeline-level wiring in `pipeline/pipeline.yaml` (which this PR complements, not invalidates), or unchanged test names. The bootstrap-skill side (parsing config.md → emitting `blis_observe:` in the generated bundle) is tracked separately in #403.

## Test plan

- [x] `python -m pytest pipeline/ .claude/skills/sim2real-{analyze,bootstrap,translate}/tests/ -q` — 1109 passed, 2 xfailed (pre-existing)
- [ ] Manual: an experiment with `blis_observe: {timeout: 3600}` in its transfer.yaml produces a PipelineRun with `timeout: "3600"` in `spec.params` and no other observe entries (so Tekton applies Pipeline defaults for the other four).
- [ ] Manual: an experiment without `blis_observe:` produces PipelineRuns with no observe entries (compat — identical to pre-PR behavior).

## Related

- Closes #337
- Depends on #338 (Pipeline-level params, merged)
- Parallel: #402 (README schema drift), #403 (bootstrap-skill emission from config.md)

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -317,6 +317,13 @@ epp_image:                  # optional
 pipeline:                   # optional — defaults applied if absent
   name: sim2real            # Pipeline resource name referenced in PipelineRuns (default: "sim2real")
   yaml: pipeline/pipeline.yaml  # path relative to repo root (default: "pipeline/pipeline.yaml")
+
+blis_observe:               # optional — per-transfer overrides for blis observe tuning
+  maxConcurrency: 10000     # all keys optional; absent keys fall through to the
+  timeout: 1800             # Pipeline-level defaults in pipeline/pipeline.yaml
+  warmupRequests: 50        # (currently 10000 / 1800 / 50 / 60s / "").
+  prewarmDuration: 60s      # Values are emitted as PipelineRun params to override
+  extraArgs: ""             # the Pipeline defaults — Tekton handles the merge.
 ```
 
 All paths are relative to the repo root and validated at Phase 1.
--- a/pipeline/lib/manifest.py
+++ b/pipeline/lib/manifest.py
@@ -237,3 +237,29 @@ def _validate_v3_fields(data: dict) -> None:
         raise ManifestError(
             f"pipeline.yaml must be a relative path, got: {pipeline['yaml']}"
         )
+
+    # blis_observe (optional): per-transfer overrides for the blis-observe
+    # tuning params the sim2real Pipeline accepts (see pipeline/pipeline.yaml).
+    # Absent or partial → omitted keys fall through to Pipeline-level defaults.
+    observe_raw = data.get("blis_observe")
+    if observe_raw is None:
+        data["blis_observe"] = {}
+    elif not isinstance(observe_raw, dict):
+        raise ManifestError("blis_observe must be a mapping")
+    else:
+        _valid_observe_keys = {
+            "maxConcurrency", "timeout", "warmupRequests",
+            "prewarmDuration", "extraArgs",
+        }
+        unknown_observe = set(observe_raw.keys()) - _valid_observe_keys
+        if unknown_observe:
+            raise ManifestError(
+                f"blis_observe contains unknown keys: {sorted(unknown_observe)}. "
+                f"Valid keys: {sorted(_valid_observe_keys)}"
+            )
+        for k, v in observe_raw.items():
+            if not isinstance(v, (str, int, float)) or isinstance(v, bool):
+                raise ManifestError(
+                    f"blis_observe.{k} must be a scalar (string or number)"
+                )
+        data["blis_observe"] = dict(observe_raw)
--- a/pipeline/lib/tekton.py
+++ b/pipeline/lib/tekton.py
@@ -42,6 +42,11 @@ def _apply_workspace_bindings(ws_names: list, bindings: dict) -> list:
     ]
 
 
+_OBSERVE_PARAM_ORDER = (
+    "maxConcurrency", "timeout", "warmupRequests", "prewarmDuration", "extraArgs",
+)
+
+
 def make_pipelinerun_scenario(
     phase: str,
     workload: dict,
@@ -56,6 +61,7 @@ def make_pipelinerun_scenario(
     blis_git_commit: str = "",
     blis_git_repo_url: str = "",
     model: str = "",
+    observe: dict | None = None,
 ) -> dict:
     """Generate a PipelineRun with resolved scenario content."""
     if spec_content is None:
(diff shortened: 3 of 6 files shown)


kalantar:

Clean implementation with both endpoints (manifest validation and PipelineRun emission) well-tested in isolation; remaining findings are polish, not blockers.

### Recommendations
- **Consider deriving the observe key allow-list from a single source of truth** (`pipeline/lib/manifest.py:250`) — manifest.py defines _valid_observe_keys = {...} and tekton.py defines _OBSERVE_PARAM_ORDER = (...) independently. They must stay in lockstep, but nothing enforces it — adding a sixth key to one and forgetting the other would silently drop the new key from generated PipelineRuns. Importing _OBSERVE_PARAM_ORDER from tekton (or extracting both to a shared constant) would remove the divergence path entirely.

PR #407, opened 2026-06-28
fix(sim2real-check): correct §1c/§1d/§1g audit methodology

### LINKED ISSUE #385: sim2real-check: audit script produces false WARN/FAIL on workload distribution and burstiness checks
The `sim2real-check` skill (`.claude/skills/sim2real-check/SKILL.md`) has two methodology bugs in Section 1 (workload parity) that produce false positives even when sim and real are in perfect agreement. Both surfaced while auditing the soft-reflective real-cluster bundle (4 workloads × 3 variants); both checks initially flagged divergence; on investigation, neither was a real sim/real issue — sim and real consume the same `workload.GenerateWorkload` and produce identical schedules.

## 1d (also 1c) — wrong truncation method in expected-value computation

**Current behavior.** The audit computes the expected mean of the input/output token distribution via Monte Carlo with rejection-sampling truncation (resample if the draw falls outside `[min, max]`).

**Actual behavior of the generator.** BLIS's `LognormalSampler` (`sim/workload/distribution.go:105-125`) uses CLAMP truncation:

```go
result := int(math.Round(val))
if s.min > 0 && result < s.min { result = s.min }
if s.max > 0 && result > s.max { result = s.max }
```

**Why it matters.** For token-length specs where a non-trivial fraction of raw samples falls below `min`, the two methods give meaningfully different means. For the `code_generation` output spec (mu=5.2711, sigma=1.2498, min=50, max=10000):

| method | mean (Monte Carlo, N=500k) |
|---|---:|
| Closed-form `exp(mu + σ²/2)` (untruncated) | 425.0 |
| **CLAMP (what BLIS does)** | **424.4** |
| REJECT (what the audit assumed) | 471.7 |
| Measured in real trace | 421.8 |

About 13.9% of raw samples fall below 50 for this distribution. Clamping piles them at exactly 50; rejection-truncation discards them. Result: every cell tripped a "−10% mean shortfall" WARN that was a definitional mismatch, not a workload divergence.

**Fix.** Compute expected moments by sampling-then-clamping using the BLIS formula above whenever the spec has bounds. Use the closed-form only when no clamping bounds are set. (Same fix applies to 1c if the input distribution clamps a non-trivial mass at its bounds.)

## 1g — aggregate CV is the wrong test for multi-client workloads

**Current behavior.** The audit computes CV of inter-arrival times across the merged trace and verdicts against the spec's `arrival.process` (CV ~0 for `constant`, ~1.0 for `poisson`).

**Why it's wrong.** When a workload has multiple clients with `process: constant` at different rates, each client is independently periodic (per-client CV = 0 exactly — `sim/workload/arrival.go:113-115` returns a fixed IAT). The aggregate stream is `concat(per_client_arrivals).sort_by(arrival_time)` (`sim/workload/generator.go:356-364`), so the merged CV is whatever the deterministic superposition produces — for two streams at 1.2 and 2.8 req/s, ~0.5. Pure Poisson superposition is approached only as N → ∞ (Palm-Khintchine).

**Numbers from baseline/cg_4.** Per-`client_id` CV: critical 0.0000, sheddable 0.0000. Aggregate CV: 0.4967. The script flagged FAIL on aggregate, even though every client is a perfect metronome.

**Fix.** Compute CV per `client_id` and verdict each client against its own `arrival.process` spec. Optionally also report aggregate CV as a diagnostic but do not gate on it.

## Notes

- Both bugs are sim/real-symmetric — sim and real consume the same `workload.GenerateWorkload`, so neither bug was masking a real divergence in this run. They are check-correctness issues only.
- `cmd/observe_cmd.go:386` — `blis observe` calls the same `workload.GenerateWorkload` as sim, confirming identical schedules from the same YAML + seed.
- Surfaced during audit of `kalantar-msb/soft-reflective` `workspace/runs/sr-mk/results` (4 workloads × 3 variants = 12 cells); audit report at `workspace/runs/sr-mk/results_charts/vllm-d/parity_report.md` for cross-reference.

## Suggested skill edits

- §1c / §1d: spec the expected-mean computation as clamp-truncation (with a helper that takes the DistSpec including min/max).
- §1g: compute CV per `client_id`; aggregate CV is informational only.

kalantar:
## Summary

Two methodology bugs in `sim2real-check` §1 (workload parity) produced false WARN/FAIL on every audit run despite sim/real actually being in agreement:

- **§1c/§1d (token distributions)** — the skill didn't pin how the *expected* mean is computed, and the audit defaulted to rejection-sampling truncation. BLIS samplers (`inference-sim/sim/workload/distribution.go`) **clamp**, not reject. For `code_generation` output (`mu=5.2711, sigma=1.2498, min=50`), ~13.9% of raw samples fall below 50; clamping piles them at 50, rejection discards them. Closed-form mean 425.0, clamp 424.4, reject 471.7, measured 421.8 — every cell tripped a "−10% mean shortfall" WARN that was a definitional mismatch, not workload divergence.

- **§1g (burstiness)** — the skill compared the CV of the merged inter-arrival stream against the spec's `arrival.process`. When a workload has multiple `constant` clients at different rates, each client is independently periodic (per-client CV = 0 exactly) but the merged stream's CV is whatever the deterministic superposition produces (~0.5 for two streams at 1.2 + 2.8 req/s). Baseline `cg_4`: per-client CV 0.000/0.000, aggregate 0.497 → spurious FAIL despite every client being a perfect metronome.

Both are check-correctness issues — sim and real consume the same `workload.GenerateWorkload`, so neither was masking a real divergence.

Closes #385

## Changes

- `.claude/skills/sim2real-check/SKILL.md` §1c/§1d: spec the expected-moment computation as sample-then-clamp Monte Carlo (N ≥ 100k) matching the BLIS sampler semantics. The closed-form `exp(mu + sigma²/2)` is valid only when no clamp bounds are set. Cite `LognormalSampler.Sample` at `distribution.go:105` so the next implementer can verify the algorithm.
- `.claude/skills/sim2real-check/SKILL.md` §1g: spec the CV computation per `client_id`, verdict each client against its own `arrival.process` spec. Aggregate (merged-stream) CV is reported as a diagnostic only. Cite the per-client independence (`arrival.go`) and the merge-sort superposition (`generator.go`).

## Reviewer attention

- The fix is in skill prompt text only — no Python code paths. `sim2real-check` has no `tests/` directory in the CI test set, so no automated tests exist or were added. The "test" is that the next audit run on a known-good run produces a PASS instead of the previously-observed false WARN/FAIL.
- The methodology spec uses pseudocode-style instructions ("group by `client_id`, sort by `arrival_time_us`, take consecutive diffs, compute CV"). That matches how every other audit step in the skill is specified.

## Sweep

`grep` for "coefficient of variation", "inter-arrival", "rejection sampl", "closed-form", "aggregate CV", "merged-stream CV" across `.claude/skills/`, `docs/`, `pipeline/`, `README.md` — only this PR's own edits hit. No stale references elsewhere.

## Test plan

- [x] `ruff check .claude/skills/ pipeline/ --select F` — clean
- [x] Methodology spec compared against actual sampler code at `inference-sim/sim/workload/distribution.go:105-125` and `arrival.go:107-115`.

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -93,16 +93,20 @@ When both exist, note any differences (especially `aggregate_rate`) and use the 
 - Show: table with each class — expected %, actual %, difference
 
 **1c. Input Token Distribution**
-- From sim: `input_distribution` params (mean, std_dev, min, max)
-- From real: compute mean, stdev, min, max of `input_tokens` column
-- Tolerance: mean within 5%, stdev within 10%
-- Show: table with expected vs actual for mean, stdev, min, max
+- From sim: `input_distribution` params (e.g. for lognormal: mu, sigma, min, max).
+- From real: compute mean, stdev, min, max of `input_tokens` column.
+- **Expected-moment derivation (critical):** use sample-then-clamp Monte Carlo matching the BLIS sampler semantics (`inference-sim/sim/workload/distribution.go`). For each draw: sample from the configured distribution → round to int → clamp to `[min, max]` when bounds are set → floor at 1. Compute the expected mean and stdev as moments of the clamped samples (N ≥ 100k for stable means).
+  - For lognormal (`LognormalSampler.Sample` at `distribution.go:105`), the closed-form `exp(mu + sigma²/2)` is correct ONLY when no clamp bounds are set. When `min > 0` (the common case — e.g. `code_generation` output at `mu=5.2711, sigma=1.2498, min=50`), a non-trivial below-min mass is piled at `min` by clamping, shifting the mean materially below the closed-form value.
+  - Do not use rejection-sampling truncation (resample if outside `[min, max]`) for expected-moment computation. BLIS clamps; rejection produces a different distribution and a measurably different mean.
+- Tolerance: mean within 5%, stdev within 10%.
+- Show: table with expected vs actual for mean, stdev, min, max. Note the expected-mean derivation (`clamped MC` vs `closed-form`) in the table caption so the reader knows which path was used.
 
 **1d. Output Token Distribution**
-- From sim: `output_distribution` params
-- From real: compute from `output_tokens` column
-- Same tolerances as input
-- Show: same table format as 1c
+- From sim: `output_distribution` params.
+- From real: compute from `output_tokens` column.
+- Use the same sample-then-clamp expected-moment derivation as 1c.
+- Same tolerances as 1c.
+- Show: same table format as 1c.
 
 **1e. Prefix Tokens**
 - From sim: check if any client has `prefix_group` or prefix config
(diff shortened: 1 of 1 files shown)


kalantar:

Both methodology bugs are fixed correctly, the skill text matches the BLIS sampler semantics in distribution.go and arrival.go, and the specialists' "blocking" concerns are speculative or overstated.

### Recommendations
- **Restore the specific line-number anchors for arrival.go and generator.go in the §1g rationale** (`.claude/skills/sim2real-check/SKILL.md:128`) — SKILL.md line 128 cites just the file names; the issue body and the verified code anchor the claim at arrival.go:113-115 (ConstantArrivalSampler.SampleIAT) and generator.go around line 356-364 (the sort.SliceStable merge). Including the line ranges preserves the verifiability the 'Why per-client' paragraph is meant to provide, with negligible cost.

Issue #377, opened 2026-06-23
feat(deploy): `slots add` / `slots remove` for live slot-pool management (local + remote)


kalantar:
## Problem

The slot-pool surface today is awkward and asymmetric between local and `--remote` modes:

- **Local mode** (`deploy.py run` on the workstation): operator edits `workspace/setup_config.json` directly. PR #373's `_refresh_namespaces` re-reads it on each dispatch cycle. Works.
- **`--remote` mode**: the in-cluster orchestrator Pod reads `/data/workspace/setup_config.json` from a per-Pod `emptyDir` populated once by an init container's `cp -r /data/config/. /data/workspace`. Workstation edits don't propagate. The CM update doesn't propagate either, because the workspace is an emptyDir snapshot frozen at Pod start. The only way to mutate it is `oc exec` / `oc cp` directly into the running Pod — undocumented, easy to fat-finger.

This was the failure mode behind the original #374 incident: a user added a third namespace to `setup_config.json` mid-run expecting #373's behavior, waited 5+ minutes, nothing happened — because the running Pod's snapshot couldn't see the edit, AND the published image was 19 days stale (separately fixed in #374).

Even with `:latest` now current and `_refresh_namespaces` actually present in the running orchestrator, the CLI surface for "I want to add a slot" / "I want to remove one" is still `oc cp`-into-a-running-pod. That's the right primitive but the wrong abstraction layer for an operator.

## Goal

Replace `oc cp` and ad-hoc `setup_config.json` editing with first-class CLI verbs:

```
deploy.py slots add <namespace>     [--no-provision] [--no-publish]
deploy.py slots remove <namespace>  [--cleanup]      [--drain-only]
deploy.py slots list                # show pool + per-slot ready/busy state
```

Symmetric across local and `--remote` modes — the operator doesn't need to know which the orchestrator is running in.

### Two new library helpers in `pipeline/lib/`

- **`provision_namespace(ns, …)`** — cluster setup for a slot: PVCs, HF secret, RBAC, Tekton tasks/pipeline. Idempotent. Each sub-step that's already correct is a no-op; each that's not is applied. On any sub-step failure, surfaced to the user at warn level (see "Surface every divergence" below).

- **`publish_slot_pool(namespaces, …)`** — write the slot list everywhere it needs to land:
  1. Update `workspace/setup_config.json` (local file) — atomic write (tmpfile + rename).
[...]

PR #411, opened 2026-06-29
fix(rbac): grant orchestrator pods + events in slot namespaces

kalantar:
## Summary

Add `pods` (get/list/watch) and `events` (get/list) to `pipeline/rbac/sim2real-runner-ns.yaml` so the orchestrator's `sim2real-runner` ServiceAccount can list pods and events in every slot namespace via the namespaced Role+RoleBinding alone — independent of whether the cluster-scoped augment (`sim2real-runner-cluster.yaml`) was applied.

When the operator running `setup.py` lacks cluster-RBAC privileges, `step_rbac` silently skips the cluster augment (`required=False` at `setup.py:405`), and the orchestrator ends up with no `pods`/`events` verb anywhere — symptom is a continuous `[WARN] get_all_pods(...) kubectl failed (rc=1)` from `_check_pod_health` and `[<workload>] pod query failed: ...` from `_check_pending_pods`.

Closes #410

## Scope clarification (vs the original issue body)

The vet step caught that the issue overstated the fix. Issue #410 listed two root causes:

1. **Role is missing `pods`/`events`.** Real. Fixed here.
2. **Namespaced binding only covers the primary ns.** Not real. `main()` at `pipeline/setup.py:825-848` already iterates `cfg.namespaces` and calls `step_rbac(ns_cfg)` per slot, so `sim2real-runner-ns.yaml` lands in every slot ns today. User verified with `kubectl get rolebinding sim2real-runner-ns -n=kalantar-1` — binding exists.

So this PR is YAML-only. No code change in `step_rbac`.

## Changes

- `pipeline/rbac/sim2real-runner-ns.yaml`: add two rules (pods get/list/watch, events get/list) to the existing Role; expand the header comment to document why each rule exists and which orchestrator code path needs it. `watch` is granted on pods even though current callsites only `list` — granting it now costs nothing and prevents a future client-go user from re-tripping this issue.
- `pipeline/tests/test_setup_pipeline.py`: new `TestSimRealRunnerNsRoleContent` class with four tests that parse the actual YAML and assert each expected rule is present with the right verbs. Guards against silent regression — if a rule is dropped, the orchestrator's health triage silently degrades to no-ops.

## Reviewer attention

- The `events` rule is also new — the original symptom called out pods, but `get_events` in `lib/health.py:281` was equally affected (silently returned `[]`, weakening triage of `FailedScheduling`/`ImagePullBackOff` warning events). Bundled here since the fix shape is identical.
- This PR is a stop-gap for #377 (`provision_namespace` per-slot helper). When that lands, it will absorb the file content unchanged.

## Sweep

`grep` for `sim2real-runner-ns`, `sim2real-runner-cluster`, `orchestrator RBAC`, `orchestrator permission` across `pipeline/`, `docs/`, `.claude/`, `README*` — no stale doc references. The only `RBAC` hits in `docs/troubleshooting.md` are about the EPP's RBAC (a different SA), unrelated to the orchestrator's.

## Test plan

- [x] Existing `TestStepRbacApply` tests still pass (6 tests).
- [x] New `TestSimRealRunnerNsRoleContent` tests pass (4 tests).
- [x] Full pytest suite: 1116 passed (was 1112).
- [ ] On-cluster smoke once merged: re-run `setup.py`, then `kubectl auth can-i list pods -n=kalantar-1 --as=system:serviceaccount:kalantar-0:sim2real-runner` → expect `yes`. Same for `events`. Then confirm `[WARN] get_all_pods(...) kubectl failed` warns no longer fire on a `deploy.py run --remote`.

### DIFF
--- a/pipeline/rbac/sim2real-runner-ns.yaml
+++ b/pipeline/rbac/sim2real-runner-ns.yaml
@@ -1,5 +1,14 @@
-# Additional namespace-scoped permissions for the sim2real-runner SA:
-# - PVC and Secret read access for orchestrator slot health checks.
+# Namespace-scoped permissions the orchestrator SA needs in each slot ns:
+# - pipelineruns (tekton.dev): create/manage workload PipelineRuns.
+# - persistentvolumeclaims, secrets: slot-readiness checks.
+# - pods: per-PipelineRun pending-pod triage (deploy.py _check_pending_pods)
+#   and broader non-Tekton pod health sweeps (lib/health.py:get_all_pods).
+# - events: triage context for pending/failed pods
+#   (lib/health.py:get_events) — feeds the pod_pending classifier.
+# Granted here rather than via the cluster bundle so the orchestrator
+# works when the operator running setup.py lacks cluster-RBAC privileges
+# (sim2real-runner-cluster.yaml is best-effort and silently skipped in
+# that case — issue #410).
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -12,6 +21,12 @@ rules:
   - apiGroups: [""]
     resources: ["persistentvolumeclaims", "secrets"]
     verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "list"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/pipeline/tests/test_setup_pipeline.py
+++ b/pipeline/tests/test_setup_pipeline.py
@@ -607,3 +607,59 @@ class TestStepRbacApply:
         assert exc.value.code == 1
         # No kubectl apply should have been issued.
         assert len(self._apply_calls(mock_run)) == 0
+
+
+class TestSimRealRunnerNsRoleContent:
+    """Content assertions on pipeline/rbac/sim2real-runner-ns.yaml.
+
+    The orchestrator SA needs every rule in this Role to be present in every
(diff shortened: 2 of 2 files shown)


kalantar:

Two-line YAML fix with content-verification tests; the inaccuracies specialists flagged are in comments newly introduced by this diff and worth a quick fix-on-the-way-in, but nothing here blocks merge.

### Recommendations
- **YAML header comment names a function that does not exist** (`pipeline/rbac/sim2real-runner-ns.yaml:4`) — The added comment says `deploy.py _check_pending_pods`, but the actual function in deploy.py:708 is `_handle_pending_pods`. Verified via `grep -n _check_pending_pods pipeline/deploy.py` (no hits) and `grep -n _handle_pending_pods pipeline/deploy.py` (line 708). Comment was introduced by this diff, so a one-word fix here keeps the RBAC-to-code trail accurate.
- **YAML comment attributes `get_events` to the `pod_pending` classifier, but the two paths are separate** (`pipeline/rbac/sim2real-runner-ns.yaml:7`) — Verified: `get_events` (health.py:281) is called only by `_check_pod_health` (deploy.py:868) and feeds `triage_pod`. `_handle_pending_pods` (deploy.py:708-748) is the only caller of `pod_pending.parse_pod_conditions` and never calls `get_events`. The added comment conflates two independent triage paths in the very file documenting why the rule exists.
- **Test docstrings repeat the same two inaccuracies** (`pipeline/tests/test_setup_pipeline.py:652`) — `test_pods_rule_present` docstring names `_check_pending_pods` (does not exist; actual: `_handle_pending_pods`), and `test_events_rule_present` (line 660) says events feeds `pod_pending classification` even though `get_events` is consumed by `triage_pod`/`_check_pod_health`, not the `pod_pending` module. Fix together with the YAML comment.

PR #413, opened 2026-06-29
fix(deploy): cancel wait-loop waits for terminal reason, not just non-Running

kalantar:
## Summary

`_cancel_and_delete_pipelinerun`'s wait loop broke on any PR status reason other than `Running`/`Started`. Tekton's `CancelledRunningFinally` matches that predicate but is transient — it means "main tasks cancelled, finally block is currently executing." Breaking out of the wait and force-deleting the PR while it's in that state killed the in-flight `llmdbenchmark-teardown` TaskRun and orphaned helm releases in the slot namespace.

Replace the predicate with an explicit `_TERMINAL_PR_REASONS` set covering Tekton's actually-terminal reasons (`Cancelled`, `PipelineRunCancelled`, `Succeeded`, `Completed`, `Failed`, `PipelineRunTimeout`, `CreateRunFailed`, `Unknown`). `CancelledRunningFinally`, `PipelineRunStopping`, and `PipelineRunStoppingTimeout` are deliberately not in the set — the wait now keeps polling through them.

Closes #412

## Why this matters (verified live)

Soft-reflective run on 2026-06-29 cancelled 16 PipelineRuns via tier-2 pod-health escalation. Every one logged `cancelled (now: CancelledRunningFinally)` and left helm releases behind:

```
$ helm list -n kalantar-0
infra-llmdbench                  kalantar-0  1  deployed  llm-d-infra-v1.4.0
qwen-qwe-079d55ee-wen3-14b-ms    kalantar-0  1  deployed  llm-d-modelservice-v0.4.15
qwen-qwe-079d55ee-wen3-14b-router kalantar-0  1  deployed  llm-d-router-gateway-dev-v0
```

`tkn pr list -n kalantar-0` returned zero PipelineRuns at the same moment — the PRs were force-deleted before their finally tasks could uninstall their charts.

## Test plan (covered)

`test_deploy_cancel.py` adds three regression tests covering exactly the failure shape:

- `test_cancel_waits_through_cancelled_running_finally` — feeds the stateful status mock a `Running → CancelledRunningFinally × 3 → Cancelled` sequence; asserts the wait observes `CancelledRunningFinally` at least 3 times without breaking, then breaks on terminal `Cancelled`, and `kubectl delete` runs exactly once after.
- `test_cancel_times_out_when_finally_never_completes` — finally never terminates; asserts the wait loop runs all 40 iterations, then the force-delete still fires (preserving the existing safety property — a stuck teardown can't pin a slot indefinitely).
- `test_cancel_breaks_on_non_cancel_terminal_reason` — `Running → Failed` short-circuits the wait.

All five existing tests in `test_deploy_cancel.py` still pass. Full suite: 1119 passed (was 1116). Lint clean.

## Reviewer attention

- The `_TERMINAL_PR_REASONS` set includes `Unknown`. Rationale: `_check_pipelinerun_status` returns `Unknown` when the PR is no longer queryable (kubectl error or PR vanished). The original predicate also broke on `Unknown` (anything not `Running`/`Started`), so including it preserves that behavior — a vanished PR is functionally terminal for our wait.
- The fix scope is intentionally narrow: only `_cancel_and_delete_pipelinerun`'s wait loop. Other status-check sites (`_reconcile_on_resume`, `_handle_timeout`) maintain their own inline tuples of terminal reasons; consolidating them onto `_TERMINAL_PR_REASONS` is a follow-up DRY cleanup, deliberately not in this PR.
- The 120s timeout itself isn't changed. If teardown legitimately exceeds that under cluster load, the orchestrator still force-deletes; that's a separate hardening item.

## Sweep

`grep` for `CancelledRunningFinally`, `CancelledRunFinally`, "cancel wait", "wait cancel" across `docs/`, `pipeline/`, `.claude/skills/`, `README*` — no stale doc references touching the cancel-wait state machine.

## Out of scope (filed separately or deferred)

- Cleaning up the orphan helm releases that have already accumulated in `kalantar-0` from this run. Fix the upstream first; existing orphans can be cleared with `deploy.py reset` or manual `helm uninstall`.
- The downstream tier-2 triage policy that cancels a pair when an unrelated leftover pod is stuck Pending in the slot ns. That's a larger design question.

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -391,6 +391,25 @@ def _cmd_build(run_dir: Path, namespace: str, skip_build: bool) -> str:
 
 # ── PipelineRun helpers ──────────────────────────────────────────────────────
 
+# Tekton PipelineRun `.status.conditions[0].reason` values that mean the run is
+# fully done — no further state transitions, finally tasks (including
+# `llmdbenchmark-teardown`) have completed.  `CancelledRunningFinally`,
+# `PipelineRunStopping`, and `PipelineRunStoppingTimeout` are explicitly
+# NOT terminal: they're transient states where finally is still executing,
+# and deleting the PipelineRun while it's in one of those states force-kills
+# the in-flight teardown and orphans its helm releases (issue #412).
+_TERMINAL_PR_REASONS = frozenset({
+    "Cancelled",
+    "PipelineRunCancelled",        # older Tekton versions
+    "Succeeded",
+    "Completed",
+    "Failed",
+    "PipelineRunTimeout",
+    "CreateRunFailed",
+    "Unknown",                     # PR not found → also terminal for our wait
+})
+
+
 def _cancel_and_delete_pipelinerun(pr_name: str, namespace: str) -> bool:
     """If a PipelineRun with the given name exists, cancel it, wait for it to
     finish cancelling, then delete it so a fresh one can be submitted.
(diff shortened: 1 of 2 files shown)


kalantar:

The wait-loop fix is correct, the new tests pin the regression, and the remaining specialist concerns are either pre-existing patterns this PR doesn't change or polish below the bar for blocking merge.

### Recommendations
- **Stale literal line numbers in two test comments** (`pipeline/tests/test_deploy_cancel.py:119`) — Lines 118-119 and 184 say "line 415" for the initial _check_pipelinerun_status call, but after the 19-line _TERMINAL_PR_REASONS block was inserted the call moved to deploy.py:434. Trivial follow-up; consider dropping the literal number so it doesn't drift again.
- **Including "Unknown" in _TERMINAL_PR_REASONS conflates "PR not found" with transient kubectl failure** (`pipeline/deploy.py:409`) — _check_pipelinerun_status (deploy.py:719-728) returns "Unknown" for any non-zero kubectl exit code, not just NotFound. The PR body explicitly preserves prior behavior here (the old `not in (Running, Started)` predicate also broke on Unknown), so this is not a regression — but if a future hardening pass wants to make the wait loop robust to transient API blips, distinguishing NotFound from Unknown in _check_pipelinerun_status is the right shape.

PR #428, opened 2026-06-30
feat(pipeline): cluster_ops file-system primitives

kalantar:
## Summary

Adds the first slice of `pipeline/lib/cluster_ops.py` — file-system-only read/write/update of `clusters/<cluster_id>/cluster_config.json`. Second child of Epic #416 (depends on #417 / layout module, just merged). The kubectl/oc primitives (`provision_namespace`, `apply_cluster_resources`) are scoped to #420.

## API (per design)

```
read_cluster_config(cluster_id)              -> dict   # {} when file absent
write_cluster_config(cluster_id, cfg)        -> None   # atomic
update_cluster_config(cluster_id, /, **updates) -> dict
```

## Implementation notes

- **Atomic write** — `tempfile.mkstemp(dir=path.parent, ...)` + `Path.replace`, matching `pipeline/lib/state_machine.py:70-77`. The tmpfile is created in the same directory as the target, so the rename is POSIX-atomic on the same filesystem. A concurrent reader sees either the prior full content or the new full content, never a torn write.

- **Deep-merge for `secret_names` / `workspaces`** — uses the existing `pipeline.lib.values.deep_merge`. For these two top-level keys the merge preserves untouched sub-keys; for all other top-level keys, the update is a full replace. Schema-defensive: if the existing value for a deep-merge key is somehow not a dict (corrupt config), the update falls back to full replace rather than crashing.

- **Positional-only `cluster_id`** — `def update_cluster_config(cluster_id: str, /, **updates)`. Without `/`, calling `update_cluster_config(\"ocp-east\", cluster_id=\"ocp-east\")` (legitimate on first write — populating the schema field) would raise `TypeError: got multiple values for argument 'cluster_id'`. The positional-only marker routes the kwarg into `**updates`. Test caught this during development.

- **Layout-only path resolution** — every path comes from `pipeline.lib.layout` (no `\"workspace/clusters/\" + cluster_id` mashing). Last bullet of the issue's acceptance.

## Tests

24 tests in `pipeline/tests/test_cluster_ops.py`, organized into three classes:

- `TestReadClusterConfig` — missing-file returns `{}`, missing-dir returns `{}`, parsed-dict round trip.
- `TestWriteClusterConfig` — creates parent dir, writes JSON, overwrites, tmpfile-cleaned-up on success AND on simulated failure, **no torn write** (prior content intact after mid-write failure), tmpfile-in-same-directory (so rename is atomic).
- `TestUpdateClusterConfig` — empty start, return-value, full-replace cases (`namespaces`, scalar), deep-merge cases (`secret_names` preserves / overrides, `workspaces` preserves / recurses), multi-key update in one call, defensive-replace when existing or supplied value isn't a dict, layout-only path resolution, positional-only signature collision-resistance.

## Acceptance (from #419)

- [x] `pipeline/lib/cluster_ops.py` created with the three functions.
- [x] Atomic write verified (tmpfile + rename; no torn writes under simulated mid-write failure — `test_atomic_rename_no_torn_write`).
- [x] `read_cluster_config` returns `{}` for absent file (no exception).
- [x] `update_cluster_config` correctly deep-merges `secret_names` and `workspaces` sub-dicts.
- [x] `pipeline/tests/test_cluster_ops.py` covers all three functions including edge cases.
- [x] `ruff check pipeline/lib/cluster_ops.py --select F` clean.
- [x] Uses `pipeline/lib/layout.py` for path resolution.
- [x] Full CI suite locally: 1168 passed, 2 xfailed (24 new, no regressions).

## Stale-reference sweep

Grepped \`docs/\`, \`.claude/skills/\`, README, CLAUDE.md for \`cluster_ops\` and \`cluster_config.json\`. Only hits are in the design doc and design proposals (historical sources). No README/CLAUDE.md updates in this PR — those are #425's deliverable.

## Notes for reviewer

- The deep-merge contract reuses \`values.deep_merge\` to keep the merge semantics consistent with \`assemble.py\`. For the actual \`cluster_config.json\` schema the inner shapes are dict-of-strings (\`secret_names\`) and dict-of-dicts (\`workspaces\`), so the recursion never reaches \`values._merge_lists\`'s k8s-identity branch.
- Read-modify-write in \`update_cluster_config\` is not inter-process-locked. The design notes no concurrent updaters are expected; later steps that introduce slot-add operations will still need to handle that constraint at the caller level if it ever becomes real.

Base: \`refactor/v2-step-0\`
Parent epic: #416
Depends on: #417 (layout module — merged)
Closes #419

### DIFF
--- a/pipeline/lib/cluster_ops.py
+++ b/pipeline/lib/cluster_ops.py
@@ -1,0 +1,109 @@
+"""File-system primitives for ``clusters/<cluster_id>/cluster_config.json``.
+
+Step 0 carves cluster-side responsibilities out of ``pipeline/setup.py`` into
+a new ``pipeline/cluster.py`` entry point backed by this library. This module
+is the first slice: read/write/update of the per-cluster config file. The
+kubectl/oc primitives (``provision_namespace``, ``apply_cluster_resources``)
+land in a separate child.
+
+Writes are atomic (tmpfile + ``Path.replace``) so a concurrent reader either
+sees the previous full content or the new full content, never a torn write.
+Path resolution is delegated to :mod:`pipeline.lib.layout` — no string
+mashing of workspace paths in this module.
+"""
+
+from __future__ import annotations
+
+import json
+import tempfile
+from pathlib import Path
+
+from pipeline.lib import layout
+from pipeline.lib.values import deep_merge
+
+
+# Top-level keys in cluster_config.json whose contents are merged into the
+# existing value rather than wholesale-replaced. Per design (see
+# docs/epics/step-0/design.md, "cluster_config.json schema"): ``secret_names``
+# is a flat string→string dict; ``workspaces`` is a dict of Tekton workspace
+# bindings. Partial updates to either should preserve untouched keys.
+_DEEP_MERGE_KEYS = ("secret_names", "workspaces")
+
+
+def read_cluster_config(cluster_id: str) -> dict:
+    """Read ``clusters/<cluster_id>/cluster_config.json``.
+
+    Returns the parsed JSON object as a dict. Returns ``{}`` if the file does
+    not exist — callers that need a not-yet-provisioned cluster to fail loudly
+    should check the result themselves.
+    """
+    path = layout.cluster_config_path(cluster_id)
+    if not path.exists():
+        return {}
+    with path.open("r", encoding="utf-8") as f:
+        return json.load(f)
+
+
+def write_cluster_config(cluster_id: str, config: dict) -> None:
+    """Atomically write ``clusters/<cluster_id>/cluster_config.json``.
+
+    Creates the cluster directory if it does not exist. Uses tmpfile +
+    ``Path.replace`` so a concurrent reader never observes a partially
+    written file: the rename is atomic on POSIX (and Windows) for paths on
+    the same filesystem, and the tmpfile is created in the same directory.
+    """
+    path = layout.cluster_config_path(cluster_id)
+    path.parent.mkdir(parents=True, exist_ok=True)
+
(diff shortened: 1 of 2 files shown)


kalantar:

A tight, well-tested first slice of cluster_ops that meets every acceptance criterion in the issue.

### Recommendations
- **Module docstring's framing about pipeline/cluster.py is forward-looking; consider tightening so first-time readers don't expect that entry point to exist yet.** (`pipeline/lib/cluster_ops.py:3`) — pipeline/cluster.py is not present on this branch (verified via ls of pipeline/). The current docstring at cluster_ops.py:3-10 does say this module is 'the first slice' and that other primitives 'land in a separate child', which is honest scoping, but a quick reader could still go looking for cluster.py. Optional polish only.
- **Drop the '(and Windows)' aside from the atomic-rename docstrings; CI is Linux-only and Path.replace atomicity on Windows is not a documented guarantee.** (`pipeline/lib/cluster_ops.py:52`) — cluster_ops.py:9 and :52 both claim 'atomic on POSIX (and Windows) for paths on the same filesystem'. The project CI runs ubuntu-latest only per CLAUDE.md, and Windows MoveFileExW/ReplaceFileW is not POSIX-equivalent. Minor accuracy nit.
- **Remove the unused monkeypatch parameter from the _isolated_experiment_root fixture, or switch to monkeypatch.setattr for automatic teardown.** (`pipeline/tests/test_cluster_ops.py:13`) — test_cluster_ops.py:13 takes monkeypatch but the body uses direct assignment (layout._EXPERIMENT_ROOT = tmp_path) with manual restore after yield. The parameter is dead and the idiomatic monkeypatch.setattr form would handle restoration automatically. Cosmetic.

PR #430, opened 2026-06-30
feat(pipeline): cluster_ops provision_namespace + apply_cluster_resources

kalantar:
## Summary

Adds the kubectl/oc-driven primitives that pair with the file-system slice landed in #428. This is Child 4 of epic #416 (Step 0 — Foundation).

- `ProvisionResult` dataclass — surfaces every divergence per sub-step (`steps_ok`, `steps_skipped`, `steps_failed`, `diverged`).
- `provision_namespace(namespace, cluster_config, *, secret_values=None, skip=())` — five-step namespace bootstrap (`namespace`, `rbac`, `secrets`, `pvc`, `tekton`) in design order. Idempotent: already-correct = no-op. Sub-steps surface as structured failures rather than silently overwriting.
- `apply_cluster_resources(cluster_id)` — applies `pipeline/pipeline.yaml` (Tekton `Pipeline` definition) across every namespace in `cluster_config["namespaces"]`. Idempotent via `kubectl apply`.
- Helpers `detect_openshift`, `check_cluster_reachable`, `secret_exists` are added to `cluster_ops`. `check_cluster_reachable` raises `ClusterUnreachableError` instead of `sys.exit` (library convention). The `setup.py` copies remain in place — they're scheduled for deletion in Child 8 (#424) per the design table.

## Reviewer attention

- **`secret_values` shape.** Design comment shows `registry_creds` as `{user, token}`. I extended to `{server, user, token}` because the docker-registry secret requires a `--docker-server` arg and the registry URL isn't in `cluster_config.json` (it stays in `setup_config.json` per design). Caller (#421 `cluster.py provision`) will resolve `server` from setup_config.
- **PVC defaults.** `cluster_config["workspaces"]` only carries claim names; size + access mode aren't part of the schema. Hardcoded `50Gi` / `ReadWriteMany` to match today's `setup.py`. Documented in the docstring + a module constant.
- **`apply_cluster_resources` namespace scope.** Tekton `Pipeline` is namespaced upstream, so "cluster-wide" here means "the same definition applied across every namespace registered for this cluster" — same wire-level behavior as today's `step_tekton`.
- **Divergence detection scope.** `kubectl apply` merges by default; deep content-divergence detection (e.g. comparing existing Secret bytes against intended) is out of scope for this child. Surfaced failures are `kubectl apply` failures + the structured RBAC cluster-forbidden / secret-absent paths. Documented in the docstring.
- **`_envsubst` is pure-Python.** Drops the system-binary dependency `setup.py` had on the `envsubst` shell tool.

## Verification

```bash
.venv/bin/python -m pytest pipeline/ .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-bootstrap/tests/ .claude/skills/sim2real-translate/tests/
# 1210 passed, 2 xfailed in 31.80s

.venv/bin/python -m ruff check pipeline/ .claude/skills/ --select F
# All checks passed!

.venv/bin/python -m ruff check pipeline/lib/cluster_ops.py --select F
# All checks passed! (per issue acceptance)
```

66 tests in `test_cluster_ops.py` (38 new) — `FakeRun` harness intercepts every subprocess call. Coverage: happy path (kubectl + oc), namespace race, skip arg (named + unknown), secret apply / reuse / skip / partial, RBAC cluster-forbidden skipped vs hard failure failed, PVC existing-skipped + storage_class injection, tekton-no-yaml skipped, `apply_cluster_resources` per-namespace fan-out + idempotency + missing-yaml raise.

## Stale-reference sweep

`grep` of `cluster_ops.*` and `ProvisionResult` across `*.md` only hit `docs/epics/step-0/design.md` — and the design doc is the spec being implemented, now matching reality. No README / CI / CLAUDE.md updates needed (no user-facing CLI yet — that lands in #421).

## Acceptance

- [x] `provision_namespace` implements all 5 sub-steps per design contract.
- [x] Secret sub-step honors absent `secret_values` entries: skip with structured entry if no existing Secret; reuse if existing Secret present.
- [x] `ProvisionResult` records `steps_ok` / `steps_skipped` / `steps_failed` accurately.
- [x] `skip` arg suppresses sub-steps named in it; remaining sub-steps proceed.
- [x] `apply_cluster_resources` applies the cluster-wide Pipeline definition idempotently.
- [x] `detect_openshift` / `check_cluster_reachable` / `secret_exists` available from `cluster_ops`.
- [x] `pipeline/tests/test_cluster_ops.py` extended with kubectl/oc mocks for both functions; covers happy path + divergence + skip arg.
- [x] `ruff check pipeline/lib/cluster_ops.py --select F` is clean.

Closes #420
Parent: #416

### DIFF
--- a/pipeline/lib/cluster_ops.py
+++ b/pipeline/lib/cluster_ops.py
@@ -11,18 +11,36 @@ is split into two slices:
 * kubectl/oc primitives — :func:`provision_namespace` (per-namespace
   resources: namespace, RBAC, secrets, PVCs, Tekton tasks) and
   :func:`apply_cluster_resources` (Tekton Pipeline definition). Each
-  primitive is idempotent: a step that's already correct is a no-op, a
-  step that diverges is surfaced as a structured failure on
-  :class:`ProvisionResult` rather than silently overwriting.
+  primitive is idempotent in the ``kubectl apply``/``oc new-project``
+  sense: applying the same manifest twice converges on the same cluster
+  state, and an already-existing object is not an error. Note that
+  ``kubectl apply`` server-side merges — this module does NOT compare
+  the live object's contents to the supplied manifest, so drift between
+  a cluster-mutated resource and the source-of-truth manifest is
+  reconciled silently. Per-step outcomes are reported via
+  :class:`ProvisionResult` (see "outcome reporting" below).
 
-Error-handling convention (per design):
+Outcome reporting:
 
-* Idempotent applies that succeed: silent.
-* Idempotent applies that surface a name collision with different content:
-  structured failure entry on ``ProvisionResult.steps_failed``.
-* Hard failures (auth, network): exceptions propagate to the caller —
-  :class:`ClusterUnreachableError` for connectivity issues,
-  ``subprocess.CalledProcessError`` for unexpected non-zero exits.
+* Successful sub-step (apply returned zero, or pre-check said
+  already-exists): recorded on ``ProvisionResult.steps_ok``.
+* Intentional suppression (caller-supplied ``skip``) or operator
+  prerequisites not provided (e.g. absent secret value + no
+  pre-installed Secret) or best-effort YAMLs that hit a tolerated
+  permission error (cluster-scoped RBAC Forbidden): recorded on
+  ``ProvisionResult.steps_skipped``. Soft divergence — flips
+  :attr:`ProvisionResult.diverged` to True without anything in
+  ``steps_failed``.
+* Non-zero exit from kubectl/oc, missing required YAML, or malformed
+  config: recorded on ``ProvisionResult.steps_failed``. Hard divergence.
+
+Hard failures (auth, network) that prevent reasoning about per-step
+outcomes raise instead — :class:`ClusterUnreachableError` for
+connectivity issues (from :func:`check_cluster_reachable`), and
+``subprocess.CalledProcessError`` from :func:`apply_cluster_resources`'s
+per-namespace applies. No exception ever escapes
+:func:`provision_namespace`: every sub-step lands in exactly one of
+``steps_ok`` / ``steps_skipped`` / ``steps_failed``.
 """
 
 from __future__ import annotations
--- a/pipeline/lib/cluster_ops.py
+++ b/pipeline/lib/cluster_ops.py
@@ -434,9 +434,9 @@ def _step_secrets(
             else:
                 skipped.append(f"{key}({name})")
             continue
-        manifest = _build_secret_manifest(key, name, namespace, value)
+        manifest, reason = _build_secret_manifest(key, name, namespace, value)
         if manifest is None:
-            return ("failed", f"unknown secret key '{key}' — no manifest builder")
+            return ("failed", f"secret '{key}' ({name}): {reason}")
         proc = _run(
             ["kubectl", "apply", "-f", "-"],
             input=manifest, check=False, capture=True,
@@ -454,52 +454,68 @@ def _build_secret_manifest(
     name: str,
     namespace: str,
     value: object,
-) -> str | None:
+) -> tuple[str | None, str | None]:
     """Render a Kubernetes Secret manifest YAML via ``kubectl create
     secret ... --dry-run=client -o yaml``.
 
-    Returns ``None`` if the key is unrecognised — caller surfaces as a
-    structured failure. The four conventional keys are:
+    Returns ``(manifest, None)`` on success and ``(None, reason)`` on any
+    failure — unknown key, incomplete docker-creds dict, or a non-zero
+    ``kubectl create --dry-run`` exit. The caller surfaces ``reason`` on
+    ``ProvisionResult.steps_failed`` so distinct failure modes get
+    distinct operator-facing messages.
+
+    The four conventional keys are:
 
     * ``hf_token`` / ``github_token`` — generic Secret with a single
       string field (``HF_TOKEN`` / ``token``).
     * ``registry_creds`` / ``dockerhub_creds`` — docker-registry Secret;
       value is a dict with ``server``, ``user``, ``token``.
+
+    All ``_run`` calls pass ``check=False`` so a non-zero dry-run is
+    reported through the return value rather than raised — provision_namespace
+    must never let a sub-step bypass :class:`ProvisionResult`.
     """
     if key == "hf_token":
-        return _run(
+        proc = _run(
             ["kubectl", "create", "secret", "generic", name,
              f"--namespace={namespace}",
              f"--from-literal=HF_TOKEN={value}",
              "--dry-run=client", "-o", "yaml"],
-            capture=True,
-        ).stdout
+            check=False, capture=True,
+        )
+        if proc.returncode != 0:
+            return None, f"kubectl create --dry-run failed: {(proc.stderr or proc.stdout).strip()}"
+        return proc.stdout, None
     if key == "github_token":
-        return _run(
+        proc = _run(
             ["kubectl", "create", "secret", "generic", name,
              f"--namespace={namespace}",
              f"--from-literal=token={value}",
              "--dry-run=client", "-o", "yaml"],
-            capture=True,
-        ).stdout
+            check=False, capture=True,
+        )
+        if proc.returncode != 0:
+            return None, f"kubectl create --dry-run failed: {(proc.stderr or proc.stdout).strip()}"
+        return proc.stdout, None
     if key in ("registry_creds", "dockerhub_creds"):
         if not isinstance(value, dict):
-            return None
-        server = value.get("server", "")
-        user = value.get("user", "")
-        token = value.get("token", "")
-        if not (server and user and token):
-            return None
-        return _run(
+            return None, f"value is not a dict (expected {{server, user, token}}, got {type(value).__name__})"
+        missing = [field for field in ("server", "user", "token") if not value.get(field)]
+        if missing:
+            return None, f"value is missing required fields: {', '.join(missing)}"
+        proc = _run(
             ["kubectl", "create", "secret", "docker-registry", name,
              f"--namespace={namespace}",
-             f"--docker-server={server}",
-             f"--docker-username={user}",
-             f"--docker-password={token}",
+             f"--docker-server={value['server']}",
+             f"--docker-username={value['user']}",
+             f"--docker-password={value['token']}",
              "--dry-run=client", "-o", "yaml"],
-            capture=True,
-        ).stdout
-    return None
+            check=False, capture=True,
+        )
+        if proc.returncode != 0:
+            return None, f"kubectl create --dry-run failed: {(proc.stderr or proc.stdout).strip()}"
+        return proc.stdout, None
+    return None, f"unknown secret key '{key}' — no manifest builder"
 
 
 def _step_pvc(
--- a/pipeline/lib/cluster_ops.py
+++ b/pipeline/lib/cluster_ops.py
@@ -1,21 +1,37 @@
-"""File-system primitives for ``clusters/<cluster_id>/cluster_config.json``.
+"""Cluster-config file primitives + kubectl/oc primitives for per-cluster setup.
 
 Step 0 carves cluster-side responsibilities out of ``pipeline/setup.py`` into
-a new ``pipeline/cluster.py`` entry point backed by this library. This module
-is the first slice: read/write/update of the per-cluster config file. The
-kubectl/oc primitives (``provision_namespace``, ``apply_cluster_resources``)
-land in a separate child.
+a new ``pipeline/cluster.py`` entry point backed by this library. The module
+is split into two slices:
 
-Writes are atomic (tmpfile + ``Path.replace``) so a concurrent reader either
-sees the previous full content or the new full content, never a torn write.
-Path resolution is delegated to :mod:`pipeline.lib.layout` — no string
-mashing of workspace paths in this module.
+* File-system primitives — :func:`read_cluster_config`,
+  :func:`write_cluster_config`, :func:`update_cluster_config`. Atomic writes
+  (tmpfile + ``Path.replace``); path resolution delegated to
+  :mod:`pipeline.lib.layout`.
+* kubectl/oc primitives — :func:`provision_namespace` (per-namespace
+  resources: namespace, RBAC, secrets, PVCs, Tekton tasks) and
+  :func:`apply_cluster_resources` (Tekton Pipeline definition). Each
+  primitive is idempotent: a step that's already correct is a no-op, a
+  step that diverges is surfaced as a structured failure on
+  :class:`ProvisionResult` rather than silently overwriting.
+
+Error-handling convention (per design):
+
+* Idempotent applies that succeed: silent.
+* Idempotent applies that surface a name collision with different content:
+  structured failure entry on ``ProvisionResult.steps_failed``.
+* Hard failures (auth, network): exceptions propagate to the caller —
+  :class:`ClusterUnreachableError` for connectivity issues,
+  ``subprocess.CalledProcessError`` for unexpected non-zero exits.
 """
 
 from __future__ import annotations
 
 import json
+import shutil
+import subprocess
 import tempfile
+from dataclasses import dataclass, field
 from pathlib import Path
 
 from pipeline.lib import layout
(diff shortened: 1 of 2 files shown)


kalantar:

One structural invariant violation in _build_secret_manifest (uses default check=True so dry-run failures escape ProvisionResult entirely) and one misleading-error path that conflates incomplete-dict with unknown-key need fixing before merge; the rest is polish.

### Important
- **_build_secret_manifest's three kubectl dry-run calls omit check=False, so any non-zero exit raises CalledProcessError out of provision_namespace and voids ProvisionResult** (`pipeline/lib/cluster_ops.py:470`) — Verified at pipeline/lib/cluster_ops.py:469-501: all three _run() invocations inside _build_secret_manifest pass capture=True but inherit check=True from _run's signature at line 160. Every other _run call inside the _step_* path (e.g. line 440, line 535) explicitly sets check=False so failures become a recorded outcome. This breaks the module's stated invariant at line 247-250: 'every sub-step lands in exactly one of steps_ok, steps_skipped, or steps_failed.' On a non-zero dry-run exit, CalledProcessError propagates through _step_secrets and out of provision_namespace, so the caller gets no ProvisionResult at all. Trivial fix: add check=False to all three calls and return None on non-zero returncode.
- **_step_secrets reports incomplete registry_creds/dockerhub_creds dicts as "unknown secret key" — the key is known, only the dict fields are missing** (`pipeline/lib/cluster_ops.py:438`) — Verified at lines 485-492: _build_secret_manifest returns None when registry_creds/dockerhub_creds has missing server/user/token. _step_secrets at line 438-439 surfaces every None as "unknown secret key '{key}' — no manifest builder". An operator with a partially-configured docker-registry secret sees an error that points them at the wrong cause. Distinguish the two None paths (unknown-key vs incomplete-dict) and produce a specific reason for each.

### Recommendations
- **Add tests for the failure branches of _step_secrets/_step_rbac/_step_pvc (kubectl apply fails, required YAML absent, missing claimName)** (`pipeline/tests/test_cluster_ops.py`) — Verified by listing test names: the suite covers happy paths and the RBAC Forbidden branch, but never exercises (a) the apply-failed branches at cluster_ops.py:444, 539, 561, (b) the required-YAML-absent branch at 391, or (c) the missing-claimName branch at 519. These are simple to stub via the existing FakeRun harness.
- **Module docstring conflates diverged with steps_failed; clarify that steps_skipped also drives diverged** (`pipeline/lib/cluster_ops.py:15`) — Line 15-16 says a divergent step is 'surfaced as a structured failure on ProvisionResult'. ProvisionResult.diverged at line 259-260 is True for steps_skipped OR steps_failed; a soft divergence (e.g. operator did not supply a secret value) flips diverged without anything in steps_failed. One-sentence wording fix.
- **Remove unused tmp_path fixture parameter from test_applies_pipeline_yaml_to_each_namespace** (`pipeline/tests/test_cluster_ops.py:828`) — Verified at line 828-837: the signature requests tmp_path but the body never references it; path isolation is handled by the autouse _isolated_experiment_root fixture. Trivial cleanup.

kalantar:
Addressed both blocking findings:

**`_build_secret_manifest` no longer voids `ProvisionResult` on dry-run failure.** All three `kubectl create --dry-run=client` calls now pass `check=False`, and the function returns `(manifest, None)` on success or `(None, reason)` on failure. A non-zero exit is recorded on `steps_failed` instead of raising out of `provision_namespace`.

**Distinct error messages for distinct secret failures.** The same `(None, reason)` shape lets `_step_secrets` surface specific reasons for each failure mode:

- unknown secret key → `secret 'futuristic_token' (...): unknown secret key 'futuristic_token' — no manifest builder`
- incomplete docker-creds dict → `secret 'registry_creds' (...): value is missing required fields: token`
- wrong-type docker-creds value → `secret 'registry_creds' (...): value is not a dict (expected {server, user, token}, got str)`
- dry-run rendering failure → `secret 'hf_token' (...): kubectl create --dry-run failed: <stderr>`

Four new tests pin these messages and the no-exception-escapes invariant: `test_unknown_secret_key_distinct_error`, `test_incomplete_docker_creds_distinct_error`, `test_non_dict_docker_creds_distinct_error`, `test_dry_run_failure_routes_through_provision_result`. Full suite is 1214 passed; ruff clean.

Recommendations from the synthesizer (additional test branches for `_step_rbac`/`_step_pvc`/`_step_tekton` failure paths, module docstring wording on `diverged`, unused `tmp_path` fixture parameter) are not addressed — they are advisory and don't block merge per the fix-issue convention.

kalantar:

The module's operator-facing docstring overstates what the code actually enforces, and a stated design invariant ("no short-circuit; every sub-step lands in steps_ok/skipped/failed") is not directly asserted by any test.

### Important
- **Module docstring claims divergence is surfaced as a structured failure rather than silently overwriting, but kubectl apply has no drift detection here** (`pipeline/lib/cluster_ops.py:14`) — Verified at lines 14-16 and 21-22: docstring promises 'a step that diverges is surfaced as a structured failure on ProvisionResult rather than silently overwriting' and that 'Idempotent applies that surface a name collision with different content' become structured failures. The code uses plain `kubectl apply` everywhere (e.g. lines 396-397, 440-442, 551-554) which server-side merges and overwrites silently; there is no content-comparison logic anywhere in the module. A future maintainer will rely on an invariant that does not exist. Reword to describe what the code actually does (non-zero kubectl exit -> structured failure).
- **No test asserts the no-short-circuit invariant — that all five sub-steps execute even after an earlier failure** (`pipeline/tests/test_cluster_ops.py:565`) — Verified `test_namespace_create_failure_recorded` only checks `('namespace', 'quota exceeded') in steps_failed` and `result.diverged is True`. The inline comment at lines 581-582 explicitly states 'Later steps still run — provision_namespace doesn't short-circuit', and the design depends on this invariant, but no test asserts that the remaining four steps (rbac/secrets/pvc/tekton) were actually attempted. A regression that adds an early return on first failure would pass every existing test. Add an assertion that `len(steps_ok) + len(steps_skipped) + len(steps_failed) == 5` after an induced early-step failure.

### Recommendations
- **Loop variable `field` in `_build_secret_manifest` shadows the `field` imported from dataclasses** (`pipeline/lib/cluster_ops.py:503`) — Verified line 503 uses `field` as the comprehension variable while line 34 imports `field` from dataclasses. List comprehension scope is isolated in Python 3 and `dataclasses.field` is only used at class-definition time (line 254-256), so there's no runtime bug — but it's a readability trap worth renaming to `f` or `fname`.
- **`apply_cluster_resources` captures kubectl output but `CalledProcessError.__str__` does not surface it, making the default failure message non-actionable** (`pipeline/lib/cluster_ops.py:628`) — Verified the `_run([...], check=True, capture=True)` call at line 628-631. CalledProcessError's default str() yields only 'Command [...] returned non-zero exit status N'; the captured kubectl stderr lives on `e.stderr` and is invisible to a caller that does `log.err(str(e))`. Docstring at 618-620 promises callers can 'surface the first failure' without guidance. Either drop `capture=True` or wrap the call to re-raise including stderr in the message.
- **`_step_tekton` apply-failure path is untested** (`pipeline/tests/test_cluster_ops.py:879`) — Verified the only tekton tests are happy-path and the no-YAMLs skipped path (line 879). The failed-apply branch at cluster_ops.py:577-578 is unexercised — a test that stubs glob to return a YAML and forces `_run` to return non-zero would close the gap.
- **`_step_pvc` failure paths (missing claimName, apply failure) are untested** (`pipeline/tests/test_cluster_ops.py:835`) — Verified `_step_pvc` at cluster_ops.py:536 returns 'failed' on missing claimName and at line 555-556 on non-zero apply, but no test exercises either path. Both are operator-visible errors with specific messages.
- **`_step_rbac` missing-required-YAML failure path is untested** (`pipeline/tests/test_cluster_ops.py:773`) — Verified every RBAC test calls `_rbac_yaml_paths_exist(monkeypatch, exists=True)` (line 467). The `if not yaml_path.exists(): if required: return ('failed', ...)` branch at cluster_ops.py:390-392 is unreached. A test using `_rbac_yaml_paths_exist(monkeypatch, exists=False)` would catch a regression that loses the submodule-init hint.
- **Inline comment cites `setup.py:399-406` but line 399 is the env dict, not the yaml_paths bundle** (`pipeline/lib/cluster_ops.py:380`) — Verified setup.py line 399 is `env = {**os.environ, ...}` and `yaml_paths = [` begins at line 400. The comment should read `setup.py:400-406`. One-line fix to keep cross-references useful.
- **Test name `test_pvc_existing_skipped` implies the pvc step lands in steps_skipped, but pre-existing PVCs return ('ok', 'applied')** (`pipeline/tests/test_cluster_ops.py:835`) — Verified `_step_pvc` at cluster_ops.py:557 returns ('ok', 'applied') when every PVC already exists. The test name and 'All PVC pre-checks hit' docstring suggest the skipped path, which only fires when workspaces is empty (line 528). Rename to `test_pvc_existing_is_noop` and consider adding `assert 'pvc' in result.steps_ok` to lock the classification.

kalantar:
Iteration 2 — addressed both important findings:

- Module docstring reworded so it describes the actual contract: `kubectl apply` server-side merges silently, so non-zero exit → `steps_failed`, soft-divergence (skipped) flips `diverged` without anything in `steps_failed`, no content-drift detection. The earlier wording implied drift detection that the code doesn't do.
- Added `test_no_short_circuit_on_early_failure` pinning the invariant that every one of the five sub-steps lands in exactly one of `steps_ok` / `steps_skipped` / `steps_failed` even after an earlier failure. Existing tests only checked the recorded failure, not the coverage of subsequent steps.

71 tests pass; ruff clean. The seven `recommendations` items (loop variable shadowing `dataclasses.field`, deferred `import re`, additional `_step_pvc`/`_step_rbac`/`_step_tekton` failure-branch tests, `apply_cluster_resources` stderr surfacing, `setup.py:399` vs `:400` reference, test-name accuracy) are advisory and not addressed — they don't block merge per fix-issue convention.

kalantar:

Two filesystem calls inside step handlers can raise unhandled exceptions that break the module's pinned "no exception escapes provision_namespace" invariant; everything else is polish.

### Important
- **_step_rbac calls yaml_path.read_text() with no exception handler, can leak OSError/UnicodeDecodeError out of provision_namespace** (`pipeline/lib/cluster_ops.py:412`) — cluster_ops.py:41-43 documents an explicit invariant: 'No exception ever escapes provision_namespace: every sub-step lands in exactly one of steps_ok / steps_skipped / steps_failed.' Commit a4d9049 just pinned that invariant. But cluster_ops.py:412 calls yaml_path.read_text() unguarded — a PermissionError or IsADirectoryError on a partially-initialised tektonc-data-collection submodule (a realistic state for a fresh operator running cluster.py provision) would escape _step_rbac, the dispatch loop, and the whole function, dropping every prior step's outcome on the floor. Wrap the read in try/except OSError and route to ('failed', ...).
- **_step_tekton calls tekton_dir.glob() with no exception handler, same invariant violation as the rbac read** (`pipeline/lib/cluster_ops.py:590`) — cluster_ops.py:590 calls sorted(tekton_dir.glob('*.yaml')) right after a .exists() probe at line 588, with no surrounding try/except. PermissionError on Path.glob is a real failure mode (e.g. a submodule directory the current user cannot read into), and it would escape provision_namespace just like the read_text path, violating the pinned invariant at cluster_ops.py:41-43. Same fix shape: wrap glob in try/except OSError and convert to ('failed', ...).

### Recommendations
- **_REPO_ROOT comment says 'two parents up' but the code uses three .parent calls** (`pipeline/lib/cluster_ops.py:68`) — cluster_ops.py:68 reads '# lives at pipeline/lib/cluster_ops.py, so the repo root is two parents up' but line 69 uses .parent.parent.parent. Verified: from pipeline/lib/cluster_ops.py, two parents lands at pipeline/, three parents lands at the repo root — the code is correct, the comment is wrong. A maintainer trusting the comment could 'fix' the code and break every filesystem reference in the module.
- **_step_secrets early-returns on the first failing secret, abandoning the rest** (`pipeline/lib/cluster_ops.py:455`) — cluster_ops.py:455-463: the loop returns on the first failed _build_secret_manifest or kubectl apply, so a misconfiguration on secret #1 hides issues with secrets #2 and #3. The rest of the module (_step_rbac accumulates skipped_names, _step_pvc continues through workspaces) embodies the 'surface every divergence' philosophy from the module docstring; secrets is the odd one out. Operators have to re-run after each fix instead of seeing all secret problems in one pass.
- **test_pvc_existing_skipped is misnamed — pre-existing PVCs land in steps_ok, not steps_skipped** (`pipeline/tests/test_cluster_ops.py:876`) — _step_pvc at cluster_ops.py:560-575: when every PVC pre-exists, the loop continues and falls through to return ('ok', 'applied'). The test name implies the step is recorded as skipped, which contradicts the actual contract documented at cluster_ops.py:25-26 ('apply returned zero, or pre-check said already-exists: recorded on steps_ok'). Rename to test_pvc_existing_is_ok and add an assert 'pvc' in result.steps_ok.
- **Five reachable failure branches in the new step handlers have no test coverage** (`pipeline/tests/test_cluster_ops.py`) — Verified against test_cluster_ops.py: there is no test for _step_tekton apply non-zero (line 595-596), no test for required RBAC YAML missing (line 409-410), no test for secret kubectl apply non-zero after a successful dry-run (line 462-463), no test for PVC binding missing claimName (line 553-554), and no test for non-required RBAC YAML hitting a Forbidden without 'clusterrole' in the message (the predicate at line 423). Each is a realistic operator-facing failure mode; the test_no_short_circuit harness already has the scaffolding to cover them.

kalantar:
Iteration 3 — addressed both important findings.

The iter-2 docstring tightening (no exception escapes provision_namespace) exposed two genuine leaks the reviewer caught:

- `_step_rbac` wrapped `yaml_path.read_text()` in `try/except OSError` / `UnicodeDecodeError`. PermissionError or a non-UTF-8 file on a partially-initialised tektonc-data-collection submodule now lands on `steps_failed` instead of propagating.
- `_step_tekton` wrapped `tekton_dir.glob("*.yaml")` in `try/except OSError`. Same hazard on an unreadable submodule directory.

Two new tests pin the routing: `test_rbac_read_text_oserror_routed_to_failed`, `test_tekton_glob_oserror_routed_to_failed`. The rbac test additionally verifies the dispatch loop continues after the failure (every sub-step in exactly one list).

73 tests pass; ruff clean. The four `recommendations` items (inline comment line offset, `_step_secrets` early-return on first failed secret, `test_pvc_existing_skipped` rename, additional failure-branch tests) are advisory and not addressed per fix-issue convention.

kalantar:

The PVC pre-check silently accepts an existing PVC regardless of spec content, which directly contradicts the issue's "name collision with different content = structured failure" language and warrants an explicit author decision before merge.

### Important
- **PVC pre-check returns ok on any existing PVC without comparing size/storageClass to the supplied spec, contradicting the issue's explicit "already-exists-with-different-content = structured failure" rule** (`pipeline/lib/cluster_ops.py:562`) — cluster_ops.py:561-567 issues `kubectl get pvc` and returns the loop's continue branch on a hit, with no spec comparison. Issue #420 states "Already-exists-with-different-content = structured failure in result, not silent succeed" and design.md:275 repeats it. The module docstring (lines 16-20) documents this as a deliberate tradeoff for `kubectl apply`, but the PVC sub-step never reaches `kubectl apply` when the resource exists, so the tradeoff justification does not cover this path. Author should either narrow the spec language or add a spec comparison before the skip-create return.

### Recommendations
- **Module docstring's "no exception ever escapes provision_namespace" claim does not hold for FileNotFoundError raised by subprocess.run when kubectl/oc is absent from PATH** (`pipeline/lib/cluster_ops.py:41`) — cluster_ops.py:41-43 promises every sub-step lands in steps_ok/skipped/failed, but the dispatch loop at lines 342-352 has no try/except around handler calls, and `_run` at line 185 calls subprocess.run which raises FileNotFoundError independent of check=False. Practically the caller calls check_cluster_reachable first which would also raise, so impact is low; the docstring would be more accurate either narrowing the claim to CalledProcessError or wrapping the dispatch in OSError catch.
- **No test for the _step_rbac required-YAML-missing branch — every test forces Path.exists to return True so the "submodule init?" guard never fires** (`pipeline/tests/test_cluster_ops.py`) — cluster_ops.py:407-410 returns ('failed', f'{yaml_path} not found (submodule init?)') only when a required YAML is absent. test_cluster_ops.py:467 and :482 (the two helpers used across the test suite) both monkeypatch Path.exists to return True, leaving the guard untested. The hint is specifically the canonical operator error for an uninitialised submodule; a regression here would only surface in real-world ops.
- **No test for _step_pvc's missing-claimName guard, which is the only branch routing malformed workspace bindings to steps_failed** (`pipeline/tests/test_cluster_ops.py`) — cluster_ops.py:558-560 returns failure when a workspaces entry has no persistentVolumeClaim.claimName. No test in pipeline/tests/test_cluster_ops.py exercises this path (verified via grep on "workspace.*has no" and "claimName" — all matches are happy-path bindings). The author could add it cheaply alongside the existing PVC tests.
- **check_cluster_reachable docstring says "403 forbidden reply" but the code does a plain substring match on the string "forbidden", which has no HTTP status code involvement** (`pipeline/lib/cluster_ops.py:213`) — cluster_ops.py:212 doc text reads "A 403 'forbidden' reply is treated as reachable", but line 225 only checks `if "forbidden" in combined`. The original setup.py:192 wording ("A `forbidden` reply proves the API server responded") is accurate; adding "403" implies status-code detection that isn't happening.
- **Unused tmp_path fixture parameter on test_applies_pipeline_yaml_to_each_namespace** (`pipeline/tests/test_cluster_ops.py:998`) — test_cluster_ops.py:998 declares tmp_path in the signature but the test body never references it; the autouse _isolated_experiment_root fixture already handles filesystem isolation. Trivial cleanup.

kalantar:
Workflow iter-4 returned NEEDS DISCUSSION on the PVC pre-check accepting existing resources without spec comparison (the issue's "already-exists-with-different-content = structured failure" language). Treating drift detection as out of scope for this child — filed as #432 to track. Merging this PR.

PR #434, opened 2026-06-30
feat(pipeline): cluster.py provision orchestrator (#421)

kalantar:
Closes #421
Part of epic #416 (Step 0 — Foundation: workspace + cluster provisioning).

## Summary

Adds `pipeline/cluster.py` with a `provision` subcommand — the first consumer of `pipeline/lib/cluster_ops.py` (landed in #430). Carves cluster-side bootstrap (namespaces, RBAC, secrets, PVCs, Tekton tasks, cluster-wide Pipeline definition) out of `setup.py` while leaving `setup.py` itself untouched. `setup.py` trimming is a later epic child (#424).

CLI matches the design contract exactly:

```
python pipeline/cluster.py provision <cluster_id> \
    --namespaces NS1,NS2,NS3 \
    [--storage-class SC] \
    [--hf-token TOKEN] [--github-token TOKEN] \
    [--registry-user USER] [--registry-token TOKEN] \
    [--dockerhub-user USER] [--dockerhub-token TOKEN] \
    [--experiment-root PATH]
```

No extra flags (no \`--openshift\`, no secret-name overrides, no \`--workspace-binding\`, no \`--no-cluster\`, no \`--redeploy-tasks\`, no \`--test-push\`).

Secret resolution: **flag > env var > prompt**. Required secrets (\`--hf-token\`, \`--registry-user\` / \`--registry-token\`) prompt when neither flag nor env supplies a value. Optional secrets (\`--github-token\`, \`--dockerhub-user\` / \`--dockerhub-token\`) never prompt — when absent, the key is omitted from \`secret_values\` and \`cluster_ops.provision_namespace\` records a structured \"skipped\" or reuses any pre-installed Secret.

## Design choices worth flagging

- **Docker-registry server defaults.** \`cluster_ops._build_secret_manifest\` requires a \`server\` field for \`registry_creds\`/\`dockerhub_creds\` payloads, but the issue's CLI surface has no \`--registry-server\` flag (per design's \"no extra flags\" rule). Resolved by env-var defaults: \`REGISTRY_SERVER\` (default \`ghcr.io\`), \`DOCKERHUB_SERVER\` (default \`docker.io\`). Both env-overridable, no flag.
- **Write-before-apply ordering.** \`cluster_ops.apply_cluster_resources\` reads \`cluster_config.json\` from disk. So \`cmd_provision\` writes the config *before* invoking apply, even though the design's textual step order has write at step 7 and apply at step 4. The on-disk result is identical either way (idempotent).
- **\`created_at\` preservation.** When an existing \`cluster_config.json\` already has \`created_at\`, it is preserved byte-for-byte; otherwise stamped via \`datetime.now(timezone.utc).strftime(\"%Y-%m-%dT%H:%M:%SZ\")\`.
- **\`dockerhub_creds\` opt-in.** \`secret_names.dockerhub_creds\` is only \`\"dockerhub-creds\"\` when both \`--dockerhub-user\` and \`--dockerhub-token\` resolve to non-empty values; otherwise \`\"\"\`. Matches the schema in the parent epic's design.

## Exit codes

| Condition | Exit |
|---|---|
| Empty/whitespace-only \`--namespaces\` | 2 |
| \`pipeline/pipeline.yaml\` missing | 1 |
| Any per-namespace \`steps_failed\` non-empty | 1 |
| \`steps_skipped\` alone (soft divergence) | 0 (reflected in summary line) |
| All namespaces \`steps_ok\` only | 0 |

Per-namespace summary line: \`<ns>: ok\` or \`<ns>: diverged: failed=<step>(<reason>), skipped=<step>(<reason>)\`. Failed items first.

## Tests

\`pipeline/tests/test_cluster_py.py\` (27 tests, mocked kubectl/oc):
- \`TestParser\` (3) — accepts required positional + \`--namespaces\`, rejects unknown subcommands, rejects unknown flags.
- \`TestParseNamespaces\` (4) — CSV split, whitespace strip, empty rejection.
- \`TestBuildClusterConfig\` (4) — hardcoded defaults, dockerhub-name conditional, \`created_at\` preservation.
- \`TestResolveSecretValues\` (7) — flag/env/prompt priority, github+dockerhub never-prompt, server env overrides, partial-dockerhub skip.
- \`TestFormatSummary\` (3) — ok / diverged / failed-before-skipped ordering.
- \`TestProvisionOrchestration\` (6) — happy path returns 0, one-namespace-failed returns 1, idempotent re-run preserves \`created_at\`, empty namespaces returns 2, missing pipeline.yaml returns 1, dockerhub-creds round-trip.

All mocks use the \`FakeRun\` pattern from \`pipeline/tests/test_cluster_ops.py\`. No network, no real cluster.

Full pipeline suite: 1177 passed, 0 regressions. \`ruff check pipeline/ --select F\` clean.

## Doc sweep

- \`CLAUDE.md\` — added a paragraph documenting \`pipeline/cluster.py\` as the new Step-0 entry point and noting that \`setup.py\` remains the legacy 8-step flow until #424 trims it. The flow diagram (\`setup.py → prepare.py → ...\`) is unchanged because cluster.py is parallel to setup.py during Step 0, not in series.
- \`pipeline/README.md\` — intentionally not updated. It is a detailed user manual that will need a wholesale rewrite once #424 lands; a partial update now would create a confusing duplicate. The PR description and CLAUDE.md mention are sufficient discoverability for the Step-0 era.
- No skill prompts reference \`cluster.py\` (it's a new file).
- CI (\`ruff check pipeline/ ...\`, \`pytest pipeline/ ...\`) already covers the new file by glob; no \`.github/workflows/test.yml\` change required.

## Out of scope

- Trimming \`setup.py\` — #424.
- \`setup_config.json\` retirement — later epic child.
- Demo against a real cluster — acceptance criterion will be exercised manually post-merge; the mocked end-to-end test covers the orchestration paths.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,8 @@ python pipeline/run.py     --experiment-root ../admission-control switch <run-na
 
 **`pipeline/run.py`** — Lists, inspects, and switches between runs. `switch` syncs generated scorer plugin files into the experiment repo's `llm-d-inference-scheduler/` directory. Pass `--experiment-root` to point at the experiment repo (default: current directory).
 
+**`pipeline/cluster.py`** — New Step-0 entry point (epic #416). Cluster-side bootstrap only: `cluster.py provision <cluster_id> --namespaces NS1,NS2,...` writes `workspace/clusters/<cluster_id>/cluster_config.json` and provisions namespaces, RBAC, Secrets, PVCs, and Tekton tasks. The cluster-side responsibilities of `setup.py` are being carved into `cluster.py`; `setup.py` remains the legacy 8-step flow until its trim PR lands (see epic #416).
+
 ## Pipeline Library (`pipeline/lib/`)
 
 | Module | Purpose |
--- a/docs/superpowers/plans/2026-06-30-issue-421-cluster-py-provision.md
+++ b/docs/superpowers/plans/2026-06-30-issue-421-cluster-py-provision.md
@@ -1,0 +1,1228 @@
+# Issue #421 — `pipeline/cluster.py provision` orchestrator
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add `pipeline/cluster.py` with a `provision` subcommand — the first consumer of `pipeline/lib/cluster_ops.py`. Carves cluster-side bootstrap out of `setup.py` while the new file leaves `setup.py` untouched (it remains the legacy 8-step flow; trimming `setup.py` is a later epic child).
+
+**Architecture:**
+
+- `pipeline/cluster.py` is a thin CLI shell. All side effects (kubectl, oc, disk) live in `pipeline.lib.cluster_ops` and `pipeline.lib.layout`.
+- Flow: parse args → resolve flags > env > prompt → build `cluster_config` dict → `write_cluster_config` (must be on disk for `apply_cluster_resources` to read) → `apply_cluster_resources` → per-namespace `provision_namespace` → print summary line each → exit non-zero on any `steps_failed`.
+- The script delegates secret resolution and OpenShift detection to library functions but never imports anything from `pipeline/setup.py` — the carve-out is one-way.
+
+**Tech Stack:** Python ≥3.10, `argparse`, `pipeline.lib.cluster_ops`, `pipeline.lib.layout`, `pytest` with `monkeypatch`-driven kubectl/oc fakes.
+
+## Global Constraints
+
+- CLI surface must match the design **exactly**: positional `cluster_id`; `--namespaces`, `--storage-class`, `--hf-token`, `--github-token`, `--registry-user`, `--registry-token`, `--dockerhub-user`, `--dockerhub-token`, `--experiment-root`. No other flags (no `--openshift`, no secret-name overrides, no `--workspace-binding`, no `--no-cluster`, no `--redeploy-tasks`, no `--test-push`).
+- Secret value resolution priority: **flag > env var > interactive prompt**. Same priority for the docker-registry `server` field where it has to be resolved (env var only; defaults to `ghcr.io` for `REGISTRY_SERVER` and `docker.io` for `DOCKERHUB_SERVER`).
+- Idempotent re-run with same inputs is a no-op. `created_at` is preserved from any existing `cluster_config.json`; only set on first write.
+- Hardcoded today's-setup.py defaults:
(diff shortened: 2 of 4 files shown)


kalantar:

Tests pass and the implementation matches the design contract; remaining items are polish, not blockers.

### Recommendations
- **Lift the _which=False monkeypatch into the class-level fixture so every orchestration test is hermetic.** (`pipeline/tests/test_cluster_py.py:264`) — Only test_happy_path_returns_zero patches cluster_ops._which. The other five TestProvisionOrchestration tests rely on the host machine not having `oc` installed; on a developer box where `oc whoami` returns 0 they will flip to the OpenShift path and emit oc commands the FakeRun was never configured for. Moving `monkeypatch.setattr(cluster_ops, '_which', lambda cmd: False)` into _freeze_experiment_root (line 264) closes the gap.
- **Use copy.deepcopy for _DEFAULT_WORKSPACES instead of dict(...).** (`pipeline/cluster.py:124`) — Line 124 takes a shallow copy of a nested dict literal. No current call site mutates the returned cfg['workspaces']['data-storage'][...] entries, so it is not a bug today, but a future test helper or migration that does will silently corrupt the module-level _DEFAULT_WORKSPACES and every subsequent config. Cheap to fix preemptively.
- **Remove the unused _full_arg_setup method.** (`pipeline/tests/test_cluster_py.py:272`) — Defined at lines 272-277 inside TestProvisionOrchestration but never called by any test. The per-test setup is duplicated inline. Removing it (or converting to a real fixture) tightens the test file without changing behavior.
- **Add a TestResolveSecretValues case where both flag and matching env var are set.** (`pipeline/tests/test_cluster_py.py:142`) — The flag>env>prompt priority contract is load-bearing but no test combines a non-empty flag with a non-empty env var. The current `or`-chain happens to do the right thing for non-empty strings, but a future refactor (e.g. swapping order or guarding on `is not None`) could pass all 27 tests while violating the documented priority. One added test pinning hf_token flag='flag-val' against env HF_TOKEN='env-val' would close the contract gap.
- **Add coverage for the CalledProcessError branch of apply_cluster_resources.** (`pipeline/tests/test_cluster_py.py:388`) — cmd_provision has two distinct except clauses around apply_cluster_resources (lines 267 and 270 in pipeline/cluster.py). test_pipeline_yaml_missing_exits_one covers FileNotFoundError; the CalledProcessError branch with its different error-message prefix (`error: cluster-wide apply failed:`) has no test. Monkeypatch apply_cluster_resources to raise subprocess.CalledProcessError and assert rc==1 plus the expected stderr prefix.
- **Soften the module docstring's `replaces` language to reflect that setup.py is intentionally untouched in this PR.** (`pipeline/cluster.py:4`) — Line 4 says provision `replaces today's cluster-side responsibilities in pipeline/setup.py`. The PR body and CLAUDE.md make clear setup.py is left as the legacy flow until #424. `carves out` or `will replace` would match the actual state; `replaces` reads as already-done.

PR #440, opened 2026-07-01
fix(cluster): self-inject repo root into sys.path (#439)

kalantar:
Closes #439.

## What changed

`pipeline/cluster.py` failed with `ModuleNotFoundError: No module named 'pipeline'` when invoked from any cwd outside the sim2real repo root. Every sibling entry point (`setup.py`, `prepare.py`, `run.py`) already self-injects its repo root into `sys.path` before importing `pipeline.lib`; `cluster.py` was missing the same guard.

Adds the four-line guard used elsewhere:

\`\`\`python
_REPO_ROOT = Path(__file__).resolve().parent.parent
if str(_REPO_ROOT) not in sys.path:
    sys.path.insert(0, str(_REPO_ROOT))
\`\`\`

Guard is idempotent — no behavior change when cwd already is the repo root.

## Regression test

New `TestScriptImportFromNonRepoCwd::test_provision_help_runs_from_tmp_cwd` in `pipeline/tests/test_cluster_py.py`. Spawns a fresh Python interpreter with `cwd=tmp_path` (outside the repo tree) and asserts `cluster.py provision --help` returns 0.

**Why subprocess and not `importlib.reload`:** pytest sets up `sys.path` to include the repo root, so any in-process import of `pipeline.cluster` from a test would succeed regardless of whether the guard exists. Only a fresh interpreter with a non-repo cwd actually exercises the fix.

## Verification

- \`ruff check pipeline/ .claude/skills/ --select F\` — clean.
- \`pytest pipeline/ .claude/skills/…\` — 1277 pass, 2 xfailed (pre-existing).
- Manual repro from an experiment repo directory now succeeds.

## Sweep

Nothing docs-shaped touched — this is a hidden `sys.path` guard, not user-visible surface. `CLAUDE.md` and `pipeline/README.md` already describe `cluster.py` invocation without implying anything about cwd requirements.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,7 +45,7 @@ python pipeline/run.py      --experiment-root ../admission-control switch <run-n
 
 **Backward compat:** Omitting `--experiment-root` defaults to the current working directory. Run all pipeline commands from the experiment repo root and the default will resolve correctly without the flag.
 
-**`pipeline/setup.py`** — One-time workspace config writer. Writes `setup_config.json` and `run_metadata.json` with operator-side fields (registry, repo name, current run, orchestrator image, pipeline_yaml path, sim2real_root). Idempotent — safe to re-run. Cluster-side bootstrap (namespaces, RBAC, secrets, PVCs, Tekton tasks, Pipeline definition) lives in `cluster.py provision`.
+**`pipeline/setup.py`** — One-time workspace config writer. Writes `setup_config.json` and `run_metadata.json` with operator-side fields (registry, repo name, current run, orchestrator image, sim2real_root). Idempotent — safe to re-run. Cluster-side bootstrap (namespaces, RBAC, secrets, PVCs, Tekton tasks, Pipeline definition, and the optional `--pipeline-yaml` manifest override) lives in `cluster.py provision`.
 
 **`pipeline/prepare.py`** — 6-phase state machine. Re-running skips completed phases (tracked in `.state.json`):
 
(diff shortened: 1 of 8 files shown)


kalantar:

CLAUDE.md's setup.py description and workspace artifact table still list pipeline_yaml as a setup_config.json field even though this PR removes it from setup.py, which contradicts the code and the project's own rule about keeping the artifact schema documented.

### Important
- **CLAUDE.md artifact table and setup.py description still document pipeline_yaml as a setup_config.json field** (`CLAUDE.md:93`) — Verified at CLAUDE.md:48 ("...orchestrator image, pipeline_yaml path, sim2real_root" listed as setup.py output fields) and CLAUDE.md:93 (setup_config.json row's field list includes pipeline_yaml). The diff removes pipeline_yaml from SetupConfig and from step_config_output's dict in pipeline/setup.py, and pipeline/README.md was correctly updated to match. CLAUDE.md was not. Line 94's cluster_config.json field list also omits pipeline_yaml, opposite direction. Project's own CLAUDE.md rule requires updating docs when artifact schema changes; two specialists independently flagged this.

### Recommendations
- **README's "every subsequent re-run of cluster.py provision <same-id>" phrasing for --pipeline-yaml persistence is easy to misread** (`pipeline/README.md:65`) — Verified at pipeline/cluster.py:143-145 that _build_cluster_config_dict only preserves created_at from the existing config; pipeline_yaml comes only from the current invocation, so a re-run of provision without --pipeline-yaml drops the key and apply_cluster_resources reverts to the built-in default. The README sentence at line 65 is technically self-consistent with the "key is not written" clause but the "every subsequent re-run" phrase suggests stickiness; a short clarification ("pass --pipeline-yaml again on every re-run") would remove the ambiguity.

PR #453, opened 2026-07-01
step-1 PR 2: sim2real assemble + delete prepare/legacy-assemble (#445)

kalantar:
Closes #445 — see the issue for full acceptance criteria.

## Summary

- New `sim2real assemble --translation HASH --cluster CLUSTER_ID --run RUN_NAME [--force]` subcommand.
- Pure helper `pipeline/lib/assemble_run.py` owns validation, slice + snapshot, deep-merge, image_tag + hf-secret injection, PipelineRun generation, and metadata writing. `pipeline/sim2real.py` gains a thin `_cmd_assemble` wrapper.
- Pinned schemas (v1): `runs/<R>/run_metadata.json` and `runs/<R>/manifest.assembly.yaml`.
- Deleted: `pipeline/prepare.py`, `pipeline/lib/{state_machine,context_builder,assemble}.py` + their four dedicated test files.
- `pipeline/tests/test_epp.py`'s two integration tests re-plumbed to use `deep_merge` directly (drops the deleted `assemble_scenarios` helper).
- Stubbed `.claude/skills/sim2real-translate/SKILL.md` with the "restored in step-2" message; deleted `test_skill_substitution.py` (asserted on placeholders the stub no longer has).
- Docs/CI sweep: CLAUDE.md, pipeline/README.md, .github/workflows/test.yml, pipeline/setup.py tail hint, docs/troubleshooting.md, .claude/skills/sim2real-bootstrap/SKILL.md, pipeline/lib/layout.py + pipeline/tests/test_pipeline_override.py docstrings.

## Acceptance mapping

| Criterion | Coverage |
|-----------|----------|
| `sim2real assemble` produces run tree | `TestAssembleCommand.test_success_produces_run_dir` + `TestAssembleRun.test_end_to_end_produces_expected_files` |
| Filter algorithms to those in `translation_output.json` (warn + skip) | `TestFilterAlgorithms` + `TestAssembleRun.test_filters_unregistered_algorithms` + `TestAssembleCommand.test_warns_on_skipped_algorithms` |
| Existing `runs/<R>/` without `--force` errors | `TestAssembleCommand.test_refuses_existing_run_without_force` + `TestAssembleRun.test_refuses_existing_run_without_force` |
| `--force` overwrites cleanly | `TestAssembleCommand.test_force_overwrites` + `TestAssembleRun.test_force_overwrites_existing_run` |
| `manifest.assembly.yaml` is verbatim assembly slice | `TestWriteManifestAssembly.test_slicer_round_trip` |
[...]

PR #459, opened 2026-07-02
fix(assemble): populate benchmarkGit*/blisGit* PipelineRun params (#458)

kalantar:
Closes #458.

## Summary

Every PipelineRun generated by \`sim2real assemble\` on \`refactor/v2-step-1\` today has empty \`benchmarkGitRepoUrl\` / \`benchmarkGitCommit\` / \`blisGitRepoUrl\` / \`blisGitCommit\` params. The cluster-side pipeline uses these four params to \`git clone\` \`llm-d-benchmark\` and \`inference-sim\` before running each benchmark, so empty values break every workload run.

Root cause: \`pipeline/lib/assemble_run.py:505-509\` had an explicit deferral (\"Step-1 does not read git submodule state — PR 3 will\") introduced by #453. #446/#447/#448/#449 all merged without picking it up.

This PR ports the discovery from the deleted \`pipeline/prepare.py:_get_submodule_shas\` into \`assemble_run.py\` as a pure helper, wires it into \`assemble()\`, and surfaces missing submodules via a new side-band attr matching the existing \`skipped_algorithms\` pattern.

## Files changed

- \`pipeline/lib/assemble_run.py\` — new \`discover_framework_submodules(repo_root: Path) -> (shas, urls, missing)\` helper + \`_REPO_ROOT\` module constant + \`_FRAMEWORK_SUBMODULE_NAMES\` tuple. Call site at \`:505-509\` now calls the helper and passes real dicts.
- \`pipeline/sim2real.py\` — extended the \`skipped_algorithms\` warning loop in \`_cmd_assemble\` with a symmetric \`missing_submodules\` loop.
- \`pipeline/tests/test_assemble_run.py\` — \`TestDiscoverFrameworkSubmodules\` (5 tests: both present, one missing, both missing, no \`.gitmodules\`, extra-entries-ignored) + 2 new integration tests on \`TestAssembleRun\` (params populated end-to-end; missing-submodule fallback).
- \`pipeline/README.md\` — bullet in \`## Assemble a run\` §Inputs read block explaining the framework-submodule dependency and the \"unknown\" fallback.
- \`docs/superpowers/plans/2026-07-02-issue-458-submodule-git-params.md\` — implementation plan preserved for reviewer context.

## Acceptance criteria mapping (#458)

- [x] \`sim2real assemble\` populates the four params from \`.gitmodules\` + framework submodule HEADs — verified end-to-end by \`test_pipelinerun_params_include_framework_submodule_state\`.
- [x] Graceful degradation when a submodule directory is missing: emit \`\"unknown\"\` + warning — \`test_missing_submodule_falls_back_to_unknown_and_warns\` asserts the value AND the \`assemble_run.missing_submodules\` side-band attr; \`pipeline/sim2real.py:_cmd_assemble\` prints the operator-facing warning.
- [x] \`pipeline/tests/test_assemble_run.py\` covers: both submodules present, one missing, both missing, URL parsing from \`.gitmodules\` — all four scenarios in \`TestDiscoverFrameworkSubmodules\`, plus an extra-entries-ignored guard.
- [ ] Real-cluster gate: full \`sim2real assemble && deploy.py run && deploy.py collect\` produces \`per_request_lifecycle_metrics.json\` — not automatable from the harness. **This is the epic's end-of-step-1 demo gate that this bug blocks. Manual verification required before merging.**
- [x] \`pipeline/README.md\` documents that framework submodules must be initialized before \`assemble\` (or that \`assemble\` auto-detects and warns) — bullet added to §Inputs read.

## Sweep notes

Grepped \`benchmarkGit|blisGit|_get_submodule_shas|submodule_shas\` across \`pipeline/\`, \`docs/\`, \`.claude/skills/\`, \`CLAUDE.md\`, \`README.md\`:

- **\`pipeline/pipeline.yaml\`**: Tekton pipeline definition — the four params are declared here and consumed downstream. Unchanged, no action.
- **\`pipeline/lib/tekton.py\`**: emits the four params into every PipelineRun spec — unchanged, no action.
- **\`pipeline/lib/assemble_run.py\`**: my new helper's docstring references legacy \`_get_submodule_shas\` as a \"matches legacy\" pointer for readers auditing the port. Intentional.
- **\`docs/superpowers/plans/*\`, \`docs/epics/*\`**: frozen historical artifacts, not swept.
- **\`.claude/skills/\`**: 0 hits.
- **\`CLAUDE.md\`, \`README.md\`**: 0 hits.

## Verification

- \`.venv/bin/python -m pytest pipeline/\` — **1158 passed**, 2 xfailed (from 1151 baseline; +7 tests: 5 discovery + 2 integration).
- \`.venv/bin/python -m pytest .claude/skills/sim2real-{analyze,bootstrap,translate}/tests/\` — 89 passed.
- \`.venv/bin/ruff check pipeline/ .claude/skills/ --select F\` — clean.
- **Real-repo smoke test**: ran \`discover_framework_submodules(<sim2real-repo-path>)\` against the actual initialized main repo — returned real SHAs (\`d059008f7abb\` for \`inference-sim\`, \`5208cf38d84c\` for \`llm-d-benchmark\`) and URLs, empty \`missing\`. Fix works end-to-end against real state.

## Non-obvious design choices worth reviewer attention

1. **URLs from \`.gitmodules\` (declarative) rather than \`git remote get-url origin\` (runtime).** Legacy \`prepare.py\` used the runtime remote; the issue's proposed solution says \`.gitmodules\`. I went with \`.gitmodules\`: declarative, doesn't depend on the submodule directory being populated, and matches how a fresh clone will look before \`git submodule update --init\`.
2. **\`\"unknown\"\` propagates to the PipelineRun verbatim** — cluster-side \`git clone\` will fail on it. This is the intended posture: \`assemble\` succeeds locally so the operator can inspect the run dir; cluster fails visibly at the right step. Matches legacy semantics exactly.
3. **Side-band warning attr \`assemble_run.missing_submodules\`** mirrors the existing \`skipped_algorithms\` protocol. \`assemble()\`'s return contract is unchanged; the CLI wrapper reads both attrs after the call.
4. **Component submodule intentionally not ported.** Legacy \`_get_submodule_shas\` also read \`manifest[\"component\"][\"path\"]\` and emitted \`shas[\"component\"]\`. The four PipelineRun params in scope for this bug are all about \`llm-d-benchmark\` and \`inference-sim\`; component image ref comes from the registered translation, not a git ref. Per issue.
5. **Worktree gotcha for future readers.** Running \`sim2real assemble\` from a fresh git worktree (rather than the primary clone) will see both submodules as missing because worktrees don't automatically populate submodules. Run \`git submodule update --init\` inside the worktree if you need real SHAs there. Not in scope for this PR but worth flagging.

### DIFF
--- a/docs/superpowers/plans/2026-07-02-issue-458-submodule-git-params.md
+++ b/docs/superpowers/plans/2026-07-02-issue-458-submodule-git-params.md
@@ -1,0 +1,711 @@
+# Fix empty benchmarkGit*/blisGit* params in sim2real assemble Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Port the framework-submodule SHA + URL discovery from the deleted `pipeline/prepare.py:_get_submodule_shas` into `pipeline/lib/assemble_run.py` so `sim2real assemble` populates `benchmarkGitRepoUrl`, `benchmarkGitCommit`, `blisGitRepoUrl`, `blisGitCommit` in generated PipelineRun YAMLs. Today they land empty and every cluster-side `git clone` step breaks.
+
+**Architecture:** Add a `discover_framework_submodules() -> tuple[dict[str,str], dict[str,str], list[str]]` helper in `assemble_run.py` that returns `(shas, urls, missing_names)`. Shas come from `git rev-parse HEAD` in each submodule directory (falls back to `"unknown"`). URLs come from parsing `.gitmodules` at the framework's `REPO_ROOT` (declarative, doesn't depend on submodule being populated). The `assemble` orchestrator calls the helper, forwards `shas` + `urls` to `generate_pipelineruns`, and stores `missing_names` on the side-band attribute `assemble_run.missing_submodules` (matches the existing `skipped_algorithms` pattern) for `sim2real.py:_cmd_assemble` to surface as operator warnings.
+
+**Tech Stack:** Python 3.10+, pytest, `configparser` (stdlib — parses `.gitmodules` since it's INI-shaped), `subprocess` (already imported transitively). No new deps.
+
+## Global Constraints
+
+- Base branch: `refactor/v2-step-1` (not `main`)
+- `assemble_run.py` is a **pure module**: no argparse, no `print`, no `sys.exit`. Warnings surface via side-band attributes (`assemble_run.skipped_algorithms`, and now `assemble_run.missing_submodules`); the CLI wrapper in `pipeline/sim2real.py` reads those and prints.
+- Framework `REPO_ROOT` is computed as `Path(__file__).resolve().parent.parent.parent` — same pattern as `pipeline/lib/cluster_ops.py:70`. Do NOT use `layout.experiment_root()` here; the two framework submodules always live in the framework repo, not the experiment repo (matches legacy `_get_submodule_shas`).
+- Framework submodule names are hardcoded: `("inference-sim", "llm-d-benchmark")`. Matches legacy. The `component` submodule handling that legacy `_get_submodule_shas` also did is out of scope (per issue).
+- `"unknown"` sentinel for missing SHAs matches legacy — the string reaches the PipelineRun spec verbatim; the cluster-side `git clone` will fail on it. That's intended: local `assemble` succeeds so the operator can inspect the run dir; the cluster run visibly fails at the right step.
+- Path discipline: every path passed to `Read`/`Edit`/`Write`/`Bash` must contain `.claude/worktrees/issue-458-submodule-git-params/`. After every edit batch, verify `git -C <parent-repo-root> status` is clean.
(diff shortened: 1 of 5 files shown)


kalantar:

Core fix is correct and well-tested; the remaining specialist findings are narrow edge cases or pre-existing artifacts that don't rise to blocking.

### Recommendations
- **Consider adding a symmetric CLI-level test for the missing_submodules stderr warning** (`pipeline/tests/test_sim2real.py:835`) — test_sim2real.py:835 has test_warns_on_skipped_algorithms which asserts the stderr text for the existing side-band warning pattern; the new missing_submodules block in sim2real.py:393-400 has no matching capsys-based test. The integration tests in test_assemble_run.py check the side-band attr directly but not the CLI print path. Purely polish given the mirrored code is trivial.
- **Consider surfacing rev-parse HEAD failures via missing_submodules instead of only via the 'unknown' value** (`pipeline/lib/assemble_run.py:91`) — assemble_run.py:82-92 catches subprocess.CalledProcessError/OSError and sets shas[name]='unknown' but does not append the name to missing, so the CLI warning at sim2real.py:393-400 stays silent. The narrow case is a submodule dir that exists with .git but where rev-parse fails; operator would still see the failure downstream when the cluster clone runs against 'unknown', so this is genuine polish, not a bug.

PR #470, opened 2026-07-02
step-2 PR 1: slicer.translation_hash_with_sources

kalantar:
Closes #465
Refs #463 (Epic: step-2 — Skill-driven translation)

## Summary

Adds `pipeline/lib/slicer.py:translation_hash_with_sources(manifest, experiment_root) -> str` — the source-content-aware hash that step-2's skill-driven `translate` will use to decide when to invalidate a translation. Extends the existing `translation_slice`/`translation_hash` substrate; leaves BYO's `_compute_translation_hash` in `sim2real.py` untouched (different producer, different inputs — as scoped in the design doc §14).

## Design

Envelope hashed:

```json
{"slice": <translation_slice(manifest)>,
 "sources": [{"name": "...", "sha256": "..."}, ... sorted by name]}
```

- Iterates `manifest["algorithms"]` sorted by name so `[a, b]` and `[b, a]` hash to the same value.
- Reads each `algorithms[i].source` file's bytes from `experiment_root` and folds its SHA-256 into `sources`.
- Missing source file → `AssembleError("source file not found: <path>")`. Imported from `pipeline.lib.assemble_run` inside the function to avoid the `assemble_run → slicer` import cycle.
- Algorithms without a `source` field (or non-dict entries) are skipped — matches the projection semantics of `translation_slice`, which also omits missing fields.
- Empty algorithms list / baseline-only manifest returns a valid hash — `sources` is `[]`.

## Acceptance criteria (from #465)

- [x] `translation_hash_with_sources(manifest, experiment_root)` exists and returns a stable 64-char SHA-256 hex.
- [x] Hash covers both `translation_slice(manifest)` and the byte contents of each `algorithms[i].source` file.
- [x] Algorithm order normalized — `[a, b]` and `[b, a]` produce the same hash (tested via `test_stable_across_algorithms_list_reordering`).
- [x] Missing source file → `AssembleError` with a clear `source file not found: <path>` message (tested via `test_missing_source_file_raises_assemble_error` and `test_missing_source_error_includes_full_path`).
- [x] Tests: both algorithms present, one missing (error), source file empty, source file binary content, multi-algorithm, hash stability across process invocations (four pinned golden hashes: single, multi, empty, unicode).

## Test matrix

Added 20 tests across two classes (`TestTranslationHashWithSources`, `TestGoldenFileStability`). Full suite: `1178 passed, 2 xfailed in 22.60s`. Lint (`ruff check pipeline/ --select F`) clean.

## Non-obvious choices worth reviewer attention

- **Deferred `AssembleError` import** inside `translation_hash_with_sources` — `pipeline/lib/assemble_run.py` already imports `slicer`, so a top-level import here would cycle. The alternative (define a `SlicerError` in slicer.py) contradicts the issue's acceptance criteria which explicitly names `AssembleError`.
- **Skipping algorithms without a `source` key** rather than raising — matches `translation_slice`'s "omit-when-absent" posture. Callers that require every algorithm to have a source can enforce that at the manifest-load layer; this hash function is intentionally tolerant.
- **`sources` element uses `{name, sha256}` (not just a bare hex list)** — carries the algorithm identity so future debug/inspect tooling can print the envelope and read it. Cost is 16 chars of JSON overhead per algorithm.
- **Sweep**: grepped `docs/`, `.claude/`, and `pipeline/` for `translation_hash_with_sources`; all matches are either the new code or forward-looking spec references in `docs/epics/step-2/design.md` that this PR fulfills. No stale references.

### DIFF
--- a/pipeline/lib/slicer.py
+++ b/pipeline/lib/slicer.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+from pathlib import Path
 from typing import Any
 
 TRANSLATION_FIELDS: list[str] = [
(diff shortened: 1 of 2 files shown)


kalantar:

Implementation satisfies every acceptance criterion with correct, well-tested hashing; the specialists' concerns are documentation nits and a defensive-coding polish, none of which block merge.

### Recommendations
- **Consider widening the source-read except clause from FileNotFoundError to OSError for consistency with assemble_run._load_yaml** (`pipeline/lib/slicer.py:150`) — pipeline/lib/assemble_run.py:148 wraps read failures with `except OSError as exc: raise AssembleError(f"cannot read {path}: {exc}")`. The new code catches only FileNotFoundError, so PermissionError / IsADirectoryError / other OSError subtypes escape as raw OS exceptions rather than the AssembleError contract the docstring implies. The narrower scope is consistent with the acceptance criteria wording ("file not found"), but broadening it would match the existing project convention.
- **test_baseline_only_manifest_hashes docstring claims equality with slice-only hash that does not hold and is not asserted** (`pipeline/tests/test_slicer.py:409`) — Verified by executing both functions on a baseline-only manifest: translation_hash and translation_hash_with_sources produce different hex strings because the envelope wraps the slice as `{"slice":...,"sources":[]}`. The test body only checks _HEX64_RE.match, so the docstring's "equals slice-only hash of an equivalent envelope" phrase is misleading to future maintainers.
- **test_single_algorithm docstring claims "differs from empty" but the assertion is not present** (`pipeline/tests/test_slicer.py:381`) — The test body only asserts _HEX64_RE.match(h); it does not compute a hash for an empty-algorithms manifest to compare against. Minor documentation overstatement — either drop the phrase or add the missing assertion.

PR #471, opened 2026-07-02
step-2 PR 2: alias resolver + schema plumbing + sim2real list translations

kalantar:
Closes #466.

## Summary

Step-2 PR 2: schema plumbing + alias resolver + `sim2real list translations`.

- **`pipeline/lib/translation_ref.py`** (new module). Shared validation regex (`^[A-Za-z0-9][A-Za-z0-9._-]*$`, max 128 chars, rejects `.` / `..` / empty; wider than step-1's `[a-z0-9-]+`), on-read shim (`read_translation_output`) that normalizes legacy step-1 top-level `image_ref`/`image_digest` into per-algorithm fields without rewriting on disk, translations iterator (`iter_translations` — logs+skips malformed files), `find_by_alias`, and `resolve_translation_ref` (validate → full-hash short-circuit → alias-exact → unique prefix).
- **`sim2real translation register`** writes the step-2 schema: `alias` at the top level (from `--algorithm`), and per-algo `image_ref` / `image_digest` / `config_path` / `source_path` / `source_sha256` inside `algorithms[0]`. Refuses on alias collision unless `--force`, which atomically clears the alias on the previous translation via tempfile + `os.replace`.
- **`sim2real assemble --translation`** now accepts an alias, a hash prefix (min 4 chars), or a full hash. Metavar `HASH` → `REF`. Errors with the design-verbatim `translation <ref> not built for algorithms: <names> — run 'sim2real build --translation <ref>' first` when any resolved algorithm has null `image_ref`. Per-algo image lookup inside the scenario-injection loop (replaces the top-level `image_ref` variable).
- **`sim2real list translations`** — new subcommand. Prints `ALIAS / HASH / SOURCE / IMAGES / CREATED` newest-first via `iter_translations`. IMAGES aggregates per-algo state (`N built`, `N pending`, `N/M built`, `N registered` for BYO).

## Schema migration

Legacy step-1 `translation_output.json` files (top-level `image_ref`, no `alias`) remain resolvable and buildable via the on-read shim — no forced migration, no `version` bump.

## Base

Base branch: `refactor/v2-step-2`. Depends on #465 (already merged into base as `2bf2054`).

## Reviewer attention

- **Resolver precedence: alias wins over prefix.** A 4-char alias `abcd` beats a hash prefix `abcd*` — design choice (design.md §Alias resolver step 4).
[...]

PR #473, opened 2026-07-02
step-2 PR 3: sim2real translate command (initial + --resume)

kalantar:
## Summary

Implements the skill-driven translation entrypoint (step-2 PR 3 of epic #463). `sim2real translate` writes `skill_input.json` + `translation_output.json` and exits at the `/sim2real-translate` checkpoint; `--resume` validates that the skill produced every declared algorithm's `<algo>_output.json`; `--force` blows the translation dir away and rewrites.

## Design

Reference: [`docs/epics/step-2/design.md#pr-3--sim2real-translate-command`](https://github.com/inference-sim/sim2real/blob/refactor/v2-step-2/docs/epics/step-2/design.md#pr-3--sim2real-translate-command). Implementation plan committed at `docs/superpowers/plans/2026-07-02-issue-467-sim2real-translate.md`.

## State machine

The 9-cell state × command table, verbatim from the design:

| State | `translate` (plain) | `translate --resume` | `translate --force` |
|---|---|---|---|
| Nothing (dir absent) | Create dir, write both checkpoint files, print hint, exit 0 | Error: `no translation to resume for hash <hash>`, exit 2 | Same as plain |
| Partial (dir exists, some `<algo>_output.json` missing) | Error: `translation <hash> incomplete`, exit 2 (never mutates) | Error: `missing outputs for: <names>`, exit 2 (never mutates) | Delete + recreate as if nothing |
| Complete (all `<algo>_output.json` present) | Print `already complete`, exit 0 (idempotent) | Same as plain-Complete | Delete + recreate; user re-runs the skill |

## Notes for reviewers

- **`translation_ref.validate_name` runs before every disk write.** Scenario doubles as the translation's alias per design; algorithm names use the same regex (both are stricter than manifest's `_PACKAGE_NAME_RE` — which enforces the manifest's own lowercase-alphanumeric-only rule at load time).
- **Manifest → skill_input bridge.** Manifest carries `context.files`; skill_input uses `context.file_paths`. `_cmd_translate` remaps at the boundary rather than changing either schema.
- **Atomic writes.** Both `translation_output.json` and `skill_input.json` go through `_atomic_write_json` (already present from PR 2). Interruption mid-write cannot leave a half-file visible.
- **What was NOT touched.** `pipeline/lib/slicer.py` (`translation_hash_with_sources` was shipped in PR 1 / #470). `pipeline/lib/translation_ref.py` (`validate_name` was shipped in PR 2 / #471). `pipeline/lib/manifest.py`, `pipeline/lib/layout.py` — untouched.
- **Sweep.** `git grep 'prepare\.py' -- '*.md'` shows references only in historical design docs / step-1 wrap-up notes — PR 6 sweeps those. `sim2real translate` in docs is all forward-looking (design docs), no updates needed for this PR.
- **CI.** Added `pipeline/tests/test_translate.py` to the explicit test-file list in `.github/workflows/test.yml`, keeping parity with the existing pattern. PR 6 does the broader audit.

## Test plan

- [x] `26 passed` in `pipeline/tests/test_translate.py` — covers all 9 state × command cells plus builder unit tests, state-detection tests, alias-validation-rejects-`../evil` test, and a divergent-recorded-algorithms test.
- [x] `1271 passed, 2 xfailed` in the full `pipeline/` suite (baseline was 1245 + 2 xfailed; +26 matches the new tests).
- [x] Manual: `sim2real translate --help` renders both `--force` and `--resume` in the mutually exclusive group.

Closes #467.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,7 @@ jobs:
             pipeline/tests/test_slicer.py \
             pipeline/tests/test_sim2real.py \
             pipeline/tests/test_assemble_run.py \
+            pipeline/tests/test_translate.py \
             .claude/skills/sim2real-analyze/tests/ \
             .claude/skills/sim2real-bootstrap/tests/ \
             .claude/skills/sim2real-translate/tests/ \
--- a/docs/superpowers/plans/2026-07-02-issue-467-sim2real-translate.md
+++ b/docs/superpowers/plans/2026-07-02-issue-467-sim2real-translate.md
@@ -1,0 +1,911 @@
+# sim2real translate command (initial + --resume) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add the `sim2real translate` subcommand to `pipeline/sim2real.py`. Two-phase, skill-checkpointed: an initial run writes `skill_input.json` + `translation_output.json` (with `image_ref: null` per algorithm) and exits at the checkpoint; `--resume` validates that `/sim2real-translate` produced the expected `generated/<algo>/<algo>_output.json` files. `--force` deletes and recreates the translation directory. Behavior on every one of the 9 state × command cells matches the state-machine table in `docs/epics/step-2/design.md#pr-3--sim2real-translate-command`.
+
+**Architecture:** Introduce a private `_translate_*` set of helpers in `pipeline/sim2real.py` (mirroring the existing `_register_translation` / `_atomic_write_json` / `_build_translation_output` pattern), and one new command handler `_cmd_translate(args)`. Reuse existing infrastructure verbatim: `slicer.translation_hash_with_sources` (PR 1), `translation_ref.validate_name` (PR 2), `translation_ref.read_translation_output` (PR 2), `layout.translation_dir` / `layout.translation_output_path` (step-0), `_atomic_write_json` (already in `sim2real.py`). The state machine is expressed as a single explicit branch on `(exists, is_complete, is_partial)` × `(plain, resume, force)` — one function, 9 cells, one exit code per cell.
+
+**Tech Stack:** Python 3.10+, pytest, PyYAML (already a dep for reading `transfer.yaml`). No new dependencies.
+
+## Global Constraints
+
+- Base branch: `refactor/v2-step-2` — worktree is `.claude/worktrees/issue-467-sim2real-translate/` under the parent repo. Every path passed to Read/Edit/Write/Bash must contain that substring.
+- **Design is authoritative.** The 9-cell state-machine table and the `skill_input.json` schema are pinned in `docs/epics/step-2/design.md`. When in doubt, re-read the design section, not this plan.
+- **Alias source.** `alias = manifest["scenario"]` per design §Commands / initial-run step 6. Must be validated via `translation_ref.validate_name` before write.
+- **Algorithm-name validation.** Every `algorithms[i].name` from `transfer.yaml` is validated via `translation_ref.validate_name` before landing in `translation_output.json` or `skill_input.json` paths.
+- **No auto-execute of the skill.** `translate` writes the checkpoint files and exits with a hint message; the operator invokes `/sim2real-translate` explicitly. Auto-fix is step-6's territory (not this PR).
+- **Completeness check** for both `--resume` and idempotent "already complete" detection is per-`algorithms[i].name`: every declared algorithm must have `generated/<algo>/<algo>_output.json` present. Extra files under `generated/` are ignored (may be from a prior algorithm-set) but logged as a warning.
+- **Atomic writes.** All JSON writes (`translation_output.json`, `skill_input.json`) go through `_atomic_write_json`. Prevents readers from observing a half-written file if the operator interrupts mid-write.
+- **No side effects on error.** Every error path exits before touching disk. Errors detected after partial writes must roll back (unlink what was written) — applies only to `--force` recreate.
+- Test invocation: `.venv/bin/python -m pytest pipeline/ -v` from the worktree root.
+- CI lint: `ruff check pipeline/ .claude/skills/ --select F`.
+
+---
+
+## File Structure
+
+| File | Change | Reason |
+|------|--------|--------|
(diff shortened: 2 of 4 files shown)


kalantar:

The skill contract file is wired to the wrong directory and the alias-uniqueness invariant the design promises isn't enforced in the translate path.

### Must fix
- **skill_input.json's translations_dir points at workspace/translations/ instead of workspace/translations/<hash>/** (`pipeline/sim2real.py:762`) — pipeline/sim2real.py:762 passes `layout.translations_dir()` (parent) to `_build_skill_input`. `layout.py` defines `translations_dir()` as `workspace/translations/` and `translation_dir(thash)` as `workspace/translations/<hash>/`. The design (docs/epics/step-2/design.md, schema for skill_input.json) pins the value as `/absolute/path/to/workspace/translations/abc123def456…` and states `algorithms[i].output_dir`, `config_output_path`, and `baseline.generated_overlay_path` are relative to `translations_dir` — with the parent dir written, the downstream skill (PR 4) will resolve output paths under `workspace/translations/generated/<algo>/` instead of the hash-specific directory. Fix by passing `layout.translation_dir(thash)` and add an assertion in `test_plain_writes_checkpoint_files` that `skin['translations_dir']` equals the hash-specific path.

### Important
- **translate does not refuse when the scenario alias is already assigned to a different translation hash** (`pipeline/sim2real.py:699`) — docs/epics/step-2/design.md explicitly states "`translate` and `register` both refuse if another translation has the same `alias` value pointing at a different hash, unless `--force` is passed." `_cmd_translation_register` (sim2real.py:350) calls `translation_ref.find_by_alias` and refuses on mismatch, but `_cmd_translate` / `_translate_write_checkpoint` (sim2real.py:670-696, 699-774) has no equivalent check — two experiments sharing a scenario name will silently coexist with the same alias attached to different hashes, violating the alias-uniqueness invariant the resolver depends on. Not tested.

### Recommendations
- **_cmd_translate does not wrap _translate_write_checkpoint in OSError handling; register's writer path does** (`pipeline/sim2real.py:690`) — sim2real.py:670-696 calls `_translate_write_checkpoint` in both the force branch and the state=nothing branch with no OSError guard, while `_cmd_translation_register` (line 554) wraps its equivalent write in `except (RuntimeError, ValueError, OSError)`. An OSError during mkdir/read_bytes/os.replace inside the checkpoint writer surfaces as a raw Python traceback rather than the `error: ...; return 2` shape used elsewhere. Rare in practice but the asymmetry with register is worth aligning.
- **test_plain_writes_checkpoint_files does not assert the context.files → context.file_paths bridge** (`pipeline/tests/test_translate.py:279`) — The bridge at sim2real.py:751-758 renames manifest `context.files` to skill_input `context.file_paths`. No integration test verifies the written JSON has `file_paths` (not `files`) when the manifest carries `context.files`. A regression that emits the wrong key would silently pass the current tests and the downstream skill would see an empty list. Worth adding a one-line assertion.

kalantar:
Addressed both blocking review items.

The `skill_input.json` bug was straightforward — the field was pointing at the parent `workspace/translations/` instead of the hash-specific directory the design pins. Fixed the call site in `_translate_write_checkpoint` and added an assertion in `test_plain_writes_checkpoint_files` that pins `skin['translations_dir']` to the resolved hash-scoped path.

For the alias-uniqueness gap, added the `translation_ref.find_by_alias` check in `_cmd_translate` — plain `translate` refuses with `alias '<scenario>' already assigned to translation <hash>; pass --force to reassign` and exits 2; `--force` calls `_clear_alias_on` on the previous hash and then rewrites, matching how `_cmd_translation_register` handles the same invariant. `--resume` skips the check because it never mutates the dir. Three new tests in `TestTranslateAliasCollision` cover the refuse / reassign / resume-ignores paths.

Not addressing this round:
- The two `recommendations` items (OSError wrapping around `_translate_write_checkpoint`, and a `context.files → context.file_paths` bridge assertion). Non-blocking per the review workflow's severity split; can be follow-ups or fold in on a later pass.

Full suite: 1274 passed, 2 xfailed. Ruff clean.

kalantar:

The alias-collision guard fires ahead of the complete-state idempotency short-circuit, so plain `translate` on a complete translation can exit 2 in the presence of a foreign alias — deviating from the acceptance criterion "already-complete → exit 0" for a narrow but reachable edge case.

### Important
- **Complete/plain `translate` is not fully idempotent — a foreign alias on a different hash causes exit 2 instead of the design-mandated "already complete" success** (`pipeline/sim2real.py:658`) — Verified at pipeline/sim2real.py:658-720: the `if not args.resume` alias-collision block returns exit 2 before the state dispatch at lines 704-713 can short-circuit on the complete state. I reproduced this with a scripted test — plain `translate` on a complete translation errored with `error: alias 'softreflective-v1' already assigned to translation bbbb...; pass --force to reassign` instead of the idempotent path. Issue #467's acceptance list requires "State machine matches design table exactly" and "idempotency (already-complete → exit 0)"; the design table's Complete/plain cell says idempotent exit 0 with no mutation. The trigger requires an existing alias-uniqueness violation across translations, which is narrow but reachable via legacy state, cross-command interactions (e.g. `translation register` using an algorithm name that collides with a later scenario), or nondeterministic `iterdir` ordering when the invariant was previously broken. Fix is a one-line guard (`and state != "complete"`) or moving the alias check into `_translate_write_checkpoint`.
- **TestTranslateAliasCollision has no case for Complete/plain + stale-alias-on-different-hash, which is precisely the cell that violates the design contract** (`pipeline/tests/test_translate.py:425`) — Verified at pipeline/tests/test_translate.py:425-489: the three tests in TestTranslateAliasCollision cover Nothing/plain, Nothing/--force, and Nothing/--resume paths. None sets up the current translation in a `complete` state before planting the foreign alias, so the deviation above went unnoticed. A test in the shape of `test_complete_plain_not_blocked_by_stale_alias` should accompany the fix.

### Recommendations
- **Module docstring is stale — still says the file was created by PR 1 with only `translation register` and doesn't mention `translate`** (`pipeline/sim2real.py:2`) — Verified at pipeline/sim2real.py:2-7: the docstring reads "Subcommands land incrementally across the step-1 epic. This file is created by PR 1 with only `translation register`. Subsequent PRs add `assemble`, `use`, `list runs`." This PR adds `translate` (step-2), and the file already carries `assemble`/`use`/`list`. Not a bug, but the top-of-file orientation for future maintainers is now misleading.
- **`_clear_alias_on`'s exception handler catches only OSError but the callee (`read_translation_output`) can also raise ValueError on malformed JSON** (`pipeline/sim2real.py:670`) — Verified at pipeline/sim2real.py:670-674 and pipeline/lib/translation_ref.py:75-104: `read_translation_output` calls `json.loads` (raises `json.JSONDecodeError`, a `ValueError` subclass) and also raises a bare `ValueError` at line 91 for non-dict payloads. Broadening the catch to `(OSError, ValueError)` keeps the clean-error contract if the foreign translation's file is corrupt. Narrow but a real gap in the error-message discipline.

PR #475, opened 2026-07-02
step-2 PR 5: sim2real build + shared build library + docs

kalantar:
## Summary

Closes #469 — step-2 PR 5 of epic #463.

Ships the `sim2real build` command for the skill-driven flow, extracts three genuinely-shared build primitives into `pipeline/lib/build.py`, refactors `deploy.py:_cmd_build` to route through the new library (no behavior change), extends `sim2real assemble` with a fail-fast check for unbuilt translations, and updates docs.

Base: `refactor/v2-step-2`.

## What changed

**New module `pipeline/lib/build.py`** — five primitives, all failure-tolerant in the "fail-safe → rebuild" direction:
- `compose_image_ref(registry, repo, tag) -> str` with non-empty validation
- `check_skopeo() -> None` — raises `BuildError` with install hint if missing
- `probe_image_digest(image_ref) -> str | None` — skopeo inspect wrapper, returns `None` on any failure (timeout, auth, missing tag, invalid JSON, missing binary)
- `dispatch_buildkit_build(image_ref, build_id, namespace, source_dir, run_dir, repo_root) -> int` — wraps `pipeline/scripts/build-epp.sh`; returns exit code (never raises on non-zero)
- `atomic_write_json(path, data)` — tempfile + `os.replace`; creates parent dirs; cleans up tempfile on failure

**Deploy refactor** — `deploy.py:_cmd_build` now calls `build.dispatch_buildkit_build` and `build.atomic_write_json` in place of inline subprocess and `write_text(json.dumps(...))`. `pipeline/tests/test_deploy_build.py` passes unchanged (parity guarantee from acceptance criterion 1).

**New `sim2real build` subcommand** —
```
sim2real build --translation ALIAS|PREFIX|HASH [--force-rebuild] [--skip-build]
```
Prereqs (fail early): skopeo on PATH · translation ref resolves · every algorithm has `generated/<algo>/<algo>_output.json` · `setup_config.json:{registry,repo_name}` non-empty · single provisioned cluster.

Per algorithm:
1. Compose `<registry>/<repo>:<translation_hash[:12]>-<algo>`.
2. Idempotent skip if image_ref+digest already recorded (unless `--force-rebuild`).
3. Pre-build `skopeo inspect` → probe hit records digest and skips build.
4. Otherwise dispatch buildkit pod via `build-epp.sh`. First non-zero exit aborts the loop.
5. Post-build `skopeo inspect` records the digest; failure → `image_digest: null` + warning.
6. Atomic writeback per algorithm — a mid-run failure preserves prior algorithms' state.

**`sim2real assemble` fail-early check** — every algorithm declared in `transfer.yaml` that is also recorded in `translation_output.json` must have a non-null `image_ref`. Missing → exit 2 with `translation <ref> not built for algorithms: <names> — run 'sim2real build --translation <ref>' first`.

**Docs** — `pipeline/README.md` gains "Translate (skill-driven)" and "Build translation images" sections plus an end-to-end skill-driven demo. `CLAUDE.md`'s pipeline library and workspace-artifact tables now cover `build.py` and the new writers of `translation_output.json`. `.github/workflows/test.yml` explicitly lists `test_build.py`.

## Sourcing note

The design doc references `transfer.yaml:epp_image.build.{hub,name}` for image-ref construction, but that field does not exist in the current v3 manifest — `pipeline/lib/manifest.py` validates `component.build.image.{hub,name}` and, in practice, real transfer.yaml files (e.g. `sr/transfer.yaml`) omit it entirely and rely on `setup_config.json:{registry,repo_name}` (which is what `deploy.py:_cmd_build` uses today). This PR sources both from `setup_config.json` for consistency with the existing deploy path. If we add `epp_image.build` to the manifest schema later, both paths can migrate together.

## Test plan

- [x] All 40 tests in `pipeline/tests/test_build.py` pass locally (`.venv/bin/python -m pytest pipeline/tests/test_build.py -v`).
- [x] All 14 tests in `pipeline/tests/test_deploy_build.py` pass unchanged after the `_cmd_build` refactor.
- [x] All 2 new `TestAssembleIncompleteTranslationCheck` tests pass; existing `TestAssembleResolvesAlias::test_assemble_accepts_alias` fixture updated to write a valid v3 stub (the assemble path now loads the manifest to run the new pre-check).
- [x] Full pipeline test suite: `1272 passed, 2 xfailed` (`.venv/bin/python -m pytest pipeline/ --ignore=pipeline/tests/test_deploy_remote.py`).
- [x] Lint clean: `ruff check pipeline/ .claude/skills/ --select F`.
- [ ] Real-cluster smoke: `sim2real translate → /sim2real-translate → sim2real translate --resume → sim2real build → sim2real assemble → deploy.py run → deploy.py collect` — deferred to the epic's demo gate. Skill-driven translate + build depend on the `/sim2real-translate` skill being re-enabled (issue #468).

## Sweep

Searched for stale refs to `prepare.py`, `_cmd_build`, `epp_image.build`, `sim2real build`. Remaining `prepare.py` references live in:
- `.claude/skills/sim2real-translate/prompts/` — handled by PR 4 (#468) which re-enables that skill.
- `docs/epics/step-{0,1,2}/design.md` and `docs/superpowers/plans/` — historical design/plan docs; leave alone.

Stale `--skip-build-epp` in `pipeline/completions.{bash,zsh}` and one instance in `pipeline/README.md` (line 807) predate this PR (the deploy flag was renamed earlier). PR 6 (#472) is the dead-code sweep, so leaving them here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,7 @@ jobs:
             pipeline/tests/test_sim2real.py \
             pipeline/tests/test_assemble_run.py \
             pipeline/tests/test_translate.py \
+            pipeline/tests/test_build.py \
             .claude/skills/sim2real-analyze/tests/ \
             .claude/skills/sim2real-bootstrap/tests/ \
             .claude/skills/sim2real-translate/tests/ \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,7 +161,10 @@ The `pipeline/` scripts form a linear dependency chain, with a one-time cluster 
 ```
 cluster.py provision  (one-time per cluster)
                    ↓
-setup.py → sim2real translation register → sim2real assemble → deploy.py
+setup.py → [BYO: sim2real translation register] OR
+           [Skill: sim2real translate → /sim2real-translate → sim2real translate --resume → sim2real build]
+                   ↓
+sim2real assemble → deploy.py
 ```
 
 Each stage communicates with the next through files written to `workspace/`. When modifying any stage, you **must** trace both directions:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,12 +21,18 @@ inference-sim to production llm-d-inference-scheduler scorer plugins.
 
 ## Transfer Pipeline
 
-The pipeline has two phases: a one-time-per-cluster bootstrap, then a per-workspace and per-run cycle. Step-1 of the `refactor/v2-step-1` epic ships the **BYO MVP** — customers register a pre-built EPP image with `sim2real translation register`, then assemble runs from it:
+The pipeline has two phases: a one-time-per-cluster bootstrap, then a per-workspace and per-run cycle. Two producers write translations with the same on-disk shape:
+
+- **BYO** (`sim2real translation register`) — the operator supplies a pre-built EPP image.
+- **Skill-driven** (`sim2real translate` → `/sim2real-translate` skill → `sim2real translate --resume` → `sim2real build`) — the operator supplies algorithm source; the skill translates it into a plugin and `sim2real build` compiles+pushes an image.
 
 ```
 cluster.py provision  (one-time per cluster)
                    ↓
-setup.py → sim2real translation register → sim2real assemble → deploy.py
+setup.py → [BYO: sim2real translation register] OR
+           [Skill: sim2real translate → /sim2real-translate → translate --resume → sim2real build]
+                   ↓
+sim2real assemble → deploy.py
 ```
 
 Run all pipeline commands from the `sim2real/` directory, pointing `--experiment-root` at the experiment repo:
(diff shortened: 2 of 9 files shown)


kalantar:

Ships cleanly with one duplication the plan called out but did not complete, plus two minor documentation ordering inconsistencies worth resolving before merge.

### Important
- **sim2real.py still ships a private _atomic_write_json alongside the new public build.atomic_write_json, keeping two identical implementations** (`pipeline/sim2real.py:62`) — sim2real.py:62-84 defines _atomic_write_json byte-for-byte equivalent to build.atomic_write_json at lib/build.py:114-133. The new _cmd_build correctly routes through build.atomic_write_json (lines 986, 1016), but the pre-existing call sites (line 100, 374, 788, 816) still call the private copy. The PR's own plan (docs/superpowers/plans/2026-07-02-issue-469-sim2real-build.md line 112) explicitly listed hoisting this helper as part of the sim2real.py work, and the project 'minimize duplication' memory rule applies directly. Not functionally broken, but the extraction was left half-done.

### Recommendations
- **CLAUDE.md 'Contributing: Pipeline Stage Contracts' diagram still shows BYO-only, contradicting the Transfer Pipeline section this PR updated to show both flows** (`CLAUDE.md:164`) — CLAUDE.md line 164 reads 'setup.py -> sim2real translation register -> sim2real assemble -> deploy.py' while lines 32-35 (updated by this PR) show both BYO and skill-driven arms. A contributor reading the contract section will not see that sim2real translate, /sim2real-translate, and sim2real build are also valid upstream stages. Small edit, but the inconsistency was introduced by this PR since the diff updated one diagram and not the other.
- **pipeline/README.md documents 'Assemble a run' before 'Build translation images', reversing the actual dependency order** (`pipeline/README.md:254`) — README.md places Assemble at line 194 and Build at line 254, but the workflow is translate -> build -> assemble (build must fill image_ref before assemble runs; see the fail-early check documented at line 240). The end-to-end demo section correctly orders these; only the standalone reference sections are inverted. Non-blocking because the assemble section explicitly documents the fail-early behavior and points to sim2real build

kalantar:

One real contract gap worth fixing before merge: two atomic-write call sites in the per-algorithm build loop don't honor the CLI's "error: <msg>; return 2" contract on filesystem faults, and the post-build case can leave the operator seeing a traceback when the image has actually been pushed.

### Important
- **Unguarded atomic_write_json calls in _cmd_build's per-algorithm loop bypass the CLI error contract** (`pipeline/sim2real.py:962`) — pipeline/sim2real.py lines 962 and 992 call build.atomic_write_json without try/except. build.py:114-133 declares the function 'Raises OSError on filesystem faults; callers surface the error and exit.' Other file I/O in the same function (lines 882-885) does wrap in try/except and returns 2. On a disk-full or permission error at line 992 the image has already been pushed successfully but the operator sees an uncaught Python traceback rather than a clean 'error: failed to record digest ... re-run sim2real build' message.

### Recommendations
- **sim2real.py module docstring is stale after this PR** (`pipeline/sim2real.py:4`) — pipeline/sim2real.py lines 4-6 still describe the file as 'created by PR 1 with only translation register. Subsequent PRs add assemble, use, list runs.' The file now hosts translate, translate --resume, and build in addition to those. A one-line refresh keeps the docstring aligned with the shipped subcommand set.
- **test_force_rebuild_ignores_probe_hit doesn't assert on the written image_ref/image_digest** (`pipeline/tests/test_build.py:457`) — pipeline/tests/test_build.py:457 verifies dispatch was called but doesn't read translation_output.json back. The parallel probe-miss test at line 432 does this. Adding the same _read_translation_output assertion would catch a regression that skipped the post-build atomic write in the force-rebuild path.
- **No test covers the 'no cluster provisioned' or 'multiple clusters' fail-early exits in sim2real build** (`pipeline/tests/test_build.py:306`) — sim2real.py:898-912 emit distinct error messages and return 2 for these paths, and they are hittable in the normal 'ran build before cluster.py provision' workflow. TestSim2realBuildPrereqs covers every other prereq exit but not these two. Two small workspace-fixture tests would close the gap.

kalantar:

Docs need three factual fixes before merge — invocations that don't exist, an incorrect completeness-check source, and a stale scope qualifier — plus the top-of-module claim in build.py that oversells the failure-tolerance of primitives that actually raise.

### Must fix
- **README Common patterns lists three deploy.py invocations that would all fail immediately** (`pipeline/README.md:807`) — Verified against deploy.py argparse (lines 3355-3436): --skip-build-epp does not exist (correct flag is --skip-build under `run` or `build` subcommands, per line 3380/3386); deploy.py --dry-run only exists on `reset` and `wipe` subparsers (lines 3424, 3431), not top-level; deploy.py --package baseline treatment is only valid on `collect`/`status`/`run`/`reset`/`wipe` subparsers, not top-level. An operator copy-pasting these gets 'unrecognized arguments'.
- **README 'Build translation images' prerequisite says transfer.yaml but code reads translation_output.json** (`pipeline/README.md:215`) — Verified in sim2real.py:_cmd_build (lines 852-873): algorithms come from tout.get('algorithms') where tout is read from translation_output.json; transfer.yaml is never opened in _cmd_build. The README prerequisite 'every algorithm declared in transfer.yaml must have translations/<hash>/generated/<algo>/<algo>_output.json on disk' misdirects an operator to the wrong file when debugging a completeness failure.

### Important
- **README section header claims 'only what ships in step-1 PR 1' but the section now documents step-2 subcommands (translate, build)** (`pipeline/README.md:117`) — Verified: the sim2real.py section (line 115 onward) now contains 'Translate (skill-driven, step-2)' at line 163 and 'Build translation images' at line 194, both step-2. The lead-in qualifier 'this section describes only what ships in step-1 PR 1' is factually false and will make a reader assume the docs are intentionally incomplete when they aren't.
- **build.py module docstring's 'Every primitive is failure-tolerant' claim contradicts compose_image_ref, check_skopeo, and dispatch's own hard raises** (`pipeline/lib/build.py:8`) — Verified: compose_image_ref (line 27-35) raises BuildError on empty inputs, check_skopeo (line 38-45) raises BuildError when skopeo is absent, and dispatch_buildkit_build (line 94-98) raises BuildError when the script is missing. The module-level docstring says 'Every primitive is failure-tolerant in the fail-safe → rebuild direction' then enumerates only three (probe, dispatch, atomic_write). A consumer scanning the docstring may skip try/except around the raise-hard primitives — the enumeration should either name the raising primitives explicitly or drop the 'Every'.

### Recommendations
- **CLAUDE.md 'CI' local-run pytest command omits test_build.py and test_translate.py that CI runs** (`CLAUDE.md:136`) — Verified: .github/workflows/test.yml lines 38-39 include both test files; CLAUDE.md's CI code block enumerates test_layout, test_cluster_ops, test_cluster_py, test_slicer, test_sim2real, test_assemble_run only. The initial `pipeline/` argument means pytest still discovers everything, so this is presentation-only drift, but a developer eyeballing the explicit list will not notice the two new test files.
- **sim2real._cmd_build does not check source_dir existence before dispatching buildkit; deploy.py's parallel path does** (`pipeline/sim2real.py:932`) — Verified: line 932 constructs source_dir = exp_root / repo_name and passes it straight to dispatch_buildkit_build (line 983) with no existence check. deploy.py:300-302 has an explicit check and clean err() for the same failure mode. When repo_name is misconfigured or the submodule is uninitialized, the build script will fail non-zero with the current 'error: build failed for <algo>' message and no indication that the source directory is missing.
- **_build_skill_input docstring says paths are relative to 'translations_dir' but caller passes the per-hash translation_dir** (`pipeline/sim2real.py:192`) — Verified: parameter is named translations_dir (plural) at line 183; the only caller at line 788 passes layout.translation_dir(thash) — a per-hash directory, not the parent translations_dir. The docstring claim 'per-algorithm output_dir and config_output_path are relative to translations_dir' would lead a maintainer to expand 'generated/softref' relative to workspace/translations/, when it is actually relative to workspace/translations/<hash>/.

kalantar:

Diff is clean; the surviving specialist findings are minor doc imprecisions or pre-existing patterns outside this PR's scope.

### Recommendations
- **CLAUDE.md forward-refs /sim2real-translate as "re-enabled in step-2" while SKILL.md is still the step-1 disabled stub** (`CLAUDE.md:67`) — Verified: .claude/skills/sim2real-translate/SKILL.md still has user-invocable: false and the step-1 disabled text; CLAUDE.md line 67 reads "DISABLED as of step-1; re-enabled in step-2 (issue #468)." An operator following CLAUDE.md today would try to invoke the skill and hit the disabled message. Phrasing like "re-enablement tracked in issue #468, landing in a subsequent PR of step-2" would remove the ambiguity.
- **CLAUDE.md --force-rebuild description only mentions the probe, not the idempotency short-circuit** (`CLAUDE.md:63`) — Verified sim2real.py:949 also checks `not args.force_rebuild` in the already-recorded short-circuit; CLAUDE.md line 63 says only "--force-rebuild skips the pre-build probe." The README documents this correctly; CLAUDE.md drift is minor but worth alignin

PR #474, opened 2026-07-02
step-2 PR 4: /sim2real-translate skill re-enable + prompt update

kalantar:
Closes #468. Part of epic #463.

## Summary

Restores the `/sim2real-translate` skill (DISABLED since step-1's BYO cutover) against the step-2 workspace layout defined in `docs/epics/step-2/design.md`. The skill now reads `workspace/translations/<hash>/skill_input.json` (written by `sim2real translate` in #473 / PR 3) and writes per-algorithm outputs to `workspace/translations/<hash>/generated/<algo>/{cmd,pkg,<algo>_config.yaml,<algo>_output.json}`, exactly as `sim2real translate --resume` probes for.

## What changed

- **`SKILL.md`**: full rewrite. Drops DISABLED preamble; `user-invocable: true`; computes translation hash via `pipeline.lib.slicer.translation_hash_with_sources` (same call site `sim2real translate` uses — no drift risk); loads `skill_input.json`; loops over `algorithms[]`, skipping any whose `<algo>_output.json` already exists (idempotent per-algo). Directs the operator to `sim2real translate --resume` next.
- **`prompts/agent-writer.md`, `prompts/agent-reviewer.md`**: path placeholders repointed from `{RUN_DIR}/generated/{ALGO_NAME}/...` to `{OUTPUT_DIR}/...` and `{BASELINE_OVERLAY_PATH}`. Snapshot directory moves to `{TRANSLATIONS_DIR}/snapshots/{ALGO_NAME}/vN` so multi-algo runs don't collide. All three `prepare.py` references (writer:82, reviewer:132, reviewer:144 — the PR #455 sweep followup) are rewritten to describe `sim2real assemble` semantics.
- **`prompts/agent-expert.md`**: the "read files in context.files" instruction now points at the inline `{CONTEXT_FILE_PATHS}` placeholder (the new `skill_input.json` inlines the operator's context set from `transfer.yaml:context.files`, rather than having the skill re-parse the manifest).
- **`scripts/copy_generated.py` + `tests/test_copy_generated.py`**: rename the `run_dir` parameter to `translation_dir`. Behavior unchanged — the parameter refers to `workspace/translations/<hash>/`, so the pre-refactor `run_dir` name was stale.
- **`CLAUDE.md`**: update the `/sim2real-translate` blurb from "DISABLED for step-1" to the new re-enabled description.

`.github/workflows/test.yml` already lists `.claude/skills/sim2real-translate/tests/` (line 41) — no CI wiring update needed.

[...]

PR #478, opened 2026-07-03
step-2: -gaie override + baselines[] plural schema + registry-creds Secret plumbing + setup.py runs/ ownership (#477, #479, #480, #481)

kalantar:
Closes #477, #479, #480, and #481. Bundling four step-2 fixes on the same branch — see the individual commits for each fix's diff. All four are conceptually independent but share the step-2 lifecycle, so landing them together simplifies the review of the interlocking test surface.

## Commit 1 — remove hardcoded -gaie override (#477)

The `sim2real-translate` writer prompt had two contradictory rules for the InferenceObjective section: (a) `spec.poolRef.name` MUST equal `${model.idLabel}-gaie` and (b) everything else in `poolRef` MUST be copied verbatim from `config.md`. The writer followed (a), so projects like `sr` (whose `config.md` declared `-router`) got an overlay pointing at a pool that would not bind. Fix: unify under "copy verbatim from `config.md`", preserving the `${model.idLabel}` templating dependency as a documentary note.

## Commit 2 — derive baseline overlay(s) from baselines[] via defaults (#479)

`sim2real translate` read a singular `baseline:` key from `transfer.yaml`, but the pinned v3 schema uses `baselines:` (plural list) with `algorithms[*].defaults` cross-references. Consequence: `skill_input.baseline` was always null, Phase 2 was always skipped, and no baseline overlay ever landed on disk. Fix: `_cmd_translate` walks `manifest[\"baselines\"]` filtered by algorithm `defaults` and emits `skill_input.baselines[]` with per-entry `generated_overlay_path` = `generated/baseline_<name>/baseline_config.yaml`. Each algorithm entry gets a resolved `baseline_overlay_path`. Assemble reads the per-baseline path, with a legacy fallback to `generated/baseline_config.yaml` so BYO translations remain resolvable. The pinned `skill_input.json` schema in `docs/epics/step-2/design.md` is updated. Tests cover unit-level and end-to-end shapes.

## Commit 3 — thread registry-creds Secret name from cluster_config to build-epp.sh (#480)

`cluster.py provision` writes the registry-credentials k8s Secret with name `registry-creds`, but `pipeline/scripts/build-epp.sh` hardcoded `registry-secret`. On clusters where both Secrets coexisted from the pre-refactor era, credential rotation via `cluster.py provision --registry-token <new>` looked successful while buildkit kept pushing with the stale token. Fix: `dispatch_buildkit_build` gains a required `registry_secret_name` kwarg; `build-epp.sh` gains a required `--registry-secret-name NAME` flag. Both `_cmd_build` implementations (`pipeline/sim2real.py` step-2, `pipeline/deploy.py` step-1) read `secret_names['registry_creds']` from `cluster_config` and forward it, failing fast when absent.

## Commit 4 — setup.py stops creating workspace/runs/ (#481)

`setup.py` created the run directory and wrote `run_metadata.json`, but `sim2real assemble` (introduced in step-1) also materializes that directory and refuses when it already exists. The documented `setup.py → sim2real assemble --run <name>` flow failed on a fresh workspace with `run directory already exists — pass --force to overwrite`. Fix: setup.py stops touching `workspace/runs/` and stops writing `current_run` / `component_image`. `sim2real assemble` becomes the sole writer of the run directory and `run_metadata.json`; `sim2real use` becomes the sole writer of `current_run`. setup.py's read-modify-write preserves any pre-existing `current_run`, so the setup → use → setup re-run cycle is stable. `pipeline/README.md` and `CLAUDE.md` are updated (split the setup_config.json writer table row into workspace fields vs. current_run pointer). `test_setup_pipeline.py` is rewritten — the `TestRunMetadataIdempotent` class is removed (vacuous now that setup.py doesn't write run_metadata) and a new `TestSetupDoesNotTouchRuns` regression class is added.

## Base correction (#477)

Issue #477 was filed with `Base: refactor/v2-step-1` but the bug is live on `refactor/v2-step-2` (the skill was DISABLED on step-1 by PR #474; step-1 epic #443 is closed). Retargeted to `refactor/v2-step-2` per user confirmation before implementation.

## Verification

- `python -m pytest pipeline/ .claude/skills/ -v` → 1415 passed, 2 xfailed
- `ruff check pipeline/ .claude/skills/ --select F` clean
- Sweep: no hardcoded `registry-secret` string, no singular-baseline references, no `-gaie` references, no setup.py writes to `workspace/runs/`, all in runtime code and user-facing docs. Archival plan/design docs under `docs/superpowers/plans/` and `docs/epics/step-0/design.md` retain the pre-refactor descriptions as history.

## Test plan

- [x] Full pytest and ruff pass locally
- [x] No leaks into the parent repo (all changes contained to the worktree)
- [x] BYO translation compat preserved via the legacy `generated/baseline_config.yaml` fallback in `assemble_run.py` (covered by dedicated tests)
- [ ] Real-cluster smoke for #480: rotate the PAT via `cluster.py provision --registry-token <new>`, then `sim2real build --translation <alias>` pushes on the new token — needs a live cluster and rotatable PAT; not automatable from the harness.
- [ ] Real-cluster smoke for #481: on a fresh workspace, `setup.py` → `sim2real assemble --run trial-1` completes without any `--force` flag — not automatable from the harness, but the two regression tests in `TestSetupDoesNotTouchRuns` capture the equivalent local-filesystem contract.

### DIFF
--- a/.claude/skills/sim2real-translate/SKILL.md
+++ b/.claude/skills/sim2real-translate/SKILL.md
@@ -153,7 +153,7 @@ test -f "$TRANSLATIONS_DIR/skill_input.json" || {
 Load per-algorithm inputs from `skill_input.json`. Every path in the file is
 interpreted relative to `experiment_root` (for `source_path`) or
 `translations_dir` (for `output_dir`, `config_output_path`,
-`baseline.generated_overlay_path`):
+`algorithms[i].baseline_overlay_path`):
 
 Path-safe emission: each `ALGO` row is TAB-delimited; each context file path is
 written to a separate line in a dedicated file (`/tmp/sim2real_translate_ctx_paths.txt`)
@@ -172,11 +172,6 @@ required = ("version", "translation_hash", "experiment_root", "translations_dir"
 missing = [k for k in required if k not in si]
 if missing:
     raise SystemExit(f"HALT: skill_input.json missing keys: {missing}")
-baseline = si.get("baseline")
-if baseline is not None:
-    print("BASELINE_OVERLAY_REL=" + baseline["generated_overlay_path"])
-else:
-    print("BASELINE_OVERLAY_REL=")
 ctx = si.get("context") or {}
 print("CONTEXT_TEXT_B64=" + base64.b64encode(
     (ctx.get("text") or "").encode()).decode())
@@ -185,22 +180,21 @@ ctx_paths_out = Path("/tmp/sim2real_translate_ctx_paths.txt")
 paths = [str(exp / p) for p in (ctx.get("file_paths") or [])]
 ctx_paths_out.write_text("".join(p + "\n" for p in paths))
 for algo in si["algorithms"]:
+    # baseline_overlay_path is None when this algorithm's ``defaults``
+    # baseline isn't in ``skill_input.baselines`` (or the manifest has
+    # no baselines at all) — emit an empty string so the writer's
+    # Phase 2 skip-check triggers.
     print("ALGO\t" + "\t".join([
         algo["name"],
         algo["source_path"],
         algo["output_dir"],
         algo["config_output_path"],
+        algo.get("baseline_overlay_path") or "",
         algo.get("notes") or "",
     ]))
 PY
 [ $? -eq 0 ] || exit 1
 
-BASELINE_OVERLAY_REL=$(grep '^BASELINE_OVERLAY_REL=' /tmp/sim2real_translate_algos.txt | cut -d= -f2-)
-if [ -n "$BASELINE_OVERLAY_REL" ]; then
-    BASELINE_OVERLAY_PATH="$TRANSLATIONS_DIR/$BASELINE_OVERLAY_REL"
-else
-    BASELINE_OVERLAY_PATH=""
-fi
 CONTEXT_TEXT_B64_VAL=$(grep '^CONTEXT_TEXT_B64=' /tmp/sim2real_translate_algos.txt | cut -d= -f2-)
 if [ -n "$CONTEXT_TEXT_B64_VAL" ]; then
     CONTEXT_TEXT=$(printf '%s' "$CONTEXT_TEXT_B64_VAL" | base64 -d) || {
@@ -220,7 +214,7 @@ script, and read context paths line-by-line so paths containing spaces are
 preserved):
 
 ```bash
-while IFS=$'\t' read -r _ name source_rel _ _ _; do
+while IFS=$'\t' read -r _ name source_rel _ _ _ _; do
     test -f "$EXPERIMENT_ROOT/$source_rel" || {
         echo "HALT: algorithm source not found: $EXPERIMENT_ROOT/$source_rel"
         exit 1
@@ -240,7 +234,7 @@ every algorithm is complete, exit with an "already complete" message:
 
 ```bash
 INCOMPLETE_ALGOS=""
-while IFS=$'\t' read -r _ name _ output_dir _ _; do
+while IFS=$'\t' read -r _ name _ output_dir _ _ _; do
     algo_out="$TRANSLATIONS_DIR/$output_dir/${name}_output.json"
     if [ ! -f "$algo_out" ]; then
         INCOMPLETE_ALGOS="$INCOMPLETE_ALGOS $name"
@@ -267,12 +261,17 @@ For each `ALGO_NAME` in `$INCOMPLETE_ALGOS`:
 
 ```bash
 # Extract this algorithm's row from /tmp/sim2real_translate_algos.txt.
-IFS=$'\t' read -r _ _ ALGO_SOURCE_REL OUTPUT_DIR_REL CONFIG_OUTPUT_REL ALGO_NOTES < <(
+IFS=$'\t' read -r _ _ ALGO_SOURCE_REL OUTPUT_DIR_REL CONFIG_OUTPUT_REL BASELINE_OVERLAY_REL ALGO_NOTES < <(
     awk -F$'\t' -v n="$ALGO_NAME" '$1=="ALGO" && $2==n {print}' /tmp/sim2real_translate_algos.txt
 )
 ALGO_SOURCE="$EXPERIMENT_ROOT/$ALGO_SOURCE_REL"
 OUTPUT_DIR="$TRANSLATIONS_DIR/$OUTPUT_DIR_REL"
 CONFIG_OUTPUT_PATH="$TRANSLATIONS_DIR/$CONFIG_OUTPUT_REL"
+if [ -n "$BASELINE_OVERLAY_REL" ]; then
+    BASELINE_OVERLAY_PATH="$TRANSLATIONS_DIR/$BASELINE_OVERLAY_REL"
+else
+    BASELINE_OVERLAY_PATH=""
+fi
 # The new skill_input schema has no algorithm_config field — parameters (if
 # any) live inside the source file. Pass "" so the prompts take their
 # "parameter-free" branches when reasoning about configurable weights.
@@ -295,8 +294,10 @@ variable:
 - `{EXPERIMENT_ROOT}` → `$EXPERIMENT_ROOT`
 - `{TRANSLATIONS_DIR}` → `$TRANSLATIONS_DIR`
 - `{OUTPUT_DIR}` → `$OUTPUT_DIR`
-- `{BASELINE_OVERLAY_PATH}` → `$BASELINE_OVERLAY_PATH` (empty string when the
-  manifest declares no baseline overlay — writer's Phase 2 handles this)
+- `{BASELINE_OVERLAY_PATH}` → `$BASELINE_OVERLAY_PATH` (per-algorithm — points
+  at the overlay for THIS algorithm's baseline via `algorithms[i].baseline_overlay_path`;
+  empty when the algorithm's `defaults` isn't in `skill_input.baselines` or the
+  manifest has no baselines — writer's Phase 2 handles this)
 - `{ALGO_SOURCE}` → `$ALGO_SOURCE`
 - `{ALGO_CONFIG}` → `$ALGO_CONFIG` (always empty string; see Step 1 comment)
 - `{ALGO_NAME}` → `$ALGO_NAME`
--- a/.claude/skills/sim2real-translate/prompts/agent-writer.md
+++ b/.claude/skills/sim2real-translate/prompts/agent-writer.md
@@ -100,18 +100,17 @@ assembly time.
 - Top-level `scenario:` list with one dict
 - The `name` field MUST match the scenario name in the experiment's baseline file exactly
   (mismatched names cause llmdbenchmark to deploy multiple scenarios instead of merging)
-- InferenceObjectives go in `extraObjects`. The format pins one rule:
-  - `spec.poolRef.name` MUST equal `${model.idLabel}-gaie` so llm-d-benchmark
-    substitutes `${model.idLabel}` at render time (requires llm-d-benchmark
-    >= PR #1103).
-
-  Everything else MUST be copied verbatim from the project's context files
-  (typically including `config.md`) at the field level. In particular: copy the entire
-  `spec.poolRef` block (including `group`, `name`, and any other keys), not
-  just `name`. Same applies to `apiVersion`, `metadata`, `spec.priority`, and
-  any other nested fields the project's `config.md` declares — if it's in
-  config.md's InferenceObjective, it MUST appear in the generated output.
-  Do not paraphrase or selectively include fields.
+- InferenceObjectives go in `extraObjects`. Copy the InferenceObjective
+  verbatim from the project's context files (typically `config.md`) at the
+  field level. In particular: copy the entire `spec.poolRef` block
+  (including `group`, `name`, and any other keys). Same applies to
+  `apiVersion`, `metadata`, `spec.priority`, and any other nested fields
+  the project's `config.md` declares — if it's in config.md's
+  InferenceObjective, it MUST appear in the generated output. Do not
+  paraphrase or selectively include fields, and do not substitute or
+  rewrite `${model.idLabel}` — llm-d-benchmark expands it at render time
+  (requires llm-d-benchmark >= PR #1103), so pass any such placeholders
+  through verbatim.
 - Plugin config in `inferenceExtension.pluginsCustomConfig` as a YAML-in-YAML string
 - Only include fields you are adding or overriding
 
--- a/.claude/skills/sim2real-translate/prompts/agent-writer.md
+++ b/.claude/skills/sim2real-translate/prompts/agent-writer.md
@@ -81,14 +81,16 @@ Example queries:
 
 Use TaskCreate: `"Phase 2: Baseline Config Derivation"` → TaskUpdate in_progress
 
-**Skip check:** If `{BASELINE_OVERLAY_PATH}` is empty (the manifest declared no baseline
-overlay for this translation), skip Phase 2 entirely and proceed to Phase 3. If
-`{BASELINE_OVERLAY_PATH}` is set and the file already exists (written by a prior algorithm's
-skill invocation in the same translation), skip Phase 2 — send:
+**Skip check:** If `{BASELINE_OVERLAY_PATH}` is empty (this algorithm's baseline is not
+in the manifest's `baselines[]` — should not happen for valid manifests), skip Phase 2
+entirely and proceed to Phase 3. If `{BASELINE_OVERLAY_PATH}` is set and the file already
+exists (written by a prior algorithm's skill invocation in the same translation whose
+`defaults` cross-references the same baseline), skip Phase 2 — send:
 ```
 SendMessage({MAIN_SESSION_NAME}, "baseline-ready: {BASELINE_OVERLAY_PATH}")
 ```
-and wait for "continue". Baseline config is shared across algorithms in the same translation.
+and wait for "continue". Algorithms that share a baseline (via `defaults`) share this
+overlay file; algorithms with distinct baselines produce distinct overlay files.
 
 Use the inputs listed in the upfront "Inputs — Read These Now" table.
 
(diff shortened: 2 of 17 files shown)


kalantar:

Surgical prompt-template fix resolves the contradiction cleanly; no code paths change, no gaie references remain in the skill, and the specialists' non-blocking notes are prompt-wording nits below the bar for PR feedback.

kalantar:

Docs drift from the pinned schema and a new BYO-compat fallback lands with no test — both need addressing before merge.

### Important
- **design.md skill_input.json example omits the new algorithms[i].baseline_overlay_path field** (`docs/epics/step-2/design.md:209`) — docs/epics/step-2/design.md lines 209-218 show the algorithms entry without baseline_overlay_path, but _build_skill_input (sim2real.py:216) unconditionally emits it and the field-semantics text at line 228 mentions baselines[i].generated_overlay_path but not algorithms[i].baseline_overlay_path — the section explicitly bills itself as the pinned schema so PR 3 and PR 4 don't drift, and this example has drifted.
- **BYO legacy overlay fallback in assemble_run.py has no test exercising the 'legacy file exists' branch** (`pipeline/lib/assemble_run.py:553`) — assemble_run.py:552-559 is new code that falls back to generated/baseline_config.yaml when the per-baseline dir is absent; test_assemble_run.py's _make_experiment never writes that file, so only the None branch is covered. A future refactor could silently regress BYO compat for pre-refactor translations — the whole point of this fallback.

### Recommendations
- **config_path in skin_baselines has no consumer — resolve the open question the issue flagged** (`pipeline/sim2real.py:795`) — sim2real.py:795 emits config_path into skill_input.baselines[] but nothing in SKILL.md or agent-writer.md reads it (verified via grep). The issue explicitly noted this as an unresolved decision to either thread through Phase 2 or drop; the PR left it unresolved.
- **SKILL.md required-keys check should include 'baselines' for schema-consistency** (`.claude/skills/sim2real-translate/SKILL.md:170`) — .claude/skills/sim2real-translate/SKILL.md:170-171 lists version/translation_hash/experiment_root/translations_dir/scenario/algorithms/context but not baselines. Impact is minor because the skill only reads algo['baseline_overlay_path'] per-row rather than top-level baselines[], but the guard is documented as the schema contract check and this key is now first-class.

kalantar:
Addressed both blocking findings from iteration 2.

`docs/epics/step-2/design.md` — the pinned skill_input.json example was missing the `algorithms[i].baseline_overlay_path` field that `_build_skill_input` has been emitting since the last commit. Added it to the JSON example on the algorithms entry, added a field-semantics bullet explaining the null case and the shared-baseline-file behavior, and listed it alongside `config_output_path` in the "paths relative to translations_dir" enumeration.

`pipeline/tests/test_assemble_run.py` — added `test_byo_legacy_baseline_overlay_at_generated_root_is_applied` covering the BYO legacy-fallback branch in `assemble_run.py:552-559`. Writes only `generated/baseline_config.yaml` (no per-baseline dir), guards the precondition, runs assemble, and asserts that a distinctive `extraObjects` entry from the legacy overlay landed on the resolved baseline. Without the fallback, `overlay_path` would be None and the assertion would fail — future refactors that regress BYO compat will trip this test.

Skipping the two recommendations (config_path with no consumer; SKILL.md required-keys omitting `baselines`) per the fix-issue rule that only must_fix + important block merge. Both are advisory.

Not addressing the silent-failure-hunter's per-lens blocking note about the fallback being silent (no warning when either overlay is absent) — the synthesizer chose not to promote it, and the fallback is intentionally silent for BYO translations where the legacy shape is normal.

Local verification: 1410 passed, 2 xfailed; ruff clean.

kalantar:

The PR's runtime logic is correct and the unit-level tests for _build_skill_input are thorough, but the project's README-update rule is violated and the diff's headline behavior — assemble reading the new per-baseline overlay path — has no assemble-level regression test.

### Important
- **pipeline/README.md not updated for the skill_input.json schema change and new overlay path layout** (`pipeline/README.md:714`) — CLAUDE.md line 158 requires README updates after any pipeline/ change affecting CLI flags, subcommand behavior, or artifact schema. This PR changes the skill_input.json schema (baseline dict -> baselines list, new algorithms[i].baseline_overlay_path), the on-disk overlay path (generated/baseline_config.yaml -> generated/baseline_<name>/baseline_config.yaml), and what assemble reads. Verified against pipeline/README.md: line 260 still lists only the legacy overlay path in Assemble's inputs, and line 714 in Scenario Overlay Format still describes a single baseline_config.yaml. The Translate Outputs section (line 184-186) omits the new baselines field entirely.
- **No assemble test writes a per-baseline overlay at the new primary path and asserts it is applied** (`pipeline/tests/test_assemble_run.py:580`) — The PR's core contract is that pipeline/lib/assemble_run.py:552 now reads generated/baseline_<name>/baseline_config.yaml instead of the hardcoded shared path. The only new assemble test (test_byo_legacy_baseline_overlay_at_generated_root_is_applied at line 580) exclusively exercises the legacy fallback branch and even asserts the per-baseline dir does not exist (line 603). If line 552 regressed to the wrong subpath or was removed, no test would catch it. _build_skill_input tests verify only that the correct string lands in JSON, not that assemble consumes it.

### Recommendations
- **Warn when neither per-baseline nor legacy overlay exists for skill-driven translations** (`pipeline/lib/assemble_run.py:559`) — assemble_run.py:559 silently falls through to overlay_path=None when both paths are absent. This is documented behavior for BYO baselines without overlays (see resolve_baseline docstring), so it is not a bug — but a stderr warning naming the baseline would surface a failed skill session that produced no overlay. Consider adding a one-line diagnostic in the fallback branch.
- **Sanitize tabs in the free-form notes field before joining into tab-delimited ALGO rows** (`.claude/skills/sim2real-translate/SKILL.md:193`) — SKILL.md line 193 emits algo.get('notes') or '' unsanitized into a tab-delimited row consumed by bash IFS=$'\t' loops at lines 217/237/264. An operator-supplied tab in notes would shift ALGO_NOTES and mis-index BASELINE_OVERLAY_REL for that algorithm. Trivial one-line fix: (algo.get('notes') or '').replace('\t', '

kalantar:
Addressed both blocking findings from iteration 3.

`pipeline/README.md` — updated three sections per CLAUDE.md's post-pipeline-change rule.

- §Translate §Outputs: expanded the skill_input.json bullet to enumerate the pinned schema (baselines list, per-algorithm baseline_overlay_path, etc.); added a `generated/baseline_<name>/` output directory bullet.
- §Assemble §Inputs: replaced the single-overlay bullet with two — the skill-driven per-baseline path and the BYO legacy path with fallback semantics.
- §Scenario Overlay Format: rewrote to describe both producer layouts side-by-side.

`pipeline/tests/test_assemble_run.py` — added `test_skill_driven_per_baseline_overlay_is_applied` covering the new primary path in `assemble_run.py`. Writes only `generated/baseline_<name>/baseline_config.yaml` with a distinctive extraObjects marker, guards the precondition that the legacy top-level file does NOT exist so the primary branch (not the fallback) is under test, runs assemble, and asserts the marker landed on the resolved baseline. If line 552 regressed, this test would fail. Pair with the existing legacy-fallback test to cover both branches.

Skipping the two recommendations (stderr warning when neither overlay exists; sanitize tabs in notes field) per the fix-issue rule that only must_fix + important block merge.

Local verification: 1411 passed, 2 xfailed; ruff clean.

kalantar:
## Review verdict: READY TO READY TO MERGE

The diff cleanly implements the plural baselines schema with defaults-driven per-algorithm overlay resolution, an appropriate legacy fallback in assemble, and end-to-end tests for both the skill-driven and BYO paths.

### Recommendations
- **Consider a two-baseline assemble test with distinct per-baseline overlays to lock in path-template correctness.** (`pipeline/tests/test_assemble_run.py:580`) — assemble_run.py:552 builds `generated/baseline_{bl_name}/baseline_config.yaml` inside a per-baseline loop; the single-baseline test at test_assemble_run.py:580 would still pass if the path template accidentally resolved to a fixed name. A two-baseline variant with distinguishing markers per baseline would close that regression class. Non-blocking polish

kalantar:

The registry-secret plumbing lands cleanly with strong tests and correct fail-early behavior at every callsite; remaining specialist findings are polish only.

### Recommendations
- **BYO overlay fallback in assemble_run.py runs silently when both the per-baseline and legacy shared overlays are absent — a one-line warn() would surface the exact class of "silent wrong path" this PR is fixing elsewhere.** (`pipeline/lib/assemble_run.py:553`) — Verified at assemble_run.py:552-559: when the per-baseline path is missing the code silently falls through to the legacy path, and if that is also missing sets overlay_path=None with no log. An operator whose skill-driven translation lost its generated/baseline_<name>/ dir will get overlay-less assembly with no signal. A warn() on the fallback and an error/warn when both are absent would match the fail-loudly posture of the rest of this PR.
- **build.py:106 docstring says an empty registry_secret_name would let build-epp.sh "read the wrong Secret and buildkit would fail to authenticate" — build-epp.sh actually validates the arg at line 48 and exits with a usage error before any kubectl or buildkit call.** (`pipeline/lib/build.py:106`) — Verified against build-epp.sh:48-51: the script enforces non-empty REGISTRY_SECRET_NAME during arg parsing and exits 1 with a Usage message before kubectl or buildkit invocation. The docstring's stated failure mode is inaccurate — the practical failure would be a shell-level usage error, not a Secret misread or auth failure.

kalantar:

Documentation contradicts itself on setup.py ownership, and the new baselines[] lookup silently returns null because the code queries a manifest key that doesn't exist in the pinned v3 schema.

### Important
- **cluster.py boundary paragraph still lists 'current run' as owned by setup.py, contradicting line 91 in the same file** (`pipeline/README.md:83`) — Verified in pipeline/README.md line 83: '...anything operator-side (registry choice, repo name, current run, orchestrator image, sim2real_root) belongs in `setup.py`'. Line 91 correctly states the current_run field is written by sim2real use. AC6 of #481 explicitly requires setup.py to no longer name a run; every other README/CLAUDE.md/writer-table location was updated. This one stale sentence contradicts the invariant this PR establishes.
- **baselines[i].config_path in skill_input.json is always null — the code queries a top-level key that doesn't exist in the v3 manifest schema** (`pipeline/sim2real.py:795`) — Verified pipeline/sim2real.py:795 uses `bl.get("config")`, but the transfer.yaml v3 schema documented in pipeline/README.md:641-647 nests baseline config paths under `sim.config` and `real.config`, and manifest.py:57-73 only validates `name` and `scenario` at the top level of each baseline entry. Every skill_input.json produced by translate will have `config_path: null`. This was introduced by commit a4dd563 in this PR. design.md line 205 shows the field with a non-null value in the schema example, and README.md line 187 documents baselines[] as carrying `config_path` — creating a false contract. Either the lookup key needs to be corrected (likely `bl.get("scenario")` given the design.md example), or the field should be dropped/documented as reserved.

kalantar:
Addressed both blocking findings from iteration 5.

`pipeline/README.md` line 83 — the cluster.py boundary sentence still named 'current run' as operator-side owned by setup.py, contradicting the updated §setup.py section on line 91. Rewrote the sentence to list only the four fields setup.py actually writes, added a follow-on clarifying that `current_run` in the same file is written by `sim2real use`, and updated the closing note from "the two never write the same file" to "the three writers never overlap on the same key".

`pipeline/sim2real.py` line 795 — `bl.get("config")` returned None for every baseline because the v3 manifest schema nests baseline config paths under `sim.config` and `real.config`, not at a top-level `config` key. Every `skill_input.json` since a4dd563 has been carrying `config_path: null` on every baseline. Dropped the field from `skin_baselines` entirely — it also had no downstream consumer, so this closes the open question flagged in the original issue #479 body ("if we keep this field it should be threaded into Phase 2; otherwise drop it"). Companion updates: removed the field from the pinned schema example in `docs/epics/step-2/design.md`, from the skill_input.json bullet in `pipeline/README.md`, and from three test fixtures in `test_translate.py` (with a negative assertion added to catch re-introduction).

Skipped the two recommendations (add `component_image` to the removed-keys parametrize list, add a direct unit test for `collect_config`'s 2-tuple return) per the fix-issue rule that only must_fix + important block merge.

Note on the workflow: the parallel[1] specialist reviewer hit its StructuredOutput retry cap during this iteration, so the synthesizer ran on 3/4 lenses. Both blocking findings came from the surviving lenses (code-reviewer, comment-analyzer), and both were valid.

Local verification: 1415 passed, 2 xfailed; ruff clean.

kalantar:

Two stale-doc references still point users at setup.py for a file setup.py no longer writes, directly undermining the ownership change this PR advertises.

### Important
- **deploy.py _cmd_build error messages still tell operators to "run setup.py" for run_metadata.json, contradicting this PR's own ownership rewrite** (`pipeline/deploy.py:280`) — Read pipeline/deploy.py lines 278-311: four err() strings direct users to "run setup.py" to fix missing/invalid run_metadata.json or empty registry/component_image. This same commit updates CLAUDE.md and pipeline/README.md to state that run_metadata.json is owned by `sim2real assemble` (confirmed at pipeline/lib/assemble_run.py:253,259 and setup.py:288-289 docstring), and the PR is directly modifying _cmd_build's signature and inserting a new check at line 300. Leaving these messages sends operators on a fruitless detour that is the exact failure mode #481 set out to remove.
- **build-epp.sh comment attributes run_metadata.json creation to a non-existent `/sim2real-setup` skill** (`pipeline/scripts/build-epp.sh:10`) — Verified `.claude/skills/` contains only sim2real-analyze, sim2real-bootstrap, sim2real-check, sim2real-translate — no sim2real-setup. This same PR rewrites the adjacent lines 11-13 of build-epp.sh's usage header (confirmed in `gh pr diff 478`), so the file is squarely in scope. run_metadata.json is now written by `sim2real assemble` (assemble_run.py:259) — the comment misleads any operator debugging the fallback path at build-epp.sh:66-74.

### Recommendations
- **agent-writer.md Phase 2 skip-check framing contradicts SKILL.md — one says the empty-overlay case "should not happen for valid manifests", the other treats it as a normal handled case** (`.claude/skills/sim2real-translate/prompts/agent-writer.md:85`) — SKILL.md:183-186 and :297-300 describe the empty `baseline_overlay_path` as a valid, handled state (algorithm's defaults not in baselines[]). agent-writer.md:85 calls the same state anomalous. Not a blocking bug but a real inconsistency in the two prompts that operators and future maintainers will read side by side.
- **No end-to-end test for a two-baseline manifest exercising the new per-baseline overlay loop twice** (`pipeline/tests/test_assemble_run.py:580`) — The per-baseline overlay dispatch introduced at assemble_run.py:545-566 is only covered end-to-end with a single-baseline manifest. A two-baseline test with distinct overlays under generated/baseline_<name>/ would confirm the loop resolves each baseline independently — currently that path is only covered by composition of unit tests. Low-risk gap, worth noting.

kalantar:
Addressed both blocking findings from iteration 6.

`pipeline/deploy.py:_cmd_build` — four err() strings in the run_metadata gating block (missing file, unparseable JSON, empty component_image, empty registry) told operators to "run setup.py" or "re-run setup.py". After #481, run_metadata.json is written by `sim2real assemble` (assemble_run.py:644-655). Rewrote all four to point at `sim2real assemble --run <name>`. Included the concrete path in the missing-file message so operators can see exactly what didn't get written.

`pipeline/scripts/build-epp.sh:10` — usage header comment attributed run_metadata.json creation to a non-existent `/sim2real-setup` skill. `.claude/skills/` contains only sim2real-{analyze,bootstrap,check,translate}. Rewrote to name `sim2real assemble`.

Scope note: `deploy.py:3324` (`_cmd_run_remote`'s ConfigMap-fallback error) also mentions setup.py, but sits outside the range the reviewer audited (lines 278-311, `_cmd_build`). Its message combines "setup.py and 'sim2real assemble'", which is partially defensible since setup.py still writes setup_config.json. Leaving it for a future targeted pass rather than folding it into this iteration.

Skipped the two recommendations (agent-writer.md Phase 2 skip-check framing; two-baseline E2E test).

Local verification: 1415 passed, 2 xfailed; ruff clean.

kalantar:

Four independent specialists ran on this PR (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer). Lens verdicts: READY / NEEDS DISCUSSION / READY / NEEDS DISCUSSION. Verifying the blocking findings against the code before promoting them to must_fix/important:

**Blocking claims found & why they were dropped:**

- `pipeline/deploy.py:293, 311` — messages say "re-run 'sim2real assemble --run <name>'" for `component_image`/`registry` fields that assemble does not populate (assemble's v1 schema at assemble_run.py:644-655 is `{version, run_name, translation_hash, cluster_id, params_hash, image_tag, assembled_at}` — no component_image/registry/repo_name). However, line 289's `component_image is None → return "skip"` short-circuit catches every case a fresh step-2 flow would produce; lines 293 and 311 are only reachable via hand-edited empty-string values or legacy step-1 files. Dead-code with misleading messages, but no runtime harm in the current step-2 flow.

- `pipeline/lib/assemble_run.py:559` — per-baseline overlay fallback silently becomes `None`. Recurring finding from five prior iterations; consistently deprioritized because the primary skill-driven path is covered end-to-end and BYO tolerates the None branch.

Neither blocker is user-reachable in the fresh step-2 flow that #481 targets. Aggressive filter → both drop to recommendations.

**Recommendations (non-blocking polish):**

1. `pipeline/deploy.py:293, 311` — dead-code branches with misleading messages introduced by an earlier review iteration. Consider deleting them in a follow-up.
2. `pipeline/lib/assemble_run.py:559` — silent None-fallback warrants a stderr warning at least. Well-scoped follow-up.
3. `pipeline/tests/test_assemble_run.py` — no multi-baseline assemble test. A swap or off-by-one in the per-baseline loop wouldn't be caught by current tests.
4. `pipeline/tests/test_setup_pipeline.py` — no chained setup→assemble integration test proving the #481 acceptance ("setup.py → assemble on fresh workspace without --force"). Individual invariants are tested on each side; composed flow is not.

**Note on workflow**: this review was run outside the standard review-changes Workflow because the PR's diff (100k chars) exceeded the fetch-artifacts stage's per-request generation budget under the workflow's watchdog. Four specialists ran directly via the Agent tool with fresh contexts and no visibility into implementation history; synthesis was performed manually against re-read code. Verdict discipline preserved.

PR #492, opened 2026-07-04
step-3 PR 2: assemble→resolve schema-completeness test (#489)

kalantar:
## Summary

Adds `TestAssembleResolveContract` in `pipeline/tests/test_assemble_run.py` — a single integration test that couples `assemble_run.assemble_run()` and `resolve.resolve_run()` on the same fixture to catch silent schema drift between what step-2's assemble writes and what step-3's resolve consumes.

Closes #489. Base: `refactor/v2-step-3` (epic #486).

## Assertions

On a fixture built via the existing `_make_experiment` pattern (plus a small inline stub for the baseline overlay — see note below):

1. Every path field the resolve JSON advertises exists on disk. Enumerated explicitly rather than walked generically so the drift signal is deterministic.
2. Value round-trip for the six fields both sides share: `run_name`, `cluster_id`, `translation_hash`, `image_tag`, `params_hash`, `assembled_at`.
3. `manifest_assembly.workloads` in the resolve view matches what the fixture wrote to `transfer.yaml`.
4. **Schema completeness** — `set(run_metadata.json keys) ⊆ {version, run_name, translation_hash, cluster_id, params_hash, image_tag, assembled_at}`. A future PR that adds a new key to `run_metadata.json` without extending resolve (and this whitelist) fires the assertion.

## Reviewer notes

- **`results.results_dir` excluded from the path-existence walk.** Resolve advertises this path unconditionally, but the `results/` directory is populated by the `collect` command later in the run lifecycle, not by `assemble`. The exclusion is documented inline in the test.
- **Baseline overlay stubbed inline.** `resolve_run` exposes `translation.baselines[].generated_overlay_path` unconditionally, but that file is a translate-time artifact that neither `_make_experiment` nor `assemble_run` creates. The test materializes an empty `baseline_baseline/baseline_config.yaml` inline so the path walk is unqualified.
- **Test placement.** Lives in `test_assemble_run.py` (matching the issue) so it can use the existing `_make_experiment` fixture verbatim.

## Sweep

Grepped for references to the new test name and the design section (`docs/epics/step-3/design.md#integration-test-assemble-resolve-parity-schema-completeness`); design description matches the implementation.

## Test plan

- [x] `test_resolve_reads_what_assemble_writes` passes.
- [x] Full `pytest pipeline/` passes (1349 passed, 2 xfailed unchanged).
- [x] `ruff check pipeline/ --select F` clean.

### DIFF
--- a/pipeline/tests/test_assemble_run.py
+++ b/pipeline/tests/test_assemble_run.py
@@ -11,7 +11,7 @@ from pathlib import Path
 import pytest
 import yaml
 
-from pipeline.lib import assemble_run, layout, slicer
+from pipeline.lib import assemble_run, layout, resolve, slicer
 
 
 def _init_fake_submodule(sub: Path) -> None:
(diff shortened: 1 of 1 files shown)


kalantar:

The integration test satisfies its stated contract and no finding rises above optional polish worth blocking on.

### Recommendations
- **Consider dropping "version" from known_keys, since resolve emits SCHEMA_VERSION rather than reading run_meta["version"]** (`pipeline/tests/test_assemble_run.py:1122`) — Verified at pipeline/lib/resolve.py:101 ("version": SCHEMA_VERSION) and pipeline/lib/assemble_run.py:647 (assemble writes version=1). The whitelist tracks what assemble writes, which is correct for the test's stated purpose, but the one key resolve never consumes from disk is version — so a future rename of that field on the file side would slip past this test. Optional polish only; the current test correctly fires on any newly-added key that appears in run_metadata.json.

PR #491, opened 2026-07-04
step-3 PR 1: sim2real resolve subcommand + hydrated-JSON view (#488)

kalantar:
Closes #488.

Introduces the new `sim2real resolve --run <name>` subcommand — the code-side helper the ported `sim2real-check` skill (child #490) will consume when invoked with `--run R`. Reusable by future skill ports (bootstrap, analyze) and by operator scripts.

## What's in the PR

- **`pipeline/lib/resolve.py`** — new module. `resolve_run(experiment_root, run_name) -> dict` implements the v1 JSON schema documented in `docs/epics/step-3/design.md`. Reads `run_metadata.json`, `translation_output.json` (via `translation_ref.read_translation_output` so legacy step-1 top-level `image_ref`/`image_digest` are normalized), and `manifest.assembly.yaml`. Filesystem probes for `phases_with_data` (predicate: one-level glob `results/<phase>/*/trace_data.csv`), `workloads_by_phase`, and the `cluster/` scenario YAMLs. `ResolveError` raised on the failure modes the design enumerates; the CLI wrapper catches it and exits 2 with the specific message.

- **`pipeline/sim2real.py`** — new `resolve` subcommand wired into `build_parser()` after `use`. Thin `_cmd_resolve` handler: `resolve_run()` → JSON to stdout (exit 0), or `ResolveError` → message to stderr (exit 2).

- **`pipeline/tests/test_resolve.py`** — 19 unit tests + 3 CLI wrapper tests. Covers happy-path v1 shape, translation section (algorithms + baselines from manifest), results section (partial-data filtering, phases_declared invariant, no-data case), cluster_scenarios enumeration, manifest_assembly slice, `cluster_config_path` null-when-absent, and every failure mode. CLI wrapper tests invoke `sim2real.py` via subprocess for real argparse routing.

- **`pipeline/README.md`** — new `sim2real resolve` entry alongside `list runs`, `list translations`, `use`. Documents the CLI, schema version, output content, and exit codes.

## Verification

- `python -m pytest pipeline/ .claude/skills/ -v` → 1437 passed, 2 xfailed (up 22 from prior 1415, matching the new tests).
- `ruff check pipeline/ .claude/skills/ --select F` → clean.
- Smoke: `sim2real resolve --run trial-3 --experiment-root <path-to-sr>` runs against the real `kalantar-msb/sr/workspace/runs/trial-3/` evidence and emits a complete JSON view.

[...]

PR #493, opened 2026-07-05
step-3 PR 3: sim2real-check SKILL.md rewrite — Step 0 dispatch + preserved checks (#490)

kalantar:
## Summary

Ports `/sim2real-check` (`.claude/skills/sim2real-check/SKILL.md`) to consume the new `sim2real resolve --run R` interface (#491) alongside the legacy `--real <path>` bundle mode. Steps 1-N (workload / config / signal / policy / runtime health) keep their evidence-based check semantics verbatim; only Step 0's mode dispatch and the downstream path-reference variables change.

Closes #490. Parent epic: #486.

## Design vs. implementation notes

The rewrite follows `docs/epics/step-3/design.md#skillmd-step-0-rewrite` closely. Three targeted deviations, all semantics-preserving, called out here so a reviewer can spot them at a glance:

1. **CLI invocation form.** The design snippet shows `sim2real resolve --run <name> --experiment-root <path>`. But `--experiment-root` is a top-level flag on `sim2real` (before the subcommand), not scoped to `resolve`. Confirmed by reading argparse in `pipeline/sim2real.py` and by running the CLI. The rewrite emits `sim2real --experiment-root <root> resolve --run <name>`.

2. **`BASELINE_CONFIG_PATH` added beyond the design's variable list.** The single-baseline path shape differs between modes: legacy has flat `$CONFIGS_DIR/baseline_config.yaml`; resolve-mode workspace nests it as `$CONFIGS_DIR/baseline_<name>/baseline_config.yaml`. Without a bridging variable, resolve-mode baseline lookups (Steps 2c, 2d, 4b, 5c.6) would resolve to a non-existent path and the demo criteria in the epic would fail. Resolves via `.translation.baselines[0].generated_overlay_path` in resolve-mode. Multi-baseline handling stays deferred to the follow-up called out in the design's out-of-scope section.

3. **`WORKLOADS_DIR` added beyond the design's variable list.** Similar bridging: legacy `<real>/workloads/` vs resolve-mode `<experiment-root>/workloads/`. Same shape as the two-mode adapter for the other variables.

4. **Plain-text confirmation prompts instead of `AskUserQuestion`.** The design prescribes `AskUserQuestion`. I have a UI-side bug on my end where clicking to focus that widget registers as picking the currently-focused option; the rewrite uses plain-text numbered prompts to sidestep it. Semantics identical — user still confirms resolved values before checks run; args are still defaults, not silent overrides.

## Variables

- **Kept:** `SIM`, `REAL`, `WORKLOAD` (interactive), `BLIS`, `GAIE`, `LLMD`
- **New from design:** `PHASES`, `WORKLOADS_BY_PHASE`, `TRANSLATION_HASH`, `IMAGE_TAG`, `MANIFEST_ASSEMBLY_PATH`, `CLUSTER_CONFIG_PATH`
- **Added (see note 2 & 3 above):** `BASELINE_CONFIG_PATH`, `WORKLOADS_DIR`

## Mechanical path substitutions in Steps 1-N

| Before | After |
|---|---|
| `<real>/generated/` | `$CONFIGS_DIR` |
| `<real>/generated/baseline_config.yaml` | `$BASELINE_CONFIG_PATH` |
| `<real>/workloads/` | `$WORKLOADS_DIR` |
| `<real>/baseline/<workload>/` and `<real>/treatment/<workload>/` | iterate `$PHASES` × `${WORKLOADS_BY_PHASE[$phase]}` referencing `$RESULTS_DIR/<phase>/<workload>/` |

Phase-iteration prose changed accordingly: where the current SKILL.md said "for each workload in the real bundle" (implying `baseline/`+`treatment/` as fixed names), the rewrite iterates `$PHASES`. Naming like "baseline", "treatment", "control" that carries semantic meaning in specific checks (4b Admission — Baseline, 4c Admission — Treatment) is preserved.

## Diff shape

- Before: 507 lines
- After: 641 lines
- Diff: **212 insertions / 78 deletions** — 290 line diff
- Design estimated ~150-200. The +134 net delta comes from the plain-text confirmation-prompt prose (replacing a two-line `AskUserQuestion` call) and the two extra path-bridge variables plus their comments.

## Sweep for stale references

Grepped for `sim2real-check` and the removed CLI flags across `**/*.md` and `**/*.py`:
- `README.md` line 128, 135 — mention `/sim2real-check` with no flags; still accurate (auto-detect mode).
- `pipeline/README.md` line 459 — describes what the resolve JSON is consumed by; still accurate.
- `pipeline/lib/resolve.py` docstring — still accurate.
- No dangling `--translation` / `--algorithm` / `--workload` refs elsewhere.

## Test plan

- [x] `pytest pipeline/ pipeline/tests/test_layout.py pipeline/tests/test_cluster_ops.py pipeline/tests/test_cluster_py.py pipeline/tests/test_slicer.py pipeline/tests/test_sim2real.py pipeline/tests/test_assemble_run.py .claude/skills/sim2real-analyze/tests/ .claude/skills/sim2real-bootstrap/tests/ .claude/skills/sim2real-translate/tests/ -v` — 1438 passed, 2 xfailed. No unit tests added for the skill (it's a prompt template — validated by the demo-evidence smoke described in the epic).
- [ ] Real-cluster smoke demo: `/sim2real-check --run <name>` against a step-2-assembled workspace run, plus `/sim2real-check --real <legacy-bundle>` against a pre-refactor bundle. Attached as evidence to the epic (#486) demo before merging the epic branch.

## Base

`refactor/v2-step-3`. All child dependencies for this issue are already on that branch: #488 (PR #491) and #489 (PR #492).

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -8,35 +8,166 @@ description: Validate a sim2real translation bundle against its simulation bundl
 $ARGUMENTS
 ```
 
-Parse user input if provided (e.g., `--sim <path> --real <path> --workload <name>`).
+Parse user input for these flags (all optional; auto-detect fills in the rest):
 
-## Step 0: Auto-Detect and Confirm Paths
+- `--sim <path>` — path to the simulation bundle (BLIS output).
+- `--real <path>` — path to a pre-refactor legacy real bundle on disk. Mutually exclusive with `--run`.
+- `--run <name>` — a workspace-registered run name. Requires the run to have been assembled by `sim2real assemble --run <name>`. Mutually exclusive with `--real`.
+- `--experiment-root <path>` — override the experiment repo root. Defaults to the current working directory. Used only in `--run` mode and in the auto-detect fallback; silently ignored in `--real` mode.
 
-Before running ANY checks, **auto-detect** paths by scanning the working directory and context, then **confirm with the user** using `AskUserQuestion`.
+Removed vs. the pre-refactor version of this skill: `--translation`, `--algorithm`, `--workload` (all redundant now — `--run` resolves them from disk).
 
-### Auto-Detection Strategy
+## Step 0: Resolve Input, Populate Variables, Confirm With User
 
-1. **Sim bundle**: Look for directories containing `README.md` + `algorithm/` + `workloads/` under `experiments/`. Common patterns: `sim2real_bundle*`, `sim2real_*bundle*`.
-2. **Real bundle**: Look for directories containing `generated/` + (`baseline/` or `treatment/`) under `experiments/`. Common patterns: `bad_and_good*`, `*admission*`, `*deployment*`.
-3. **BLIS codebase**: The current working directory (check for `sim/` dir and `go.mod`).
-4. **GAIE codebase**: Look for `gateway-api-inference-extension` under `tmp/`, `../`, or nearby.
-5. **llm-d scheduler**: Look for `llm-d-inference-scheduler` under `tmp/`, `../`, or nearby.
-6. **Workloads**: List subdirectories under `<real>/baseline/` and `<real>/treatment/`.
+Before running ANY checks, resolve the input into a common set of named shell variables, then confirm with the user before proceeding. Every downstream check subsection references these variables — Steps 1-N are mode-agnostic.
 
-Use `Glob` and `Bash` (ls) to find candidates. Then present findings to the user:
+### Mutex enforcement
+
+If both `--run` and `--real` are set, exit with:
 
 ```
-I found the following paths. Please confirm or correct:
-
-Sim bundle:      <detected or "not found">
-Real bundle:     <detected or "not found">
-BLIS codebase:   <detected or "not found">
-GAIE codebase:   <detected or "not found">
-llm-d scheduler: <detected or "not found">
-Workload(s):     <detected or "all">
+--run and --real are mutually exclusive; --run resolves a workspace-registered run, --real reads a filesystem bundle path.
 ```
 
-Use `AskUserQuestion` to let them confirm or provide corrections. Do NOT proceed until confirmed.
+### Mode dispatch
+
+Three modes: `--run` (resolve), `--real` (legacy), or neither (auto-detect + user picker). Each mode populates the same variable set.
+
+```bash
+# --experiment-root <path> resolves to a directory. Default to the current
+# working directory when the flag is absent.
+EXPERIMENT_ROOT="${EXPERIMENT_ROOT:-$(pwd)}"
+
+if [ -n "$RUN" ]; then
+    # Resolve-mode: pull the hydrated JSON view of the run and parse it.
+    # --experiment-root is a top-level flag on sim2real (before the
+    # subcommand), not a subcommand-scoped flag.
+    sim2real --experiment-root "$EXPERIMENT_ROOT" resolve --run "$RUN" \
+        > /tmp/sim2real_check_resolved.json
+    CONFIGS_DIR=$(jq -r '.translation.generated_dir' /tmp/sim2real_check_resolved.json)
+    RESULTS_DIR=$(jq -r '.results.results_dir' /tmp/sim2real_check_resolved.json)
+    PHASES=$(jq -r '.results.phases_with_data | join(" ")' /tmp/sim2real_check_resolved.json)
+    # WORKLOADS_BY_PHASE is a bash associative array keyed by phase name.
+    declare -A WORKLOADS_BY_PHASE
+    while IFS=$'\t' read -r phase wls; do
+        WORKLOADS_BY_PHASE["$phase"]="$wls"
+    done < <(jq -r '.results.workloads_by_phase | to_entries[] | "\(.key)\t\(.value | join(" "))"' /tmp/sim2real_check_resolved.json)
+    TRANSLATION_HASH=$(jq -r '.translation.hash' /tmp/sim2real_check_resolved.json)
+    IMAGE_TAG=$(jq -r '.image_tag' /tmp/sim2real_check_resolved.json)
+    MANIFEST_ASSEMBLY_PATH=$(jq -r '.manifest_assembly.path // ""' /tmp/sim2real_check_resolved.json)
+    CLUSTER_CONFIG_PATH=$(jq -r '.cluster_config_path // ""' /tmp/sim2real_check_resolved.json)
+    # BASELINE_CONFIG_PATH resolves the first baseline overlay in the
+    # workspace's nested-under-baseline_<name>/ shape. Empty if the run has
+    # no baseline. Downstream check subsections that need the baseline
+    # EPP config reference this variable.
+    BASELINE_CONFIG_PATH=$(jq -r '.translation.baselines[0].generated_overlay_path // ""' /tmp/sim2real_check_resolved.json)
+    # WORKLOADS_DIR is the experiment repo's workload YAML directory.
+    # In resolve-mode there is no per-run copy — the manifest.assembly.yaml
+    # names the source files, all under $EXPERIMENT_ROOT/workloads/.
+    WORKLOADS_DIR=$(jq -r '.experiment_root' /tmp/sim2real_check_resolved.json)/workloads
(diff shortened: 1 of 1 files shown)


kalantar:

The mode-dispatch rewrite lands cleanly but leaves several variables the downstream checks reference (SIM, BLIS, GAIE, LLMD) unset in the --run and --real branches, and swallows resolve/ls failures without surfacing them to the operator.

### Must fix
- **SIM is never populated in any of the three mode branches, yet Steps 1-5 dereference $SIM heavily.** (`.claude/skills/sim2real-check/SKILL.md:151`) — Lines 41-118 of SKILL.md contain the entire mode dispatch and no branch assigns SIM. The confirmation prompt at line 151 shows 'Sim bundle (SIM): <path>' as if it were resolved. Steps 1a, 2a, 2d, 3, 4d, 5b all reference $SIM/README.md, $SIM/config.md, $SIM/algorithm/, $SIM/workloads/ — none of these are reachable when the operator invokes with --run or --real. The header (line 11) tells the operator all flags are optional and 'auto-detect fills in the rest', so an operator running `/sim2real-check --run trial-3` will proceed with SIM empty and every $SIM-based check will fail on missing files.
- **BLIS, GAIE, and LLMD detection lives only in the auto-detect branch but Sections 3 and 4 hard-require them and the confirmation prompt shows them in every mode.** (`.claude/skills/sim2real-check/SKILL.md:117`) — Lines 116-119 place BLIS/GAIE/LLMD detection inside the 'Auto-detect + user picker' subsection. The --run branch (lines 41-67) and --real branch (lines 68-103) never invoke that detection. The confirmation template at lines 162-164 displays BLIS/GAIE/LLMD for all modes with a '(not found)' placeholder. Sections 3 (Signal Parity) and 4 (Policy Parity) require grepping GAIE/LLMD codebases for code-proof claims. In --run or --real invocation, these variables stay unset, the operator sees '(not found)' in the confirmation, and if they type ok the code-proof checks silently have no codebase to grep. Pre-refactor SKILL.md ran this detection unconditionally in Step 0.
- **sim2real resolve exit status is unchecked; a non-zero exit silently yields blank variables and an empty confirmation prompt.** (`.claude/skills/sim2real-check/SKILL.md:45`) — Lines 45-46 pipe resolve stdout to /tmp/sim2real_check_resolved.json with no exit-code guard. If resolve fails (unknown run, missing workspace, schema drift while the step-3 branch stabilizes), the file contains partial output or a Python traceback text, jq -r on missing fields prints 'null', and the operator sees a confirmation with mostly-blank fields with no indication of the underlying error. Since this SKILL.md is an executable prompt template that a Claude agent will run verbatim, the missing check is a real failure mode, not a code-quality nit.
- **Legacy-mode phase discovery hides a missing RESULTS_DIR behind `2>/dev/null`, producing zero PHASES and a report with no evidence.** (`.claude/skills/sim2real-check/SKILL.md:82`) — Line 82 is `PHASES=$(ls "$RESULTS_DIR" 2>/dev/null | while read d ...)`. If $REAL is a mistyped path so neither $REAL/results nor $REAL/ exists, or if $REAL is a legacy bundle whose shape doesn't match either probed layout, ls errors are suppressed. PHASES ends up empty, WORKLOADS_BY_PHASE stays empty, and every Steps 1-5 iteration `for phase in $PHASES` iterates zero times. The operator receives a report that looks complete but contains no actual checks — a classic silent-empty-loop failure mode that the pre-refactor skill didn't have.

### Important
- **The line-18 description of removed flags conflates 'removed as CLI flag' with 'removed entirely' for --workload.** (`.claude/skills/sim2real-check/SKILL.md:18`) — Line 18 states --translation, --algorithm, --workload are all 'removed vs. the pre-refactor version'. Line 165 of the confirmation template still shows 'Workload filter (WORKLOAD): <name or (all)>' as an interactive-only override, and line 149 documents `WORKLOAD=w7` as an override syntax. An operator reading only the header will not realize WORKLOAD persists interactively and may be confused by the confirmation template. This is a documentation-only fix but the SKILL.md is the sole spec for how the agent behaves.
- **WORKLOADS_DIR is hardcoded to $EXPERIMENT_ROOT/workloads/ in resolve-mode, but manifest.assembly.yaml workload paths are arbitrary relative-to-experiment-root paths.** (`.claude/skills/sim2real-check/SKILL.md:74`) — Line 67 sets `WORKLOADS_DIR=$(jq -r '.experiment_root' ...)/workloads` and line 74-75 comments assert all workload YAMLs live 'under $EXPERIMENT_ROOT/workloads/'. Reviewing pipeline/lib/assemble_run.py shows workload paths from transfer.yaml are resolved as `exp_root / wl_path_str` where the path is arbitrary. The convention holds in current fixtures but is not a schema guarantee. Section 1's WORKLOAD PARITY lookup will silently miss workload YAMLs if the experiment repo uses a different layout. Loosening the comment or driving WORKLOADS_DIR from manifest.assembly.yaml is the correct fix.

kalantar:
Round-1 review feedback addressed in 466c249. All six items were valid FIXes on the shell pseudocode inside Step 0 — no re-architecture, just tightening the mode-dispatch block and adding one new shared step for the four codebase variables the checks need in every mode.

**Must-fix**

- SIM and BLIS/GAIE/LLMD are now populated by a new "Auxiliary detection (all modes)" section that runs after the mode dispatch, regardless of whether the operator invoked with `--run`, `--real`, or nothing. Previously these lived only under the auto-detect subsection, which the resolve and legacy branches never entered. SIM is now blocked at the confirmation prompt if auto-detect can't find it; GAIE/LLMD/BLIS-dependent checks emit INAPPLICABLE if their codebase is unavailable.
- `sim2real resolve` now has an explicit exit-code guard, and the three mandatory fields (`CONFIGS_DIR`, `RESULTS_DIR`, `PHASES`) use `jq -re` so schema drift fails loudly instead of coalescing to empty strings. Also moved from a shared `/tmp/sim2real_check_resolved.json` to per-invocation `mktemp` + EXIT trap so a stale file from a prior failed run can't be silently reused.
- Legacy mode now validates that `--real` and `RESULTS_DIR` exist before scanning, and errors out if `PHASES` ends up empty after the denylist filter. Previously a mistyped path or an unrecognized bundle shape produced a clean-looking report containing zero checks.

**Important**

- Line 18 now clarifies that `--workload` was removed as a CLI flag but WORKLOAD persists as an interactive-only override at the confirmation prompt — matching the confirmation template and the `WORKLOAD=w7` override example that were already elsewhere in the file.
- WORKLOADS_DIR comment softened: workload paths in `manifest.assembly.yaml` are arbitrary relative-to-experiment-root per `pipeline/lib/assemble_run.py:_load_workload`. The `$EXPERIMENT_ROOT/workloads/` default is now documented as a project convention, and operators can override at the prompt when the experiment repo uses a different layout.

Verified: `ruff check pipeline/ .claude/skills/ --select F` clean; `pytest pipeline/ ...` 1438 passed / 2 xfailed.

kalantar:

The SKILL.md rewrite faithfully implements the epic's Step 0 dispatch and preserves Steps 1-N semantics via mechanical path substitutions, and the specialists' remaining concerns don't survive verification against the resolve.py schema contract or the prompt-template nature of the artifact.

### Recommendations
- **Normalize TRANSLATION_HASH and IMAGE_TAG jq reads with `// ""` to avoid displaying the literal string "null" at the confirmation prompt on schema drift** (`.claude/skills/sim2real-check/SKILL.md:75`) — Verified against jq behavior: `echo '{}' | jq -r '.translation.hash'` prints `null`. resolve.py guarantees these fields when exit=0, so this is defense in depth, but the confirmation prompt would render `Translation hash: null` under drift — a small clarity win to align with the `// ""` pattern already used on lines 77-78 and 83.
- **Emit a WARN when a legacy phase directory yields zero workload subdirectories rather than silently populating an empty list** (`.claude/skills/sim2real-check/SKILL.md:133`) — Verified at line 132-135: the `ls -d ...*/  2>/dev/null | xargs` idiom produces an empty string when the phase has no subdirs, and downstream `for wl in ${WORKLOADS_BY_PHASE[$phase]}` loops iterate zero times with no operator-visible signal. Since the legacy branch already has an early-exit for missing PHASES (line 122), the same pattern extended to per-phase workload emptiness would keep the diagnostic story consistent.

PR #501, opened 2026-07-05
step-4 PR 1: schema `byo: true` marker + validator + command-specific guards

kalantar:
Closes #497. Parent epic #496 (step-4). Base: `refactor/v2-step-4`.

## Summary

Adds the per-algorithm `byo: true` marker to the v3 `transfer.yaml` schema, loosens the manifest validator so BYO entries can omit `source:` (and all-BYO manifests can omit `component:`), and adds explicit command-specific guards so BLIS-only commands fail clearly on BYO-shaped inputs.

Non-BYO entries still require `source:`. Non-BYO manifests still require `component:`. Mixed manifests still require `component:`.

## Changes

- `pipeline/lib/manifest.py` — accept optional `algorithms[i].byo: bool`; skip `source:` requirement when `byo is True`; reject non-bool values with a message naming the field and its type; loosen `_validate_v3_fields` so `component:` is required only when at least one algorithm is non-BYO. Error message for the still-required case is now `"component is required when non-BYO algorithms are specified"` — the existing regex-based test still matches.
- `pipeline/sim2real.py` `_cmd_translate` — refuse to translate when any algorithm carries `byo: true`. Error names the algorithm and points at `sim2real translation register`. Without this guard the schema loosening would let a BYO manifest silently produce a zero-source checkpoint via `slicer.translation_hash_with_sources` (which skips entries whose `source:` is None).
- `pipeline/sim2real.py` `_cmd_build` — load `transfer.yaml` up front (with the same `.` / `config/` fallback `_cmd_translate` uses) and error if `component:` is absent. If the manifest is unparseable, defer to the downstream path so its error wins rather than swallowing it here.
- `pipeline/tests/test_manifest.py` — 11 new tests: `byo` acceptance, source loosening on BYO, non-bool rejection (parametrized over 5 shapes), `component:` optional for all-BYO, `component:` still required for mixed, and a downstream-consumer regression test called out in the acceptance criteria.
- `pipeline/tests/test_translate.py` — 2 new tests: all-BYO and mixed manifests both error with the mandated message.
- `pipeline/tests/test_build.py` — 3 new tests: transfer.yaml at repo root, transfer.yaml at `config/`, and no manifest (guard is a no-op).
- `pipeline/README.md` — one paragraph documenting the `byo:` marker under the manifest-shape section.
- `docs/superpowers/plans/2026-07-05-issue-497-byo-marker-schema.md` — plan doc used to drive implementation, following the convention of earlier issue plans in the same directory.

## Verification

- `.venv/bin/python -m pytest pipeline/` — 1365 passed, 2 xfailed (unrelated).
- `.venv/bin/ruff check pipeline/ --select F` — clean.
- Grep sweep for the old error string `"component is required when algorithms are specified"` returns only self-references inside the plan doc explanation.

## Reviewer notes

- The `_cmd_build` guard adds a `manifest.load_manifest` call at the top of the command. This is a new dependency for `_cmd_build` (previously it read only the workspace translation dir, never `transfer.yaml`). If loading `transfer.yaml` from `_cmd_build` is undesirable, the alternative would be to move the guard into the caller-facing register path — but the issue's acceptance criterion mandates the specific `"nothing to build — this transfer.yaml declares no component"` message, which requires reading the manifest.
- The `_cmd_translate` guard runs after the existing `translation_ref.validate_name` loop and before `slicer.translation_hash_with_sources`. Placing it earlier would mean the name-validation regression tests catch BYO manifests via the name-invalid path, which is less specific than the mandated error message.
- The BYO discriminator is per-entry (`algorithms[i].byo`) rather than a top-level flag. This matches the design doc's "Explicit BYO discriminator" decision and supports future mixed-mode manifests (some BLIS, some BYO) without a schema-level split.

### DIFF
--- a/docs/superpowers/plans/2026-07-05-issue-497-byo-marker-schema.md
+++ b/docs/superpowers/plans/2026-07-05-issue-497-byo-marker-schema.md
@@ -1,0 +1,702 @@
+# Issue #497: schema `byo: true` marker + validator + command-specific guards
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add per-algorithm `byo: true` marker to the v3 `transfer.yaml` schema, loosen the validator so BYO entries can omit `source:` and all-BYO manifests can omit `component:`, and add explicit guards in `sim2real translate` / `sim2real build` so BLIS-only commands fail with actionable messages on BYO-shaped manifests.
+
+**Architecture:** The change lands in three places. (1) `pipeline/lib/manifest.py` gets a new optional `byo` bool per algorithm entry, and two validator loosens — the per-algo required-field loop skips `source` when `byo is True`, and `_validate_v3_fields`'s `component` requirement changes from "any algorithm" to "any non-BYO algorithm". (2) `pipeline/sim2real.py` gains two per-command guards: `_cmd_translate` errors on any BYO algorithm, and `_cmd_build` loads `transfer.yaml` (with the same discovery fallback `_cmd_translate` uses) and errors if `component:` is absent. (3) Tests + README note.
+
+**Tech Stack:** Python 3, PyYAML, pytest.
+
+## Global Constraints
+
+- Non-BYO entries still require `source:`. Non-BYO manifests still require `component:`. BLIS validation stays strict.
+- Error messages verbatim from the issue acceptance criteria — do not paraphrase.
+- Preserve existing tests (regression); no drift in surface area outside what the issue names.
+- Lint: `ruff check pipeline/ --select F` must stay clean.
+- File paths: this session runs inside the worktree at `.claude/worktrees/issue-497-byo-marker-schema/`. Every path in every step must include that substring.
+
+---
+
+### Task 1: Schema — accept, type-check, and document the `byo: true` marker
+
+**Files:**
+- Modify: `pipeline/lib/manifest.py:80-90` (per-algorithm required-field loop)
+- Test: `pipeline/tests/test_manifest.py` (add tests after the existing algorithms-block tests)
+
+**Interfaces:**
+- Consumes: nothing new
+- Produces: loader accepts `algorithms[i].byo: true|false` (bool); type-check errors on non-bool; when `byo is True`, `source:` becomes optional for that entry; `name` and `defaults` remain required for every entry
+
+- [ ] **Step 1: Write failing tests for `byo` acceptance, type check, and BYO-entry source loosening**
+
+Append to `pipeline/tests/test_manifest.py`:
+
+```python
+# ── byo: true per-algorithm marker (v3 schema addition) ─────────────────
+
+def test_byo_true_algorithm_loads_without_source(tmp_path):
+    """`algorithms[i].byo: true` makes `source:` optional for that entry.
+
+    Non-BYO manifests still need `component:`, so keep the MINIMAL_V3
+    component block; loosening `component:` is exercised separately.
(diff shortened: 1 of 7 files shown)


kalantar:

The README's v3 schema bullet still declares `component` as an unconditional required field, contradicting the loosening this PR ships one section below.

### Important
- **README line 37 lists `component` as a required v3 schema field without qualification, contradicting the loosening this PR introduces** (`pipeline/README.md:37`) — Verified against `pipeline/README.md` line 37: `v3 schema with \`component\`, \`baselines\`, \`algorithms\`, \`workloads\` fields`. This PR loosens `component:` to optional for all-BYO manifests (manifest.py `_validate_v3_fields`) and adds a `byo:` subsection immediately below explaining that, but the first-encounter schema description still says `component` unconditionally. Since documenting the schema loosening in the README is part of this PR's stated deliverable, the first-line description needs to reflect it — otherwise a reader hitting the schema bullet before the subsection is misled about what is and isn't required.

### Recommendations
- **`_cmd_build`'s new BYO guard catches only `ManifestError`, letting an `OSError` from `load_manifest.path.read_text()` escape unhandled to `main()`** (`pipeline/sim2real.py:892`) — Verified: `pipeline/lib/manifest.py:31-34` wraps `path.read_text()` in a `try/except yaml.YAMLError` only, so `PermissionError` or a TOCTOU race after the outer `manifest_path.exists()` check would propagate. `pipeline/sim2real.py:889-892` catches only `_manifest.ManifestError`, and `main()` (line 1374) has no broad handler — the result is a Python traceback instead of the `error: ...` line every other path in this CLI produces. Note this same narrow-except pattern already existed for `_cmd_translate`'s primary `load_manifest` call and this PR does not change that, so it's a pre-existing consistency issue rather than a new bug; worth flagging as polish (widen to `(ManifestError, OSError)`) but not a merge blocker.

kalantar:
Addressed the important item from the review: rewrote pipeline/README.md line 37 to qualify component: as required only when at least one algorithm is BLIS-shape, and cross-linked to the byo: subsection below.

Skipped the recommendation about widening the _cmd_build BYO guard's except clause to also cover OSError — the synthesizer classified it as advisory (not merge-blocking), and the same narrow-except pattern already exists for _cmd_translate's load_manifest call. If you want it as a small follow-up polish across both call sites in one pass, happy to file that separately.

168 tests pass.

kalantar:

Acceptance criteria are covered, tests are comprehensive, ruff is clean; only cosmetic polish items remain.

### Recommendations
- **README asserts a `.go` file-type constraint the code does not enforce.** (`pipeline/README.md:50`) — pipeline/lib/slicer.py:157-162 reads `algorithms[i].source` via `read_bytes()` with no extension check, and the translate tests use `.py` sources. The README line 50 wording 'no `.go` source for the skill to translate' overstates the constraint — 'no `source:` file' would match the validator error text and the actual behavior.
- **`_cmd_build` BYO guard fires with a misleading 'all algorithms are BYO' message when a manifest declares no algorithms at all.** (`pipeline/sim2real.py:894`) — manifest.py's `_validate_v3_fields` accepts a manifest with `algorithms: []` and no `component:` (has_blis_algorithm is False), and sim2real.py:894's guard triggers on `"component" not in _mf`. So a baseline-only transfer.yaml prints 'all algorithms are BYO; images are pre-built' though there are zero algorithms. Tightening the guard to also require `_mf.get('algorithms')` would make the message accurate for the shape it fires on.
- **`_cmd_build` docstring 'Prerequisites (checked in order)' no longer reflects the actual order.** (`pipeline/sim2real.py:856`) — sim2real.py:856 lists skopeo as step 1, but the BYO guard at lines 881-900 executes before `check_skopeo` at line 906. A reader relying on the docstring for control flow order will be misled. Prepending a step 0 for the BYO guard would restore accuracy.

PR #503, opened 2026-07-05
step-4 PR 2: sim2real translation register batched-algorithm mode

kalantar:
## Summary

Extends `sim2real translation register` to accept N algorithms in one call via a repeatable `--algorithm <name>=<image-ref>@<config-path>` flag, so the upcoming `/sim2real-bootstrap --byo` (#499) can emit ONE register command for a multi-algorithm BYO run and `sim2real assemble --translation <hash>` continues to consume a single translation dir.

**New:**
- Repeatable `--algorithm NAME=IMAGE@CONFIG` grammar. First-`=` + rightmost-`@` split rule (documented in `pipeline/README.md`).
- Batched translation-hash formula: `sha256(canonical-json(sorted-by-name list of {name, image, config-sha}))`. Order-invariant.
- `registered.json`'s image-ref fields move under a per-algorithm `algorithms` list; the on-read shim in `translation_ref.read_translation_output` continues to normalize legacy step-1 top-level fields, so any pre-existing step-1 shape on disk remains readable.

**Kept working (deprecated):**
- `--algorithm NAME --image REF --config PATH` as the N=1 legacy form. Deprecation warning emitted to stderr. Removal is a step-5+ follow-up.

**Non-goals:** no changes to `sim2real assemble`, `sim2real translate`, `sim2real build`, or the manifest schema (that's PR-1 / #497).

## Notes for reviewers

- The N=1 hash under the new formula differs from N=1 under the step-1 formula (design section on the batched hash formula covers this — no long-lived registrations exist yet, so the break is acceptable).
- Alias policy: N=1 keeps `alias = algorithm_name` (matches step-1). N>1 has `alias = null` — batched translations are referenced by hash. Alias-collision checks still fire per-algorithm regardless of N, so a batched register whose members shadow an existing alias errors with the same shape as single-algo.
- Overlay paths containing `@` cannot be represented (structural consequence of rightmost-`@` split). Overlay paths containing `=` are supported.
- `_build_registered`'s shape has changed (image-refs moved into a per-algo list). No other module writes `registered.json`; the shim only affects reads of `translation_output.json`.

## Docs sweep

- Updated `pipeline/README.md`: rewrote the "Register a translation" section with new `--algorithm NAME=IMAGE@CONFIG` grammar as preferred, deprecated form preserved as legacy, hash formula documented, workspace-artifacts table updated to note batched shape.
- Updated the quick-start "3. Register a translation" example at line ~494 to use the new form and added a cross-reference to the full section.
- `.claude/skills/sim2real-bootstrap/SKILL.md` still shows the old form — its refresh is scoped to PR-4 (#500) per the epic.
- `docs/epics/step-*/design.md` — historical epic design docs left untouched.
- `docs/superpowers/plans/` — historical planning docs left untouched.
- `pipeline/lib/*.py` error-message references to `sim2real translation register` are informational strings, not grammar docs — left as-is.

Closes #498

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -27,7 +27,7 @@ python pipeline/cluster.py provision <cluster_id> --namespaces NS1,NS2,...
 # Per-workspace + per-run cycle:
 python pipeline/setup.py       --experiment-root ../admission-control
 python pipeline/sim2real.py translation register \
-    --algorithm <name> --image <ref> --config <treatment-overlay-path>
+    --algorithm <name>=<image-ref>@<config-path>
 python pipeline/sim2real.py assemble \
     --translation <hash> --cluster <cluster_id> --run <run_name>
 python pipeline/deploy.py      --experiment-root ../admission-control
@@ -168,7 +168,7 @@ python pipeline/sim2real.py translation register \
 **`translation_hash` derivation (BYO, batched):** SHA-256 hex over canonical JSON of a list of `{name, image, config}` dicts sorted by `name`, where `config` is the SHA-256 of the overlay file content and `image` is the digest ref if present (otherwise the raw ref string). Order-invariant — registering the same algorithms in a different order produces the same hash.
 
 ```
-sha256(canonical-json(sorted-by-name list of {name, image_digest_or_ref, config_sha256}))
+sha256(canonical-json(sorted-by-name list of {name, image, config}))
 ```
 
 **Idempotency:** re-registering the same set of (algorithm, image, config-content) tuples — in any order — is a no-op. The existing translation directory is detected, a warning is printed, and exit is 0.
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -178,7 +178,7 @@ sha256(canonical-json(sorted-by-name list of {name, image, config}))
 - `--config` file missing or malformed YAML → exit 2, no writes.
 - Existing translation directory records different algorithm names (hash collision) → exit 2, no writes.
 - `--registered-hash` given and does not match computed → exit 2, no writes.
-- Another translation already owns an `--algorithm` alias with a different hash → exit 2, no writes; re-run with `--force` to reassign (N=1 only).
+- Another translation already owns an `--algorithm` alias (an algorithm name from the batch collides with an existing translation's alias, on any N) → exit 2, no writes; re-run with `--force` to clear the previous owners' aliases before proceeding.
 - Duplicate algorithm names within a single invocation → exit 2, no writes.
 
 ---
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -129,7 +129,7 @@ Pass `--algorithm` once per algorithm. Each value is a single `NAME=IMAGE@CONFIG
 # N=2 example
 python pipeline/sim2real.py translation register \
     --algorithm softreflective=ghcr.io/org/sr-router:v1@overlays/sr.yaml \
-    --algorithm eager=ghcr.io/org/eager-router:v1@sha256:abc123@overlays/eager.yaml \
+    --algorithm eager=ghcr.io/org/eager-router@sha256:<64-hex-digest>@overlays/eager.yaml \
     [--baseline-config path/to/baseline-overlay.yaml] \
     [--registered-hash <expected-sha256-hex>] \
     [--experiment-root PATH]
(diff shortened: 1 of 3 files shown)


kalantar:

Two documentation issues in the README materially misrepresent the shipped behavior of the batched `--algorithm` flag and should be corrected before merge; the code itself is sound and well-tested.

### Important
- **README describes `--force` as "Only meaningful for N=1" but the alias-collision check runs per-algorithm for any N** (`pipeline/README.md:143`) — Verified pipeline/sim2real.py:426-434 loops `for p in prepared` and raises `RuntimeError('already assigned')` on any per-algo collision regardless of batch size; test_batched_alias_collision_on_one_of_N_algos at test_sim2real.py:1237 confirms N=2 registrations trigger the same error. The README flag-table entry at pipeline/README.md:143 is therefore incorrect — an operator running a batched register whose members happen to match an existing alias will hit the error and, per the README, wrongly believe `--force` doesn't apply to them.
- **README N=2 example uses a truncated 6-hex digest that will emit a null-digest warning verbatim** (`pipeline/README.md:132`) — The example at pipeline/README.md:132 uses `@sha256:abc123`, but `_extract_digest_from_ref` at pipeline/sim2real.py:60 requires exactly 64 hex chars — anything shorter returns None and the register command emits `warning: image_digest for 'eager' recorded as null` (verified against the warning loop at sim2real.py:761-767). An operator copying the documented example verbatim gets an unexpected warning; the fix is a full 64-char placeholder or a tag-only example.

### Recommendations
- **Add a test for the algorithm-set mismatch (`ValueError` on hash-collision) path at sim2real.py:404** (`pipeline/tests/test_sim2real.py`) — The defensive guard at pipeline/sim2real.py:404 raising `ValueError('algorithm-set mismatch')` is unreachable by any current test — the old test that exercised it was removed. A future refactor that changes the exception type or drops the check would go undetected. Low risk but easy to plug.
- **`_compute_translation_hash` docstring names input key `config_sha` while the canonical JSON emits key `config`; note the rename to prevent hash-reproduction confusion** (`pipeline/sim2real.py:136`) — Verified at pipeline/sim2real.py:138 (docstring) vs pipeline/sim2real.py:151 (canonical form). The wire-format key is `config`, not `config_sha`. Anyone attempting to reproduce a translation hash from the docstring alone would compute a different hash.

kalantar:
Both important-tier findings from the last review are addressed in `d3419d8`:

- The `--force` flag description previously said "Only meaningful for N=1", which contradicts the per-algorithm collision loop in `_register_translation`. The table entry now correctly explains that `--force` applies per algorithm for any N and only clears the colliding aliases on previous owners (the batched translation itself still has `alias=null`).
- The N=2 example had a truncated 6-hex digest (`@sha256:abc123`), which would have made an operator's paste emit a null-digest warning. Replaced with `<64-hex-digest>` as an unambiguous placeholder.

The two `recommendations` (add a test for the algorithm-set-mismatch guard; note the `config_sha` → `config` key rename in the docstring) are not being applied in this PR — they're follow-up polish.

Tests: 99/99 pass; ruff clean.

kalantar:

A doc line in the failure-modes bullet directly contradicts both the flag-table entry and the code's actual behavior on --force with batched invocations, and should be corrected before merge.

### Important
- **README failure-modes bullet says '--force to reassign (N=1 only)' but --force applies per-algorithm for any N** (`pipeline/README.md:181`) — pipeline/README.md line 181 says '...re-run with --force to reassign (N=1 only)'. The same file's flag-table at line 140 says --force 'Applies per-algorithm for any N', and pipeline/sim2real.py lines 425-434 iterate `prepared` and call `_clear_alias_on(other)` for every colliding algorithm regardless of batch size (verified by test_batched_alias_collision_on_one_of_N_algos). The parenthetical is false and will mislead operators of batched registrations.

### Recommendations
- **The algorithm-set mismatch ValueError guard has no test after the old collision test was removed** (`pipeline/tests/test_sim2real.py`) — The prior test_hash_collision_different_algorithm_errors was deleted and its replacement (test_idempotent_different_order_same_hash) exercises a different scenario. Grepping the new test file for 'algorithm-set mismatch' or 'algorithm name mismatch' returns no hits, so the guard at pipeline/sim2real.py lines 405-409 is unreachable from tests. A regression would only bite an operator with a corrupted workspace, so this is polish rather than blocking.
- **README failure-modes bullet still says '--config file missing' but --config is only the deprecated flag** (`pipeline/README.md:178`) — pipeline/README.md line 178 references `--config` even though in the new form the config path is inside the --algorithm triple and the actual error is 'error: config file not found: {path}'. Reword to something like 'Algorithm config file not found or not valid YAML'.
- **test_happy_path_n1_new_form uses a tag-only image ref but doesn't assert the null-digest warning fires** (`pipeline/tests/test_sim2real.py:603`) — The test only asserts 'deprecated' not in stderr; the positive per-algorithm null-digest warning (pipeline/sim2real.py lines 745-751) is emitted here but never checked. test_digest_ref_no_null_warning only covers the negative cas

kalantar:
Fixed the single important-tier finding from the last review in `b00ba83`. The failure-modes bullet in `pipeline/README.md` still said "re-run with `--force` to reassign (N=1 only)", which contradicts the flag table's "Applies per-algorithm for any N" description and the actual per-algorithm collision loop in `_register_translation`. Reworded the bullet to align.

The three recommendations from this pass (add a test for the algorithm-set-mismatch guard; reword the `--config file missing` bullet to reflect that the config path now lives inside the `--algorithm` triple; add a null-digest positive-case assertion in `test_happy_path_n1_new_form`) are follow-up polish and not applied in this PR.

Tests: 99/99 pass; ruff clean.

PR #515, opened 2026-07-07
pairkey: parser module + _load_pairs iteration integration (#509)

kalantar:
Closes #509. Foundation PR (PR 1) for the step-5 epic ([#502](https://github.com/inference-sim/sim2real/issues/502)).

## Scope

- **New module** `pipeline/lib/pairkey.py`:
  - `PairKeyParts` frozen dataclass with `workload`, `package`, `iteration`, and `to_key()` (canonical `wl-<workload>|<package>|iN`).
  - `parse_pair_key(key)` — strict per the design grammar; legacy keys without `|iN` suffix parse as `iteration=1`; malformed keys raise `ValueError` with a message naming the offending input.
  - `parse_iteration_spec(spec)` — list + range syntax (`"2"`, `"1,3"`, `"1-3"`, `"1,3-5"`); rejects `0`, negative, reversed ranges, non-integer tokens, and leading zeros.
- **`pipeline/deploy.py`**:
  - `_is_pair_key` unchanged (metadata-exclusion gate).
  - `_load_pairs` refactored to a thin wrapper over `_load_pairs_with_errors` (returns just the pairs dict).
  - `_load_pairs_with_errors(cluster_dir)` → `(pairs, malformed_count)`. Each entry whose filename-derived key parses under the grammar gains an `iteration` field.
- **Tests**: `pipeline/tests/test_pairkey.py` (60 tests) and `pipeline/tests/test_load_pairs.py` (10 tests). CI workflow and CLAUDE.md updated to include both.

`assemble`, `pipeline.yaml`, and user-facing CLI flags are untouched — those land in PRs 2, 3, 4.

## Deviation from the design's literal loader-layer policy — reviewer attention

The step-5 design [§Strand 1](https://github.com/inference-sim/sim2real/blob/refactor/v2-step-5/docs/epics/step-5/design.md#strand-1--pair-key-parser-rework) says `_load_pairs` should drop malformed keys with a WARN and return only valid pairs. Applied literally to the current tree, this would drop every existing pair-key today: `assemble` currently writes `pipelinerun-<wl>-<pkg>.yaml` (dash-separated), which produces keys like `wl-smoke-baseline` that fail the new pipe-separated grammar. Dropping them would break every downstream `deploy.py` command and every test in `test_deploy_run.py`, `test_deploy_pairs.py`, `test_deploy_reset.py`, etc., before PR 2 lands.

This PR keeps unparseable entries in the returned dict (without an `iteration` field) and reports the count via `_load_pairs_with_errors`. Once PR 2 reshapes `assemble` filenames to the new grammar and existing runs are re-assembled, `malformed_count` should trend to zero and the drop-with-WARN semantics from the design can be tightened without behavior change. Rationale is documented inline on `_load_pairs_with_errors`. No per-key WARN is emitted, since it would fire on every legacy key and read as noise until PR 2 lands.

Alternative considered and rejected: update `assemble` in this PR to write pipe-separated filenames. The design explicitly puts that in PR 2 and the epic ordering commits to landing PR 1 first, so I stayed in scope.

## Verification

```
ruff check pipeline/ .claude/skills/ --select F           # All checks passed
python -m pytest pipeline/ ...                            # 1508 passed, 2 xfailed
```

The full CI-equivalent test list (matching `.github/workflows/test.yml`) passes locally, including the 70 new tests.

## Stale-reference sweep

Grepped `pairkey`, `PairKeyParts`, `parse_pair_key`, `parse_iteration_spec`, `_load_pairs_with_errors` across `**/*.md`, `docs/`, `.claude/skills/`, and `README*`. Only hits are inside `docs/epics/step-5/design.md` (source-of-truth spec — I did not retrofit it) and this PR's own files. `pipeline/README.md`'s pair-key section will be updated in PR 6 per the epic's ordering; nothing referring to a pre-parser world goes stale from PR 1.

## Base

Base: `refactor/v2-step-5` (per the epic's `Base:` line).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,6 +38,8 @@ jobs:
             pipeline/tests/test_translation_ref.py \
             pipeline/tests/test_translate.py \
             pipeline/tests/test_build.py \
+            pipeline/tests/test_pairkey.py \
+            pipeline/tests/test_load_pairs.py \
             .claude/skills/sim2real-analyze/tests/ \
             .claude/skills/sim2real-bootstrap/tests/ \
             .claude/skills/sim2real-translate/tests/ \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,6 +144,8 @@ python -m pytest pipeline/ \
   pipeline/tests/test_translation_ref.py \
   pipeline/tests/test_translate.py \
   pipeline/tests/test_build.py \
+  pipeline/tests/test_pairkey.py \
+  pipeline/tests/test_load_pairs.py \
   .claude/skills/sim2real-analyze/tests/ \
   .claude/skills/sim2real-bootstrap/tests/ \
   .claude/skills/sim2real-translate/tests/ \
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -51,6 +51,7 @@ def _c(code: str, text: str) -> str:
 
 from pipeline.lib import build, cluster_ops, layout
 from pipeline.lib.log import info, ok, warn, err
+from pipeline.lib.pairkey import parse_pair_key
 from pipeline.lib.redact import redact_yaml_file, redact_yaml_tree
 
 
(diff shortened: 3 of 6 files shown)


kalantar:

Foundational parser module lands cleanly; the WARN/drop deviation flagged by three specialists is explicitly documented and reasoned in both the PR body and the loader's docstring, and the remaining items are minor polish that does not warrant a review comment.

PR #520, opened 2026-07-07
deploy: --iteration filter (list+range) on filter-aware subcommands (#512)

kalantar:
## Summary

Adds `--iteration` argparse entries to `run`, `status`, `collect`, `reset`, and `wipe`, wired through `_apply_run_filters` using `parse_iteration_spec` from `pipeline/lib/pairkey.py` (landed in #509 via #515). Composes with existing `--workload` / `--package` / `--status` / `--only` under AND semantics.

Closes #512. Part of epic #502.

## What changed

- **`pipeline/deploy.py`**:
  - Import `parse_iteration_spec` alongside `parse_pair_key`.
  - New module-level helper `_key_iteration(key)` — parses key iteration; legacy dash-shape keys collapse to iteration=1 (matches the step-5 design table row for the mid-rollout state where PR 2's canonical filename layout hasn't landed).
  - `_apply_run_filters`: parses `--iteration` up-front; malformed input aborts with a clear error before other filter work. Applies as an intersection filter in both the `--only` branch and the general branch; included in the "no filters → empty set" gate.
  - `_resolve_scope`: includes `iteration` in `filters_given`.
  - `_report_filter_mismatch`: surfaces `--iteration '<spec>'` in the mismatched-parts list and prints a `Valid --iteration values` line.
  - `_cmd_status`'s empty-progress path: includes iteration in its `filters_given` diagnostic.
  - `_cmd_collect`: `--iteration` alone now triggers the scoped path; threaded through `_ScopeArgs`.
  - `--iteration SPEC` added to the five subparsers.

- **Tests**:
  - `pipeline/tests/test_deploy_run.py`: 12 new unit tests on `_apply_run_filters` — single/list/range/mixed syntax, composition with `--workload` / `--package` / `--status` / `--only`, malformed input (`0`, `5-1`, `abc`), and legacy dash-shape → i1 semantics. Plus 4 integration tests on `_cmd_status` including the one named in the issue (`status --iteration 2,3`).
  - `pipeline/tests/test_deploy_collect.py`: 3 integration tests on `_cmd_collect` including the one named in the issue (`collect --iteration 1-3` narrows the extract target set) plus a no-match abort and a malformed-input abort.

## Notes for reviewers

- **Legacy dash-shape keys** (e.g. `wl-smoke-baseline`, produced by pre-PR-2 assemble) do not parse under the new grammar. The design's step-5 table row specifies these should be treated as implicit i1 for filter purposes. `_key_iteration` implements that: catches `ValueError` and returns 1. Once PR 2 lands and runs re-assemble, this fallback becomes a no-op.
- **`--only` composition**: the issue calls out AND semantics with `--only`. Implemented by applying the iteration filter as a secondary narrowing step on the `--only`-resolved set. Existing `--only` behavior (short-circuits `--workload`/`--package`/`--status`) is preserved for those three; only `--iteration` now further narrows an `--only` selection.
- **Docs sweep**: `pipeline/README.md`'s flag tables list `--workload` / `--package` / `--status` but not `--iteration` yet. The epic (#502) puts docs in PR 6 (#514), which lands last across all step-5 code. Not updated here.
- No changes to `pipeline.yaml`, `assemble`, or the `iN/` result-path model — those are PR 2 (#510) and PR 3 (#511).

## Base

`refactor/v2-step-5`. Dependency #509 satisfied by PR #515 already merged into this base.

## Test plan

- [x] `python -m pytest pipeline/ -v` — 1438 passed
- [x] CI-shape run (`pipeline/` + selected skill tests) — 1527 passed
- [x] `ruff check pipeline/ .claude/skills/ --select F` — all checks passed
- [x] Named integration tests exercise `status --iteration 2,3` and `collect --iteration 1-3`
- [x] Malformed input (`--iteration 0`, `5-1`, `abc`) exits with code 1 and a clear message

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -51,13 +51,31 @@ def _c(code: str, text: str) -> str:
 
 from pipeline.lib import build, cluster_ops, layout
 from pipeline.lib.log import info, ok, warn, err
-from pipeline.lib.pairkey import parse_pair_key
+from pipeline.lib.pairkey import parse_iteration_spec, parse_pair_key
 from pipeline.lib.redact import redact_yaml_file, redact_yaml_tree
 
 
 def _is_pair_key(key: str) -> bool:
     """Return True if key is a real pair entry (not metadata)."""
     return not key.startswith("_")
+
+
+def _key_iteration(key: str) -> int:
+    """Return the iteration number encoded in a pair key.
+
+    Canonical grammar (``wl-<workload>|<package>[|i<N>]``) yields the
+    parsed ``iteration``; keys without an ``|iN`` suffix parse as ``1``.
+    Legacy dash-shape keys that predate the step-5 grammar (e.g.
+    ``wl-smoke-baseline`` from a pre-PR-2 run) do not parse; they map to
+    ``1`` here to match the design's step-5 semantics — the
+    single-replica shape reads as implicit ``i1``. ``--iteration 1``
+    therefore includes legacy pairs and higher values exclude them,
+    which is what the design table specifies for the mid-rollout state.
+    """
+    try:
+        return parse_pair_key(key).iteration
+    except ValueError:
+        return 1
 
 
 def _format_capacity(effective: int, probed: int, reserved: int,
@@ -684,6 +702,7 @@ def _cmd_status(args, run_dir: Path,
             getattr(args, "workload", None) is not None,
             getattr(args, "package", None) is not None,
             getattr(args, "status", None) is not None,
+            getattr(args, "iteration", None) is not None,
         ])
         if filters_given:
             suffix += " — filters ignored"
@@ -1499,14 +1518,16 @@ def _cmd_collect(args, run_dir: Path, cluster_config: dict):
             "cluster.")
         sys.exit(1)
 
-    # ── Pair-level scoping (--only / --workload / --package) ─────────────
+    # ── Pair-level scoping (--only / --workload / --package / --iteration) ─
     scope_only = getattr(args, "only", None)
     scope_workload = getattr(args, "workload", None)
     scope_package = _parse_list(getattr(args, "package", None))
-    scoped = scope_only is not None or scope_workload is not None
+    scope_iteration = getattr(args, "iteration", None)
+    scoped = (scope_only is not None or scope_workload is not None
+              or scope_iteration is not None)
 
     if scoped and not progress:
-        err("--only/--workload require progress data to resolve pairs, but none was found.")
+        err("--only/--workload/--iteration require progress data to resolve pairs, but none was found.")
         sys.exit(1)
 
     if scoped and progress:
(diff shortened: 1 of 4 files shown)


kalantar:

**Summary:** Remote-mode `deploy.py run` silently drops `--iteration` because `_collect_run_flags` was not extended when the new flag was added.

### must_fix

- **`pipeline/deploy.py:3165` — `_collect_run_flags` does not forward `--iteration` to the in-cluster orchestrator Job** (source lens: pr-test-analyzer)
  - The tuple at `deploy.py:3165` iterates only `("only", "workload", "package", "status")`. `run_p` gets `--iteration` added at line 3511, and `_cmd_run_remote` calls `_collect_run_flags(args)` at line 3421 to build the flag list for the Job. `deploy.py run --remote --iteration 2` therefore launches a Job with no `--iteration` argument and runs every iteration — a silent contract violation on the `run` subcommand named in the issue's acceptance criteria. No test covers the remote-mode forwarding path for `iteration`.

### important

_(none)_

### recommendations

- **`pipeline/deploy.py:2293` — `Valid --iteration values:` uses `','.join(...)` while the three preceding lines use `', '.join(...)`** (source lens: code-reviewer)
  - Lines 2290-2292 print workloads/packages/statuses with a comma-space separator; line 2293 prints iteration numbers without the space. The mismatch report output ends up visually inconsistent for a one-character fix.

---

_Posted manually after the workflow's post step failed on macOS `base64 -d` (BSD base64 uses `-D`). Verdict content is verbatim from the synthesizer._

kalantar:
Addressed the review. The blocker was real: `_collect_run_flags` iterates a hard-coded tuple of flags to forward to the remote Job, and `--iteration` wasn't in the tuple — so `deploy.py run --remote --iteration N` would silently ignore the filter. Added `"iteration"` to that tuple and switched the getattr call to use a default so callers with a partial Namespace don't need to enumerate the new attr. Three new tests in `test_deploy_remote.py` cover the forwarding path: a single iteration, a range verbatim, and absent-by-default.

Also fixed the cosmetic separator inconsistency in `_report_filter_mismatch` — the `Valid --iteration values:` line now uses `, ` to match the three sibling lines above it.

Full test suite: 1441 passed. Lint clean.

kalantar:

Implementation cleanly satisfies the acceptance criteria; only one minor comment polish nit surfaced during verification.

### Recommendations
- **Inline comment in test_apply_run_filters_iteration_mixed reads as if the fixture only has i1 and i3** (`pipeline/tests/test_deploy_run.py:923`) — _ITER_PROGRESS at lines 854-863 contains i1, i2, and i3 keys. The comment 'Only i1 and i3 exist in fixture; i4/i5 don't exist so they're just absent.' can be read as a claim about the fixture contents, which is false. Rephrasing to 'spec {1,3,4,5} keeps i1/i3; i2 is excluded by spec, i4/i5 not in fixture' would remove the ambiguity for future maintainers.

PR #521, opened 2026-07-07
assemble: additive-merge + --replicas consumption (grow-only) (#510)

kalantar:
Closes #510.

## Summary

Grows `pipeline/lib/assemble_run.py` from step-1's single-replica shape to
`--replicas N` support with the additive-merge (grow-only) invariant.

- New `--replicas N` argparse arg on `sim2real assemble` (default: 1).
- `manifest.assembly.yaml` gains a `replicas: N` top-level field.
- `params_hash` excludes `replicas` so bumps don't trip drift detection.
- PipelineRun filename shape becomes `pipelinerun-<workload-safe>|<package>|i<N>.yaml`.
- k8s `metadata.name` gains `-i<N>` suffix (always).
- Existing-run decision tree: legacy detection, drift detection, grow-only
  guard, no-op idempotence, additive-merge write.

## Scope decisions (from design doc)

1. **Unified pipe shape.** No `--replicas` defaults to `--replicas 1`,
   which emits `|i1`. Legacy hyphen shape is not preserved as a first-class
   output (rationale: refactored code hasn't shipped; unifying now avoids
   forked shapes downstream).
2. **`manifest.assembly.yaml.replicas` is the source of truth** for current
   replica count on re-assemble (single source; matches step-0's schema).
3. **`params_hash` excludes `replicas`** so drift detection is orthogonal
   to replica-count changes. Preserves "drift detection unchanged" from the
   design.
4. **`_load_pairs` untouched.** Pipe-shape filenames naturally produce
   canonical grammar-matching keys via its existing filename-stem logic.
   No changes needed for PR 2.
5. **`--force` bypasses drift but NOT grow-only.** Follows the design doc's
   "No flag bypasses grow-only in step-5". Shrink is tracked in #506.

## Stale-reference sweep

- `pipeline/lib/assemble_run.py` — `generate_pipelineruns` docstring updated
  to reflect pipe shape.
- `pipeline/README.md` — old filename shape referenced in one place; leaving
  unchanged per epic PR 6 (docs PR) ordering.
- `pipeline/deploy.py:_load_pairs` — "deviation" comment about legacy shape
  is still accurate during the rollout; left unchanged.

## Design doc

See `docs/superpowers/specs/2026-07-07-issue-510-assemble-replicas-design.md`
for full decision-tree table and error messages.

## Base

Base: `refactor/v2-step-5`. Parent epic: #502.


kalantar:

The grow-only invariant has a real bypass path when drift, --force, and a shrink request coincide, and two smaller issues introduced by this PR (comment renumbering, unwrapped parse calls) should be cleaned up in the same revision.

### Must fix
- **--force + content drift + replicas < prior_replicas silently rebuilds at the smaller replica count, bypassing the grow-only guard** (`pipeline/lib/assemble_run.py:662`) — Verified at pipeline/lib/assemble_run.py:662-682: the drift branch (line 662) runs shutil.rmtree(run_dir) at line 668 when force=True, then falls through past the block to the fresh-assemble path — which recreates the run at whatever `replicas` was passed. The shrink guard at line 670 is only reached inside the else-branch (no drift). The spec at docs/superpowers/specs/2026-07-07-issue-510-assemble-replicas-design.md:21 and the decision table at line 83 explicitly state '--force does NOT bypass' the grow-only guard, and test_assemble_replicas.py:98 only exercises the no-drift + no-force shrink path, so this regression is untested.

### Important
- **Section number labels 4 and 5 each appear twice in assemble_run's body, defeating the numbered navigation** (`pipeline/lib/assemble_run.py:740`) — Verified at pipeline/lib/assemble_run.py: this PR inserted new sections '# 3.' (line 627), '# 4. Load translation index' (line 702), and '# 5. Snapshot' (line 733) but the pre-existing '# 4. Resolve scenarios' (line 740), '# 5. Inject hf secret' (line 795), '# 6.' (line 802), '# 7.' (line 805), '# 8.' (line 841) were not renumbered. The remaining five comments should shift to 6-10.
- **Decision-tree reads run_metadata.json and manifest.assembly.yaml with bare json.loads / yaml.safe_load, so a corrupt workspace file escapes _cmd_assemble's AssembleError-only handler as a raw traceback** (`pipeline/lib/assemble_run.py:641`) — Verified at pipeline/lib/assemble_run.py:641-642 — both parses run without try/except. pipeline/sim2real.py:1164 catches only _assemble_run_lib.AssembleError from assemble_run, matching the pattern the rest of the module already follows (see the ManifestError wrap at assemble_run.py:622-625 and the translation_output.json wrap at 703-708). An interrupted prior write or hand-edit produces json.JSONDecodeError / yaml.YAMLError that surfaces as an unactionable Python traceback instead of a message pointing at --force.

kalantar:
Addressed the three blocking findings in d82d2d6:

- **Grow-only ordering (must_fix):** moved the shrink guard to run before the drift comparison, so `--force` cannot bypass it when drift and a shrink request coincide. Added `test_shrink_with_force_and_drift_still_refuses` to pin the invariant.
- **Section-number renumbering:** shifted the pre-existing `# 4..# 8` comments to `# 6..# 10` so the numbered navigation is monotonic through the whole function.
- **Corrupt state file handling:** wrapped the `yaml.safe_load` / `json.loads` reads of `manifest.assembly.yaml` and `run_metadata.json` in try/except that re-raises as `AssembleError` (or rmtree+rebuild under `--force`). Added three tests: corrupt manifest refuses without force, corrupt run_metadata refuses without force, corrupt state with `--force` rebuilds cleanly.

Full pipeline suite: 1449 passed, 2 xfailed. Ruff clean.

Deferred (from the non-blocking findings) — should land in later PRs of the step-5 epic:
- `pipeline/deploy.py:_discover_phases` still uses hyphen-split extraction that produces wrong output for pipe-shape filenames. Falls back-path only; PR 3 owns deploy.py.
- Additional test-coverage suggestions for `run_metadata.json[replicas]` and `resolve_run()['replicas']`.
- Stale reference in `pipeline/deploy.py:_load_pairs_with_errors` ("PR 2 has not yet landed" clause).
- Plan/spec cosmetic inconsistencies (test count, signature default notation).

kalantar:
Extracted the shared scenario-resolution pipeline into a private `_resolve_packages` helper (returns a typed `_ResolvedPackages` NamedTuple). Both `assemble_run` and `_additive_grow` dispatch through it, so the three previously-dropped AssembleError guards — `bundle_path is None`, `base_name not in resolved_baselines`, and the `read_translation_output` try/except — now live in exactly one place and cover the grow path too.

Also folded in the unbuilt-algo check (needed before `inject_image_tag` runs, which would otherwise crash on a null image_ref, escaping the AssembleError boundary) and dropped the dead `cluster_id` param from `_additive_grow`. The numbered section comments in `assemble_run` renumber to 1..8 to reflect the collapse.

Added two regression tests on the grow path: deleted scenario file → clean AssembleError; corrupt translation_output.json → clean AssembleError.

Full suite: 1451 passed, 2 xfailed. Ruff clean.

Non-blocking recommendations from the iteration-2 review remain deferred to the epic docs PR and later cleanup PRs: `pipeline/README.md` staleness, `_additive_grow` run_metadata.json try/except, `assemble_run` docstring step-list refresh, `test_grow_from_3_to_5_*` could also assert `run_metadata.json[replicas]==5`.

kalantar:

Solid additive-merge implementation with 18 dedicated tests exercising every branch of the decision tree; remaining specialist findings are polish, not blockers.

### Recommendations
- **Module docstring is stale in two ways after --replicas lands: still says 'one PipelineRun per (workload, package)' and 'params_hash over the assembly-slice bytes'.** (`pipeline/lib/assemble_run.py:6`) — Verified at lines 4-8: the header still describes the pre-replicas emit shape and the pre-canonicalization hash. compute_params_hash at lines 252-266 now parses YAML, pops replicas, and hashes sort_keys=True canonical form, and generate_pipelineruns loops over an `iterations` range producing N files per pair. Anyone reading the module intro will form the wrong mental model.
- **test_grow_from_3_to_5 asserts manifest.assembly.yaml['replicas']==5 but never reads run_metadata.json on the additive-grow path.** (`pipeline/tests/test_assemble_replicas.py:67`) — Verified: the test at lines 67-96 only checks manifest.assembly.yaml. _additive_grow at assemble_run.py:659-663 also rewrites run_metadata.json['replicas'] and 'assembled_at' — only exercised by the force-rebuild path (test_drift_with_force_rmtree_rebuilds), not by the additive-grow code path itself. A regression in the JSON update would slip through.

PR #522, opened 2026-07-07
pipeline.yaml: thread replica param + PipelineRun name validator (#511)

kalantar:
Closes #511.

Part of epic #502 (step-5 — Replicas + iteration filtering). PR 3 of 6.

## Summary

Threads a `replica` PipelineRun param through `pipeline.yaml` so each iteration's results land under `runs/<R>/results/<phase>/<workload>/i<N>/`, and adds a fail-fast PipelineRun name-length validator. This unlocks step-5's N-replica dispatch (already possible via #521's assemble) writing to per-iteration directories, which #513 (`sim2real-check` iN/ traversal) then consumes.

## Changes

- **`pipeline/lib/tekton.py`**
  - New `RESULTS_DIR_TEMPLATE` constant + `build_results_dir(run, phase, workload, replica)` helper — single source of truth for the resultsDir template shape.
  - New `validate_pipelinerun_name(name)` — raises `ValueError` if `len(name) > 253` (RFC 1123 DNS-subdomain limit). Called from `make_pipelinerun_scenario` at PR-name construction, so an over-long run/workload combination fails at assemble time, not at Tekton admission.
  - `make_pipelinerun_scenario` now emits a `replica` PipelineRun param whose value is `str(iteration)` (Tekton has no int type).

- **`pipeline/pipeline.yaml`**
  - Adds pipeline-level `replica` param (`type: string`, `default: "1"`).
  - Threads `$(params.replica)` into `resultsDir` for every task that writes to results: `prepare-results-dir`, `stream-epp-logs`, `stream-gpu-stats`, `run-workload-blis-observe-binary`, `collect-results`. `resultsDir` shifts from `<run>/<phase>/<workload>` to `<run>/<phase>/<workload>/i<N>`.

- **Docs sweep** — `pipeline/README.md` and `CLAUDE.md` results-path references updated for the new `iN/` layer. Comprehensive doc coverage of `--replicas` / `--iteration` / additive-assemble semantics remains #514's scope (step-5 PR 6).

## Design decisions

1. **`replica` param baked into the PR YAML at assemble time**, not injected at dispatch. Each iteration already has its own pre-generated `pipelinerun-*.yaml` (from #521's assemble). `namespace` is dynamic per slot and must be rewritten at dispatch; `replica` is static per file and belongs in the YAML. This meets the issue's "pass `replica` as a PipelineRun param at dispatch" criterion trivially — the YAML that gets `kubectl apply`'d already carries it — and keeps `deploy.py:_cmd_run` untouched for the replica concern.

2. **`plans_dir` (pipeline.yaml line 100) stays per-workload.** `plans_dir` is the llm-d-benchmark plans directory (input), not results (output). Iterations reuse the same plans; no need to duplicate.

3. **`_apply_run_filters` iteration term** — already delivered in #520 (PR 4 landed before PR 3). No change needed.

4. **`_cmd_wipe` / `_cmd_collect` local paths unchanged.** `_cmd_collect` rsyncs the PVC tree transparently — new `iN/` layer appears in the destination without code change. `_cmd_wipe` deletes the whole `(pkg, wl)` subtree, which still works (removes all `iN/` under it). Per-iteration wipe (`wipe --iteration N` narrowing wipe to a specific iteration's subdir) is a UX concern deferred to a follow-up.

## Tests

- `pipeline/tests/test_tekton.py` — new tests for `build_results_dir`, `validate_pipelinerun_name`, oversized-name rejection, and the `replica` param on `make_pipelinerun_scenario` output.
- `pipeline/tests/test_pipeline_yaml.py::TestReplicaParamThreading` — asserts pipeline declares the `replica` param + every `resultsDir` in `pipeline.yaml` matches `RESULTS_DIR_TEMPLATE`. This is the grep-audit-in-code that catches drift when the YAML is edited by hand.

Full suite: **1487 passed, 2 xfailed** (no regressions). Ruff pyflakes clean.

## Acceptance checklist

- [x] N-replica assemble → N× PipelineRuns dispatched per (workload, phase) — already delivered by #521; each PR YAML now also carries `replica`.
- [x] Results land under `runs/R/results/<phase>/<workload>/iN/` — pipeline.yaml threading.
- [x] PR name > 253 chars → `validate_pipelinerun_name` raises before Tekton is contacted.
- [x] Grep audit of `resultsDir` in `pipeline.yaml` — captured in `TestReplicaParamThreading::test_every_results_dir_matches_canonical_template` (walks every task).
- [x] `replica` param dispatched as a string, not an int (Tekton API compat).
- [x] Rendered-pipeline test passes.
- [ ] Integration test: full 3-replica run against a real cluster. **Deferred** — no live-cluster fixture in CI. The rendered-pipeline test + assemble unit tests cover the shape end-to-end at unit level.

## Doc sweep

Grepped `pipeline/README.md`, `CLAUDE.md`, `docs/` for stale results paths. Updated:
- `pipeline/README.md` — collect destination, step-1 BYO success criterion, step-2 skill-driven success gate.
- `CLAUDE.md` — gpu_logs row in workspace-artifacts table.

Left untouched: step-1/step-2 design docs and proposals (historical; describe pre-step-5 world), the `deploy.py wipe` line (still functionally accurate at the `(pkg, wl)` level).

## Base

Merges into `refactor/v2-step-5` (epic branch), not `main`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,6 +41,7 @@ jobs:
             pipeline/tests/test_build.py \
             pipeline/tests/test_pairkey.py \
             pipeline/tests/test_load_pairs.py \
+            pipeline/tests/test_collect_internals.py \
             .claude/skills/sim2real-analyze/tests/ \
             .claude/skills/sim2real-bootstrap/tests/ \
             .claude/skills/sim2real-translate/tests/ \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,7 +112,7 @@ All artifacts live under `<experiment-root>/workspace/` (gitignored). When no `-
 | `runs/<run>/cluster/<algo>.yaml` | `sim2real assemble` | `deploy.py` |
 | `runs/<run>/cluster/pipelinerun-*.yaml` | `sim2real assemble` | `deploy.py run` |
 | `runs/<run>/results/{phase}/` | `deploy.py collect` | `/sim2real-analyze` skill, `deploy.py wipe` |
-| `runs/<run>/results/{phase}/<workload>/gpu_logs/<node>.log` | `deploy.py collect` (pulled from PVC) | analysis / debugging |
+| `runs/<run>/results/{phase}/<workload>/i<N>/gpu_logs/<node>.log` | `deploy.py collect` (pulled from PVC) | analysis / debugging |
 | ConfigMap `sim2real-progress-{run}` | `deploy.py run`, `deploy.py reset` | All `deploy.py` subcommands |
 | `runs/<run>/plans/<phase>/<workload>/` | `deploy.py run` | workload tasks |
 
--- a/docs/superpowers/plans/2026-07-07-issue-511-pipeline-replica-threading.md
+++ b/docs/superpowers/plans/2026-07-07-issue-511-pipeline-replica-threading.md
@@ -1,0 +1,536 @@
+# Issue #511 Implementation Plan: `pipeline.yaml` threading + `_cmd_run` dispatch + PipelineRun name validator
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Thread a `replica` PipelineRun param through `pipeline.yaml` so each iteration's results land under `runs/<R>/results/<phase>/<workload>/i<N>/`, and add a fail-fast PipelineRun name-length validator.
+
+**Architecture:** Bake the `replica` param into each PipelineRun YAML at assemble time (via `make_pipelinerun_scenario`), rather than injecting at dispatch. Each iteration already has its own pre-generated PR YAML file (from #521), so passing `replica` as a static param in that file is the natural fit — dispatch stays a simple `kubectl apply` of the pre-generated file. Add a `build_results_dir(...)` helper as the single source of truth for the resultsDir template contract, used by `pipeline.yaml`'s test to assert every `resultsDir` value matches. Add `validate_pipelinerun_name(...)` called from `make_pipelinerun_scenario` so an over-long name fails at construction time (assemble), not at Tekton dispatch.
+
+**Tech Stack:** Python 3.10+, PyYAML, pytest, Tekton v1 API (string params).
+
+## Global Constraints
+
+- **Tekton param type**: `replica` is `type: string` with `default: "1"` — Tekton has no int type. Value must be emitted as a Python `str` in the PipelineRun params list.
+- **k8s DNS-label limit**: PipelineRun `metadata.name` must be ≤253 chars (RFC 1123 DNS subdomain). Enforce at construction; do not defer to Tekton rejection.
+- **Legacy compat**: `iteration=1` (default) must produce identical PR YAML output shape as before (aside from the new `replica: "1"` param), so single-replica flows continue to work.
+- **CI paths listed in `.github/workflows/test.yml`**: any new test file must be added there — CI only covers explicitly listed paths.
+- **Path discipline**: every file operation targets absolute paths under `.claude/worktrees/issue-511-pipeline-yaml-replica-threading/`.
+
+## Design decisions locked in
+
+1. **`replica` param injected at assemble time** (in `make_pipelinerun_scenario`), not at dispatch. Rationale: each iteration has its own PR YAML file, `replica` is static per file (unlike `namespace`, which is dynamic per slot). Baking it in keeps dispatch (`_cmd_run`) untouched for the replica concern. Acceptance criterion "replica dispatched as a string, not an int (Tekton API compat)" is trivially met — the YAML emits `"1"`.
+2. **`plans_dir` (line 97 of pipeline.yaml) stays per-workload, not per-iteration.** Rationale: `plans_dir` is llm-d-benchmark input (workload plans), not benchmark results output. Iterations reuse the same plans; no need to duplicate.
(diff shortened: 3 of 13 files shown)


kalantar:

The core pipeline.yaml threading and validator are correct, but two silent regressions in deploy.py collect and a mishandled ValueError at the assemble CLI boundary should be resolved (or explicitly filed as follow-ups) before merge.

### Important
- **`_extract_phases_from_pvc` --skip-logs path silently produces empty results under the new iN/ layout** (`pipeline/deploy.py:1233`) — deploy.py:1233-1262 kubectl-cps explicit filenames at /data/{run}/{phase}/{wl}/{fname}, but this PR moves them under {wl}/iN/. Every cp fails; the `and "no such file" not in r.stderr.lower()` guard at 1237/1245/1253/1261 swallows the error and reports success. The PR body claims collect works transparently — for --skip-logs, that claim is silently false.
- **mtime probe keys results by the wrong directory, silently disabling the incremental collect optimization the README still advertises** (`pipeline/deploy.py:1074`) — _probe_remote_mtimes at deploy.py:1074 does `Path(parts[1]).parent.name`. After this PR, `find` returns paths like /data/{run}/{phase}/{wl}/iN/trace_data.csv — the parent is `iN`, not the workload. remote_mtimes.get(wl_name) always returns None, so _is_up_to_date always returns False and every collect re-downloads the entire workload subtree. README.md:389 (edited in this PR) still asserts "Repeated collects are incremental", which is now silently untrue.
- **validate_pipelinerun_name's ValueError escapes the AssembleError contract, so an oversized name yields a raw traceback instead of the intended clean CLI error** (`pipeline/sim2real.py:1164`) — tekton.py:37-48 raises ValueError; assemble_run.py:693 docstring states "Raises AssembleError on any validation failure"; sim2real.py:1164 only catches AssembleError before printing `error: {exc}`. The whole point of the validator is user-facing fail-fast; the current wiring undermines that goal. One-line fix (wrap at generate_pipelineruns, or widen the except).

kalantar:
Addressed the three blocking findings from the automated review.

**Wrapped the tekton validator's `ValueError` at the assemble boundary** (`pipeline/lib/assemble_run.py`). `validate_pipelinerun_name` raises `ValueError` when a PR name exceeds 253 chars, and `sim2real.py`'s assemble command only catches `AssembleError` — so an oversized name landed as a raw Python traceback instead of the intended clean `error: ...` message. `generate_pipelineruns` now converts to `AssembleError` at the `make_pipelinerun_scenario` call site, and there's a test asserting the wrap holds.

**Fixed `_probe_remote_mtimes` keying** (`pipeline/deploy.py`). It walked `Path.parent.name` — the direct parent of `trace_data.csv`, which is now `i<N>` rather than the workload. Every `remote_mtimes.get(wl_name)` returned `None`, silently disabling incremental collect. The probe now walks two levels up (to the workload dir) and keeps the max mtime across iterations.

**Split `_is_up_to_date` into two helpers**. The old single-file check is preserved for `_extract_phase_plans` (plan YAMLs, unchanged use case). A new `_is_workload_up_to_date` scans every `i<N>/trace_data.csv` under a workload dir and returns True only when the oldest local iteration is at least as new as the newest remote copy. The three workload-level call sites use the new helper.

**Rewrote the `--skip-logs` branch of `_extract_phases_from_pvc`** to discover `i<N>/` subdirs per workload via a second-level `ls`, then copy trace files, `epp_logs/`, `gpu_logs/`, and `resources/` into `dest_dir/wl_name/i<N>/`. The old explicit-filename cp at `.../{wl}/{fname}` no longer matched anything on the PVC after this PR's pipeline.yaml change, and the `"no such file"` guard was swallowing the failure.

New file: `pipeline/tests/test_collect_internals.py` (10 tests covering the probe and workload-level freshness check). Existing `test_deploy_collect.py` fixtures updated for the `i<N>/` layout — the mocked `ls` responses now return `i1` at the second-level walk, and the stale-log-dir wipe test checks under `i1/`. Added the new file to `.github/workflows/test.yml`.

Full suite: 1499 passed, 2 xfailed. Ruff clean.

kalantar:

Acceptance criteria met with only doc/test-hygiene polish items surviving verification.

### Recommendations
- **_extract_phases_from_pvc docstring still shows the pre-step-5 flat PVC layout** (`pipeline/deploy.py:1148`) — Read deploy.py:1148-1156: the Data-layout block lists /data/{runName}/{phase}/{workloadName}/trace_data.csv (no iN layer) and the skip-line still says 'skipping those whose local trace_data.csv is already up to date'. The code below (lines 1268-1379) writes and probes the iN/ layer. A future reader will get a wrong mental model of the on-PVC contract.
- **Three test_deploy_collect.py tests still patch deploy._is_up_to_date, which _extract_phases_from_pvc no longer calls** (`pipeline/tests/test_deploy_collect.py:1824`) — Verified in deploy.py that all workload-level freshness checks in _extract_phases_from_pvc now go through _is_workload_up_to_date (lines 1269, 1385, 1437); _is_up_to_date only survives inside _extract_phase_plans (line 1582). The monkeypatches at test_deploy_collect.py:1824, 1877, 1923 patch the wrong symbol and are silently dead — the code path is only reached because _probe_remote_mtimes is stubbed to {}.
- **_is_workload_up_to_date docstring claims a 'missing-iterations-locally → False' guarantee the code does not enforce** (`pipeline/deploy.py:1121`) — Read deploy.py:1114-1137: the function only globs local i*/trace_data.csv and compares min(local mtimes) >= remote_mtime; it never counts iterations against the remote. The docstring's 'If some iterations are missing locally OR any local iteration is older than remote, returns False' overstates the contract. Practically safe because kubectl cp preserves mtimes so a re-dispatched iN bumps remote_mtime above any stale local iN, but the docstring should not promise what the code doesn't check.

PR #523, opened 2026-07-07
sim2real-check: iN/ traversal + per-iteration verdicts + exit code (#513)

kalantar:
## Summary

Grows `sim2real-check` by one path segment (`iN/` traversal) and adds per-iteration verdicts + exit-code contract, per epic step-5 (#502).

Closes #513.

## What changed

- **New Python enumerator** — `.claude/skills/sim2real-check/scripts/enumerate_iterations.py`. Reads `run_metadata.json` + `manifest.assembly.yaml` + `translation_output.json`; classifies every `(phase, workload, iteration)` triple as `PRESENT` / `MISSING` / `SKIP` by cross-referencing disk state; emits structured JSON on stdout and exits per contract.
- **SKILL.md** — adds "Step 0.5: Enumerate iterations" that invokes the enumerator (or synthesizes the same JSON shape in-shell for `--real` mode), prints a divergence-warning header when the run shape is mixed or malformed iteration-dir names are present, and threads each row's `results_dir` into every subsequent check subsection via a "Path substitution contract". Adds a "Per-iteration verdict rollup" section at the end that folds signal outcomes over PRESENT rows and encodes the automation-facing exit code.
- **Tests** — `.claude/skills/sim2real-check/tests/test_enumerate_iterations.py`. Eleven pytest cases covering the six acceptance-criteria fixtures (all-present replica, missing i2, algorithm-not-in-translation SKIP, legacy shape implicit i1, mixed shape divergence, malformed iteration-dir count) plus `main()` exit-code assertions and invocation-error paths.
- **CI** — `.github/workflows/test.yml` and the `CLAUDE.md` CI section gain `.claude/skills/sim2real-check/tests/` in the pytest path list.

## Design decisions

- **Source of declared iterations:** `run_metadata.json.replicas` (defaults to 1 when absent). Locally on the check-skill side we read this rather than querying the cluster ConfigMap directly — `replicas` is the manifest-authoritative source that seeds the ConfigMap at `deploy.py run` time. The design's "State carriers" table backs this: manifest = declared intent.
- **SKIP semantic:** an algorithm phase present in `manifest.assembly.yaml.algorithms` but absent from `translation_output.json.algorithms`. Baseline phases are never SKIP. Matches step-3's design intent ("algorithm not in translation").
- **Shape detection is per-run, not per-pair:** if any pair has `iN/` subdirs and any other pair has a direct `trace_data.csv`, the run is `mixed` and the enumerator emits a divergence warning. In mixed mode every declared-but-absent iteration lands as MISSING. This is the read-only diagnostic the design's Source-of-truth section calls out (`sim2real-check` is the exception to the fail-fast rule — the operator uses check to see the state).
- **`resolve.py` is untouched.** The enumerator reads `manifest.assembly.yaml` + `translation_output.json` directly rather than routing through `sim2real resolve`, so no schema drift in the resolve output. The check skill still calls `resolve` for the CONFIGS/RESULTS/TRANSLATION variables at Step 0 (unchanged) and adds the enumerator alongside at Step 0.5.

## Stale-reference sweep

Grepped `docs/`, `.claude/skills/`, `pipeline/README.md`, `CLAUDE.md` for stale pre-iN/ path shapes and pre-verdict wording:

- `CLAUDE.md` already carries the `/i<N>/gpu_logs/<node>.log` path shape (from #522). No update needed.
- `docs/epics/step-5/design.md` is the design source; my implementation matches. No update needed.
- SKILL.md's own check-subsection references to `$RESULTS_DIR/<phase>/<workload>/…` are prose that the new "Path substitution contract" (inserted right before Section 1) remaps to `<row.results_dir>/…`. This is deliberately not a mechanical rewrite of every subsection — the subsections are LLM-consumed prose and doing the substitution mechanically would explode the diff without changing behavior.

## Test evidence

Enumerator suite:

```
$ python -m pytest .claude/skills/sim2real-check/tests/test_enumerate_iterations.py -v
...
11 passed in 0.07s
```

Full CI locally (`.github/workflows/test.yml` path list):

```
$ ruff check pipeline/ .claude/skills/ --select F
All checks passed!
$ python -m pytest pipeline/ ... .claude/skills/sim2real-check/tests/ -q
1599 passed, 2 xfailed in 25.22s
```

End-to-end smoke against a synthesized 3-replica run with i2/ missing on the treatment side:

```
$ python .claude/skills/sim2real-check/scripts/enumerate_iterations.py \
      --run trial --experiment-root /tmp/…
{
  "shape": "replica",
  "declared_replicas": 3,
  "counts": {"PRESENT": 5, "MISSING": 1, "SKIP": 0},
  "exit_code": 1,
  ...
}
exit=1
```

## Reviewer attention

- **SKILL.md path substitution contract.** The one-paragraph note that "every `$RESULTS_DIR/<phase>/<workload>/…` reference below reads as `<row.results_dir>/…`" is a deliberate design choice — the subsections are LLM-consumed prose, and mechanically substituting the shape in every location would double the diff without any semantic gain. If reviewers would prefer an in-place mechanical rewrite of every subsection path, say so and I'll follow up.
- **`$SIM2REAL_ROOT` derivation.** Step 0.5 derives the framework root via a Python one-liner that imports `pipeline.lib.layout` and takes `parents[2]`. Fallback is `$(pwd)` — safe when the operator runs the check from the framework repo. If the resolve output should carry `sim2real_root` explicitly, that's a `resolve.py` schema-v2 change and out of scope here.
- **Exit-code contract.** The enumerator's `exit_code` is `0`/`1` based on `MISSING`-only; `FAIL` is added by the LLM's signal checks at rollup time. The SKILL.md rollup section documents that the overall skill exit code is `max(enumerator_exit, any_FAIL_present)`. Consumers relying on the exit code should note this two-source composition.

## Base

`refactor/v2-step-5`. Not `main`.

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -220,6 +220,118 @@ Workload filter (WORKLOAD):      <name or "(all)">
 
 Do NOT proceed until the user replies `ok` (or equivalent). Additionally, do NOT proceed if `SIM` is unset — every check subsection dereferences it. If auto-detect failed to find a sim bundle and the operator did not pass `--sim`, block on `SIM=<path>` at the prompt.
 
+### Step 0.5: Enumerate declared iterations (both modes)
+
+Once the confirmation prompt has been answered `ok`, enumerate the run's declared iteration range and cross-reference with disk state. This step must run **before** any check subsection — every subsection below iterates over the resulting `PRESENT` rows.
+
+**Resolve-mode (`--run`):** invoke the enumerator against the workspace-registered run:
+
+```bash
+# $SIM2REAL_ROOT is the framework repo root (the directory that contains
+# pipeline/ and .claude/). Derive it from the sim2real CLI's own path;
+# fall back to the current working directory if the derivation is
+# ambiguous (edge case for operators running from a repo checkout).
+SIM2REAL_ROOT=$(python -c 'from pipeline.lib import layout; import pathlib, sys; print(pathlib.Path(layout.__file__).resolve().parents[2])' 2>/dev/null || pwd)
+
+ENUM_JSON=$(mktemp -t sim2real_check_enum.XXXXXX.json)
+# Chain onto RESOLVED_JSON's trap so both temp files are cleaned up.
+trap 'rm -f "$RESOLVED_JSON" "$ENUM_JSON"' EXIT
+if ! python "$SIM2REAL_ROOT/.claude/skills/sim2real-check/scripts/enumerate_iterations.py" \
+        --run "$RUN" --experiment-root "$EXPERIMENT_ROOT" > "$ENUM_JSON"; then
+    ENUM_EXIT=$?
+    if [ "$ENUM_EXIT" -eq 2 ]; then
+        # Invocation error — enumerator already wrote a specific message
+        # to stderr. Bail so the operator can fix the input.
+        exit 2
+    fi
+    # Exit 1 means "MISSING rows present" — that's a real-run diagnostic,
+    # not an invocation error. Keep going and let the final rollup
+    # surface the MISSING rows in the summary table.
+fi
+SHAPE=$(jq -r '.shape' "$ENUM_JSON")
+DECLARED_REPLICAS=$(jq -r '.declared_replicas' "$ENUM_JSON")
+DIVERGENCE_WARNINGS=$(jq -r '.divergence_warnings | join("\n  ")' "$ENUM_JSON")
+MALFORMED_ITER_COUNT=$(jq -r '.malformed_iter_dir_count' "$ENUM_JSON")
+```
+
+**Legacy-mode (`--real`):** the enumerator is `--run`-only (it reads `workspace/runs/<name>/`). In legacy mode, synthesize the enumeration in-shell: every `(phase, workload)` under `$RESULTS_DIR` becomes a single row with `iteration=1`, `status=PRESENT` if `$RESULTS_DIR/<phase>/<workload>/trace_data.csv` exists. Do not emit `MISSING` for legacy bundles.
+
+```bash
+# Legacy-mode synthesis: build a JSON payload with the same schema as
+# the enumerator's output so downstream steps can consume $ENUM_JSON
+# uniformly across both modes.
+ENUM_JSON=$(mktemp -t sim2real_check_enum.XXXXXX.json)
+{
+    echo '{"run":"'"$REAL"'","shape":"legacy","declared_replicas":1,'
+    echo '"divergence_warnings":[],"malformed_iter_dir_count":0,"rows":['
+    first=1
+    for phase in $PHASES; do
+        for wl in ${WORKLOADS_BY_PHASE[$phase]}; do
+            [ -f "$RESULTS_DIR/$phase/$wl/trace_data.csv" ] || continue
+            [ $first -eq 0 ] && echo ','
+            first=0
+            printf '{"phase":"%s","workload":"%s","iteration":1,"status":"PRESENT","results_dir":"%s","note":null}' \
+                "$phase" "$wl" "$RESULTS_DIR/$phase/$wl"
+        done
+    done
+    echo '],"counts":{"PRESENT":0,"MISSING":0,"SKIP":0},"exit_code":0}'
+} > "$ENUM_JSON"
+SHAPE=legacy
+DECLARED_REPLICAS=1
+DIVERGENCE_WARNINGS=""
+MALFORMED_ITER_COUNT=0
+```
+
+**Diagnostic header** — print immediately (before any check subsection runs) so the operator sees mixed/corrupt state up front:
+
+```bash
+if [ -n "$DIVERGENCE_WARNINGS" ] || [ "$MALFORMED_ITER_COUNT" -gt 0 ]; then
+    echo "── Run shape diagnostic ─────────────────────────────────────"
+    echo "  shape:                    $SHAPE"
+    echo "  declared_replicas:        $DECLARED_REPLICAS"
+    echo "  malformed_iter_dir_count: $MALFORMED_ITER_COUNT"
+    if [ -n "$DIVERGENCE_WARNINGS" ]; then
+        echo "  divergence_warnings:"
+        echo "    $DIVERGENCE_WARNINGS"
+    fi
+    echo "─────────────────────────────────────────────────────────────"
+fi
+```
+
+**`$ENUM_JSON` schema (v1):**
+
+```json
+{
+  "run": "trial",
+  "shape": "replica" | "legacy" | "mixed",
+  "declared_replicas": 3,
+  "divergence_warnings": ["..."],
+  "malformed_iter_dir_count": 0,
+  "rows": [
+    {"phase": "baseline", "workload": "wl-chat", "iteration": 1,
+     "status": "PRESENT", "results_dir": "<abs path>", "note": null},
+    {"phase": "sim2real-ac", "workload": "wl-chat", "iteration": 2,
+     "status": "MISSING", "results_dir": null,
+     "note": "no results/i2/ directory"},
+    {"phase": "sim2real-routing", "workload": "wl-chat", "iteration": 1,
+     "status": "SKIP", "results_dir": null,
+     "note": "algorithm not in translation"}
+  ],
+  "counts": {"PRESENT": N, "MISSING": M, "SKIP": S},
+  "exit_code": 0
+}
+```
+
+Each check subsection below iterates over the `PRESENT` rows in `$ENUM_JSON`. Use each row's `results_dir` as the working path — it already carries the `iN/` segment in replica shape and points at the direct workload dir in legacy shape. `MISSING` and `SKIP` rows do not run subsections; they land in the final rollup with their enumerator-assigned status verbatim.
+
+**Iterate PRESENT rows:**
+
+```bash
+# Emit one line per PRESENT row: "<phase>\t<workload>\t<iteration>\t<results_dir>"
+jq -r '.rows[] | select(.status == "PRESENT") |
+       [.phase, .workload, .iteration, .results_dir] | @tsv' "$ENUM_JSON"
+```
+
 ## Goal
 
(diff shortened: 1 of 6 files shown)


kalantar:

The enumerator's legacy branch emits MISSING rows in direct contradiction to its own docstring and the sibling SKILL.md legacy-synth path, so incomplete-but-legacy runs get exit 1 instead of the documented "PRESENT/SKIP only" report.

### Must fix
- **Legacy shape emits MISSING when a workload dir lacks trace_data.csv, contradicting the docstring and diverging from the SKILL.md legacy synth** (`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:162`) — Verified by execution: a run with wl-chat containing direct trace_data.csv and wl-code as an empty dir yields shape='legacy', counts={'PRESENT':2,'MISSING':2}, exit_code=1. The module docstring at line 12 states 'MISSING - Not applicable to legacy runs' and SKILL.md line 270 uses `[ -f ... ] || continue` (silent skip) for the exact same scenario in --real mode. The elif branch at lines 162-171 is the offender. No test exercises the legacy-with-uncollected-workload path (test_legacy_shape_implicit_i1 always writes trace_data.csv), so the suite passes. Concrete downstream harm: exit code 1 vs 0 flips the automation-facing contract for a run that is legitimately partial-legacy.

### Recommendations
- **SKILL.md bash captures $? inside `if ! cmd; then` where it is always 0, making the exit-2 discrimination dead code** (`.claude/skills/sim2real-check/SKILL.md:241`) — Verified with `bash -c 'if ! (exit 2); then echo $?; fi'` returning 0. In SKILL.md at line 241, `ENUM_EXIT=$?` inside `if ! python ...; then` is always 0, so `[ "$ENUM_EXIT" -eq 2 ]` never fires and the intended bail-on-invocation-error path is unreachable. Capture the exit code outside the negated-if construct.
- **Non-dict JSON in run_metadata.json or translation_output.json raises AttributeError that main() does not catch, violating the exit-2 contract** (`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:299`) — Verified by execution: seeding `translation_output.json` with `[]` (a list) causes `tout.get('algorithms')` at line 108 to raise AttributeError, which propagates past `except EnumError` in main() and exits with Python's default code 1. SKILL.md treats exit 1 as 'MISSING rows present, keep going,' so downstream jq operations then run on an empty $ENUM_JSON. _read_manifest at line 325 already guards this with an isinstance check; _read_json should do the same.

kalantar:
Applied the review's must_fix: the enumerator's legacy branch no longer emits `MISSING` when a workload dir has no `trace_data.csv`. The elif branch that appended `MISSING` rows for legacy shape now `continue`s silently, matching the module docstring's "Not applicable to legacy runs" and the SKILL.md `--real`-mode synth path (`[ -f … ] || continue`).

Added `test_legacy_shape_uncollected_workload_is_silent` — a two-workload legacy run where one workload has no `trace_data.csv` — asserting `exit_code == 0` and no MISSING rows. The pre-existing `test_legacy_shape_implicit_i1` test still passes (it collected data for every workload, so the elif was never reached).

Did not act on the two recommendations (advisory only per the review verdict):
- SKILL.md `if ! cmd; then ENUM_EXIT=$?` capturing 0 — the bash is LLM-interpreted rather than operator-executed, so the practical impact is limited. Worth a separate cleanup if we start distributing the snippet.
- `_read_json` not guarding against non-dict JSON roots — the exit-2 contract is documented on the enumerator's own interface; the SKILL.md consumer treats exit 1 as "keep going," so a corrupt-JSON path would degrade to a silent no-op rather than a hard bail. Same suggestion, same rationale: separate cleanup.

Local verification: 12/12 enumerator tests pass; full CI test list (1600 passed, 2 xfailed); ruff --select F clean.

kalantar:

The exit-2 bail-out in Step 0.5 is unreachable due to a bash `$?` capture bug, and several docstring/schema statements contradict the actual contract or the test-verified behavior.

### Must fix
- **`ENUM_EXIT=$?` after `if ! python ...` always reads 0 — the exit-2 bail-out is dead code** (`.claude/skills/sim2real-check/SKILL.md:241`) — Empirically confirmed with `bash -c 'if ! python3 -c "raise SystemExit(2)"; then echo $?; fi'` — prints 0, not 2. When enumerate_iterations.py returns 2 for an invocation error, the `if [ "$ENUM_EXIT" -eq 2 ]` branch at SKILL.md:242 never fires; execution falls through to the 'MISSING rows' continuation path and downstream jq calls (lines 251-254) run against empty or partial `$ENUM_JSON`. This directly violates the exit-code contract documented at SKILL.md:778-784 and at enumerate_iterations.py:21-26.

### Important
- **$ENUM_JSON schema example shows `"exit_code": 0` alongside a MISSING row, contradicting the documented contract** (`.claude/skills/sim2real-check/SKILL.md:321`) — The rows array (SKILL.md:313-315) contains a MISSING row, but per enumerate_iterations.py:182 and SKILL.md:780-781 any MISSING row must yield exit_code 1. Automation authors reading this schema example as reference will encode the wrong expectation.
- **Module docstring and SKILL.md rollup both list "unreadable ConfigMap" as an exit-2 cause, but the enumerator reads no ConfigMap** (`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:25`) — Grepping the script confirms no ConfigMap access — it reads run_metadata.json, manifest.assembly.yaml, and translation_output.json. SKILL.md:782 repeats the same misleading phrasing. A maintainer diagnosing an exit-2 will look in the wrong place.
- **_detect_shape docstring says legacy pairs in mixed mode report MISSING for `i2..iN`, but actual behavior is all `i1..iN` MISSING** (`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:212`) — In mixed shape, enumerate_run:143 uses the replica-shaped iteration path unconditionally, so a legacy pair (with only direct `trace_data.csv`) has no `iN/` subdir for any N including i1. test_mixed_shape_divergence_warning at test_enumerate_iterations.py:291 confirms MISSING count is 2 for a 2-replica pair, i.e. the entire declared range including i1.
- **No test asserts `results_dir` values on PRESENT rows** (`.claude/skills/sim2real-check/tests/test_enumerate_iterations.py`) — SKILL.md:361 and 371 state that `results_dir` must carry the `iN/` segment in replica shape and the direct workload dir in legacy shape — this is the working path every downstream check subsection reads. No test in test_enumerate_iterations.py accesses `r.results_dir`, so a regression that dropped or duplicated the `iN/` segment would pass CI while silently misdirecting all subsections.

### Recommendations
- **Add a main()-level test for the corrupt-JSON exit-2 path** (`.claude/skills/sim2real-check/tests/test_enumerate_iterations.py`) — _read_json converts OSError/JSONDecodeError to EnumError → exit 2 via main(), but only the missing-run branch is exercised (test_main_exit_code_on_missing_run). A regression that removed the except clause would produce a raw traceback exit 1 without any test catching it.
- **_int_or_default silently defaults to 1 for null / non-numeric `replicas`** (`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:288`) — A run_metadata.json with `"replicas": null` (or zero, or negative) silently collapses the declared range to 1 with no diagnostic — a 3-replica run would enumerate only i1 and report a false clean pass. A warning to stderr (and raising EnumError for <= 0) would prevent the silent narrowing.

kalantar:
Applied the iteration-2 must_fix and all four important findings.

- **SKILL.md Step 0.5 bash**: the `if ! cmd; then ENUM_EXIT=$?` pattern was dead code (bash negation collapses `$?` to 0 inside the then-block). Rewrote it as an unconditional invocation followed by an immediate `ENUM_EXIT=$?` capture, so the exit-2 bail-out is reachable. Verified against `python3 -c "raise SystemExit(2)"` and friends: captured values now match the real exit codes verbatim.

- **Schema example**: the `$ENUM_JSON` illustration carries a MISSING row, so `exit_code` is now `1` (was `0`, contradicting the contract). Added a one-liner restating the contract inline.

- **"unreadable ConfigMap" phrasing**: replaced in both the Python docstring (line 25) and the SKILL.md rollup (line 782) with "unreadable workspace files (`run_metadata.json`, `manifest.assembly.yaml`, `translation_output.json`)". The enumerator reads only workspace files — no ConfigMap access exists in either place.

- **`_detect_shape` docstring**: updated to say legacy pairs in mixed mode report MISSING for the entire declared range (`i1..iN`), not just `i2..iN`. In mixed mode the iteration path is replica-shaped for every pair, so a legacy pair with only a direct `trace_data.csv` has no `iN/` subdir for any N. The existing `test_mixed_shape_divergence_warning` already confirmed this behavior; the docstring is now consistent with it.

- **`results_dir` assertions in tests**: added spot-checks to `test_replica_all_present` (asserts one PRESENT row's `results_dir` ends with `results/<phase>/<workload>/i2`) and `test_legacy_shape_implicit_i1` (asserts a legacy-shape `results_dir` ends with `results/<phase>/<workload>` and does NOT have an `/i1` suffix). A regression that dropped or duplicated the `iN/` segment would now fail CI.

Verification: 12/12 enumerator tests pass; ruff `--select F` clean.

Recommendations (advisory only, not addressed): main()-level test for the corrupt-JSON exit-2 path, and `_int_or_default` warning on non-numeric `replicas`. Both worth follow-ups if the enumerator gains more surface area, but not blocking here.

kalantar:

Well-scoped enumerator with clean tests, correct exit-code plumbing, and appropriate README/CI updates; specialist findings are all doc-consistency or edge-case robustness nits, none of which block landing on the epic branch.

*(Verdict posted manually — the workflow's post-comment agent hit a base64 tool error on this iteration after the synthesizer produced the verdict. Full transcript available if useful.)*

### Recommendations (advisory only, do not block merge)

- **`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:152`** — PRESENT/MISSING predicate checks `trace_data.csv` while docstring, MISSING note, and issue spec say the boundary is `iN/` directory existence.

  Lines 8-12 of the module docstring and the MISSING note built at line 176 (`no results/i{it}/ directory`) both frame the boundary as directory presence, but line 152 uses `(disk_dir / 'trace_data.csv').exists()`. A partial collection that creates i2/ without trace_data.csv gets reported as MISSING with a note that lies about the directory. Fix is a one-line predicate change or a text-only doc/note alignment; either resolves the internal inconsistency.

  *(Source lens: code-reviewer.)*

PR #528, opened 2026-07-07
chore(deps): sync llm-d-benchmark, bootstrap templates, pipeline EPP injection, and per-algo overlay for v0.9.0 / v0.7.0

kalantar:
Closes #527. Closes #526. Closes #530.

## Summary

Four coordinated changes to make refactor/v2 correct for llm-d-router v0.9.0 + llm-d-benchmark v0.7.0:

1. Bump `llm-d-benchmark` submodule from `5208cf38` (pre-v0.7.0) to `76473d05` (v0.7.0).
2. Update `.claude/skills/sim2real-bootstrap/templates/defaults/`:
   - Delete `llm-d-rbac.yaml` — v0.9.0 routerlib chart installs the equivalent Role + RoleBinding natively.
   - `epp-verbosity.yaml`: `inferenceExtension.verbosity` → `router.epp.verbosity`.
3. Rewrite `pipeline/lib/epp.py` and adjacent code to emit `router.epp.image` (chart-native shape with `registry` split from `repository`) instead of `images.inferenceScheduler`. v0.7.0 no longer reads the old key.
4. **Fix regression** in `sim2real.py:_cmd_build` — the refactor/v2 rewrite omitted the per-algorithm source-overlay lifecycle that `deploy.py:_cmd_build` on main had. Every algorithm was built from the same (arbitrary) source tree, so N tags → N images with the same binary → EPP crash-loops with "plugin type not registered." Restored via `restore_baseline` + `restore_treatment` (from `pipeline.lib.source_toggle`) around each buildkit dispatch, wrapped in a try/finally that always restores the tree.

## Commits

- `26aa625` chore(deps): bump llm-d-benchmark submodule to v0.7.0
- `499d65c` chore(bootstrap): update templates/defaults for v0.9.0 router chart
- `5ac52c2` pipeline: emit router.epp.image instead of images.inferenceScheduler (#526)
- `7e056b3` sim2real build: apply per-algo source overlay before dispatch (#530)

## What's NOT in scope (follow-up)

- **#529** — `sim2real-check`'s SKILL.md tells the reviewing LLM to parse EPP config at `inferenceExtension.pluginsCustomConfig` and Priority Bands via `extraObjects`. Under v0.9.0 those paths are stale; check parity report is unreliable until this lands. (translate skill already writes the chart-native shape.)

## Test plan

- [x] `git submodule status llm-d-benchmark` reports `76473d05`.
- [x] `.claude/skills/sim2real-bootstrap/templates/defaults/llm-d-rbac.yaml` no longer present.
- [x] `.claude/skills/sim2real-bootstrap/templates/defaults/epp-verbosity.yaml` sets `router.epp.verbosity`.
- [x] `pipeline/lib/epp.py` and `assemble_run.py:inject_image_tag` write `router.epp.image` with split registry/repository.
- [x] `sim2real.py:_cmd_build` applies per-algo overlay before dispatch and restores the tree afterwards (try/finally).
- [x] `python3 -m pytest pipeline/` — 1398 passed, 2 xfailed (pre-existing).
- [x] `python3 -m ruff check pipeline/ --select F` — clean.
- [ ] End-to-end verification: multi-algo translation → `sim2real build --force-rebuild` → each image contains its own plugin registration → deployed EPP starts. Blocked on a real cluster run; #530 fix is on the branch.

## Sweep summary

Grepped for `inferenceScheduler` across the whole worktree after edits. Only remaining hits are in `docs/superpowers/plans/2026-07-01-issue-445-sim2real-assemble.md` (historical planning doc for a merged issue — intentionally left as-is). All active code paths updated.


kalantar:
Addressed the review verdict in `8b4e5ed`.

**Fail loud on post-build cleanup errors** — the finally-block `restore_baseline` in `sim2real.py:_cmd_build` used to print a warning and continue, so a partial cleanup failure could leave the source tree dirty and produce a false-success run with the wrong binary compiled into subsequent images (exactly the silent A/A-image class this PR was written to prevent). It now prints an `error:`, sets `any_failure = True`, and breaks the loop so the run exits 2.

**Unit coverage for the overlay lifecycle** — added `TestCmdBuildOverlayLifecycle` in `test_sim2real.py` with four tests: overlay applied per-algo before dispatch (captures source_dir state at dispatch time and asserts each algo has its own overlay with the other absent — direct regression guard for #530), finally-block runs on buildkit failure, finally-block failure fails loud and does not proceed to the next algo, missing `algo_output.json` returns exit 2.

**Docstring fix on `inject_image_tag`** — the digest-ref sentence still described the pre-refactor behavior. Reworded to state that digest refs are split at the last `/` the same as tag refs (the `@sha256:...` suffix stays attached to the bare repository component) and that `tag` is always `""` for digests. Also documented the bare-image case.

**Bare-image test for `inject_image_tag`** — mirrors the new `test_bare_repository_no_registry` in `test_epp.py` for the sibling `inject_image_ref` function.

pipeline/ pytest: 1403 pass (5 new), 2 xfailed unchanged. `ruff --select F` clean.

The `code-reviewer` lens's scope-boundary finding (#526 work bundled into a #527 PR) is a call for you to make — I didn't split the PR because #526 is a hard dependency on the v0.9.0 upgrade actually working and the PR body already declares it as closing both issues. Happy to split into two PRs if you'd prefer that boundary — say the word.

PR #533, opened 2026-07-08
fix(assemble): --force rebuilds run_dir on every existing-run branch (#532)

kalantar:
## Summary

Closes #532. Base: `refactor/v2`.

`sim2real assemble --force` silently no-op'd in two branches of the existing-run decision tree — same-hash + same-replicas (true no-op) and same-hash + grow-replicas (additive-grow). This PR makes `--force` uniformly rebuild the run directory (`shutil.rmtree` + fresh assemble) in every branch inside `if run_dir.exists():`. Only shrink is preserved; the shrink guard runs first and `--force` cannot bypass it (tracked in #506).

## Change surface

- `pipeline/lib/assemble_run.py` — inside `assemble_run()`, added `if force: shutil.rmtree(run_dir)` inside the equal-replicas branch (previously true-no-op return) and `elif force: shutil.rmtree(run_dir)` for the grow branch (previously always additive-grow). Existing behavior when `force=False` unchanged.
- `pipeline/tests/test_assemble_replicas.py` — two new tests:
  - `test_reassemble_at_same_replica_count_with_force_rebuilds` — counterpart to the existing `test_reassemble_at_same_replica_count_is_noop`, asserts sentinel file seeded between calls is deleted (proves `rmtree`) and iN files still present.
  - `test_grow_with_force_rebuilds_instead_of_additive_grow` — counterpart to `test_grow_from_3_to_5_preserves_i1_i3_and_adds_i4_i5`, uses the same sentinel technique to prove full rebuild vs additive-grow.
- `pipeline/README.md` — new "**`--force` behavior**" paragraph clarifying that `--force` bypasses no-op and additive-grow but not shrink. The pre-existing drift and legacy-run notes already covered `--force` for those two branches.

## Why this is a bug (not a feature request)

The pre-existing code already treated `--force` as "rmtree + rebuild" in the drift, corrupt-metadata, legacy-shape, and missing-metadata branches (see `pipeline/lib/assemble_run.py` around lines 796–848). The equal-replicas and additive-grow branches were the only ones that silently discarded `--force`. The README's line 351 already documented `--force` as "recursively deletes workspace/runs/<run>/ before re-materializing it" — the pre-fix code contradicted this. This PR aligns code with documented intent and closes the discovery gap.

## Reviewer attention

- The shrink guard at line 829 still runs before every other check, unconditionally. `test_shrink_with_force_and_drift_still_refuses` (pre-existing) covers this; no regression risk.
- The `if force: shutil.rmtree(run_dir)` inserted here duplicates a pattern that appears three other times in this function. A future cleanup could hoist the `--force` check to the top of `if run_dir.exists():` (after the shrink guard), collapsing five duplicated `rmtree` calls into one. Chose not to do that here to keep this fix surgical.
- Additive-grow-with-`--force` is now interpreted as "full rebuild", not "grow-with-side-effects". Per issue-scope comment: `--force` is the "I know what I'm doing, nuke it" escape hatch; silently routing to grow when the operator asked to rebuild is the same class of confusion as the equal-replicas no-op.

## Stale-reference sweep

Grepped for "true no-op", "no-op idempotence", "additive-grow", "--force" across `pipeline/`, `docs/`, `README.md`, and `.claude/skills/`:

- `pipeline/README.md:330` — updated. Added `--force` behavior paragraph and reframed the three-way listing as "without `--force`".
- `docs/epics/step-5/design.md:78,107,201` — accurate as historical design records; describes without-`--force` behavior which is unchanged. Not modified.
- `pipeline/lib/assemble_run.py:857` — my own edit. Comment "True no-op" is now inside the `else: force=False` branch, still accurate.
- `pipeline/tests/test_assemble_replicas.py:4` — module docstring says "no-op idempotence" among tested behaviors. Still true — the no-op path with `force=False` is unchanged and the new tests exercise the counterpart behavior.

## Test plan

- [x] `python -m pytest pipeline/tests/test_assemble_replicas.py -v` — 20 pass (2 new + 18 pre-existing).
- [x] `python -m pytest pipeline/` — 1555 pass, 2 xfailed, no regressions.
- [x] `ruff check pipeline/ --select F` — clean.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -325,11 +325,13 @@ Then each treatment scenario has `router.epp.image` set from that algorithm's ow
 
 **Replicas.** `--replicas N` (default 1) is the number of iterations per (workload, package) pair. Each iteration gets its own PipelineRun (`{phase}-{workload}-{run}-iN`, with `_` → `-` normalization) and its own results subdirectory (`results/<phase>/<workload>/iN/`). `manifest.assembly.yaml` carries a top-level `replicas: N`, and `run_metadata.json` carries the same value as a schema field. Both are set on every assemble.
 
-**Additive-merge (grow-only).** Re-assembling an existing run with `--replicas` interacts with the prior state as follows:
+**Additive-merge (grow-only).** Re-assembling an existing run with `--replicas` interacts with the prior state as follows (without `--force`):
 
 - `N == prior_replicas` — true no-op. No files rewritten; the assemble returns silently.
 - `N > prior_replicas` — additive grow. Existing PipelineRun files (`i1..i{prior}`) are preserved byte-for-byte and by mtime; new files are emitted for `i{prior+1}..iN`. `manifest.assembly.yaml` and `run_metadata.json` are rewritten with the new `replicas` count; `params_hash` is preserved (drift check passed).
-- `N < prior_replicas` — refused with `run '<name>' already has <prior> replicas; refusing to shrink to <N>. Replica shrink is tracked in #506.` This guard runs BEFORE the drift check, so `--force` does NOT bypass it.
+- `N < prior_replicas` — refused with `run '<name>' already has <prior> replicas; refusing to shrink to <N>. Replica shrink is tracked in #506.` This guard runs BEFORE every other check, so `--force` does NOT bypass it.
+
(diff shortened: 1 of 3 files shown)


kalantar:
# Review verdict

**Verdict:** READY TO MERGE

**Summary:** The fix correctly closes the silent no-op holes with tests that directly cover both new branches, and every specialist objection either targets a pre-existing pattern the PR doesn't change, is invalidated by code the specialist didn't see, or is a doc nit not worth blocking on.

**Must-fix:** (none)
**Important:** (none)
**Recommendations:** (none)

---

Independent four-specialist review (run `wf_bacd9ba4-350`): code-reviewer, comment-analyzer, pr-test-analyzer, silent-failure-hunter — each in a fresh context. Adversarial synthesizer re-read the code, considered every finding raised, and returned all three tiers empty.

_(Verdict posted manually; the workflow's auto-post step failed because `gh` inferred the repo from the operator's cwd, `kalantar-msb/sr`, instead of `inference-sim/sim2real`.)_

PR #539, opened 2026-07-08
fix(sim2real-translate): update skill to current harness team/session API

kalantar:
## Summary

Update `sim2real-translate` skill to the current Claude Code harness API. Four fixes from the issue plus one additional agent-name-collision fix caught during implementation, all in a single file. Closes #482.

## Changes (all in `.claude/skills/sim2real-translate/SKILL.md`)

1. **Drop deprecated team API.** Remove `TeamCreate` / `TeamDelete` from allowed-tools; drop the `TeamCreate(...)` call and the `team_name=...` arg on all three `Agent()` spawns. The harness now says: *"team_name — Deprecated; ignored. The session has a single implicit team."*

2. **Fix parent-session address.** `{MAIN_SESSION_NAME}` substitution changed from `"main-session"` (invalid recipient) to `"main"` per current `SendMessage` docs.

3. **Free agent names between algorithms.** Replace the two shutdown blocks (`SendMessage("*", "shutdown")` x3 + `TeamDelete()`) with `TaskStop(task_id="writer"/"reviewer"/"expert")`. The `SendMessage("*", "shutdown")` calls were dead code — no prompt file listens for that message. `TaskStop` is the actual stopping mechanism.

4. **Also fix name collision inside a single algorithm's "continue N more rounds" branch.** Previously spawned a new writer without stopping the old one, so the second would be silently renamed `writer-2` and the reviewer's outgoing messages to `writer` would land in the wrong inbox. Added `TaskStop(task_id="writer")` before the respawn.

5. **Make writer's background execution explicit.** Writer's `Agent()` call now has `run_in_background=true` (previously omitted; worked only because the harness default is now background).

## Reviewer attention

- **Prompt files intentionally unchanged.** The only `{*_AGENT_NAME}` placeholder used anywhere is `{EXPERT_AGENT_NAME}` (writer and reviewer both send queries to the expert). No `{REVIEWER_AGENT_NAME}` or `{WRITER_AGENT_NAME}` placeholder exists in any prompt. Issue's option A (`TaskStop` between algos → names freed → next spawn gets clean `writer`/`reviewer`/`expert` names) is sufficient without adding new substitutions.
- **`SendMessage("*", "shutdown")` calls removed alongside `TeamDelete()`.** Grep confirmed no prompt file references `"shutdown"` as a signal to handle — the messages were being sent into a void. If future prompts want a shutdown signal, adding it back is one line.
- **`team_name=...` removal on `Agent()` calls.** Coherent with the team-API deletion. The arg still parses (harness ignores it silently), but keeping stale-but-tolerated arguments invites the same drift confusion this PR is fixing.

## Sweep for stale references

Grep across `.claude/`, `pipeline/`, `docs/`, `CLAUDE.md`, `README.md` for `TeamCreate`, `TeamDelete`, `main-session`: zero hits outside this PR. Prompt files (`prompts/agent-*.md`) reference only `{EXPERT_AGENT_NAME}` and no team API.

## Test plan

- [x] `python -m pytest .claude/skills/sim2real-translate/tests/ -v` — 37 passed.
- [ ] Field verification: run `/sim2real-translate` against a multi-algorithm experiment; confirm no orphan agents remain after the skill completes; confirm agent names don't drift to `writer-2` / `expert-2` across algorithms.

Closes #482

### DIFF
--- a/.claude/skills/sim2real-translate/SKILL.md
+++ b/.claude/skills/sim2real-translate/SKILL.md
@@ -13,12 +13,11 @@ argument-hint: "[--experiment-root PATH] [--rounds N]"
 user-invocable: true
 allowed-tools:
   - Agent
-  - TeamCreate
-  - TeamDelete
   - SendMessage
   - TaskCreate
   - TaskUpdate
   - TaskList
+  - TaskStop
   - Bash(python3 *)
   - Bash(.venv/bin/python *)
   - Bash(cd * && *)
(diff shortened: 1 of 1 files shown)


kalantar:

The four issue-482 acceptance criteria land cleanly and no verified finding rises to blocking merge.

### Recommendations
- **Post-spawn prose still contrasts Expert/Reviewer as background against Writer as foreground.** (`.claude/skills/sim2real-translate/SKILL.md:327`) — Lines 327-328 read "Expert and Reviewer initialize in the background while Writer begins Phase 2" — accurate before this PR, slightly stale now that line 324 gives the Writer `run_in_background=true` as well. Not a bug (the concurrent-timing description still holds and the reader sees identical run_in_background=true on all three spawns just above), but tightening the sentence would keep the doc drift-free.

PR #541, opened 2026-07-08
feat(translate): surface alias alongside hash in stdout messages

kalantar:
## Summary

Update `sim2real translate` stdout to surface the alias (== `transfer.yaml:scenario`) alongside the hash in all three state messages, and change the "next command" hint to use `sim2real build --translation <alias>` so it's copy-pasteable rather than just a status report. Closes #483.

## Changes

- **`pipeline/sim2real.py`** — three print sites updated:
  - **Checkpoint written** (initial `translate`): `translation <hash> (alias: <scenario>) checkpoint written — run '/sim2real-translate' then 'sim2real translate --resume'`
  - **`--resume` on complete state**: `translation <hash> (alias: <scenario>) complete — run 'sim2real build --translation <scenario>' next`
  - **Plain on complete state (idempotent)**: `translation <hash> (alias: <scenario>) already complete — run 'sim2real build --translation <scenario>' next`
- **`pipeline/tests/test_translate.py`** — three named tests updated to assert the alias appears in stdout and the "next command" hint uses `--translation <alias>`:
  - `test_plain_prints_checkpoint_message`
  - `test_plain_prints_already_complete`
  - `test_resume_prints_already_complete`
- **`pipeline/README.md`** — Complete-state row of the translate state table shows the per-column messages verbatim with alias placeholder.

## Reviewer attention

- **Wording split between plain and `--resume` on complete state is deliberate** per the issue spec: plain says "already complete" (idempotent no-op); `--resume` says "complete" (validation passed). The `test_resume_prints_already_complete` test name refers to the *state* being validated (state was already complete when `--resume` ran), not the message text — kept as-is; the test now asserts `"complete"` rather than `"already complete"` with a comment explaining the deliberate difference.
- **Alias == scenario**: `_cmd_translate` reads `scenario = manifest.get("scenario")` at line 831 and passes it to `_translate_write_checkpoint`. It's in scope at all three print sites. Alias validation and cross-hash uniqueness (`translation_ref.find_by_alias`) already run earlier in the flow, so the value printed matches what is stored in `translation_output.json:alias`.
- **`--translation <alias>` hint is valid**: `sim2real build`'s `--translation` flag accepts alias / hash-prefix / full hash via `translation_ref.resolve_translation_ref`. The hint text isn't a docs-only claim.
- **No schema or on-disk changes**: `translation_output.json` is unchanged; the alias is only surfaced in stdout.

## Sweep for stale references

Grepped `pipeline/`, `.claude/`, `docs/`, `README.md`, `CLAUDE.md` for `already complete`, `checkpoint written — run`, and `complete — run 'sim2real build'`:

- `.claude/skills/sim2real-translate/SKILL.md:232, 245` — different code path (skill-side idempotency check with its own message advising `sim2real translate --resume` after the skill run). Left unchanged.
- `docs/epics/step-2/design.md:268, 293` — historical design record describing pre-#483 spec. Left unchanged (matches step-5 design.md precedent from prior PRs — historical design records are not living source of truth).
- `docs/superpowers/plans/2026-07-02-issue-467-sim2real-translate.md:18` — historical implementation plan. Left unchanged.

## Test plan

- [x] `python -m pytest pipeline/tests/test_translate.py -v` — 33 passed.
- [x] `python -m pytest pipeline/` — 1555 passed, 2 xfailed (baseline).

Closes #483

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -211,7 +211,7 @@ Two-phase state machine — the operator picks the lane explicitly (plain vs `--
 |---|---|---|---|
 | Absent | Create dir, write `skill_input.json` + `translation_output.json` with `image_ref: null`. Print checkpoint. Exit 0. | Error: `no translation to resume`. Exit 2. | Same as plain. |
 | Partial (checkpoint files present, some `generated/<algo>/<algo>_output.json` missing) | Error: `translation incomplete — run '/sim2real-translate' then 'translate --resume'`. Exit 2. Never mutates. | Reports missing algorithms by name. Exit 2. Never mutates. | Delete + recreate as if absent. |
-| Complete (all `<algo>_output.json` present) | Print `translation already complete — run 'sim2real build' next` and exit 0. | Same. | Delete + recreate; operator re-runs the skill. |
+| Complete (all `<algo>_output.json` present) | Print `translation <hash> (alias: <alias>) already complete — run 'sim2real build --translation <alias>' next` and exit 0. | Print `translation <hash> (alias: <alias>) complete — run 'sim2real build --translation <alias>' next` and exit 0. | Delete + recreate; operator re-runs the skill. |
 
 The translation hash is derived from `transfer.yaml`'s translation slice (scenario, component, context, per-algorithm sources) plus the SHA-256 of each `algorithms[i].source` file's bytes (see `pipeline/lib/slicer.py:translation_hash_with_sources`). Two runs of `translate` with the same `transfer.yaml` and the same source files produce the same hash and reuse the same `translations/<hash>/` directory.
 
--- a/pipeline/sim2real.py
+++ b/pipeline/sim2real.py
@@ -919,7 +919,10 @@ def _cmd_translate(args) -> int:
                 file=sys.stderr,
             )
             return 2
-        print(f"translation {thash} already complete — run 'sim2real build' next")
+        print(
+            f"translation {thash} (alias: {scenario}) complete — "
+            f"run 'sim2real build --translation {scenario}' next"
+        )
         return 0
 
     if args.force:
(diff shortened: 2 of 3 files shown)


kalantar:

Production code and README changes are correct, but the test that documents the deliberate "complete" vs "already complete" wording split does not actually enforce it — a one-line assertion tighten before merge.

### Important
- **test_resume_prints_already_complete's assertion is too weak to enforce the deliberate wording split it documents** (`pipeline/tests/test_translate.py:477`) — Verified at pipeline/tests/test_translate.py:477: `assert "complete" in out` passes for both "complete" and "already complete" (substring match — confirmed `'complete' in 'already complete'` is True). The comment at lines 474-476 and the PR body explicitly call out the deliberate split between the resume path ("complete") and the plain path ("already complete") per issue #483, and the test was renamed/updated to guard it, but the current assertion would pass if the resume path silently regressed to "already complete". One-line fix: add `assert "already complete" not in out`.

### Recommendations
- **README's Absent row still says only "Print checkpoint. Exit 0." while the Complete row was updated to quote the full alias-inclusive message** (`pipeline/README.md:212`) — Verified at pipeline/README.md:212-214: the Complete row now shows the full new message with alias placeholder, but the Absent row keeps a bare summary. Since the checkpoint message was also changed to include the alias in this PR (pipeline/sim2real.py:1049), the two rows are now unevenly detailed. Not blocking — the Absent summary is still accurate — but a one-line update would keep the table internally consistent.

kalantar:
Tightened the resume-on-complete test to also assert `"already complete" not in out` (664c228). The old substring-only check was passing on both wordings, so a silent regression to the plain-path phrasing wouldn't have been caught.

The recommendation about the README Absent-state row being unevenly summarized against the Complete-state row is noted but not addressed here — separate polish call.

PR #543, opened 2026-07-08
feat(sim2real-check): detect llm-d-router and BLIS via SIM2REAL

kalantar:
## Summary

Update `sim2real-check`'s auxiliary-detection step to match the current deployment reality — `llm-d-inference-scheduler` has been sunset in favor of `llm-d-router` (which has absorbed `gateway-api-inference-extension`), and this framework repo carries its own BLIS via the `inference-sim` submodule. Closes #494.

## Changes (single file: `.claude/skills/sim2real-check/SKILL.md`)

1. **`LLMD` auto-detect** (line 159): search `llm-d-router` only. Legacy `llm-d-inference-scheduler` fallback removed.
2. **`GAIE` auto-detect** (line 158): still tries standalone `gateway-api-inference-extension` first, but when that's absent and `LLMD` resolves, `GAIE` is set to the same router path. `gateway-api-inference-extension` has been merged into `llm-d-router`, so the router codebase serves both roles when consolidated.
3. **`BLIS` auto-detect** (line 157): new first choice `$SIM2REAL_ROOT/inference-sim` — the framework repo's own BLIS submodule (see `.gitmodules`). Existing predicates (cwd with `sim/` + `go.mod`, then `../`, then `tmp/`) become fallbacks.
4. **Labels** (lines 154, 215, 362): "llm-d scheduler" → "llm-d router" / "llm-d-router" in the variable description, confirmation prompt row, and reference-paths section.

## Reviewer attention

- **Amendment from user overrides the issue's original fallback design.** The issue as filed proposed adding `llm-d-router` as a fallback to `llm-d-inference-scheduler` in the LLMD predicate. The user's direction on this fix is stronger: scheduler is gone entirely, so it's not a fallback but a replacement.
- **`$GAIE` and `$LLMD` may resolve to the same directory.** In the consolidated case (router present, standalone gateway-api-inference-extension absent), both point at the same path. §3/§4 code-proof greps that reference `$GAIE` and `$LLMD` separately will then execute against the same tree — this is harmless (grep twice, get the same match twice) and preserves the skill's ability to work in the non-consolidated case (separate `gateway-api-inference-extension` checkout still exists). No prose change to §3/§4 needed.
- **`$SIM2REAL_ROOT` discovery** is already set later in the skill (Step 0.5, line 234) via `python -c 'from pipeline.lib import layout; ...'`. The BLIS predicate at line 157 references `$SIM2REAL_ROOT` and gracefully falls back if it's unset at auxiliary-detection time.
- **No test changes.** The skill is a prompt template. `tests/test_enumerate_iterations.py` covers the `scripts/enumerate_iterations.py` python helper, not the SKILL.md prose — unaffected by this diff (12 tests still pass).

## Sweep for stale references

Grepped `.claude/skills/`, `pipeline/`, `docs/`, `README.md`, `CLAUDE.md` for `llm-d-inference-scheduler`:

- Inside `.claude/skills/sim2real-check/SKILL.md:159` — one occurrence remains, inside my new prose *explaining* that scheduler has been sunset. Intentional.
- `.claude/skills/sim2real-translate/SKILL.md:5` — description text mentions scheduler as example target component. **Out of scope for this issue** (per issue's `sim2real-check`-only scoping); flagged for follow-up.
- `.claude/skills/sim2real-translate/scripts/copy_generated.py:27` — docstring similarly names scheduler as example. **Out of scope**; flagged for follow-up.
- `pipeline/deploy.py`, `pipeline/setup.py`, `pipeline/README.md`, `pipeline/tests/test_*.py`, `pipeline/scripts/build-epp.sh`, `CLAUDE.md:6` — multiple references to `llm-d-inference-scheduler` in code, tests, and root docs. **Out of scope for this issue** per its explicit "Skill-only change; no code changes to `pipeline/`" scoping and the user's amendment which focused on the skill. Recommend a follow-up issue if the broader rename is desired.

## Test plan

- [x] `python -m pytest .claude/skills/sim2real-check/tests/ -v` — 12 pass (unchanged).
- [ ] Field verification: run `/sim2real-check --run <name>` in an environment where `llm-d-router` is cloned; confirm the confirmation prompt shows `llm-d-router (LLMD): <path>` and the §3/§4 code-proof checks run against that path.
- [ ] Field verification: run in an environment where the framework's `inference-sim/` submodule is initialized but no `sim/`+`go.mod` exists elsewhere; confirm `BLIS` is picked up from `$SIM2REAL_ROOT/inference-sim`.

Closes #494

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -154,9 +154,11 @@ fi
 Regardless of which mode-dispatch branch ran, the checks below need four additional variables — `SIM` (the simulation bundle), `BLIS` (the simulator codebase), `GAIE` (the gateway-api-inference-extension codebase), and `LLMD` (the llm-d-router codebase). The mode-dispatch block does not populate them. Run this detection *after* the mode dispatch completes, in every mode, using the same predicates as the pre-refactor skill:
 
 - **Sim bundle** (`SIM`): if `--sim` was passed, use it verbatim. Otherwise scan `$EXPERIMENT_ROOT/experiments/*/` for directories containing `README.md` + `algorithm/` + `workloads/`. First match wins. Common patterns: `sim2real_bundle*`, `sim2real_*bundle*`. Leave unset if no candidate is found.
-- **BLIS codebase** (`BLIS`): first try `$SIM2REAL_ROOT/inference-sim` — the framework repo's own BLIS submodule (`.gitmodules`) — if `$SIM2REAL_ROOT` is set and that directory contains `sim/` + `go.mod`. Fall back to the current working directory if it contains `sim/` + `go.mod`, otherwise search `../` and `tmp/` for a directory with the same shape.
-- **GAIE codebase** (`GAIE`): look for `gateway-api-inference-extension` under `tmp/`, `../`, or nearby. If not found and `LLMD` (below) resolves to an `llm-d-router` checkout, set `GAIE` to that same path — `gateway-api-inference-extension` has been merged into `llm-d-router`, so the router codebase serves both roles.
+- **BLIS codebase** (`BLIS`): first derive `$SIM2REAL_ROOT` inline via the same discovery Step 0.5 uses (`python -c 'from pipeline.lib import layout; import pathlib; print(pathlib.Path(layout.__file__).resolve().parents[2])'`); then try `$SIM2REAL_ROOT/inference-sim` — the framework repo's own BLIS submodule (`.gitmodules`) — if that directory contains `sim/` + `go.mod`. Fall back to the current working directory if it contains `sim/` + `go.mod`, otherwise search `../` and `tmp/` for a directory with the same shape.
+- **GAIE codebase** (`GAIE`): look for `gateway-api-inference-extension` under `tmp/`, `../`, or nearby.
 - **llm-d router** (`LLMD`): look for `llm-d-router` under `tmp/`, `../`, or nearby. (Legacy `llm-d-inference-scheduler` checkouts are no longer detected — that project has been sunset in favor of `llm-d-router`.)
+
+After all four variables have been resolved, apply this GAIE alias: **if `GAIE` is unset and `LLMD` points at an `llm-d-router` checkout, set `GAIE=$LLMD`**. `gateway-api-inference-extension` has been merged into `llm-d-router`, so the router codebase serves both roles when no standalone `gateway-api-inference-extension` checkout is on disk.
 
 **Required vs. optional at check time:**
 
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -151,12 +151,12 @@ fi
 
 ### Auxiliary detection (all modes)
 
-Regardless of which mode-dispatch branch ran, the checks below need four additional variables — `SIM` (the simulation bundle), `BLIS` (the simulator codebase), `GAIE` (the gateway-api-inference-extension codebase), and `LLMD` (the llm-d-inference-scheduler codebase). The mode-dispatch block does not populate them. Run this detection *after* the mode dispatch completes, in every mode, using the same predicates as the pre-refactor skill:
+Regardless of which mode-dispatch branch ran, the checks below need four additional variables — `SIM` (the simulation bundle), `BLIS` (the simulator codebase), `GAIE` (the gateway-api-inference-extension codebase), and `LLMD` (the llm-d-router codebase). The mode-dispatch block does not populate them. Run this detection *after* the mode dispatch completes, in every mode, using the same predicates as the pre-refactor skill:
 
 - **Sim bundle** (`SIM`): if `--sim` was passed, use it verbatim. Otherwise scan `$EXPERIMENT_ROOT/experiments/*/` for directories containing `README.md` + `algorithm/` + `workloads/`. First match wins. Common patterns: `sim2real_bundle*`, `sim2real_*bundle*`. Leave unset if no candidate is found.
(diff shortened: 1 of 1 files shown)


kalantar:

Two ordering bugs in the new auxiliary-detection prose make the added GAIE-alias fallback and BLIS submodule-primary path silently unreachable in the sequence the document itself prescribes.

### Important
- **GAIE alias references LLMD but is evaluated before LLMD in the bullet sequence.** (`.claude/skills/sim2real-check/SKILL.md:158`) — Read lines 152-159: bullets are ordered SIM -> BLIS -> GAIE -> LLMD. The GAIE bullet says 'If not found and LLMD (below) resolves to an llm-d-router checkout, set GAIE to that same path.' A top-to-bottom evaluation sees LLMD unset when the GAIE condition fires, so the newly added alias never triggers and operators with only an llm-d-router checkout continue to see GAIE=(not found) and INAPPLICABLE code-proof verdicts — the exact failure this bullet was added to prevent. Fix by either reordering bullets so LLMD is resolved before GAIE, or introducing the alias as an explicit post-pass step ('after all four are resolved, if GAIE is unset and LLMD is an llm-d-router checkout, set GAIE=LLMD').
- **BLIS primary path guards on $SIM2REAL_ROOT, which is only derived later at Step 0.5 (line 234).** (`.claude/skills/sim2real-check/SKILL.md:157`) — The BLIS bullet says 'first try $SIM2REAL_ROOT/inference-sim ... if $SIM2REAL_ROOT is set'. Confirmed at line 234 that $SIM2REAL_ROOT is derived inside Step 0.5, which runs after the confirmation prompt — auxiliary detection at lines 152-159 executes earlier. Unless the operator exports SIM2REAL_ROOT before invoking the skill, the 'first try' branch is dead and the intended new primary path never activates; detection falls through to the pre-PR predicate. Either derive $SIM2REAL_ROOT at auxiliary-detection time using the same python snippet as line 234, or move the BLIS primary-path check to after Step 0.5, or explicitly document that the new path only activates when the operator sets the env var externally.

kalantar:
Fixed the two ordering bugs the review flagged (84a5190):

- BLIS predicate now derives `$SIM2REAL_ROOT` inline using the same python snippet Step 0.5 uses, so the `$SIM2REAL_ROOT/inference-sim` first guess is actually reachable during auxiliary detection. Before, it was dead code unless the operator exported the env var externally (Step 0.5 sets it later in the flow).
- GAIE alias moved out of the GAIE bullet into an explicit post-pass step. The bullets now resolve top-to-bottom, and the alias ("if `GAIE` is unset and `LLMD` points at an `llm-d-router` checkout, set `GAIE=$LLMD`") runs after all four variables are populated — so it can actually see `LLMD`'s value.

The other findings the review lens surfaced (removing `llm-d-inference-scheduler` entirely; broadening BLIS/GAIE scope beyond the issue's original text) reflect user-authorized amendments to the issue's scope given at `/fix-issue` invocation time — not reverted. The synthesizer did not promote those to blocking for the same reason.

kalantar:

Two explicit acceptance criteria from issue #494 are contradicted by the diff — the legacy scheduler is dropped from LLMD detection rather than preserved as first choice, and the confirmation label names only the router. A third change (GAIE alias) adds an unverified architectural claim outside the issue's scope.

### Must fix
- **LLMD detection drops llm-d-inference-scheduler entirely instead of falling back to it** (`.claude/skills/sim2real-check/SKILL.md:159`) — SKILL.md line 159 now reads 'look for llm-d-router under tmp/, ../, or nearby. (Legacy llm-d-inference-scheduler checkouts are no longer detected — that project has been sunset in favor of llm-d-router.)' Issue #494 lists 'Existing llm-d-inference-scheduler detection unchanged when it is present' as an explicit acceptance criterion, and proposes an ordered fallback (scheduler first, router second). Operators with only llm-d-inference-scheduler on disk will see LLMD=(not found) where the prior skill populated it, and every §3/§4 code-proof check will emit INAPPLICABLE — a regression the issue was written to prevent.
- **Confirmation prompt label names only llm-d-router, not both accepted codebases** (`.claude/skills/sim2real-check/SKILL.md:217`) — SKILL.md line 217 renders 'llm-d-router (LLMD): <path>'. Issue #494's acceptance criteria state 'The confirmation prompt line names both accepted codebases so the operator understands what will be searched,' with the example 'llm-d scheduler / router (LLMD): /path/to/llm-d-router'. If detection is corrected to preserve scheduler support, this label must also identify both alternatives so operators recognize which codebase was matched.

### Important
- **New GAIE=$LLMD alias rests on an unverified architectural claim and is out of scope for #494** (`.claude/skills/sim2real-check/SKILL.md:161`) — SKILL.md line 161 adds 'After all four variables have been resolved, apply this GAIE alias: if GAIE is unset and LLMD points at an llm-d-router checkout, set GAIE=$LLMD. gateway-api-inference-extension has been merged into llm-d-router...' Issue #494 scopes the change strictly to LLMD detection; nothing in the issue, CLAUDE.md, or the linked references in the issue background documents that GAIE was subsumed into llm-d-router. If the claim is wrong or partial (e.g., different package paths), §3/§4 greps that target GAIE-specific symbols against $GAIE will silently return no matches and emit PASS/FAIL against the wrong tree rather than INAPPLICABLE, with no diagnostic to the operator. The aliasing itself is also invisible at the confirmation prompt (both GAIE and LLMD lines will show identical paths with no annotation of why).

kalantar:
Review workflow iteration 3: **READY TO MERGE**.

Summary (verbatim): *"A tight prompt-template change with explicit fallback prose that already handles the failure modes the specialists worried about."*

`must_fix`, `important`, and `recommendations` all empty.

(Posted manually — the workflow's comment-post step choked on a macOS `base64` shell portability issue after the review agents had already returned clean.)

PR #545, opened 2026-07-08
docs(check): parse router.epp.pluginsCustomConfig and router.inferenceObjectives (v0.9.0 chart-native)

kalantar:
Closes #529.

## Summary

`sim2real-check`'s SKILL.md told the reviewing LLM to parse EPP plugin
config from `inferenceExtension.pluginsCustomConfig.custom-plugins.yaml`
and priority bands from `extraObjects` filtered to `kind:
InferenceObjective`. Both paths are stale under v0.7.0/v0.9.0 chart-
native shape: `sim2real assemble` now emits plugin config under
`router.epp.pluginsCustomConfig` and priority bands as a compact list
under `router.inferenceObjectives` (the chart renders each entry into a
`kind: InferenceObjective` CR at deploy time).

Result before this PR: the "EPP & InferenceObjective Config Parity vs
config.md" section either bailed out with "no config found" or reported
spurious mismatches on every v0.9.0-shaped run.

## What changed

Six edits, all in `.claude/skills/sim2real-check/SKILL.md`, within the
existing §2c / §2d / §5c.6 subsections:

1. **Cross-variant diff table (§2c):** replaced the `extraObjects
   (InferenceObjective CRDs)` row with a `router.inferenceObjectives
   (priority bands)` row.
2. **Source-of-truth mapping (§2d), baseline row:** swapped to
   `router.epp.pluginsCustomConfig.custom-plugins.yaml`, with a legacy
   fallback for pre-v0.7.0 runs.
3. **Source-of-truth mapping (§2d), InferenceObjective row:** swapped
   to `router.inferenceObjectives`, with a legacy fallback.
4. **Comparison field list (§2d):** for `router.inferenceObjectives`
   entries, compare `name` and `priority` per entry. `poolRef.name` /
   `poolRef.group` are hardcoded by the v0.9.0 chart (`Release.Name` +
   `inference.networking.k8s.io`) and validated once against `config.md`
   rather than per-item. Legacy shape still handled.
5. **§5c.6 conditional trigger:** runs if either the chart-native path
   or the legacy path declares at least one objective.
6. **§5c.6 inputs:** read `(name, priority)` from
   `router.inferenceObjectives` first; fall back to `(name,
   spec.priority)` from the legacy path.

## Design notes

- **Backward compat (acceptance criterion 3):** shape sniff, not
  explicit deprecation. Each stale reference is replaced with the
  chart-native path plus an "otherwise falls back to legacy" clause.
  Pre-v0.7.0 runs continue to work; the check produces useful output
  on runs of either shape.
- `extraObjects` continues to legitimately hold non-InferenceObjective
  entries (`EnvoyFilter`, etc.) in the same scenario. The edits are
  scoped to the InferenceObjective-specific references — no global
  `extraObjects` rename.

## Verification

Empirical check against a real v0.9.0-shaped scenario at
`kalantar-msb/sr/workspace/runs/trial-9/cluster/softreflective.yaml`:

- `router.epp.pluginsCustomConfig.custom-plugins.yaml` resolves (541
  chars of EndpointPickerConfig YAML-in-YAML).
- `router.inferenceObjectives` = `[{name: critical, priority: 100},
  {name: sheddable, priority: -50}]`.
- Legacy `inferenceExtension.pluginsCustomConfig` path is absent as
  expected.

## Stale-doc sweep

Swept `docs/`, `.claude/skills/`, `pipeline/README.md`, `CLAUDE.md` for
`inferenceExtension.pluginsCustomConfig` and `extraObjects.*Inference
Objective`:

**In scope (fixed here):** the 6 hits in `sim2real-check/SKILL.md`.

**Out of scope — same drift in adjacent locations, deferred:**

- `.claude/skills/sim2real-translate/prompts/agent-writer.md` — 4
  mentions of `inferenceExtension.pluginsCustomConfig`
- `.claude/skills/sim2real-translate/prompts/agent-reviewer.md` — 2
  mentions
- `.claude/skills/sim2real-bootstrap/generate_scenarios.README.md:90`
- `pipeline/README.md:850-880` — the "Scenario Overlay Format" /
  "Plugin config" doc block uses `inferenceExtension.pluginsCustomConfig`
  as the example shape.

Issue #529 explicitly retracted the translate-side claim (verified on
disk: latest translation output for `sr/` already uses `router.epp`
shape). The translate prompts are inconsistent with the output but the
output is chart-native, so it's a latent inconsistency rather than a
live bug. Worth a follow-up issue; not folded in here to keep this
PR scoped to what #529 called out.

**Unrelated hits — left alone:**

- `extraObjects` references to `EnvoyFilter` and similar non-Inference
  Objective entries in real assembled scenarios (the field still
  legitimately holds these).

## Test plan

- [x] No code touched — existing test suite unaffected.
- [x] Empirical check that new paths resolve on a real v0.9.0 scenario
      (see Verification section).

### DIFF
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -672,7 +672,7 @@ Step 5c.2 (priority bands recognized) is a downstream symptom check — it sees 
 
 **Five criteria — all must PASS for the check to PASS:**
 
-1. **Coverage.** Every configured `InferenceObjective` name appears at least once in the `Request handled` lines, resolving to its configured `spec.priority`. *Every configured objective receives traffic.*
+1. **Coverage.** Every configured `InferenceObjective` name appears at least once in the `Request handled` lines, resolving to its configured priority (`priority` from `router.inferenceObjectives` on chart-native shape, or `spec.priority` from `extraObjects` on legacy shape — see Inputs above). *Every configured objective receives traffic.*
 2. **Determinism.** Each distinct `objectiveKey` value resolves to a singleton `priority`. No `objectiveKey` fans out to multiple priorities. *No mis-mapping mid-run.*
 3. **Workload resolution rate.** `count(Request handled lines with objectiveKey != "")` ≥ `count(rows in trace_data.csv)` per cell. *Every workload request carried the header and resolved to a configured objective.*
 4. **Priority value set.** The set of `priority` values seen across all `Request handled` lines is a subset of `{configured priorities} ∪ {0}`. The 0 is the documented `defaultPriority` fallback at `director.go:222`. *No stray priorities, e.g. from a misloaded CR.*
--- a/.claude/skills/sim2real-check/SKILL.md
+++ b/.claude/skills/sim2real-check/SKILL.md
@@ -475,7 +475,7 @@ The real bundle may contain multiple experiment variants — one per phase in `$
 | `plugins` (list of plugin types) | Mostly — see below | Only the flow control policy plugin entry |
 | `schedulingProfiles` (scorers, weights, picker) | Yes | — |
 | `flowControl` | — | Yes — this is the independent variable |
-| `extraObjects` (InferenceObjective CRDs) | Yes (priority bands) | — |
+| `router.inferenceObjectives` (priority bands) | Yes | — |
 | Instance count (from server_logs) | Yes | — |
 | vLLM config (from server logs) | Yes | — |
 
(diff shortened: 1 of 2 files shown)


kalantar:

Edit 6 updated the Inputs paragraph to read the priority value from a shape-dependent field, but the criterion immediately below still hardcodes the legacy field name, leaving §5c.6 internally inconsistent.

### Important
- **§5c.6 criterion 1 still says `spec.priority` after the adjacent Inputs paragraph was updated to read `priority` from `router.inferenceObjectives`.** (`.claude/skills/sim2real-check/SKILL.md:675`) — Verified against the head-of-branch file: line 671 now says `Read the configured set of (name, priority) pairs from ... router.inferenceObjectives ...; fall back to (name, spec.priority) ... extraObjects ...`, but line 675 unconditionally says `resolving to its configured spec.priority`. For chart-native runs the configured priority lives at `priority`, not `spec.priority`, so criterion 1's field reference is inconsistent with the Inputs paragraph three lines above. Two specialists caught this independently and the fix is a one-line wording change to match the shape-sniff already used at line 671 (e.g. `priority (chart-native) or spec.priority (legacy)`).

kalantar:
Addressed the important finding from the review workflow.

**Fixed** — `.claude/skills/sim2real-check/SKILL.md:675`: criterion 1 (Coverage) previously read `resolving to its configured spec.priority`, which contradicted the Inputs paragraph three lines above (line 671) that now reads `priority` from `router.inferenceObjectives` on chart-native shape and falls back to `spec.priority` on legacy. Rewrote criterion 1 to use the same shape-sniff phrasing, so a chart-native run compares against `priority` and a legacy run compares against `spec.priority` — consistent with Inputs. Commit `5f1bfa1`.

The three non-blocking observations (§2c cross-variant table has no legacy fallback annotation; fallback paths don't surface which branch was taken; poolRef validation has no defined behavior when config.md is silent) are worth considering as follow-up refinements to §5c.6 / §2c robustness. Not folded in here to keep this PR scoped to the specific stale-path fix #529 called out.

PR #548, opened 2026-07-08
feat(sim2real-analyze): standard analyses catalog

kalantar:
## Summary

Introduces a standard analyses catalog under `.claude/skills/sim2real-analyze/analyses/` so the interactive loop can offer a curated menu instead of the generic "what would you like to analyze next?" prompt. Addresses analysis-format drift across team members running ad-hoc analyses on the same run.

Closes #536.

## What's in the catalog

Two entries, one per runner type — proves the pattern works for both:

- `latency-table` — `runner: script`, `output: table`. Migrated from `scripts/compute_table.py`; behavior unchanged. This is the default that runs on skill entry.
- `per-request-scatter` — `runner: prompt`, `output: png`. Ported from `kalantar-msb/soft-reflective:workspace/prompts/per-request-scatter.md` and adapted for sim2real's single-run, `{baseline, treatment}` layout. The `slo_class` column doesn't exist in current sim2real trace CSVs (confirmed by grep), so coloring defaults to per-phase; the prompt keeps SLO-class coloring as a fallback for future compatibility.

## Front-matter schema

```yaml
---
name: <slug>
title: <human title>
when-to-use: <one-line trigger>
inputs: run
output: html | png | table
runner: script | prompt
script: <filename>   # required when runner == script
---
```

`scripts/list_analyses.py` discovers `analyses/*.md`, parses front-matter with PyYAML (already in requirements), validates required fields, and emits JSON. Malformed entries are logged to stderr and skipped so a single broken entry doesn't take the catalog down.

## SKILL.md changes

- Step 3 now invokes `analyses/latency_table.py` (behavior identical to the previous `scripts/compute_table.py`).
- Step 4 enumerates the catalog on entry and offers a numbered menu. Free-text is matched against each entry's `when-to-use`; on a strong match the skill invokes the entry, otherwise it falls back to the existing one-off-script flow.
- New "Standard analyses catalog" section documents how to add an entry.

## What warrants reviewer attention

1. **`compute_table.py` → `analyses/latency_table.py` rename.** The `REPO_ROOT = Path(__file__).resolve().parents[4]` math is preserved (same depth from repo root: `.claude/skills/sim2real-analyze/analyses/latency_table.py`). The test file uses `import latency_table as compute_table` to keep the existing 60+ test references working without a mechanical rename — reviewer's call on whether to also rename the local binding.
2. **Per-request-scatter port adaptation.** The soft-reflective source assumes `<bundle>/<run>/<phase>/<workload>/` with `<run>` as a replicate label; the sim2real port assumes `workspace/runs/<name>/results/<phase>/<workload>/` and `phase ∈ {baseline, treatment}`. The prompt body diverged from the source in that dimension — reviewer should verify the port's `INPUT DATA` block matches sim2real's actual layout.
3. **Free-text matching lives in the LLM, not a scorer.** `when-to-use` is prose that the skill's LLM reads at catalog-enumeration time and matches against user input. No mechanical matcher, no unit-testable ranking. This trades unit-testability for zero infrastructure. The alternative (a keyword scorer or embedding matcher) can be added later without changing the catalog format.

## Sweep for stale references

Grepped `**/*.md`, `**/*.py`, `**/*.yml`, `**/*.yaml` for `compute_table`, `scripts/compute_table`, `test_compute_table`. Findings:

- `test_latency_table.py` — 60+ hits are the local `compute_table` binding after `import latency_table as compute_table`. Intentional; not stale.
- `docs/proposals/incremental-implementation-plan-addendum.md:325,328` and `docs/proposals/incremental-implementation-plan.md:157` — the deferred design docs that motivated this issue. They describe step-7 as a future refactor and reference `compute_table.py` as "today's baked-in table." Both statements are now partially superseded by this PR (which delivers the catalog + scaffolding half of step-7 with dynamic-phase / `sim2real resolve` deferred). Left unchanged in this PR — reviewer's call whether the docs should be updated to say "partially delivered by #<this PR>" here or in a follow-up when the deferred rest-of-step-7 lands.
- `.github/workflows/test.yml` and `CLAUDE.md` — both already enumerate `.claude/skills/sim2real-analyze/tests/` at directory level, so the new `test_list_analyses.py` is picked up automatically. No CI change needed.

## Test plan

- [x] `python3 -m pytest pipeline/ .claude/skills/*/tests/` — 1758 passed, 1 skipped, 2 xfailed (identical to main baseline)
- [x] `python3 .claude/skills/sim2real-analyze/scripts/list_analyses.py` on the real catalog — emits both entries with all required fields
- [x] `python3 .claude/skills/sim2real-analyze/analyses/latency_table.py --run <name>` — smoke-tested via the migrated test suite (all 33 pre-existing tests pass)
- [ ] Manual smoke of `/sim2real-analyze` interactive loop against a real run — not yet exercised in this session (no local run available), reviewer or follow-up

## Deferred (not in this PR)

Per the reshaped issue body, the following remain deferred with the larger workspace refactor and are explicit non-goals for this issue:

- Dynamic phase enumeration beyond `{baseline, treatment}`
- `sim2real resolve --run R` integration
- Multi-replicate bundle analyses (soft-reflective's `latency-pairwise-*`, `claim-verification-report`, heatmaps)
- The CDF trio, throughput-over-time, and error-rate analyses (trivial per-file follow-ups once the scaffolding lands)

### DIFF
--- a/.claude/skills/sim2real-analyze/SKILL.md
+++ b/.claude/skills/sim2real-analyze/SKILL.md
@@ -1,9 +1,9 @@
 ---
 name: sim2real-analyze
 description: |
-  Analyze sim2real pipeline run results. Shows per-workload latency comparison tables
-  (TTFT/TPOT/E2E baseline vs treatment) and handles any user analysis request: charts,
-  distributions, HTML reports, cross-run comparisons.
+  Analyze sim2real pipeline run results. Offers a catalog of standard analyses
+  (per-workload latency comparison, per-request scatter) plus a free-form loop
+  for custom charts, distributions, HTML reports, and cross-run comparisons.
 argument-hint: "[--run NAME]"
 user-invocable: true
 allowed-tools:
@@ -58,8 +58,10 @@ If the user describes a specific analysis → skip to Step 4 with that request.
 
 ## Step 3 — Compute and print comparison table
 
+Run the default catalog entry (`latency-table`):
+
 ```bash
-python .claude/skills/sim2real-analyze/scripts/compute_table.py --run <name>
+python .claude/skills/sim2real-analyze/analyses/latency_table.py --run <name>
 ```
 
 Print the full output to the user. If the script exits with code 1, surface the error and stop.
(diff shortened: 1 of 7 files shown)


kalantar:
## Independent Review Verdict

**READY TO MERGE**

Diff is coherent and well-tested; the specialists' findings are all polish-level, and only two clear enough to note in passing.

### Recommendations (advisory — do not block merge)

- **`.claude/skills/sim2real-analyze/analyses/per-request-scatter.md:51` — TPOT formula contradicts its own validity clause.** Lines 51-52 spell out `TPOT = ... / max(output_tokens-1, 1) / 1000` and immediately annotate `(only valid when output_tokens > 1)`. The `max()` guard silences the div-by-zero but does not filter single-token rows, so an LLM implementing the formula literally will emit invalid values for `output_tokens == 1` while the parenthetical says it should not. `latency_table.py` already gets this right by pre-filtering rows to `output_tokens > 1` before dividing. *(source: comment-analyzer)*
- **`.claude/skills/sim2real-analyze/scripts/list_analyses.py:75` — `_validate` accepts an `entry_path` argument it never uses.** Signature at line 75 takes `entry_path: Path` but the function body (lines 76-97) never references it; the only call site at line 117 passes `md` for no effect. Either drop the parameter or wire it into the returned error messages so future maintainers do not think the slot is already available for path-contextual diagnostics. *(source: comment-analyzer)*

*(Verdict recovered from workflow journal after the automated post step tripped an API filter.)*

PR #552, opened 2026-07-09
feat(progress): scope ConfigMap by (scenario, run) to end cross-experiment-root bleed (#551)

kalantar:
Closes #551.

## Problem

Progress ConfigMaps collided across experiment roots that share a run name.
Two experiment repos with a `trial-1` run pointed at the same
`sim2real-progress-trial-1` ConfigMap, so state from one leaked into the
other — the specific symptom that motivated this issue was 23 stale
"pending" entries from an earlier experiment root surfacing as pending
work in `deploy.py run --status` for a new experiment.

Kubernetes identity is `(kind, namespace, name)` — labels alone cannot
disambiguate; the name itself has to carry the scoping.

## Fix

Include the already-required `scenario` field from `transfer.yaml` in the
ConfigMap name. New shape: `sim2real-progress-<scenario>-<run>`.

- `sim2real assemble` writes `scenario` into `run_metadata.json` (extracted
  from the manifest that's already loaded).
- `deploy.py` reads `run_metadata.json:scenario` via a new
  `_make_progress_store(namespace, run_dir)` helper, replacing 6 direct
  `ConfigMapProgressStore(...)` construction sites in `_cmd_status`,
  `_cmd_collect`, `_cmd_run`, `_cmd_reset`, `_cmd_wipe`, and
  `_cmd_run_remote`.
- `ConfigMapProgressStore` gains a `scenario` kwarg and composes the new
  name.
- Discovery labels `sim2real.scenario` and `sim2real.run` are added on
  `save()` so operators can `kubectl get cm -l sim2real.run=<r>` when
  poking around.

## Legacy migration

Existing operators have `sim2real-progress-<run>` ConfigMaps in flight.
On `load()`, if the new-name CM is `NotFound`, look up the legacy name;
if it exists, migrate its data into the new-name CM and delete the legacy
CM. One-shot, invisible, safe under crash between steps (a partial
migration is picked up on the next load, since the new-name CM already
has the data before the delete runs).

## Wiring choice (vs the issue body's proposal)

The issue's proposal said "deploy.py reads transfer.yaml" — but
`deploy.py` doesn't currently read `transfer.yaml`, and coupling it to
the experiment repo layout would be a regression. Instead the scenario
flows through `run_metadata.json`, which `deploy.py` already reads for
`cluster_id`. Result: `deploy.py` still consumes only workspace
artifacts, and `sim2real assemble` remains the sole reader of
`transfer.yaml`.

## Defense-in-depth (AC5)

If a run somehow reaches `deploy.py` without a `scenario` in its
`run_metadata.json` (e.g., a legacy run assembled before this PR that
hasn't been re-assembled), `_make_progress_store` exits cleanly with:

    run_metadata.json is missing 'scenario' — re-run
    'sim2real assemble --run <name>' to record it (added in issue #551).

Additive-grow of a legacy run refreshes the field from the manifest,
so `sim2real assemble --run <name> --replicas <n>` on an in-flight
legacy run also migrates the metadata.

## Verification checklist (from #551)

- [x] AC1 `sim2real-progress-<scenario>-<run>` created with the correct
  name on fresh runs — `test_configmap_name_includes_scenario_and_run`
- [x] AC2 Second experiment root with the same run name lands in a
  distinct ConfigMap — implicit in the name formula; regression-guarded
  by the sanitize + name-composition tests
- [x] AC3 All 6 subcommand call sites resolve to the new name — all
  routed through `_make_progress_store`; existing subcommand tests pass
- [x] AC4 Legacy migration —
  `test_load_migrates_legacy_when_new_name_notfound` covers the full
  copy-then-delete sequence
- [x] AC5 `deploy.py` fails cleanly when scenario is missing —
  `test_missing_scenario_exits`, `test_empty_scenario_exits`
- [x] Optional labels — `test_save_emits_discovery_labels_...`,
  `test_save_no_labels_when_scenario_missing`

## Docs swept

Grepped for `sim2real-progress-{run}` / `sim2real-progress-<run>` and
`run_metadata.json` schema references. Updated:

- `pipeline/README.md` — `run_metadata.json` schema entry gets `scenario`;
  4 ConfigMap-name references updated to the new shape (one keeps the
  legacy shape as documentation of the migration).
- `CLAUDE.md` — workspace-artifacts table entry updated.

Historical plan docs (`docs/superpowers/plans/2026-06-30-*`,
`2026-07-01-*`) left alone as snapshots.

## Test plan
- [x] `python -m pytest pipeline/` — 1586 passed, 2 xfailed (pre-existing).
- [x] CI-parity suite (all files listed in `.github/workflows/test.yml`)
  runs green locally.

**Base:** `main`.


kalantar:
## Code Review — NEEDS CHANGES

The core ConfigMap re-scoping and legacy migration are solid, but the new scenario surface has a dead-code fallback with a misleading comment, uncovered test paths, and a stale ConfigMap name in the Quick Start prose that will misdirect operators.

### Important (blocks merge)

**1. Stale ConfigMap name format in Quick Start prose** — `pipeline/README.md:609`

Lines 425, 462, 716, 797-799 use `sim2real-progress-{scenario}-{run}` but line 609 still reads "Progress lands in the run-scoped `sim2real-progress-<run-name>` ConfigMap." An operator following Quick Start will look for a ConfigMap that does not exist under the new naming.

_(source: comment-analyzer)_

**2. Dead-code fallback in `resolve.py` with a factually wrong comment** — `pipeline/lib/resolve.py:114`

`_TRANSLATION_TOP_KEYS = ('scenario', 'component', 'context')` in `pipeline/lib/slicer.py:56,96` — `assembly_slice()` explicitly skips those keys, so `manifest.assembly.yaml` never contains `scenario`. The fallback `manifest_assembly.get("scenario")` in `resolve_run()` cannot fire; the comment claims a legacy recovery path that doesn't exist. `pipeline/tests/test_assemble_run.py:192` (`assert "scenario" not in parsed`) directly confirms the slice omits it.

_(source: code-reviewer / silent-failure-hunter / comment-analyzer)_

**3. `resolve_run()`'s new `scenario` field on the primary path is never asserted** — `pipeline/tests/test_resolve.py:88`

`_make_run` writes `run_metadata.json` without a `scenario` key. Grepping `test_resolve.py` for `result["scenario"]` returns nothing. Every test exercises only the (dead) `manifest_assembly` fallback because the fixture manually inserts `scenario` into `manifest.assembly.yaml` at line 122. A regression breaking the `run_meta` primary path would pass the current suite.

_(source: pr-test-analyzer / code-reviewer)_

**4. `_additive_grow` scenario backfill has no test** — `pipeline/tests/test_assemble_replicas.py:1`

`pipeline/lib/assemble_run.py:724` sets `rm["scenario"] = manifest.get("scenario", "") or ""` inside the additive-grow write. No test starts from a pre-#551 `run_metadata.json` (no `scenario`), grows, and asserts `scenario` is now populated. Since `deploy.py` hard-fails on empty scenario, a silent regression here would break the operator-facing migration story for legacy runs.

_(source: pr-test-analyzer)_

### Recommendations (non-blocking)

**5. Legacy delete failure during migration is reported as "Cluster unreachable"** — `pipeline/lib/progress.py:103`

`load()` calls `save()` (writes the new-name CM successfully) then `_delete_cm()`; a delete failure raises `RuntimeError`, which `_load_progress` converts to `ProgressUnavailable` under `allow_unreachable=True` (`deploy.py:284-286`) and callers print "Cluster unreachable" (`deploy.py:729, 1659`). The cluster is reachable and the migration is complete at that point — only the legacy-CM cleanup leaked. Consider swallowing the delete failure as a warning.

_(source: silent-failure-hunter)_

---

_This review ran the review-changes Workflow (four specialist reviewers → adversarial synthesizer, each in a fresh context). Posted manually because the workflow's final PR-post step tripped on a `base64 -w` flag that macOS `base64` doesn't accept._

kalantar:

Implementation matches the design as revised during review; the migration policy change from auto to manual is intentional and documented, and remaining specialist findings are either re-litigation of decided design points or minor polish.

### Recommendations
- **Labels test only exercises an already-clean scenario, so a regression to using raw self._scenario in save() would pass unnoticed.** (`pipeline/tests/test_progress.py:172`) — test_save_emits_discovery_labels_when_scenario_and_run_supplied uses scenario='softr' and run_name='trial-1', both of which are already sanitized-identical. save() at progress.py:114-118 reads self._sanitized_scenario/self._sanitized_run; substituting self._scenario/self._run_name would still satisfy the current asserts. Adding one case with scenario='Soft_R' asserting labels['sim2real.scenario'] == 'soft-r' would close that hole.
- **_make_progress_store does not wrap the ConfigMapProgressStore constructor, so a >253-char combined name produces a raw traceback instead of the helper's usual clean err()/exit(1).** (`pipeline/deploy.py:237`) — Every other failure mode in _make_progress_store (missing file, invalid JSON, blank scenario) exits cleanly via err() + sys.exit(1). The final return at deploy.py:237-239 leaves the ValueError from progress.py:61-66 uncaught at all six call sites. Corner case (requires very long scenario+run), but the operator-friendly contract is inconsistent.

PR #558, opened 2026-07-09
fix(remote): encode pipe-shape pipelinerun filenames in run-inputs ConfigMap keys (#557)

kalantar:
Closes #557.

## What changed

`deploy.py run --remote` was failing at "Applying run-inputs ConfigMap" with the error `data[cluster--pipelinerun-<w>|<p>|iN.yaml]: Invalid value ... regex used for validation is '[-._a-zA-Z0-9]+'`. Kubernetes' ConfigMap-data-key validator rejects `|`, which the pipe-shape pair-key grammar (`lib/pairkey.py`) puts into every pipelinerun filename on disk.

Fix is a round-trip encoding at the two matching sites in `pipeline/lib/remote.py`:

- **Write side** (`build_run_inputs_configmap`, previously line 79): filename → CM key via `_encode_filename_for_cm`, which currently replaces `|` with `.7C` (URL-hex style; `.` is legal in CM keys, `%` is not).
- **Read side** (`_configmap_items`, previously lines 105–107): CM key → filename via `_decode_filename_from_cm`, so the projected volume `path:` still points at `runs/<R>/cluster/pipelinerun-<w>|<p>|iN.yaml` inside the pod. Round-trip is byte-exact.

Helpers use an `_CM_KEY_ESCAPES` dict so if another disallowed character ever shows up in cluster/ filenames, it's a one-line addition — no changes to the call sites. Only `|` is in scope today.

## Tests

Added to `pipeline/tests/test_remote.py`:

- Direct helper tests: `_encode_filename_for_cm` maps `|` → `.7C`; no-pipe input passes through; `_decode_filename_from_cm` is the exact inverse.
- Parametrized round-trip over pipe-shape and no-pipe fixtures — asserts encoded output matches `_CM_KEY_ALLOWED_RE` and decoding recovers the original.
- Integration: `build_run_inputs_configmap` fed a `cluster/` dir with pipe-shape filenames produces zero keys containing `|`, and every key matches the k8s allowed regex.
- End-to-end round-trip: `build_run_inputs_configmap` → `_configmap_items` reconstructs the original pipe-shape filename in the volume `path:`.
- A regex-drift guard test on `_CM_KEY_ALLOWED_RE` so a future upstream change would surface loudly.

Full pipeline test suite: 1602 passed, 2 xfailed (unchanged from `main`). Ruff clean.

## Reviewer attention

- **Encoding choice.** `.7C` was picked over `__pipe__` for compactness and URL-percent-encoding parity. Could equally be `__pipe__` — it's isolated in `_CM_KEY_ESCAPES` so easy to swap. Argument for `.7C`: standard convention, terse; against: `.` interleaves with filename dots. If reviewers prefer `__pipe__` (or another sentinel), the change is one dict entry plus a test string update.
- **Test fixture caveat.** `_write_defaults` seeds `pipelinerun-smoke-baseline.yaml` in the `cluster/` dir; the new round-trip integration test explicitly includes that entry in its expected set (see comment inline). Not a real footgun; called out for future readers.
- **Non-goal.** This PR does not rename pipelinerun filenames on disk to avoid `|` — that would require reversing the pair-key grammar migration and is much larger. Encoding-in-CM-only is the smaller surface.

## Stale-reference sweep

Grepped `**/*.md` and `pipeline/*.py` for `cluster--`, `run-inputs`, `sim2real-run-inputs`. One hit outside `remote.py` — `pipeline/deploy.py:3622` code comment `"cluster--*.yaml entries"` — still accurate since the `cluster--` prefix survives; only the suffix's `|` is now encoded. No doc, skill, or plan-file references to the CM-key format itself. Nothing to update.

### DIFF
--- a/pipeline/lib/remote.py
+++ b/pipeline/lib/remote.py
@@ -1,5 +1,6 @@
 """Remote run support — Kubernetes resource generation for --remote mode."""
 
+import re
 from pathlib import Path
 
 CONFIGMAP_NAME = "sim2real-run-inputs"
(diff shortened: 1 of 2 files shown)


kalantar:

The fix is correct and tested, but the new helper's docstring promises caller validation that the caller doesn't perform, leaving the exact kubectl-rejection failure mode open to recur on any future illegal filename character.

### Important
- **`_encode_filename_for_cm` docstring promises caller validation, but `build_run_inputs_configmap` never performs it.** (`pipeline/lib/remote.py:120`) — The docstring at pipeline/lib/remote.py:38 explicitly says 'Callers should verify the result is legal before use; see `_CM_KEY_ALLOWED_RE`.' but the sole caller (line 120) does not. If any future cluster/*.yaml filename introduces a non-`|` illegal character not present in `_CM_KEY_ESCAPES` (e.g., a space, `@`, `/`), the PR reintroduces exactly the silent kubectl-rejection failure mode it is meant to eliminate — no Python-level error, just an opaque `Invalid value` at apply time. Three of four specialists independently flagged this and the fix is a one-line assertion after encoding (or dropping the caller-verify clause from the docstring). Either reconcile the docstring with reality or add the assertion.

### Recommendations
- **Drift-guard test docstring overstates what the test can detect.** (`pipeline/tests/test_remote.py:474`) — Verified at pipeline/tests/test_remote.py:474: `test_cm_key_allowed_re_matches_k8s_validator` only exercises the local `_CM_KEY_ALLOWED_RE` against hardcoded strings — it has no dependency on the Kubernetes source, so it cannot 'surface the drift' if upstream changes. The PR body repeats the same claim. Rewrite the docstring to describe what the test actually does (regression samples for the expected pattern) rather than promising drift detection.

PR #561, opened 2026-07-09
feat(sim2real-analyze): add five deferred single-run catalog entries

kalantar:
## Summary

Closes #560. Fills in the catalog scaffolded by #548 with the five entries that PR body called out as "trivial per-file follow-ups once the scaffolding lands":

- `ttft-cdf`, `tpot-cdf`, `e2e-cdf` — one panel per workload, baseline vs treatment overlaid empirical CDF, three metric wrappers over a shared `_cdf.cdf_main`.
- `throughput-over-time` — sends/sec vs time, phases as lines, workloads as panels.
- `error-rate` — per-workload table of total / errors / rate / top non-ok statuses. The only entry that does NOT filter `status == "ok"` (errors are the metric).

`list_analyses.py` output goes from 2 → 7 entries with no warnings.

## Design decisions worth reviewer attention

1. **`_common.py` + `_cdf.py` shared helpers.** Rather than duplicate `resolve_run` / `display_name` / `load_csv` / `discover_workloads` across five scripts, I extracted them into `analyses/_common.py`. Underscore-prefixed so it's clearly a helper, not a catalog entry (the discovery machinery already ignores non-`.md` files, but the prefix documents intent). CDF plotting lives in `analyses/_cdf.py` so `error_rate.py` doesn't drag in matplotlib. Alternative: inline everything into each of the five scripts (matches `latency_table.py`'s self-contained pattern). Reviewer's call — happy to inline if that's the house preference.

2. **`throughput-over-time` uses `send_time_us`, not `arrival_time_us`.** The SKILL.md pedagogical snippet used `arrival_time_us`, but `arrival_time_us` isn't in `_common.REQUIRED_COLS` (which mirrors `latency_table.py`'s `REQUIRED_COLS`). Using `send_time_us` keeps the column-set contract consistent across the catalog. Semantic difference: sends-per-second measures load offered from the client, which is the natural view for "did the two phases see the same load shape?". Documented in `throughput-over-time.md`.

3. **`test_list_analyses.py::test_real_catalog_has_two_entries` relaxed to subset check.** The old assertion was `names == {latency-table, per-request-scatter}` — hard-coded exact-count. Renamed to `test_real_catalog_contains_expected_entries` with the seven-entry set and `missing = expected - names; assert not missing`. This tolerates future additions without another test edit, but a *removal* still fails loudly.

4. **SKILL.md sweep.** The throughput-over-time example snippet (lines 258–284 pre-PR) is now redundant with the catalog entry — deleted. The TTFT histogram example is preserved as the one-off-script template for the fallback flow, with a note explaining when to reach for the one-off vs the CDF entry.

## What warrants extra scrutiny

- **Test coverage.** New tests cover: `_empirical_cdf` unit, each metric wrapper produces a valid PNG, TPOT-all-single-token exits cleanly, all-error rows exits cleanly, unknown metric raises, config-default run resolution. Throughput adds `_rate_series` bucketing tests. Error-rate adds `_format_pct` / `_format_row` unit tests plus a happy path with mixed statuses. If a reviewer wants finer-grained coverage of a particular script's plotting output (matplotlib artefact inspection beyond "file exists and non-empty"), say so — I punted on that in favor of the smoke-test standard `test_latency_table.py` uses.

- **Import contract.** Scripts `import _common` / `import _cdf` at the top level; this works when run as `python analyses/foo.py` (sys.path[0] = the script dir) and when tests import them (conftest.py prepends `analyses/` to sys.path). If the skill runner ever invokes them differently (e.g., as a module) this will need revisiting — but the SKILL.md contract is `python .../analyses/<script> --run <name>`, so it holds.

## Verification

- [x] `python3 .claude/skills/sim2real-analyze/scripts/list_analyses.py` → emits 7 entries, no stderr warnings
- [x] `ruff check pipeline/ .claude/skills/ --select F` → clean
- [x] `pytest` full CI suite (pipeline + all `.claude/skills/*/tests/`) → 1817 passed, 1 skipped, 2 xfailed
- [x] Smoke test — each new script runs end-to-end on a synthetic workspace produced by tests; PNGs render at nonzero size, text file writes
- [ ] Manual smoke of `/sim2real-analyze` interactive loop against a real run — not exercised in this session (no local run available), reviewer or follow-up

## Deferred, still

Bundle analyses (`latency-pairwise-*`, `claim-verification-report`, heatmaps) remain deferred — they need the unlanded `sim2real resolve` refactor. Explicitly out of scope per #560's Non-goals.

## Test plan

- [ ] Reviewer confirms `_common.py` extraction is the right level (vs inline duplication)
- [ ] Reviewer sanity-checks `send_time_us` vs `arrival_time_us` throughput semantic
- [ ] Reviewer runs `/sim2real-analyze` menu against a real run and confirms all five entries are offered

### DIFF
--- a/.claude/skills/sim2real-analyze/SKILL.md
+++ b/.claude/skills/sim2real-analyze/SKILL.md
@@ -235,7 +235,7 @@ histogram at a different bin size or a workload subset.
 For the standard CDF view use `ttft-cdf` from the catalog. Reach for a
 per-workload histogram when you need to see bin density directly (e.g.,
 "is the TTFT distribution bimodal at 50ms increments?") rather than
-the smoothed cumulative view.
+the empirical step-function CDF the catalog entry produces.
 
 ```python
 import pandas as pd
--- a/.claude/skills/sim2real-analyze/SKILL.md
+++ b/.claude/skills/sim2real-analyze/SKILL.md
@@ -223,9 +223,19 @@ All timestamps are **microseconds**. Metrics:
 - TPOT = (last_chunk_time_us - first_chunk_time_us) / (output_tokens - 1) / 1000 ms (output_tokens > 1 only)
 - E2E  = (last_chunk_time_us - send_time_us) / 1000 ms
 
-## Example analysis scripts
+## Example one-off analysis script
 
-### TTFT distribution histogram
+Common analyses live in the catalog (see previous section). This
+example shows the shape of a one-off script for the fallback flow when
+you need something the catalog doesn't cover — for instance, a
+histogram at a different bin size or a workload subset.
+
+### TTFT distribution histogram (one-off example)
+
+For the standard CDF view use `ttft-cdf` from the catalog. Reach for a
+per-workload histogram when you need to see bin density directly (e.g.,
+"is the TTFT distribution bimodal at 50ms increments?") rather than
+the smoothed cumulative view.
 
 ```python
 import pandas as pd
@@ -251,34 +261,6 @@ for ax, wl in zip(axes, workloads):
 
 plt.tight_layout()
 out = f"workspace/runs/{run}/results_charts/ttft_distribution.png"
-plt.savefig(out)
-print(f"Saved: {out}")
-```
-
-### Throughput over time
-
-```python
-import pandas as pd
-import matplotlib.pyplot as plt
-
-run = "<name>"
-wl = "workload_fm8_short_output_highrate"
-
-fig, ax = plt.subplots(figsize=(12, 4))
-for phase in ["baseline", "treatment"]:
-    df = pd.read_csv(f"workspace/runs/{run}/results/{phase}/{wl}/trace_data.csv")
-    df = df[df["status"] == "ok"]  # only compute metrics for successful requests
-    t0 = df["arrival_time_us"].min()
-    df["t_sec"] = (df["arrival_time_us"] - t0) / 1e6
-    counts = df.groupby(df["t_sec"].astype(int)).size()
-    ax.plot(counts.index, counts.values, label=phase)
-
-ax.set_xlabel("Time (s)")
-ax.set_ylabel("Requests/s")
-ax.set_title(f"Throughput over time — {wl.replace('workload_','').replace('_','-')}")
-ax.legend()
-plt.tight_layout()
-out = f"workspace/runs/{run}/results_charts/throughput_over_time.png"
 plt.savefig(out)
 print(f"Saved: {out}")
 ```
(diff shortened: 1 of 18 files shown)


kalantar:

Catalog entries are functionally sound and well-tested, but throughput-over-time.md misrepresents what the chart measures and one error-rate test is mislabeled.

### must_fix

_(none)_

### important (blocks merge)

**1. Throughput doc claims `load offered to the server` but code filters to `status == "ok"`**  
`.claude/skills/sim2real-analyze/analyses/throughput-over-time.md:50` · source_lens: comment-analyzer

`throughput_over_time.py` builds `ok = [r for r in rows if r["status"] == "ok"]` and then computes `t0 = min(times_us)` inside `_rate_series(ok)`. The doc's "`send_time_us` measures load offered" claim (line 50) and the "Time origin is the first send" claim (line 18) both describe raw sends. When phases have different error rates, an operator following this doc will misattribute a shape divergence to load-gen rather than to error behavior. Fix is a one-line caveat in the doc or dropping the `ok` filter for this specific chart.

### recommendations (advisory)

**1. `test_main_empty_csv_is_skipped` name and docstring describe a scenario the test does not exercise**  
`.claude/skills/sim2real-analyze/tests/test_error_rate.py:96` · source_lens: pr-test-analyzer, comment-analyzer

The docstring reads `Workloads with empty rows in both phases are skipped, not fatal`, but the test body adds a second workload (`workload_beta`) with one valid `ok` row in each phase and asserts both workloads appear. `_common.load_csv` actually `sys.exit(1)`s on an empty CSV via the `if not text.strip()` guard, so the described scenario is not just untested — it is the opposite of the real behavior. Rename to reflect what the test does (multiple workloads report cleanly) or add a real empty-file case.

### Notes on dropped findings

- Two lenses (pr-test-analyzer, code-reviewer) flagged "no test file for `throughput_over_time.py`" as blocking. Dropped by the synthesizer on code re-read: `tests/test_throughput_over_time.py` exists in the diff.
- silent-failure-hunter flagged three findings against `_common.load_csv` (UnicodeDecodeError not caught; OSError emits wrong-category message; broad `except Exception`). Dropped — these are pre-existing patterns carried over from `latency_table.py`; this PR relocates them into a shared helper but does not introduce them.

_Posted manually because the workflow's `gh pr comment` step failed on a macOS base64 flag incompatibility. All 7 review agents completed successfully; only the final post step errored._

kalantar:

Well-structured catalog additions with reasonable test coverage and clean shared extraction; specialist blockers either restate pre-existing patterns copied from latency_table.py or overstate a test-coverage claim, leaving only a minor doc-wording nit.

### Recommendations
- **"smoothed cumulative view" mischaracterizes the empirical CDF, which is a step function** (`.claude/skills/sim2real-analyze/SKILL.md:238`) — SKILL.md line 234-238 contrasts the histogram against "the smoothed cumulative view", but `_cdf.py:28-39` (`_empirical_cdf`) computes y = k/n step points and its module docstring explicitly calls it "the standard textbook empirical CDF" — no kernel smoothing or interpolation is applied. "Empirical" or "step-function" is the precise word.

PR #563, opened 2026-07-09
fix(deploy,remote): mirror EXPERIMENT_ROOT into layout at startup

### LINKED ISSUE #562: deploy.py --remote: orchestrator pod misresolves cluster_config via cwd, then emits a local-cluster hint the pod can't act on
## Symptom

Fresh `deploy.py --run trial-1 run --remote` against a workspace that is fully provisioned locally: the local caller applies the run-inputs ConfigMap, applies the orchestrator Job, and reports the pod started. The pod exits 1 within a few seconds with:

```
━━━ sim2real-deploy ━━━

20:56:49 [ERROR] No namespaces configured. Run cluster.py provision with --namespaces.
```

The local caller had `namespaces` set in `workspace/clusters/<id>/cluster_config.json` (otherwise the pre-flight check at `pipeline/deploy.py:3535-3539` would have short-circuited before applying anything). So the pod is reading an *empty* view of the same config the local process just uploaded — and then telling the operator to fix it by running a command the pod can't run.

## Root cause

`pipeline/deploy.py:3809` (main dispatcher) calls `_load_run_cluster_config(run_dir)`, which is the per-run dispatch introduced by #446. That helper:

```python
# pipeline/deploy.py:174
def _load_run_cluster_config(run_dir: Path) -> dict:
    ...
    cluster_id = meta.get("cluster_id") if isinstance(meta, dict) else None
    ...
    return cluster_ops.read_cluster_config(cluster_id)   # line 203
```

Never calls `layout.set_experiment_root(...)`. `cluster_ops.read_cluster_config` resolves the path via `layout.cluster_config_path(cluster_id)`:

```python
# pipeline/lib/cluster_ops.py:94
def read_cluster_config(cluster_id: str) -> dict:
    path = layout.cluster_config_path(cluster_id)
    if not path.exists():
        return {}                      # <-- silent empty
    ...
```

And `layout.cluster_config_path` → `layout.cluster_dir` → `layout.clusters_dir` → `layout.workspace_dir` → `layout.experiment_root()`:

```python
# pipeline/lib/layout.py:45
def experiment_root() -> Path:
    return _EXPERIMENT_ROOT if _EXPERIMENT_ROOT is not None else Path.cwd()
```

**In the orchestrator pod**, the entrypoint is `python .../deploy.py --experiment-root /data --run <N> run --skip-build`. `main` sets its own `EXPERIMENT_ROOT` module global (line 3796-3797) but does **not** call `layout.set_experiment_root(EXPERIMENT_ROOT)`. So `layout._EXPERIMENT_ROOT` stays `None`, and `layout.experiment_root()` falls back to `Path.cwd()`. Whatever `WORKDIR` the container image ships is used as the workspace root — not `/data`.

Result: `read_cluster_config` looks for `<cwd>/workspace/clusters/<id>/cluster_config.json`, doesn't find it, returns `{}`. `_cmd_run` at line 2741-2743 sees `namespaces = []` and exits.

**Locally this happens to work** because operators run `deploy.py` from the experiment root, so `Path.cwd()` and `EXPERIMENT_ROOT` coincide. The bug only surfaces when `--experiment-root` diverges from `cwd`, which is exactly the remote-pod case.

Compare with the *legacy* `_load_cluster_config` (line 150), which does the right thing:

```python
# pipeline/deploy.py:158
layout.set_experiment_root(EXPERIMENT_ROOT)
cluster_ids = layout.list_cluster_ids()
```

`_load_run_cluster_config` (the actual code path from `main`) dropped that setup step.

## Misleading error-message class

Every namespace/cluster-not-configured error emits `"Run cluster.py provision with --namespaces."` — a hint for the *local* operator to run a local command. When the same code runs inside the orchestrator pod, that hint is unreachable: the pod can't invoke `cluster.py`, and the real problem is either (a) the path-resolution bug above, or (b) missing config in the uploaded ConfigMap. Neither is fixed by running `cluster.py provision`.

All occurrences in `pipeline/deploy.py`:

- `pipeline/deploy.py:740` — `_cmd_status`
- `pipeline/deploy.py:1670` — inside a runtime function; also namespace-not-found
- `pipeline/deploy.py:2743` — `_cmd_run` (the one the pod actually hits)
- `pipeline/deploy.py:2763` — `_cmd_run`, second namespace check
- `pipeline/deploy.py:3205` — another handler
- `pipeline/deploy.py:3259` — another handler
- `pipeline/deploy.py:3538` — `_cmd_run_remote` pre-flight (only fires locally, message is appropriate here)
- `pipeline/deploy.py:3814` — `main` build branch
- `pipeline/deploy.py:3854` — `main` stop branch

At least the three inside `_cmd_run` (2743, 2763) can fire *inside* the orchestrator pod. If the path bug is fixed and the config-in-pod is still empty for some other reason, the pod will still emit "Run cluster.py provision" — misleading. See suggestion below.

Related structural risk: `pipeline/deploy.py:3814` and `3854` are on the `main` dispatch path and are reachable from a pod that receives a `build` or `stop` subcommand. Today the orchestrator Job only launches `run --skip-build`, so the pod won't hit them — but if the pod's args ever change, they will fire silently with the same misleading hint.

## Adjacent audit findings

Beyond the primary bug, several other places in the codebase share the "resolve path via `layout` without ensuring `set_experiment_root` was called" shape. These are lower-severity — some are only reached from local commands today, but each is a latent version of the same trap:

- `pipeline/sim2real.py:1178` — `layout.list_cluster_ids()` inside a pre-flight check.
- `pipeline/sim2real.py:1490` — `layout.experiment_root()` used to construct a run dir path.
- `pipeline/sim2real.py:1194` — `cluster_ops.read_cluster_config(cluster_ids[0])` — same downstream layout dependency.
- `pipeline/lib/assemble_run.py:787` — `cluster_ops.read_cluster_config(cluster_id)`; `assemble_run.assemble_run()` receives `experiment_root` as a parameter (line 251) but the callers (`sim2real.py:1439`) don't call `layout.set_experiment_root` before invoking it. This is only reached from `sim2real assemble`, which today only runs locally.
- `pipeline/deploy.py:2711` — `_refresh_namespaces` calls `_load_cluster_config()`, which does the right thing. Included here only to note the two-loader split.

The pattern is: **any function that uses `cluster_ops`, `layout.*` path helpers, or `layout.experiment_root()` without calling `layout.set_experiment_root(EXPERIMENT_ROOT)` first is a landmine** the moment its caller is invoked from a directory that is not the experiment root. Today only the pod hits that condition; tomorrow anything invoked from CI or a wrapper script does.

## Suggested direction (not a spec — for the implementer to refine)

1. **Fix the primary bug.** Call `layout.set_experiment_root(EXPERIMENT_ROOT)` from `main()` in `pipeline/deploy.py` (right after line 3797 where the global is set), so every downstream use of `layout.*` and `cluster_ops.*` resolves against `--experiment-root`, not `cwd`. This is a one-line fix and it eliminates the pod symptom for the current case. Alternatively, fix `_load_run_cluster_config` itself to call `set_experiment_root` — matches the legacy `_load_cluster_config` pattern, keeps the responsibility inside the loader — but the `main()` fix is broader and cheaper.

2. **Rewrite the misleading error hints on the paths reachable inside the pod.** Lines 2743 and 2763 should either:
   - Detect that they are running inside the orchestrator (e.g. an env var set by `build_orchestrator_job`) and emit a hint like `"cluster_config.json in the run-inputs ConfigMap has no namespaces — re-run 'sim2real assemble --run <N>' or check the run's run_metadata.json:cluster_id"`, OR
   - Emit both hints and let the operator pick which applies.
   
   The current message optimizes for the local case (which is where the check should have already caught the problem) and misleads the pod case (which is the case that actually gets here).

3. **Consider making layout resolution eager.** `layout.experiment_root()` falling back to `Path.cwd()` is the trap. If the fallback instead raised `RuntimeError("layout.set_experiment_root() not called")` when `_EXPERIMENT_ROOT is None`, every consumer would fail loudly at the boundary rather than silently reading the wrong config. This turns latent bugs into loud ones and would have caught the current issue at commit time. Higher blast radius — some tests probably rely on the `Path.cwd()` fallback — but worth weighing.

4. **Audit the adjacent findings above and either fix each site or add a lint that flags `cluster_ops.read_cluster_config(...)` calls not preceded by `layout.set_experiment_root(...)`.**

## Reproducer

Only reproducible with `--remote` today, because that's the only path where `Path.cwd()` diverges from `--experiment-root`. The operator's original invocation:

```
python $SIM2REAL/pipeline/deploy.py --run trial-1 run --remote
```

with local `workspace/clusters/<id>/cluster_config.json` containing a valid `namespaces` list. Local pre-flight passes; orchestrator pod fails within ~5s with the misleading hint above.

## Acceptance criteria

- After the fix, `deploy.py --run <N> run --remote` against a fully-provisioned workspace succeeds through the pod's `_cmd_run` initialization (i.e. the pod does not emit "No namespaces configured" for a config that was present in the uploaded ConfigMap).
- If the pod DOES legitimately get an empty `cluster_config` (e.g. the run-inputs ConfigMap really was missing the namespaces list), the error message it emits is actionable from inside the pod's context — not a hint to run a local command.
- A test exercises the `EXPERIMENT_ROOT != cwd` case for `_load_run_cluster_config` so this doesn't regress silently again.

## Source

- `pipeline/deploy.py:174` (`_load_run_cluster_config`) — the buggy loader
- `pipeline/deploy.py:3796-3809` (main dispatcher) — where the fix belongs
- `pipeline/deploy.py:2733-2743` (`_cmd_run`) — where the misleading error fires
- `pipeline/lib/cluster_ops.py:94` (`read_cluster_config`) — the silent-empty fallback that hides the underlying resolution failure
- `pipeline/lib/layout.py:30-51` (`set_experiment_root` / `experiment_root`) — the `Path.cwd()` fallback trap
- Introduced by refactor/v2 → main merge #535 (per-run dispatch, #446)

kalantar:
Closes #562.

## What was broken

`deploy.py --run <N> run --remote` against a fully-provisioned workspace: local pre-flight passes, orchestrator Job launches, then the pod exits ~5s later with:

```
[ERROR] No namespaces configured. Run cluster.py provision with --namespaces.
```

The ConfigMap **was** mounted and the initContainer **did** copy `cluster_config.json` to `/data/workspace/clusters/<id>/cluster_config.json`. The pod just wasn't looking there: `Dockerfile:25` pins `WORKDIR /app`, the entrypoint receives `--experiment-root /data`, and `deploy.py:main` set its own module-level `EXPERIMENT_ROOT` **but never called `layout.set_experiment_root(EXPERIMENT_ROOT)`**. Every downstream `layout.experiment_root()` call fell back to `Path.cwd() = /app`, so `cluster_ops.read_cluster_config` searched `/app/workspace/…` (empty) and returned `{}` silently. `_cmd_run` then saw no namespaces and emitted a hint the pod can't act on.

## What changed

**1. `pipeline/deploy.py`** — extract `_init_experiment_root(args)` (lines 174–194) that resolves the arg **and** calls `layout.set_experiment_root(...)` in one shot. `main()` calls it (line 3853) instead of inlining the resolution. This is the one-line fix; the extraction just makes it independently testable.

**2. `pipeline/deploy.py`** — introduce `_no_namespaces_hint()` (lines 145–172) which detects the `SIM2REAL_ORCHESTRATOR_POD=1` env var and emits a hint pointing at `sim2real assemble` when inside the pod, keeping the historical `cluster.py provision` hint locally. Applied to the two pod-reachable sites in `_cmd_run` (line 2800 and 2820). The other seven sites of the same error string live on local-only code paths — unchanged per the issue's scoping note.

**3. `pipeline/lib/remote.py`** — `build_orchestrator_job` sets `SIM2REAL_ORCHESTRATOR_POD=1` in the container env so the marker is in place when the pod's deploy.py runs.

## What warrants reviewer attention

1. **`_init_experiment_root` vs an inline one-liner in `main()`.** I extracted it so the specific "cwd != experiment_root" regression is unit-testable without invoking `main()` (which needs a real workspace, real args, and does a lot before the check). Alternative: put the two lines inline in `main()` and skip the extraction. The extraction adds ~20 lines (including docstring) but buys three focused regression tests. Reviewer's call.

2. **Env-var marker approach for pod detection.** Alternative was checking `cwd != EXPERIMENT_ROOT` at error time — but post-fix those are equal in the pod, so we'd lose the disambiguator. The env var is set by the same place that builds the Job spec (`build_orchestrator_job` in `remote.py`), so the linkage is co-located and testable. Downside: two files must stay in sync (the constant `_ORCHESTRATOR_POD_ENV` in `deploy.py` and the string in `remote.py`). Considered exporting the constant from `remote.py` but that creates an import cycle risk; kept them as parallel literals with a test asserting the produced Job spec has the exact value.

3. **Scope: only 2 of 9 "run cluster.py provision" sites changed.** The issue's "Suggested direction 2" targets the pod-reachable sites (`_cmd_run` at 2800/2820). The other seven live on `_cmd_status`, `_cmd_collect`, `_cmd_reset`, `_cmd_wipe`, `_cmd_run_remote`, and the `main` branches for `build` and `stop` — all local-only paths where the historical hint is correct. Deliberately left alone; the issue's "adjacent audit findings" section flagged them as separate scope decisions.

## What I did NOT do

- **Did NOT audit adjacent trap sites** (`sim2real.py:1178,1194,1490`, `assemble_run.py:787`). They're only reachable from local CLI paths today, so the primary bug is fixed. Each is a latent version of the same trap; recorded in the issue as "audit findings", left for a follow-up.
- **Did NOT change `layout.experiment_root()` fallback behavior.** The issue flagged this as higher blast-radius; keeping it out of this PR.

## Verification

- `python3 -m pytest pipeline/ .claude/skills/*/tests/` → **1824 passed, 1 skipped, 2 xfailed** (delta +7 = new tests here)
- `ruff check pipeline/ .claude/skills/ --select F` → clean
- Sweep for stale refs: the exact old error string appears in seven `docs/superpowers/plans/*.md` files — historical planning-doc snapshots of past code state, not live docs. Left alone. No SKILL.md, README, or CLAUDE.md mentions the string or the new env var.

## Test plan

- [x] `_init_experiment_root` mirrors arg into layout when cwd differs — regression guard for the exact bug
- [x] `_init_experiment_root` falls back to cwd when arg absent / empty
- [x] Job spec includes `SIM2REAL_ORCHESTRATOR_POD=1` env var
- [x] `_no_namespaces_hint` returns local hint (with `cluster.py provision`) when env var absent
- [x] `_no_namespaces_hint` returns pod hint (with `sim2real assemble`, no `cluster.py`) when env var == "1"
- [x] `_no_namespaces_hint` only flips on literal "1" — not on truthy-looking "0", "false", or ""
- [ ] Manual smoke: `deploy.py --run <N> run --remote` against a real cluster — reviewer or follow-up

### DIFF
--- a/pipeline/deploy.py
+++ b/pipeline/deploy.py
@@ -138,6 +138,63 @@ def _discover_phases(cluster_dir: "Path") -> list[str]:
 
 # ── Setup config ─────────────────────────────────────────────────────────────
 
+# Set to "1" in build_orchestrator_job's env spec so downstream deploy.py
+# error paths can distinguish "user is running this locally" from "running
+# inside the orchestrator pod" and emit a hint the operator can act on
+# (issue #562).
+_ORCHESTRATOR_POD_ENV = "SIM2REAL_ORCHESTRATOR_POD"
+
+
+def _in_orchestrator_pod() -> bool:
+    return os.environ.get(_ORCHESTRATOR_POD_ENV) == "1"
+
+
+def _no_namespaces_hint() -> str:
+    """Return the operator-actionable "no namespaces" message for this context.
+
+    In the pod: the cluster_config that was uploaded via the run-inputs
+    ConfigMap didn't carry a ``namespaces`` list — an ``assemble``/pod-input
+    problem, not a cluster-provisioning problem. Running ``cluster.py`` here
+    is impossible; the pointer at ``sim2real assemble`` is what recovers.
+
+    Locally: the historical hint at ``cluster.py provision --namespaces``
+    is correct — this is where the operator adds namespaces to the
+    cluster's ``cluster_config.json``.
+    """
+    if _in_orchestrator_pod():
+        return (
+            "No namespaces in the run-inputs cluster_config — the uploaded "
+            "ConfigMap is missing them. Re-run 'sim2real assemble --run <N>' "
+            "locally (ensure the target cluster has namespaces provisioned) "
+            "and then 'deploy.py run --remote' again."
+        )
+    return "No namespaces configured. Run cluster.py provision with --namespaces."
+
+
+def _init_experiment_root(args) -> Path:
+    """Resolve EXPERIMENT_ROOT from CLI args and mirror it into ``layout``.
+
+    Every subsequent path resolution — ``layout.workspace_dir()``,
+    ``layout.cluster_config_path()``, ``cluster_ops.read_cluster_config()``
+    — reads from ``layout._EXPERIMENT_ROOT`` via ``layout.experiment_root()``.
+    That accessor falls back to ``Path.cwd()`` when the module global is
+    unset, which is subtly wrong in any context where ``--experiment-root``
+    differs from the current working directory — most visibly the
+    orchestrator pod, whose Dockerfile pins ``WORKDIR /app`` while the
+    entrypoint receives ``--experiment-root /data`` (issue #562).
+
+    Wiring layout at startup means every downstream consumer sees the same
+    experiment root the deploy.py module global does. ``set_experiment_root``
+    is idempotent, so re-calling in main is safe.
+    """
+    global EXPERIMENT_ROOT
+    EXPERIMENT_ROOT = (
+        Path(args.experiment_root).resolve() if args.experiment_root else Path.cwd()
+    )
+    layout.set_experiment_root(EXPERIMENT_ROOT)
+    return EXPERIMENT_ROOT
+
+
 def _load_setup_config() -> dict:
     path = EXPERIMENT_ROOT / "workspace" / "setup_config.json"
     if not path.exists():
@@ -2740,7 +2797,7 @@ def _cmd_run(args, run_dir: Path, cluster_config: dict) -> None:
 
     namespaces = cluster_config.get("namespaces") or []
     if not namespaces or not namespaces[0]:
-        err("No namespaces configured. Run cluster.py provision with --namespaces."); sys.exit(1)
+        err(_no_namespaces_hint()); sys.exit(1)
 
     registry_secret_name = (
         (cluster_config.get("secret_names") or {}).get("registry_creds") or ""
@@ -2760,7 +2817,7 @@ def _cmd_run(args, run_dir: Path, cluster_config: dict) -> None:
     cluster_dir = run_dir / "cluster"
     primary_ns = _configmap_namespace(cluster_config, namespaces)
     if not primary_ns:
-        err("No namespace configured. Run cluster.py provision with --namespaces."); sys.exit(1)
+        err(_no_namespaces_hint()); sys.exit(1)
     store = _make_progress_store(primary_ns, run_dir)
 
     # Derive GPU resource type from baseline scenario
@@ -3793,8 +3850,7 @@ def main():
     if not machine_readable:
         print(_c("36", "\n━━━ sim2real-deploy ━━━\n"))
 
-    global EXPERIMENT_ROOT
-    EXPERIMENT_ROOT = Path(args.experiment_root).resolve() if args.experiment_root else Path.cwd()
+    _init_experiment_root(args)
 
     setup_config = _load_setup_config()
 
--- a/pipeline/lib/remote.py
+++ b/pipeline/lib/remote.py
@@ -191,7 +191,13 @@ def build_orchestrator_job(
                         "name": "orchestrator",
                         "image": image,
                         "args": args,
-                        "env": [{"name": "PYTHONUNBUFFERED", "value": "1"}],
+                        "env": [
+                            {"name": "PYTHONUNBUFFERED", "value": "1"},
+                            # Marker so downstream deploy.py code paths can
+                            # emit pod-appropriate error hints instead of the
+                            # local "run cluster.py provision" message (#562).
+                            {"name": "SIM2REAL_ORCHESTRATOR_POD", "value": "1"},
+                        ],
                         "volumeMounts": [
                             {"name": "workspace", "mountPath": workspace_path},
                         ],
--- a/pipeline/tests/test_deploy_remote.py
+++ b/pipeline/tests/test_deploy_remote.py
@@ -882,3 +882,80 @@ def test_run_remote_skips_validation_when_configmap_unreachable(monkeypatch, tmp
     combined = captured.out + captured.err
     assert "ConfigMap unreachable" in combined
     assert "skipping pre-flight filter validation" in combined
+
+
+# ── _init_experiment_root — mirror args into layout (#562) ──────────────────
+
+def test_init_experiment_root_mirrors_arg_into_layout(tmp_path, monkeypatch):
+    """When --experiment-root is passed, layout.experiment_root() returns it,
+    NOT Path.cwd(). Regression guard for #562: without this, an orchestrator
+    pod (WORKDIR /app, --experiment-root /data) would resolve every
+    layout-mediated path against /app and read empty cluster_config.
+    """
+    from pipeline.lib import layout
+    # Force cwd to differ from experiment_root so the fallback would
+    # produce the wrong answer if the fix were removed.
+    other = tmp_path / "elsewhere"
+    other.mkdir()
+    monkeypatch.chdir(other)
+    exp_root = tmp_path / "workspace-root"
+    exp_root.mkdir()
+
+    args = argparse.Namespace(experiment_root=str(exp_root))
+    resolved = mod._init_experiment_root(args)
+
+    assert resolved == exp_root
+    assert layout.experiment_root() == exp_root
+    assert layout.experiment_root() != other
+
+
+def test_init_experiment_root_falls_back_to_cwd_when_arg_absent(tmp_path, monkeypatch):
+    """No --experiment-root → uses cwd for both the module global and layout."""
+    from pipeline.lib import layout
+    monkeypatch.chdir(tmp_path)
+    args = argparse.Namespace(experiment_root=None)
+    resolved = mod._init_experiment_root(args)
+
+    assert resolved == tmp_path
+    assert layout.experiment_root() == tmp_path
+
+
+def test_init_experiment_root_empty_string_treated_as_absent(tmp_path, monkeypatch):
+    """argparse-style empty-string arg (falsy) should fall back to cwd, not
+    resolve to Path('')."""
+    from pipeline.lib import layout
+    monkeypatch.chdir(tmp_path)
+    args = argparse.Namespace(experiment_root="")
+    resolved = mod._init_experiment_root(args)
+    assert resolved == tmp_path
+    assert layout.experiment_root() == tmp_path
+
+
+# ── _no_namespaces_hint — pod-vs-local context detection (#562) ─────────────
+
+def test_no_namespaces_hint_local_context(monkeypatch):
+    monkeypatch.delenv("SIM2REAL_ORCHESTRATOR_POD", raising=False)
+    msg = mod._no_namespaces_hint()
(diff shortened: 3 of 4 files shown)


kalantar:

The layout-mirroring fix and pod-context hint are correct, but acceptance criterion #3 is not covered end-to-end and the new tests leak module-global state in a way every other test file in this repo explicitly guards against.

### must_fix

_(none)_

### important (blocks merge)

**1. No test exercises `_load_run_cluster_config` with cwd != experiment_root, despite an explicit acceptance criterion asking for one**
`pipeline/tests/test_deploy_remote.py:887` · source_lens: code-reviewer, silent-failure-hunter, pr-test-analyzer

Issue #562's acceptance criteria state verbatim: *"A test exercises the EXPERIMENT_ROOT != cwd case for `_load_run_cluster_config` so this doesn't regress silently again."* The three added tests verify `_init_experiment_root` sets `layout.experiment_root` correctly, but none write a `workspace/clusters/<id>/cluster_config.json` in `tmp_path`, chdir elsewhere, and assert `_load_run_cluster_config` actually reads it. A future refactor that has `_load_run_cluster_config` compute paths from `mod.EXPERIMENT_ROOT` directly would silently reintroduce the exact class of bug the issue calls out; that is the specific regression path the AC exists to close.

**2. Three new tests mutate `layout._EXPERIMENT_ROOT` without a teardown fixture, violating the pattern every other test module in this repo follows**
`pipeline/tests/test_deploy_remote.py:889` · source_lens: code-reviewer, silent-failure-hunter, pr-test-analyzer

Verified: `test_layout.py`, `test_build.py`, `test_resolve.py`, `test_translate.py`, `test_cluster_ops.py`, `test_sim2real.py`, `test_assemble_run.py`, and `test_cluster_py.py` all define `@pytest.fixture(autouse=True)` fixtures that reset `layout._EXPERIMENT_ROOT = None` after each test. `test_deploy_remote.py` has no such fixture. The new tests call `mod._init_experiment_root(...)` which invokes `layout.set_experiment_root()` — after they run, `layout._EXPERIMENT_ROOT` holds a stale `tmp_path` pointer. Any later test in the same session that assumes the None-fallback-to-cwd behavior will silently read the wrong root, producing order-dependent failures.

### recommendations

_(none)_

_Posted manually because the workflow's `gh pr comment` step failed on a macOS base64 flag incompatibility. All 7 review agents completed successfully; only the final post step errored._

kalantar:

The mirroring fix is correct and well-tested; the two remaining hardcoded local-only messages sit on paths the orchestrator Job cannot dispatch to today.

### Recommendations
- **The build and stop arms in main() still emit the local-only hint; converting them to _no_namespaces_hint() would close the same class of bug.** (`pipeline/deploy.py:3870`) — Verified deploy.py:3870 and 3910 still emit the literal 'Run cluster.py provision with --namespaces.' while the neighbouring _cmd_run sites at 2800/2820 use _no_namespaces_hint(). build_orchestrator_job (remote.py:165-169) hardcodes the pod's args to 'run --skip-build', so these arms are unreachable from the pod today — not a bug, but the two-line follow-up would eliminate the latent trap the PR's own scope is written against.
- **SIM2REAL_ORCHESTRATOR_POD is owned by the reader (deploy.py) while the writer (remote.py) hardcodes the literal — flip ownership so remote.py exports the constant.** (`pipeline/lib/remote.py:199`) — Verified deploy.py:145 defines _ORCHESTRATOR_POD_ENV = 'SIM2REAL_ORCHESTRATOR_POD' while remote.py:199 uses the bare string; deploy.py already imports from remote.py so the direction is safe. Moving the constant to remote.py and importing it in deploy.py removes a silent-divergence window that no existing test would catch (each side is asserted against the string independently).

PR #565, opened 2026-07-10
fix(deploy,collect): decide per-iteration, stop wiping the workload dir

kalantar:
Closes #564.

## Summary

`deploy.py collect` used to make its up-to-date / wipe decisions per workload. When the two replicas of one `(phase, workload)` pair dispatched across cluster slots, the second slot's collect either skipped the workload (because the first slot's copy had already made the local mtime look fresh) or wiped the whole workload directory before re-copying only its own iteration. In the reported reproduction (`trial-1`), 25 of 72 iterations were lost silently.

The fix rewrites the incremental-collect logic to operate at iteration granularity throughout, and stops wiping the workload directory as a unit:

- `_probe_remote_mtimes(...)` now returns `{workload: {iN: mtime}}` (was `{workload: max_mtime}`), so an iteration's mtime is compared against its own remote copy.
- `_is_workload_up_to_date` is replaced by `_is_iteration_up_to_date(iN_dir, remote_mtime)`.
- The two full-copy branches (workload-scoped and unscoped) share a new helper `_copy_workload_iterations_full(...)` that enumerates iterations on the current slot's PVC and copies each `i<N>/` individually. The workload directory is never wiped; only the specific `i<N>/` being re-copied is cleared first, so iterations already fetched from other slots stay in place.
- The `--skip-logs` branch's up-to-date gate moves inside the per-iteration loop, and the log-directory wipe moves with it so skipped iterations keep their contents.

## Test plan

- [x] Unit tests for `_probe_remote_mtimes` new nested shape (in `test_collect_internals.py` and `test_deploy_collect.py`)
- [x] Unit tests for `_is_iteration_up_to_date` (fresh / stale / boundary / missing dir / missing csv)
- [x] Unit tests for `_copy_workload_iterations_full`: cross-slot preservation, selective skip, error paths (ls fails / no iN on PVC)
- [x] Integration test through `_extract_phases_from_pvc`: run two sequential collects, one per emulated slot, and assert both iterations land on disk (the direct #564 repro)
- [x] Existing stale-file / idempotency / filter tests updated for the post-fix layout (stale fixtures moved under `i1/`; second-level `ls` mocked)
- [x] Full CI-scoped test suite: `1831 passed, 1 skipped, 2 xfailed`
- [x] Lint: `ruff check pipeline/ .claude/skills/ --select F` — clean
- [x] Docs sweep: `pipeline/README.md` `deploy.py collect` paragraph rewritten to describe per-iteration incrementality; historical plan docs under `docs/superpowers/plans/` intentionally left alone

## Reviewer attention

- The three collect branches (`skip_logs`, `elif workload`, `else` unscoped) collapse a fair amount of code onto the new helper — verify the callback semantics (`on_workload_done`) still fire once per workload with the aggregated error, matching pre-fix behavior.
- The skip_logs branch preserves its original per-iteration file list (`trace_data.csv`, `trace_header.yaml`, `epp_stream_done`, `gpu_stream_done`, `epp_logs/`, `gpu_logs/`, `resources/`) — no files were dropped or added; only the enclosing skip logic changed.
- `test_extract_phases_preserves_iterations_across_slot_collects` is the regression guard for future refactors; if it turns brittle, that's a signal the underlying invariant needs attention.


kalantar:

Core per-iteration fix is correct and well-tested with a direct #564 regression guard; remaining findings are docstring inaccuracies and low-value polish.

### Recommendations
- **_extract_phases_from_pvc docstring still describes the pre-i<N>/ flat PVC layout** (`pipeline/deploy.py:1313`) — Verified at pipeline/deploy.py:1313-1315 — the layout example shows /data/{runName}/{phase}/{workloadName}/trace_header.yaml and trace_data.csv directly under the workload dir, but every code path in this function (including the helpers the PR added) operates on the i<N>/ level. Two independent lenses converged on this. Since the PR substantially reworked this function, the layout block is worth fixing for the next maintainer.
- **Docstring claim that skip_logs "copies only trace files" is factually wrong** (`pipeline/deploy.py:1337`) — Verified at pipeline/deploy.py:1337-1338 versus lines 1482-1519: the skip_logs branch copies trace files plus epp_logs/, gpu_logs/, resources/, epp_stream_done, and gpu_stream_done. Only server_logs (vLLM) is skipped. The current docstring says "skipping vLLM and EPP log files" which is the exact opposite of what happens — a reader relying on this to reason about what `--skip-logs` collects will be misled.

Issue #59, opened 2026-05-08
deploy.py: add 'pairs' subcommand to list available keys (Layer 3)


kalantar:
## Parent issue

Part of #56 — deploy.py filter discoverability

## Layer 3: `deploy.py pairs` subcommand

A new subcommand that dumps available pair keys, workloads, and packages. Works even before `progress.json` exists by falling back to scanning `cluster/pipelinerun-*.yaml`.

### Usage

```bash
# Human-readable table
python pipeline/deploy.py pairs

# Machine-readable (for completion scripts, Layer 4)
python pipeline/deploy.py pairs --keys-only
python pipeline/deploy.py pairs --workloads-only
python pipeline/deploy.py pairs --packages-only
```

### Acceptance criteria

- `deploy.py pairs` prints a table of pair keys with their workload and package
- Works without `progress.json` (scans cluster/ directory directly)
- Machine-readable output modes for scripting / tab completion

PR #575, opened 2026-07-15
feat(pipeline): stream-metrics sidecar — closes #574

kalantar:
## Summary

Adds a third streaming sidecar next to `stream-epp-logs` and `stream-gpu-stats`. `stream-metrics` scrapes each pipelinerun's EPP and vLLM decode pods' Prometheus `/metrics` endpoints on a 2 s cadence and writes `metrics/timeseries.csv` into every cell, giving offline analysis access to `WaitingQueueSize`, `KVCacheUsagePercent`, `flow_control_pool_saturation`, and the rest of the router/engine Prometheus surface without cluster reachability at analysis time.

Companion PR: [inference-sim/tektonc-data-collection#59](https://github.com/inference-sim/tektonc-data-collection/pull/59) — already merged. This PR bumps the submodule pointer to include it.

Closes #574.

## What it does

- **New Tekton Task `stream-metrics`** (lives in the submodule, wired in via `pipeline/pipeline.yaml`) — discovers EPP + vLLM decode pods via label selectors, scrapes `/metrics` on `eppPort=9090` / `vllmPort=8000` (both overridable), appends rows to `metrics/timeseries.csv`, re-discovers targets every 30 s, exits on the `metrics_stream_done` sentinel written by `collect-results`. Non-fatal on failure.
- **CSV columns match `trace_data.csv`'s `_us` convention** — `timestamp_us,target,metric_name,labels,value`. `labels` is the raw Prometheus label-string (RFC-4180-quoted); values that are `NaN` / `+Inf` are dropped so `value` stays a clean float column.
- **`deploy.py collect`** now pulls `metrics/` + `metrics_stream_done` from the PVC and wipes stale `metrics/` in the same batch as the other log dirs.

## Allowlist + discovery

Default `metricsAllowlist` is a curated prefix-match set:

- **vLLM:** `num_requests_waiting`, `num_requests_running`, `gpu_cache_usage_perc`, `num_preemptions_total`, `request_queue_time_seconds`, `iteration_tokens_total`, `gpu_prefix_cache_hits_total`, `gpu_prefix_cache_queries_total`.
- **EPP:** `flow_control_pool_saturation`.

Empty string = capture everything (research escape hatch).

**On the first successful scrape of each target**, the sidecar writes `metrics/<target>.discovered.txt` — sorted-unique list of every metric name that target exposed, no allowlist filter. Operators read this file after any run to widen the allowlist from evidence that lives with the data.

## Deviations from the issue's proposal (called out here so they're not hunt-and-peck)

1. **`timestamp_us` (integer microseconds since epoch), not `timestamp_us` with sub-second precision.** Busybox `date`'s `%N` support is inconsistent; scrape interval is 2 s so per-second is honest precision. Column name/unit still matches the trace side.
2. **`labels` (raw Prometheus label-string), not `labels_json`.** Correct JSON escaping in awk is fragile; pandas parses the raw form via one regex.
3. **Allowlist by default with a discovery-file aid, not "capture everything".** Curated 9-prefix default keeps the CSV compact; `.discovered.txt` gives operators the menu for widening.
4. **Same pod-discovery selectors as `stream-epp-logs.yaml`** (`llm-d.ai/igw-mode=llm-d-router-gateway` + `inferencepool` fallback for EPP; `llm-d.ai/role=decode` for vLLM), not the smoketest's mode-keyed selector. Both are task params — overridable per deployment.
5. **RBAC unchanged** — `helm-installer` already has `pods list` in-namespace via `tektonc-data-collection/tekton/roles-ns.yaml:54-56`, which is all Option A discovery needs.

## Stale-reference sweep

Grepped `.claude/`, `pipeline/`, `docs/`, and `CLAUDE.md` for the streamer/sentinel/log-dir names.

- `pipeline/README.md`: added streamers list + "Metric capture" subsection.
- `CLAUDE.md`: added `metrics/timeseries.csv` and `metrics/<target>.discovered.txt` rows to the workspace artifact table.
- `.claude/skills/sim2real-check/SKILL.md:367`: cell-contents description previously listed only `epp_stream_done`; updated to name all three sentinels + `metrics/`.
- `.claude/skills/sim2real-analyze/SKILL.md`: extended the data-reference block to include `metrics/timeseries.csv` and `metrics/<target>.discovered.txt`.

## Test plan

- [x] Full CI pytest matching `.github/workflows/test.yml` — 1876 passed, 1 skipped, 2 xfailed.
- [x] `TestStreamMetricsTask` (6 cases) + `stream-metrics` added to `TestPrepareResultsDirTask.WRITERS` — verifies task presence, taskRef, `runAfter` shape, workspace binding, param surface, and `resultsDir` parity with the workload.
- [x] `test_collect_skip_logs_invokes_metrics_copy` — verifies `deploy.py collect` issues a `kubectl cp` for `metrics/` and the `metrics_stream_done` sentinel.
- [x] `test_collect_skip_logs_clears_stale_log_dirs` extended — verifies stale `metrics/` is wiped alongside the other log dirs.
- [ ] Live-cluster smoke — deferred; requires an existing cluster. The task is non-fatal by design so it should never block a workload from producing traces even if scraping fails.


kalantar:

Parent-repo wiring for the stream-metrics sidecar is correct and consistent with the established sibling patterns; specialists' concerns either target pre-existing patterns this PR doesn't change, or are about a phantom pipeline.yaml hunk that doesn't exist in the actual committed diff.

### Recommendations
- **sim2real-analyze/SKILL.md treatment subtree lists metrics/timeseries.csv but omits metrics/<target>.discovered.txt shown under baseline** (`.claude/skills/sim2real-analyze/SKILL.md:191`) — Read lines 175-193: the baseline block enumerates both timeseries.csv and <target>.discovered.txt (line 186), while the treatment block only lists timeseries.csv (line 191). Both cells run the same sidecar and produce the same files, so the treatment subtree should mention the discovery file too for parity.

Issue #60, opened 2026-05-08
deploy.py: zsh/bash tab completion for filter arguments (Layer 4)


kalantar:
## Parent issue

Part of #56 — deploy.py filter discoverability

## Layer 4: Shell tab completion

Generate completion scripts that provide live tab-completion for `--only`, `--workload`, `--package`, and `--status` by querying the workspace.

### Files

- `pipeline/completions.zsh`
- `pipeline/completions.bash`

### Behavior

- Completes subcommands: `run`, `status`, `collect`, `cleanup`, `pairs`
- For `--only`: calls `deploy.py pairs --keys-only` to get valid pair keys
- For `--workload`: calls `deploy.py pairs --workloads-only`
- For `--package`: calls `deploy.py pairs --packages-only`
- For `--status`: offers `pending`, `running`, `done`, `failed`, `timed-out`

### Setup

```bash
# zsh
source $SIM2REAL/pipeline/completions.zsh

# bash
source $SIM2REAL/pipeline/completions.bash
```

### Acceptance criteria

- Tab completion works for all filter arguments
- Depends on Layer 3 (`pairs` subcommand) for data
- Gracefully handles missing workspace (no errors, just no completions)

### Notes

This layer is optional — implement only if Layers 1-3 prove insufficient in practice.

PR #577, opened 2026-07-15
fix(pipeline): bump submodule for stream-metrics EPP auth + cleanups — closes #576

kalantar:
## Summary

Bumps the `tektonc-data-collection` submodule to pick up three sidecar fixes in `stream-metrics.yaml`. Closes #576.

Companion PR: [inference-sim/tektonc-data-collection#60](https://github.com/inference-sim/tektonc-data-collection/pull/60) — already merged. This PR moves the submodule pointer from `187708a` to `77e5c7b`.

## What the sidecar fixes do

- **Bug 1 (data loss).** EPP `/metrics` scrapes were returning `HTTP/1.1 401 Unauthorized` because the shell script called `wget` with no `Authorization` header. GAIE / llm-d-inference-scheduler serves `/metrics` on port 9090 with secure-serving on by default. The fix reads the auto-projected SA token at `/var/run/secrets/kubernetes.io/serviceaccount/token` (present in every Tekton step pod) and passes `Authorization: Bearer <token>` on `epp:*` target scrapes only. vLLM `/metrics` on port 8000 remains unauthenticated as before. Result: `metrics/timeseries.csv` will now contain EPP rows — including `flow_control_pool_saturation`, the derived saturation value the flow-controller consumes and the primary motivation for adding the sidecar in #574.
- **Bug 2 (noise).** busybox `wget -q` doesn't suppress HTTP-error stderr, so every failed scrape emitted `wget: server returned error: HTTP/1.1 …` to the container log. Redirected wget stderr to `/dev/null`; failure state is still captured in the per-target `${name}.err` file with the exit code.
- **Bug 3 (cosmetic).** Phase 1's `Discovered scrape targets (EPP selector: …):` log line always printed empty because `EPP_SELECTOR_USED` was assigned inside a `$(discover_targets)` subshell and never survived. Now published via `/tmp/targets.epp_selector` — the same file-based handoff pattern the function already uses for the target lists themselves.

## Diff shape

```
 tektonc-data-collection | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

Pure submodule pointer bump; no parent-repo code changes.

## Test plan

- [x] `python -m pytest pipeline/tests/test_pipeline_yaml.py pipeline/tests/test_deploy_collect.py` — 116 passed. No new tests added — the fix is entirely inside a shell script in the submodule; unit-testing shell behavior against a live cluster is out of scope for this PR.
- [ ] **Live-cluster smoke.** After merge, re-apply the updated task set to `kalantar-0` (or re-run `cluster.py provision`) and retry a workload. Expected:
  - EPP scrapes succeed (no `HTTP 401` in pod log).
  - `metrics/timeseries.csv` contains rows with `target=epp:<pod>` for allowlisted metric prefixes, including `flow_control_pool_saturation`.
  - `metrics/epp:<pod>.discovered.txt` is written on first scrape.
  - Pod log clean of `wget: server returned error…` lines.
  - Phase 1 log line reads `Discovered scrape targets (EPP selector: llm-d.ai/igw-mode=llm-d-router-gateway):` (or the fallback value), not empty.
  - stream-metrics exits on `metrics_stream_done` sentinel — assumes the updated `collect-results` task is also re-applied so the sentinel actually gets written.

## Not in this PR

If the live-cluster smoke shows `403 Forbidden` instead of the previous `401 Unauthorized`, GAIE is enforcing SubjectAccessReview on `/metrics` and the pipeline SA needs a ClusterRole granting `get` on `nonResourceURLs: ["/metrics"]`. That's the standard Prometheus-scraper RBAC pattern; would land as a follow-up submodule PR touching `tekton/roles-cluster.yaml`.

## Stale-reference sweep

Grepped for `stream-metrics`, `TOKEN`, `EPP_SELECTOR_USED`, `bearer`, `Authorization` across `.claude/`, `pipeline/`, `docs/`, and `CLAUDE.md`. Nothing worth updating — the fixes are internal to the shell script; no documented behavior surface changed. The `pipeline/README.md` "Metric capture" subsection added in #575 still describes the sidecar correctly (allowlist, discovery-file convention, live-inspection recipe are all unaffected).


kalantar:

Submodule pointer bump lands three correct, well-commented fixes for the EPP 401 data loss, wget stderr flood, and empty startup selector log; no specialist finding rises above shrug-level for a submodule-bump PR.

Issue #61, opened 2026-05-08
Resource-availability-aware orchestration for deploy.py run


kalantar:
## Problem

`deploy.py run` dispatches PipelineRuns to namespace slots as soon as a slot is free, with no awareness of whether the cluster can actually schedule the resulting pods. On shared clusters with limited GPUs, this leads to:

1. Pods enter Pending with Unschedulable conditions
2. The 4-hour timeout fires
3. The orchestrator retries blindly, burning another 4 hours
4. Repeat until GPUs happen to free up

The orchestrator wastes hours waiting on runs that were never going to schedule, while also failing to capitalize on windows when capacity becomes available.

## Design

Six components work together to make dispatching resource-aware:

```
┌─────────────────────────────────────────────────────────┐
│                  Orchestrator Loop                       │
│                                                         │
│  ┌──────────────┐    ┌──────────────┐                   │
│  │ Capacity     │───▶│ Dispatch     │                   │
│  │ Probe        │    │ Gate         │                   │
│  └──────────────┘    └──────┬───────┘                   │
│                             │                           │
│         ┌───────────────────┼───────────────┐           │
│         ▼                   ▼               ▼           │
│  ┌─────────────┐   ┌──────────────┐  ┌───────────┐     │
│  │ Early       │   │ Backoff      │  │ Capacity- │     │
│  │ Reclaim     │   │ Controller   │  │ Propor-   │     │
│  │             │   │              │  │ tional    │     │
│  └─────────────┘   └──────────────┘  │ Dispatch  │     │
│                                      └───────────┘     │
│                                                         │
│  ┌──────────────────────────────────────────────────┐   │
│  │ Workload GPU Cost (per-PipelineRun)              │   │
│  └──────────────────────────────────────────────────┘   │
└─────────────────────────────────────────────────────────┘
```

### 1. Capacity Probe

Query the cluster for available GPU capacity:

```
free_gpus = sum(node.allocatable[gpu_resource_type]) - sum(pod.requested[gpu_resource_type])
```

The **GPU resource type** (e.g., `nvidia.com/gpu`, `amd.com/gpu`, `habana.ai/gaudi`) is derived from the resolved scenario config — the same place workload GPU cost comes from. Each scenario specifies the resource name its pods will request; the probe uses that to query the correct resource.

[...]

PR #580, opened 2026-07-16
feat(pipeline): adopt collect_metrics.sh + provision EPP metrics RBAC — closes #579

kalantar:
## Summary

Adopts `collect_metrics.sh` from `llm-d-benchmark` for EPP + vLLM metrics scraping, replacing the bespoke `stream-metrics` shell/awk scraper landed in #575 / fixed in #577. Provisions the EPP-side auth RBAC at `cluster.py provision` / `cluster.py slot add` time so sim2real's non-admin pipeline can consume it without changes to `llmdbenchmark-standup`.

Closes #579. Supersedes #578 (closed with a pointer here).

Companion upstream PR: [inference-sim/tektonc-data-collection#61](https://github.com/inference-sim/tektonc-data-collection/pull/61) — merged as `2381e8e`.

## What sim2real gains

- **Working EPP metrics.** The reason our trial-3 EPP scrapes 401'd wasn't a client-side auth bug — it was that `llmdbenchmark-standup`'s non-admin patch strips the chart's monitoring stack (`router.monitoring.prometheus.enabled: false`), so the cluster-scoped RBAC EPP needs to validate bearer tokens was never provisioned. This PR provisions the equivalent RBAC directly at `cluster.py provision` time. See #579 for the full analysis and the `_patch_router_for_non_admin` reference.
- **Less code to maintain.** ~330 lines of sim2real-owned shell/awk replaced by a ~180-line wrapper that pulls the upstream scraper.
- **More metrics coverage.** DCGM GPU metrics, container cgroup metrics, and richer EPP families (`inference_pool_*`, `inference_extension_*`, `llm_d_inference_scheduler_*`) — all now captured "for free" via `collect_metrics.sh`.

## Submodule pointer bump — `tektonc-data-collection@2381e8e`

- **`roles-cluster.yaml`** — new `sim2real-metrics-reader` ClusterRole (three rules: `tokenreviews:create`, `subjectaccessreviews:create`, `nonResourceURLs:/metrics:get`) + per-namespace `sim2real-metrics-reader-${NAMESPACE}` ClusterRoleBinding with `Group: system:serviceaccounts:${NAMESPACE}` subject.
- **`roles-ns.yaml`** — new `inference-gateway-sa-metrics-reader-secret` Secret (`type: kubernetes.io/service-account-token`, annotated to `helm-installer`). Kubernetes auto-populates `data.token`.
- **`tasks/stream-metrics.yaml`** — full rewrite. `python:3.12-slim-bookworm` image; `apt-get install curl`, download pinned `kubectl v1.34.1`, `curl` fetches `collect_metrics.sh` + `process_metrics.py` at `harnessSha`, then `start` / poll-sentinel / `stop` / `process`.

## sim2real-side changes

- **`pipeline/pipeline.yaml`** — new `harnessSha` param on the `stream-metrics` task ref, pinned to `76473d0` (the SHA sim2real currently vendors in its `llm-d-benchmark` submodule). Same pinning pattern `install-inference-perf.yaml` uses for `inference-perf`'s git commit. Bump this in lock-step with the submodule pointer on future updates.
- **`pipeline/deploy.py`** — no code change. The recursive `kubectl cp metrics/` stanza (added in #575) already captures the new `metrics/{raw,processed}/` layout `collect_metrics.sh` writes.
- **`pipeline/tests/test_deploy_collect.py`** — docstring update on `test_collect_skip_logs_invokes_metrics_copy` to describe the new layout. Assertion unchanged.
- **`pipeline/README.md`** — "Metric capture" subsection rewritten. New layout, EPP auth via the reader Secret, how to widen the metrics summary (edit `AGGREGATE_METRICS` in `process_metrics.py` upstream + pin `harnessSha`).
- **`CLAUDE.md`** — workspace artifact rows: `timeseries.csv` + `discovered.txt` replaced by `metrics/raw/*.log` + `metrics/processed/*.json`.
- **`.claude/skills/sim2real-analyze/SKILL.md`** — data-reference block updated.
- **`docs/superpowers/plans/2026-07-15-issue-574-stream-metrics.md`** — added a "superseded by #579" preamble (kept as historical record).
- **`docs/superpowers/plans/2026-07-15-issue-579-adopt-collect-metrics.md`** — new plan doc.

## Reviewer attention

Three areas worth eyes:

1. **`pipeline/pipeline.yaml:157` — the pinned SHA.** Manual maintenance burden: submodule bump needs a matching `harnessSha` update. Same pattern as `install-inference-perf`'s `GIT_COMMIT` env var, so precedent is fine; noting explicitly so it's not surprising later.
2. **RBAC blast radius.** The new ClusterRoleBinding subject is `Group: system:serviceaccounts:<namespace>` — every SA in the pool namespace inherits TokenReview + SAR + `/metrics get` on any target. Tighter alternatives (bind to a specific SA) were rejected because the EPP SA name is chart-derived per release. Blast radius stays scoped to one namespace per binding.
3. **`stream-metrics.yaml` non-fatal posture.** Every apt/curl/download step is `|| exit 0` so a network failure logs a warning and the workload stays green. Deliberate — matches the existing streamers' posture. Alternative (fail hard) was rejected as it would let a transient network hiccup break every run.

## Stale-reference sweep

Grepped `.claude/`, `pipeline/`, `docs/`, `CLAUDE.md` for `timeseries.csv`, `discovered.txt`, `metricsAllowlist`, `EPP_SELECTOR_USED`, `eppSelectorFallback`. All remaining hits are inside plan docs (the current #579 plan describing the change; the #574 plan now marked superseded). No stale refs in code or user-facing docs.

## Test plan

- [x] Full pytest suite matching `.github/workflows/test.yml` — **1876 passed, 1 skipped, 2 xfailed**.
- [x] `envsubst` + `yaml.safe_load_all` on the merged submodule's `roles-cluster.yaml` and `roles-ns.yaml` — clean; correct object counts.
- [x] `python -c "import yaml; yaml.safe_load(open('tektonc-data-collection/tekton/tasks/stream-metrics.yaml'))"` — clean.
- [ ] **Live cluster.** Deferred. The operator will re-add a slot (or re-apply the updated `roles-*.yaml` + `stream-metrics.yaml`) to pick up the new RBAC + Secret, then re-run a workload.

### DIFF
--- a/.claude/skills/sim2real-analyze/SKILL.md
+++ b/.claude/skills/sim2real-analyze/SKILL.md
@@ -180,15 +180,19 @@ workspace/runs/<name>/
                           #                       last_chunk_time_us, output_tokens,
                           #                       arrival_time_us, input_tokens, status, ...
         trace_header.yaml # model, time_unit (microseconds), workload_spec, server config
-        metrics/timeseries.csv       # EPP + vLLM Prometheus scrapes (optional — present
-                                     # when stream-metrics ran); columns:
-                                     # timestamp_us, target, metric_name, labels, value
-        metrics/<target>.discovered.txt  # sorted-unique metric names each target exposed
+        metrics/raw/<pod>_<ts>_metrics.log   # Prometheus text-exposition dumps, one file
+                                             # per pod per scrape (optional — present when
+                                             # stream-metrics ran). Produced by
+                                             # collect_metrics.sh which stream-metrics wraps.
+        metrics/processed/metrics_summary.json    # post-run percentiles over the metrics
+                                                  # named in process_metrics.py:AGGREGATE_METRICS
+        metrics/processed/replica_status_timeseries.json  # replica state/scale over the run
     treatment/
       workload_<name>/
         trace_data.csv
         trace_header.yaml
-        metrics/timeseries.csv
+        metrics/raw/...
+        metrics/processed/...
   deploy_comparison_table.txt  # written by analyses/latency_table.py
   results_charts/              # your analysis outputs go here
 ```
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -115,8 +115,9 @@ All artifacts live under `<experiment-root>/workspace/` (gitignored). When no `-
 | `runs/<run>/cluster/pipelinerun-*.yaml` | `sim2real assemble` | `deploy.py run` |
 | `runs/<run>/results/{phase}/` | `deploy.py collect` | `/sim2real-analyze` skill, `deploy.py wipe` |
 | `runs/<run>/results/{phase}/<workload>/i<N>/gpu_logs/<node>.log` | `deploy.py collect` (pulled from PVC) | analysis / debugging |
-| `runs/<run>/results/{phase}/<workload>/i<N>/metrics/timeseries.csv` | `deploy.py collect` (pulled from PVC) | analysis — EPP + vLLM `/metrics` time-series, columns `timestamp_us,target,metric_name,labels,value` (allowlist-filtered by `stream-metrics`); see `pipeline/README.md#metric-capture-stream-metrics` for defaults + widening |
-| `runs/<run>/results/{phase}/<workload>/i<N>/metrics/<target>.discovered.txt` | `deploy.py collect` (pulled from PVC) | ops — sorted-unique metric names the target exposed on first scrape; source of truth for widening `metricsAllowlist` |
+| `runs/<run>/results/{phase}/<workload>/i<N>/metrics/raw/<pod>_<ts>_metrics.log` | `deploy.py collect` (pulled from PVC) | analysis — raw Prometheus text-exposition dumps, one file per pod per scrape (produced by `collect_metrics.sh`, which `stream-metrics` wraps — see `pipeline/README.md#metric-capture-stream-metrics`) |
+| `runs/<run>/results/{phase}/<workload>/i<N>/metrics/processed/metrics_summary.json` | `deploy.py collect` (pulled from PVC) | analysis — post-run percentiles over the metrics named in `process_metrics.py:AGGREGATE_METRICS` |
+| `runs/<run>/results/{phase}/<workload>/i<N>/metrics/processed/replica_status_timeseries.json` | `deploy.py collect` (pulled from PVC) | analysis — replica state / scale over the workload window |
 | ConfigMap `sim2real-progress-{scenario}-{run}` | `deploy.py run`, `deploy.py reset` | All `deploy.py` subcommands |
 | `runs/<run>/plans/<phase>/<workload>/` | `deploy.py run` | workload tasks |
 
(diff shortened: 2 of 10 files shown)


kalantar:
Addressed the review's must_fix on `slot remove` teardown.

`cmd_slot_remove` now deletes the two per-namespace metrics-reader resources before updating the pool config:

- `ClusterRoleBinding/sim2real-metrics-reader-<ns>` (cluster-scoped)
- `Secret/inference-gateway-sa-metrics-reader-secret` in `<ns>`

Both kubectl calls use `--ignore-not-found` so a partially-provisioned slot removes cleanly. The delete logic lives in a new `cluster_ops.deprovision_metrics_rbac` helper that returns `(ok, err)`; the caller surfaces failures as warnings but still updates the pool config so on-disk state can't get stuck out of sync with cluster reality.

Two tests: `test_removes_non_primary` now asserts the two expected deletes fire (with `--ignore-not-found`) and no others. `test_removes_non_primary_survives_partial_teardown_failure` proves a kubectl-delete failure warns but still updates the pool.

Other in-namespace resources (Role, RoleBinding, SA, PVCs, deployed stacks) are still preserved — those are shared with drained-but-not-yet-torn-down workloads. Only the metrics-reader pair (the ones this PR introduces) get removed.

Suite green: 1877 passed, 1 skipped, 2 xfailed; ruff clean.

kalantar:
Addressed both important findings from the iter-2 review.

The teardown helper now attempts both deletes regardless of the first's outcome — if the CRB delete fails, the Secret delete is still tried. Errors from either are accumulated and returned together. That was the concerning pattern in the previous pass: the Secret is the more sensitive residual (a kubelet-populated bearer token for helm-installer), so short-circuiting past it on any transient CRB error was exactly backwards.

The slot-remove stdout message is now gated on success: on the ok path it still says "metrics-reader RBAC + Secret cleaned"; on the failure path it says "metrics-reader teardown had errors — see stderr" instead. No more contradicting the stderr warning with a stdout success line.

Extended `test_removes_non_primary_survives_partial_teardown_failure` to assert both properties: (a) the Secret delete IS attempted after the CRB delete fails, and (b) stdout does NOT contain "cleaned" on the failure path.

Suite green: 1877 passed / 1 skipped / 2 xfailed; ruff clean.

kalantar:

Acceptance criteria are met, tests pass, and no verified finding rises above optional polish.

### Recommendations
- **cmd_slot_remove docstring points at the wrong function for where these resources were provisioned.** (`pipeline/cluster.py:519`) — cluster.py:519 says '... ClusterRoleBinding + Secret provisioned in :func:`cluster_ops.deprovision_metrics_rbac`' — but deprovision_metrics_rbac only deletes. Provisioning happens in provision_namespace's _step_rbac (roles-cluster.yaml + roles-ns.yaml). A future reader chasing the reference will land on the wrong function.
- **collect_metrics.sh process failure emits no WARNING line in the Tekton task log.** (`tektonc-data-collection/tekton/tasks/stream-metrics.yaml:180`) — In the vendored submodule file (ref 2381e8e), Phase 5 runs 'collect_metrics.sh process >> ${LOG} 2>&1' with no '|| echo WARNING…' — inconsistent with every Phase 0-2 install/fetch step. Not a broken contract (the task description scopes the non-fatal-warning promise to install/fetch/discovery, and Phase 6's file count implicitly surfaces the failure), but adopting the same '|| echo WARNING' pattern would make processing failures grep-visible in the Tekton logs without changing exit behaviour.

PR #583, opened 2026-07-16
fix(check): resolve + enumerate blind to replica-shape runs with path-string workloads (#572)

kalantar:
## Summary

Two silent-empty predicates surfaced by `/sim2real-check --run=<replica-run>`:

- **`pipeline/lib/resolve.py:_build_results_section`** used a one-level predicate (`wl/trace_data.csv`). Replica-shape runs — the shape `sim2real assemble` produces when `manifest.assembly.yaml:replicas > 1` — store `trace_data.csv` under `wl/iN/`, so `phases_with_data` returned `[]` and `workloads_by_phase` returned `{}` despite data on disk. Extracted a `_workload_has_data` helper that accepts either legacy flat shape or replica shape (`iN` dirs matching `^i[1-9][0-9]*$`, same grammar as the enumerator's `_ITER_DIR_RE`).

- **`.claude/skills/sim2real-check/scripts/enumerate_iterations.py:_names`** filtered to dict entries with a `name:` field. `manifest.assembly.yaml` written by `sim2real assemble` lists workloads as file-path strings (`workloads/foo.yaml`), which were silently dropped — the enumerator returned `{counts: {PRESENT: 0, MISSING: 0, SKIP: 0}}` on any replica run. Added a `str` branch that emits `Path(s).stem` so identifiers match the on-disk directory names `_detect_shape` compares against.

## Verification against real data

Repro from `kalantar-msb/softr` `trial-2` (3 phases × 12 workloads × 2 iterations = 72 `trace_data.csv` on disk):

| | Before | After |
|---|---|---|
| `sim2real resolve --run trial-2` → `phases_with_data` | `[]` | `["baseline","constantceiling","softreflective"]` |
| `sim2real resolve --run trial-2` → `workloads_by_phase[phase]` | `{}` | 12 workloads each |
| `enumerate_iterations.py --run trial-2` → `counts.PRESENT` | `0` | `72` |

## Test plan

- [x] `pipeline/tests/test_resolve.py`: added replica-shape, mixed-shape, and malformed-iN-dir cases (extended `_make_run` with `results_shape` / `replicas` kwargs).
- [x] `.claude/skills/sim2real-check/tests/test_enumerate_iterations.py`: `_names` unit tests for dict / str / mixed / junk / bare-string inputs, plus an end-to-end path-string-workloads test on a replica-shape run.
- [x] Full CI test set: 1887 passed, 1 skipped, 2 xfailed (no regressions).
- [x] `ruff check pipeline/ .claude/skills/ --select F`: clean.
- [x] End-to-end verification on `kalantar-msb/softr trial-2` (see table above).

## Reviewer attention

- The two predicates are symmetric in the grammar they accept (`^i[1-9][0-9]*$`). `resolve.py` maintains its own `_ITER_DIR_RE` rather than importing the enumerator's — they're separate packages and the enumerator isn't on `sys.path` for the library. Kept-in-sync comment added on the resolve-side constant.
- The strict-predicate guard test (`test_results_section_ignores_malformed_iter_dirs`) exists to prevent a naive fix (`any(child/trace_data.csv for child in iterdir)`) that would false-positive on `plans/` or `metrics/` sibling dirs — the concern the pre-fix docstring called out.
- `_names` fix widens semantics: string entries now silently derive identifiers from `Path.stem`. For workloads this matches how `_detect_shape` matches directory names. Baselines and algorithms have always used the dict shape and continue to work unchanged.

## Docs sweep

- Updated `docs/epics/step-3/design.md` (`phases_with_data` predicate description) to reflect the new shape-tolerant behavior.
- No `README.md` / `pipeline/README.md` references to the affected predicates.

Closes #572

Base: main

### DIFF
--- a/.claude/skills/sim2real-check/scripts/enumerate_iterations.py
+++ b/.claude/skills/sim2real-check/scripts/enumerate_iterations.py
@@ -278,7 +278,20 @@ def _detect_shape(
 
 
 def _names(items) -> list[str]:
-    """Extract ``name`` field from a list of manifest dicts."""
+    """Extract workload/baseline/algorithm identifiers from a manifest list.
+
+    Two entry shapes are supported so the same helper serves all three
+    manifest sections:
+
+    - Dict entries with a ``name:`` field (baselines, algorithms, and
+      workloads in older/hand-edited manifests) — emit ``name``.
+    - String entries (workloads written by ``sim2real assemble`` as file
+      paths, e.g. ``workloads/code_generation_4.yaml``) — emit
+      ``Path(s).stem`` so the identifier matches the on-disk directory
+      name that ``_detect_shape`` compares against. Bug #572.
+
+    Empty strings and non-str/non-dict entries are silently skipped.
+    """
     if not isinstance(items, list):
         return []
     out = []
@@ -287,6 +300,10 @@ def _names(items) -> list[str]:
             name = it.get("name")
             if isinstance(name, str) and name:
                 out.append(name)
+        elif isinstance(it, str) and it:
+            stem = Path(it).stem
+            if stem:
+                out.append(stem)
     return out
 
 
--- a/.claude/skills/sim2real-check/tests/test_enumerate_iterations.py
+++ b/.claude/skills/sim2real-check/tests/test_enumerate_iterations.py
@@ -22,6 +22,7 @@ def _make_run(
     workloads: list[str],
     translation_algorithms: list[str] | None = None,
     disk_layout: dict | None = None,
+    workloads_as_paths: bool = False,
 ) -> Path:
     """Materialize a run under ``root/workspace/`` with the given shape.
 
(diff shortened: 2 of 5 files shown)


kalantar:

Both bugs are fixed with focused tests; the only surviving concern is a factually inaccurate "replicas > 1" phrase in the new docstring and design.md, which is docs-only polish.

### Recommendations
- **"replicas > 1" is the wrong threshold — the replica (iN/) shape is written for all step-5 runs, including replicas=1** (`pipeline/lib/resolve.py:50`) — Verified assemble_run.py:948 uses iterations=range(1, replicas + 1); with replicas=1 that still writes i1/trace_data.csv. The pre-step-5 legacy shape is distinguished by the absence of a top-level replicas key (assemble_run.py:828-834), not by replicas>1. Same wording appears in docs/epics/step-3/design.md:164. Docs-only; correcting it prevents a future maintainer from concluding single-replica runs can skip the new predicate.

Issue #841, opened 2026-08-24
assemble: no signal when a framework-defaults fragment has no effect on the resolved scenario


kalantar:
## Summary

A framework-defaults fragment that has no effect on the resolved scenario looks
exactly like one that is working. Nothing surfaces the difference. Split out of
#839 Part 4, which found two such fragments out of ten in `pd-infocomm-2`.

## Evidence that this is a real class, not a one-off

- **`vllm-logging`** was entirely inert for its whole life. It set
  `decode.vllm.additionalFlags: ["--no-disable-uvicorn-access-log"]`, but
  `additionalFlags` is a list of scalars, so `pipeline/lib/values.py:_merge_lists`
  Tier 1 has the overlay *replace* the base — and both `baselines/<name>.yaml` and the
  translate overlay set that same list downstream of the fragment. Its other key,
  `loggingLevel: INFO`, was already the framework default. The fragment changed the
  scenario text and nothing else. Fixed in #839.
- **`vllm-keepalive`** becomes inert the moment the llm-d-benchmark pin advances past
  the `VLLM_HTTP_TIMEOUT_KEEP_ALIVE` env-macro. Tracked in #838, but nothing would
  *notice* — the fragment would just stop mattering.

Both were found by hand, during unrelated debugging, after the runs they affected had
already completed.

## Sketch

For each fragment, resolve the baseline with and without it and report any whose
removal changes nothing:

```python
ref = resolve(disable=base)
for frag in fragments:
    if resolve(disable=base + [frag]) == ref:
        warn(f"{frag}: no effect on the resolved scenario")
```

Both primitives already exist in `pipeline/lib/assemble_run.py`:
`load_defaults_overlay` (line 162) and `resolve_baseline` (line 322).

## Two things to settle before implementing

**1. Which assertion.** The sketch above answers "does removing this fragment change
the resolved scenario?" That is not the same as "does this fragment do anything?" —
it fires whenever a downstream layer sets the same key to the same value, which is
a legitimate configuration, not a bug. Concretely, it would fire on `epponly.yaml`'s
`router.proxy.args` if a baseline also carried them.

Only the narrower assertion — *this fragment's keys are overwritten downstream, or
land somewhere nothing reads* — catches the `additionalFlags` bug class. Worth
deciding whether to implement the cheap version and accept the false positives as
warnings, or the precise version and accept that it needs to know which keys the
chart reads.

**2. Where it lives.** There is no `sim2real check` subcommand today —
[...]

Issue #840, opened 2026-08-24
bootstrap: framework defaults fragments are emitted unconditionally, but several are topology-, model-, or plugin-specific


kalantar:
## Summary

`sim2real-bootstrap` copies every fragment in `templates/defaults/` into the
experiment's `baselines/defaults/` and, in BLIS mode, emits `defaults.disable: []`
— so all of them are on. Several of the fragments are not universally correct, and
the only opt-out is for the operator to notice and add a stem to `defaults.disable`
after the fact. Split out of #839 Part 3, which is the debugging session that
surfaced it.

Note the two bootstrap modes already disagree here: BYO puts every stem into
`defaults.disable` (see `.claude/skills/sim2real-bootstrap/tests/test_byo.py:161`),
BLIS emits an empty list. Whatever this issue settles should make them consistent.

## The fragments that are not universally correct

These fall into two distinct groups, and they probably want different mechanisms.

### Topology / model / plugin-conditional — currently always on, not always right

- **`preserve-request-id`** is an Istio `EnvoyFilter` with a hardcoded workload
  selector (`gateway.networking.k8s.io/gateway-name: infra-llmdbench-inference-gateway`).
  Under `epponly` + `gateway.externallyManaged` no such Gateway is deployed, so the
  filter matches nothing, and it needs Istio CRDs present to apply at all. It is also
  the only fragment using `extraObjects`, which makes it the only one that can fail at
  *apply* time rather than being silently inert. `pd-infocomm-2` disables it and had to.
- **`model-pvc-size: 1Ti`** (added by #839) is right for Llama-3.3-70B and wasteful for
  a small model. Better derived from model size than emitted as a constant.
- **`tokenizer-sidecar`** (added by #839) is needed only when the arms declare a
  `token-producer` plugin, and it hardcodes `hf-secret` and `HF_HOME=/tmp/hf`.
- **`epponly`** (added by #839) is a topology decision, not a default.

### Diagnostics that perturb what they measure — want on for bring-up, off for measurement

`epp-verbosity` and `vllm-logging` both belong here. Per-request logging is not free
at high QPS and the EPP sits in the request path. Being fragments rather than bundle
content is exactly what gives them an off switch, so the argument here is for keeping
them in this layer — possibly with a way to flip the whole diagnostic group at once
rather than stem by stem.

## What needs deciding

1. Does conditional emission belong in bootstrap (don't copy the fragment when the
[...]

Issue #838, opened 2026-08-24
bootstrap: remove the vllm-keepalive defaults fragment once the llm-d-benchmark pin includes the VLLM_HTTP_TIMEOUT_KEEP_ALIVE env-macro


kalantar:
## Summary

`baselines/defaults/vllm-keepalive.yaml` (a bootstrap default fragment, currently
hand-authored in the `pd-infocomm-2` bundle and slated to fold into
`sim2real-bootstrap/templates/defaults/`) sets `VLLM_HTTP_TIMEOUT_KEEP_ALIVE=120`
on the prefill pod via `extraEnvVars`.

llm-d-benchmark now emits that variable itself, with the same default, on **both**
roles. Once this repo's llm-d-benchmark pin advances past that commit, the
fragment becomes redundant and should be deleted. It is **not** redundant today.

## Why the setting is needed at all

Both sides of the mismatch are verified in source:

- the routing sidecar pools idle connections for 90s — `t.IdleConnTimeout = 90 * time.Second`
  on a cloned `http.DefaultTransport` with unlimited idle conns
  (llm-d-router v0.9.0, `pkg/sidecar/proxy/proxy.go:415`)
- vLLM closes them after 5s — `VLLM_HTTP_TIMEOUT_KEEP_ALIVE: int = 5` in `vllm/envs.py`,
  *"Timeout in seconds for keeping HTTP connections alive in API server"*

A Go transport holding sockets for 90s against a server that drops them at 5s reuses
connections the server has already closed; the reuse fails with a TCP RST.

## Why it is redundant *later* but not *now*

llm-d-benchmark `config/templates/jinja/_macros.j2:231-232`:

```jinja
- name: VLLM_HTTP_TIMEOUT_KEEP_ALIVE
  value: "{{ config.vllm.httpTimeoutKeepAlive | default(120) }}"
```

Unconditional, into every model-server container, both roles, `default(120)` — the
same value the fragment sets. Added in llm-d-benchmark `28cfc06`, *"Set
VLLM_HTTP_TIMEOUT_KEEP_ALIVE=120 as env-macro default"* (#1675, 2026-07-27).

But:

| | ref | macro present? |
|---|---|---|
| llm-d-benchmark pinned by this repo | `76473d0` = `v0.7.0-27` (2026-06-25) | **no** — `git grep -i http_timeout_keep_alive` over the whole tree returns nothing |
| macro added | `28cfc06` = `v0.7.0-106` (2026-07-27) | yes |

[...]

PR #842, opened 2026-08-24
fix(bootstrap): sync baselines/defaults templates; fix two inert fragments (closes #839)

### LINKED ISSUE #839: bootstrap: sync baselines/defaults templates — two shipped fragments are no-ops, five are missing, and emission should not be unconditional
## Summary

`sim2real-bootstrap/templates/defaults/` ships 5 fragments. The `pd-infocomm-2`
bundle now carries **10**, and two of the shipped five have **diverged** from the
bundle's corrected copies. A bootstrap re-run today would (a) drop five fragments
the bundle needs and (b) reinstate two known-bad ones.

## Part 1 — two shipped templates are now wrong (do these first)

| template | shipped | bundle | why |
|---|---|---|---|
| `vllm-logging.yaml` | sets `decode.vllm.additionalFlags: ["--no-disable-uvicorn-access-log"]` + `decode.vllm.loggingLevel: INFO` | sets `vllmCommon.flags.disableUvicornAccessLog: false` | **both shipped keys are no-ops** |
| `epp-verbosity.yaml` | `router.epp.verbosity: "5"` | `"3"` | 5 is unusable log volume in practice |

### `vllm-logging.yaml` — the shipped version does nothing

- `decode.vllm.additionalFlags` is a scalar list, so `_merge_lists` **replaces**
  rather than appends. `baselines/<name>.yaml` sets that list, and the translate
  overlay replaces it again — the fragment's entry is discarded before either. The
  flag that actually reaches the rendered command comes from the translate overlay,
  which documents precisely this in its own comment.
- `decode.vllm.loggingLevel: INFO` is already the framework default twice over:
  `defaults.yaml:51` (`logging_level: &logging_level INFO`, referenced at `:876`)
  plus a `default('INFO')` fallback at `_macros.j2:229-230`.

It was also decode-only, while the framework's suppression
(`defaults.yaml:789 disableUvicornAccessLog: true`, emitted at `_macros.j2:154/:397`)
applies to **both** roles — so prefill served its requests with **no access log**.
In `pd-infocomm-2` that was the one pod whose request path was under investigation;
the uvicorn access line is what surfaced the prefill-leg `min_tokens`/`max_tokens=1`
rejections.

Correct form (restores vLLM's own default, `disable_uvicorn_access_log: bool = False`,
on both roles):

```yaml
scenario:
- name: defaults
  vllmCommon:
    flags:
      disableUvicornAccessLog: false
```

### `epp-verbosity.yaml` — 5 is too noisy to run with

EPP levels are named constants (`llm-d-router pkg/common/observability/logging/const.go:20-23`):
`DEFAULT=2, VERBOSE=3, DEBUG=4, TRACE=5`. Call sites in v0.9.0: DEFAULT 52,
VERBOSE 27, **DEBUG 144**, TRACE 58. `"5"` enables all 202 DEBUG+TRACE sites, and
the EPP is in the request path so the logging costs CPU as well as disk.

`"3"` is what the operator has been setting by hand every run anyway. Worth
recording in the template's comment: 3 stops **below** the level where the P/D
decider explains itself (`prefix_based_pd_decider.go:97` uses `V(logging.DEBUG)`),
so the escalation for prefill-selection problems is `"4"`, not `"5"` — 4 keeps the
decider diagnostics and still drops the 58 TRACE sites.

## Part 2 — five fragments to add

All five are hand-authored in `pd-infocomm-2/baselines/defaults/` and thoroughly
commented; they can be copied near-verbatim.

| fragment | sets | notes |
|---|---|---|
| `epponly.yaml` | `gateway.className: epponly`, `router.proxy.resources`, `router.proxy.args` | see below |
| `model-pvc-size.yaml` | `storage.modelPvc.size: 1Ti` | model-dependent — see Part 3 |
| `tokenizer-sidecar.yaml` | `router.tokenizer.enabled: true` + HF_TOKEN/HF_HOME env | plugin- and cluster-dependent — see Part 3 |
| `vllm-keepalive.yaml` | `prefill.extraEnvVars` `VLLM_HTTP_TIMEOUT_KEEP_ALIVE=120` | delete instead once the benchmark pin advances — #838 |
| `enforce-eager.yaml` | `vllmCommon.flags.enforceEager: false` | **placement under review, do not port yet** — see Part 3 |

### `epponly.yaml` carries the Envoy sidecar args

`router.proxy.args` was originally in the bundle's `baselines/<name>.yaml` and has
been moved into this fragment, because `--concurrency 8` is chosen to match the
`cpu: "8"` limit the same fragment sets — split across two files, nothing linked
them.

Note for whoever ports it: that args block is **upstream boilerplate**, appearing
byte-for-byte (comment included) in **8 of the 9** shipped llm-d-benchmark guide
scenarios (`config/scenarios/guides/`: optimized-baseline, flow-control,
agentic-serving, predicted-latency-routing, tiered-prefix-cache,
precise-prefix-cache-routing, pd-disaggregation, wide-ep-lws). That every shipped
scenario overrides the same two chart defaults identically (`--log-level trace`,
and no `--concurrency` so Envoy sizes from `hardware_concurrency()`) is an argument
for them being *framework* defaults. Worth a separate upstream issue against
llm-d/llm-d-benchmark; not filed yet.

## Parts 3 and 4 moved out

The original filing carried two further sections. Both are now their own issues, and
this issue is narrowed to the template sync:

- **#840** — fragments are emitted unconditionally, but `preserve-request-id`,
  `model-pvc-size`, `tokenizer-sidecar`, and `epponly` are topology-, model-, or
  plugin-specific. Design question; also needs BLIS and BYO to agree on the default
  enable/disable posture (BYO already disables every stem, BLIS enables every stem).
- **#841** — nothing surfaces a fragment that has no effect on the resolved scenario.
  Needs both the assertion and its placement settled first (there is no
  `sim2real check` subcommand today).

**`enforce-eager` is still deliberately excluded from this port.** It is the one
fragment that is measurement-critical rather than diagnostic — it controls CUDA graphs
and therefore moves ITL, the metric the arms are compared on. It arguably belongs in
`config.md` as a stated experiment decision (cf. #832), but the original reason it was
made a fragment is being reconsidered. Leave it in the bundle until that is settled.

## Additional findings from vetting

Verified against HEAD `3914556`. Four things the original filing did not cover, all in
scope for this issue:

**There is a third inventory location, and it contains a wrong constant.**
`docs/troubleshooting.md` carries a per-fragment value table at `:13-17` (both Part 1
edits make rows stale), a `defaults.disable` example at `:24`, and this at `:127`:

> `verbosity: "5"` (V(5) = VERBOSE) satisfies this requirement

V(5) is TRACE, not VERBOSE — wrong independently of this issue. The same line cites
"`/sim2real-check` §4.2"; the actual section is §5c.6. All three need fixing here.

**The 5 → 3 change lands exactly on a documented threshold, and clears it.**
`/sim2real-check` §5c.6 requires the EPP at `-v=3` or higher for the
InferenceObjective resolution check (`objectiveKey` is attached under
`V(logutil.VERBOSE)`, i.e. 3). `"3"` satisfies it, at zero margin. Worth an explicit
note in the fragment comment so the next person to lower it knows what breaks.

**`router.proxy` and `routing.proxy` are different components, and both fragments
are intended to coexist.** Porting `epponly.yaml` adds `router.proxy.{resources,args}`
— routerlib's Envoy sidecar inside the EPP pod (`routerlib/values.yaml:76 proxy:`).
The already-shipped `routing-proxy-resources.yaml` sets `routing.proxy.resources` —
llm-d-benchmark's nixl routing sidecar beside the model servers
(`defaults.yaml:708 router:` vs `:772 routing:`). Nothing currently distinguishes them
and the names invite deduplication. Both fragment comments should say which sidecar
they size.

**Line references need re-checking against the pin.** The cites above resolve to
`defaults.yaml:890` and `_macros.j2:100/366` and `:154/399` on llm-d-benchmark
`8369bc7`, not the `:789` / `:364` / `:397` in Part 1 — sim2real pins `76473d0`, so
re-derive before writing them into a comment. Key names are correct as filed.

## Checklist

- [ ] fix `templates/defaults/vllm-logging.yaml` to the typed `vllmCommon.flags.disableUvicornAccessLog: false` form (Part 1)
- [ ] set `templates/defaults/epp-verbosity.yaml` to `"3"`, document the 2/3/4/5 scale and the §5c.6 `-v=3` floor
- [ ] add `epponly`, `model-pvc-size`, `tokenizer-sidecar`, `vllm-keepalive` (Part 2)
- [ ] note in both `epponly.yaml` and `routing-proxy-resources.yaml` which sidecar each one sizes
- [ ] update the two inventory lists in `SKILL.md` — the `# Available fragments` comment in the `transfer.yaml` template (`:449-454`) and the generated-tree listing (`:518-522`)
- [ ] update `docs/troubleshooting.md` — the fragment table (`:13-17`), the disable example (`:24`), and the `V(5) = VERBOSE` / `§4.2` errors at `:127`
- [ ] hold `enforce-eager` pending the config.md-vs-fragment decision (#832)

Not in scope, tracked separately: conditional emission (#840), inert-fragment detection (#841).

## Related

- #840 — conditional emission (was Part 3)
- #841 — inert-fragment detection (was Part 4)
- #838 — remove `vllm-keepalive` once the llm-d-benchmark pin includes the env-macro
- #832 — `enforce_eager` missing from config.md's "values the simulation cannot supply"
- #830, #831 — bootstrap emission fixes from the same debugging session

kalantar:
Closes #839. Base: `main`.

`templates/defaults/` shipped 5 fragments; the `pd-infocomm-2` bundle carries 10, and two of the shipped five had diverged from the bundle's corrected copies. A bootstrap re-run today would drop fragments the bundle needs and reinstate two known-bad ones.

## Part 1 — the two fragments that were wrong

**`vllm-logging.yaml` did nothing.** It set `decode.vllm.additionalFlags: ["--no-disable-uvicorn-access-log"]` plus `loggingLevel: INFO`. Both keys were inert:

- `additionalFlags` is a list of scalars, so `pipeline/lib/values.py:_merge_lists` Tier 1 has the overlay *replace* the base. The defaults overlay is the **first** layer in `resolve_baseline`'s chain, and both `baselines/<name>.yaml` and the translate overlay set that same list downstream — so the fragment's entry was discarded before it could reach the rendered command.
- `loggingLevel: INFO` is already the framework default twice over (`defaults.yaml:51`, plus a `default('INFO')` fallback at `_macros.j2:230`).

It was also decode-only, while the framework's suppression (`defaults.yaml:789`, emitted at `_macros.j2:154`/`:397`) covers **both** roles — so prefill served its requests with no access log at all. Now uses the typed `vllmCommon.flags.disableUvicornAccessLog: false`, which the macro reads per-role.

**`epp-verbosity.yaml` was set to an unusable level.** EPP levels are named constants (`llm-d-router pkg/common/observability/logging/const.go:20-23`): `DEFAULT=2, VERBOSE=3, DEBUG=4, TRACE=5`. `"5"` turns on every DEBUG and TRACE site at once, and the EPP is in the request path so it costs CPU as well as disk. Now `"3"`, with the 2/3/4/5 scale documented in the fragment and in `docs/troubleshooting.md`.

`"3"` is the **floor**, not a preference: `/sim2real-check` §5c.6 reads `objectiveKey` off `Request handled` lines, attached under `V(logging.VERBOSE)` — exactly 3. At `"3"` the check runs; below it, INAPPLICABLE. Both the fragment header and the troubleshooting note say so, so the next person to lower it knows the cost.

## Part 2 — four fragments added

`epponly`, `model-pvc-size`, `tokenizer-sidecar`, `vllm-keepalive`. Ported from the bundle's hand-authored copies, with comments trimmed to the mechanism, the source cites, and each fragment's applicability — dropping the bundle-specific correction logs, which are that repo's debugging history rather than something every future experiment repo needs.

`enforce-eager` is **held**, per the issue: it controls CUDA graphs and therefore moves ITL, the metric the arms are compared on, so it is a stated experiment decision rather than a silent default (#832).

## Reviewer attention

**A test caught a real problem mid-implementation.** Nothing previously validated fragment *content* — `test_byo.py` only checks that the emitted inventory matches the directory listing, and it derives that by glob. So a fragment with a malformed shape or an unmergeable key would ship to every bootstrapped experiment silently. The new `tests/test_defaults_templates.py` asserts shape (single `scenario` entry named `defaults`, nothing but `scenario`/`shared` at top level, sets something beyond `name`), merge safety (all nine merge through the real `deep_merge` without a `_merge_lists` ValueError), and no two fragments writing the same leaf key.

Its scalar-list guard immediately failed on `epponly.yaml`: `router.proxy.args` **is itself a scalar list**. So an experiment whose `baselines/<name>.yaml` sets `router.proxy.args` — the natural place, since that's where the upstream guide scenarios put it — silently takes the whole list, dropping `--concurrency 8`, while *keeping* the fragment's `limits.cpu: "8"`. That is precisely the args/CPU-limit coupling the fragment was created to hold together, breaking one layer up.

I did not weaken the guard to green the build. Scalar-list keys are now allowlisted per fragment with a stated reason, the allowlist is checked for staleness in both directions, and `epponly.yaml`'s header carries an explicit warning. If you think the args belong in the baseline instead, that's a live design question — say so and I'll file it.

**`router.proxy` and `routing.proxy` are different components.** `epponly.yaml` sets `router.proxy.{resources,args}` — routerlib's Envoy sidecar *inside the EPP pod* (`config/charts/routerlib/values.yaml:76`). The already-shipped `routing-proxy-resources.yaml` sets `routing.proxy.resources` — llm-d-benchmark's nixl routing sidecar *beside each model server* (`defaults.yaml:627 router:` vs `:691 routing:`). Distinct containers, distinct pods, distinct jobs. Nothing previously distinguished them and the names invite deduplication, so both fragment headers now say which sidecar they size. The resolved overlay shows them coexisting as separate blocks.

**One upstream claim I did not propagate.** The bundle's `epponly.yaml` justified its args block as overriding chart defaults of `--log-level trace` and no `--concurrency` — text inherited verbatim from 8 of the 9 upstream guide scenarios. That is stale as of routerlib v0.9.0, whose envoy preset already passes `--log-level warn` **and** `--cpuset-threads` (`llm-d-router-standalone/values.yaml:253-265`). `--cpuset-threads` sizes Envoy's worker pool from the cgroup, which is strictly more adaptive than a hardcoded `--concurrency 8` — and copying the upstream block drops it. I kept the args at byte-parity with upstream rather than silently deciding that tradeoff, and recorded it in the fragment header as something to revisit.

## Line references

All of the issue's cites verify against the **pinned** llm-d-benchmark (`76473d0`, per the submodule SHA), not `main`: `defaults.yaml:51`/`:307`/`:649`/`:789`, `_macros.j2:100`/`:154`/`:230`/`:364`/`:397`, `step.py:331`. Also confirmed at the pin: `VLLM_HTTP_TIMEOUT_KEEP_ALIVE` appears nowhere, so `vllm-keepalive` is load-bearing today (#838 tracks its removal); all 9 guide scenarios override `modelPvc.size` to 1Ti; 8 of 9 carry the proxy args block. The issue body's note asking to re-derive these was my error from checking the wrong commit — I've since corrected it there.

## Stale-reference sweep

Swept `**/*.md`, `docs/`, `.claude/skills/`, and `README*` for the five old stems, hardcoded fragment counts, `verbosity: "5"`, `--no-disable-uvicorn-access-log`, and `-v=3`. Updated:

- `SKILL.md` — both inventory lists (`transfer.yaml` template comment, generated file tree), the Task 4b prose (which described fragments as "EPP llm-d.ai RBAC, request-id preservation, …" — RBAC was removed from the shipped defaults some time ago), and the templates row in the reference table.
- `docs/troubleshooting.md` — a **third** inventory location the issue didn't mention. Its fragment table is now 9 rows with a "universally correct?" column, and the EPP-verbosity note at what was line 127 said `verbosity: "5"` **(V(5) = VERBOSE)**, which was wrong independently of this PR (V(5) is TRACE), and cited "`/sim2real-check` §4.2" when the section is §5c.6. Both fixed. Its vLLM snippet also still showed the now-known-inert `additionalFlags` form; replaced with the typed key plus a note on why.

Left alone: `pipeline/README.md` and `sim2real-check/SKILL.md` reference the mechanism generically or state the `-v=3` requirement correctly — no stems, no counts, nothing stale.

No CI workflow change needed: the new test file lives in `.claude/skills/sim2real-bootstrap/tests/`, already listed.

## Verification

```
ruff check pipeline/ .claude/skills/ --select F     ->  All checks passed
python -m pytest pipeline/ + all 5 skill test dirs  ->  2352 passed, 1 skipped, 2 xfailed
                                                        coverage 97.68% (fail-under 90)
```

Also ran the real `load_defaults_overlay` over the template directory to confirm all nine merge into one `scenario[0]` named `defaults` with no key loss.

## Out of scope

Split out of #839 during vetting, both filed:

- **#840** — fragments are emitted unconditionally, but `epponly`, `model-pvc-size`, `tokenizer-sidecar`, and `preserve-request-id` are topology-, model-, or plugin-specific. Also needs BLIS and BYO to agree on the default posture — BYO already pre-disables every stem, BLIS enables every stem.
- **#841** — nothing surfaces a fragment with no effect on the resolved scenario. The new tests cover the statically-decidable slice (a key that provably cannot survive the merge chain); the resolve-diff check is #841's.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -346,10 +346,10 @@ ls "$EXPERIMENT_ROOT"/workloads/*.yaml
 
 **action:** shell
 
-Framework workarounds documented in `docs/troubleshooting.md` (EPP llm-d.ai
-RBAC, request-id preservation, EPP/vLLM verbosity, routing-proxy resource
-requests) are applied automatically by `sim2real assemble` from
-`<experiment-root>/baselines/defaults/`. Copy the framework templates into
+Framework workarounds documented in `docs/troubleshooting.md` (request-id
+preservation, EPP/vLLM verbosity, sidecar resource requests, model-PVC sizing,
+topology and tokenizer wiring) are applied automatically by `sim2real assemble`
+from `<experiment-root>/baselines/defaults/`. Copy the framework templates into
 the experiment so each experiment is self-contained and reproducible.
 
 ```bash
@@ -361,6 +361,13 @@ ls "$EXPERIMENT_ROOT/baselines/defaults/"
 Each fragment is a partial scenario YAML — operators can edit individual
 fragments in place to tweak them for the experiment, or list a fragment
 stem under `defaults.disable` in `transfer.yaml` to opt out entirely.
+
+**Not every fragment is universally correct.** `epponly` is a topology decision,
+`model-pvc-size` is sized for a 70B-class model, `tokenizer-sidecar` matters only
+when the arms declare a `token-producer` plugin, and `preserve-request-id` matches
+nothing under an epponly / externally-managed gateway. Each fragment's header
+comment states its own applicability and when to disable it. Making emission
+conditional rather than leaving that to the operator is tracked as #840.
 
 ---
 
@@ -448,9 +455,13 @@ defaults:
   disable: []
   # Available fragments (filename stems in baselines/defaults/):
   #   - epp-verbosity
+  #   - epponly
   #   - externally-managed-gateway
+  #   - model-pvc-size
   #   - preserve-request-id
   #   - routing-proxy-resources
+  #   - tokenizer-sidecar
+  #   - vllm-keepalive
   #   - vllm-logging
 ```
 
@@ -515,11 +526,15 @@ Exit code 0 and all fields printed = success.
   baselines/
     <name>.yaml                  <- task-3
     defaults/                    <- task-4b
-      preserve-request-id.yaml
       epp-verbosity.yaml
-      vllm-logging.yaml
+      epponly.yaml
       externally-managed-gateway.yaml
+      model-pvc-size.yaml
+      preserve-request-id.yaml
       routing-proxy-resources.yaml
+      tokenizer-sidecar.yaml
+      vllm-keepalive.yaml
+      vllm-logging.yaml
   <component-name>/             <- task-2 (submodule)
   algorithms/
     <algorithm>.go               <- pre-existing
@@ -674,4 +689,4 @@ This skill ships with supporting files in its directory. Invoke in place — do 
 | `generate_scenarios.py` | Converts JSON config (`top3_selection.json`) → scenario YAMLs. Use when JSON input exists (BLIS). |
 | `generate_scenarios.README.md` | Coverage map for the JSON-input path. Documents field mappings, omission rules, and gaps. |
 | `byo.py` | Implements the `--byo` branch — argument parsing, YAML validation, path-safe copy operations, `transfer.yaml` emission, batched `sim2real translation register` command generation. Invoked by SKILL.md's dispatch when `--byo` (or any BYO-only flag) is passed. |
-| `templates/defaults/*.yaml` | Framework-owned baseline workaround fragments (RBAC, request-id, verbosity). Copied into `<experiment-root>/baselines/defaults/` at BLIS task-4b and at BYO run time, so every experiment is self-contained and reproducible. |
+| `templates/defaults/*.yaml` | Framework-owned baseline workaround fragments (request-id, verbosity, sidecar sizing, model-PVC size, topology, tokenizer). Copied into `<experiment-root>/baselines/defaults/` at BLIS task-4b and at BYO run time, so every experiment is self-contained and reproducible. Shape and merge-safety are asserted by `tests/test_defaults_templates.py`. |
--- a/.claude/skills/sim2real-bootstrap/templates/defaults/epp-verbosity.yaml
+++ b/.claude/skills/sim2real-bootstrap/templates/defaults/epp-verbosity.yaml
@@ -1,8 +1,38 @@
-# Workaround: bump EPP log verbosity for diagnosability.
+# Raise EPP log verbosity above the framework default for diagnosability.
 # See docs/troubleshooting.md "Increasing logging verbosity / EPP".
+#
+# THE SCALE. llm-d-router's EPP logs through named constants, not raw numbers
+# (pkg/common/observability/logging/const.go:20-23, v0.9.0):
+#   DEFAULT = 2    VERBOSE = 3    DEBUG = 4    TRACE = 5
+# llm-d-benchmark defaults `router.epp.verbosity` to "1"
+# (config/templates/values/defaults.yaml:649 at the pinned SHA), i.e. below even
+# DEFAULT, so the EPP is near-silent out of the box.
+#
+# WHY "3" AND NOT "5". DEBUG(4) is by far the largest bucket of call sites in the
+# EPP, and the EPP sits in the request path -- so the logging costs CPU on every
+# request as well as disk. "5" turns on every DEBUG and TRACE site at once, which
+# in practice is more log volume than a run can carry.
+#
+# "3" IS THE FLOOR, NOT A PREFERENCE. /sim2real-check S5c.6 (InferenceObjective
+# resolution at runtime) reads `objectiveKey` and `priority` off `Request handled`
+# lines, which are attached under V(logging.VERBOSE) -- i.e. exactly 3. At "3"
+# that check runs; below "3" it reports INAPPLICABLE and the run loses its
+# objective-resolution evidence. Do not lower this without accepting that.
+#
+# ESCALATING. "3" stops BELOW the level where the P/D decider explains itself
+# (prefix_based_pd_decider.go uses V(logging.DEBUG)), so the escalation for
+# prefill-selection problems is "4" -- which keeps the decider diagnostics and
+# still drops every TRACE site. Reach for "5" only for a specific TRACE-level
+# call site you have already identified.
+#
+# WHY THIS IS A FRAGMENT AND NOT PART OF baselines/<name>.yaml. Verbosity is a
+# diagnostic, not part of the measured mechanism, so nothing in config.md declares
+# it. Keeping it here means `transfer.yaml: defaults.disable` can switch it off
+# for a measurement run; a key in baselines/<name>.yaml has no such switch.
+#
 # scenario[0].name is realigned to the experiment baseline at merge time.
 scenario:
 - name: defaults
   router:
     epp:
-      verbosity: "5"
+      verbosity: "3"
--- a/.claude/skills/sim2real-bootstrap/templates/defaults/epponly.yaml
+++ b/.claude/skills/sim2real-bootstrap/templates/defaults/epponly.yaml
@@ -1,0 +1,82 @@
+# Standalone router topology: no Kubernetes Gateway is deployed; the EPP runs with
+# an Envoy sidecar that serves HTTP directly. Sizes and configures that sidecar,
+# because under this topology it becomes the WHOLE data plane.
+#
+# THIS FRAGMENT SIZES A DIFFERENT SIDECAR THAN routing-proxy-resources.yaml.
+# The two look like duplicates and are not:
+#   router.proxy   (here)  -> the Envoy sidecar INSIDE THE EPP POD. Maps onto the
+#                             routerlib chart's top-level `proxy:`
+#                             (config/charts/routerlib/values.yaml:76, v0.9.0).
+#   routing.proxy  (there) -> llm-d-benchmark's nixl routing sidecar BESIDE EACH
+#                             MODEL SERVER, the peer of vllmCommon.kvTransfer.
+# They are distinct containers in distinct pods with distinct jobs. Both fragments
+# are intended to coexist; do not "deduplicate" them.
+#
+# THE TWO KEYS BELOW ARE COUPLED, which is why they live in one file:
+# `--concurrency 8` is chosen to match the `limits.cpu: "8"` set immediately below.
+# Change one without the other and you either starve Envoy's workers or
+# oversubscribe them.
+#
+# WARNING -- THE COUPLING CAN BREAK SILENTLY. `args` is a list of SCALARS, so
+# pipeline/lib/values.py `_merge_lists` Tier 1 has any downstream layer REPLACE it
+# wholesale rather than merge into it. The defaults overlay is the first layer in
+# resolve_baseline's chain, so if your `baselines/<name>.yaml` or the registered
+# treatment overlay sets `router.proxy.args` at all, it takes the whole list -- and
+# `--concurrency 8` goes with it -- while `limits.cpu: "8"` below still applies.
+# That is the failure this file was written to prevent, reappearing one layer up.
+# Set the args HERE, or set BOTH keys in your baseline; never split them.
+# (`.claude/skills/sim2real-bootstrap/tests/test_defaults_templates.py` allowlists
+# this one key so a new fragment cannot add another scalar list unnoticed.)
+#
+# THE ARGS BLOCK IS UPSTREAM BOILERPLATE, not a local invention. It appears
+# byte-for-byte -- comment included -- in 8 of the 9 shipped llm-d-benchmark guide
+# scenarios at the pinned SHA (config/scenarios/guides/: optimized-baseline,
+# flow-control, agentic-serving, predicted-latency-routing, tiered-prefix-cache,
+# precise-prefix-cache-routing, pd-disaggregation, wide-ep-lws; only
+# workload-autoscaling omits it). That every shipped scenario overrides the same
+# chart defaults identically is an argument for them being framework defaults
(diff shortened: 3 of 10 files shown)


kalantar:
## Automated review — `review-changes` workflow

> **Provenance, stated plainly:** this is machine-generated review output, not a human sign-off, and it is posted by the same session that authored the PR. Four specialist reviewers ran in isolated fresh contexts with no access to the authoring session's reasoning, then an adversarial verifier-synthesizer re-read the code before ruling. The workflow's own posting step was blocked by a safety classifier (it base64-encoded the payload and told the posting agent not to inspect it — fair to block), so the verdict was read from the run journal and is reproduced here in the clear. One input to human review, not a substitute for it.

**Verdict: READY TO MERGE**

> Unusually well-verified change — every citation and merge-mechanics claim I spot-checked against the pinned submodule and pipeline/lib/values.py held up, the new guard test covers the actual root-cause bug class, and all 233 skill tests pass.

`must_fix`: none. `important`: none.

### Per-lens results

| Lens | Verdict | Findings |
|---|---|---|
| `code-reviewer` | READY | 0 |
| `comment-analyzer` | READY | 0 |
| `silent-failure-hunter` | READY | 0 |
| `pr-test-analyzer` | **NEEDS CHANGES** | 4, all self-marked `non-blocking` |

One lens dissented. The synthesizer, after re-reading the code, folded its four findings into a single advisory recommendation and ruled READY TO MERGE. Both are recorded here so the disagreement stays visible.

### Recommendation (advisory — does not block merge)

**Consider a single assertion that epp-verbosity's value meets the documented §5c.6 floor.** — `tests/test_defaults_templates.py`, via `pr-test-analyzer`

> The new tests are entirely glob-driven and structural, so nothing enforces the one value in the diff that carries a downstream contract: docs/troubleshooting.md and the fragment header both state /sim2real-check §5c.6 goes INAPPLICABLE below "3". A one-line `int(verbosity) >= 3` check would encode that floor rather than duplicate the constant. The other Part-1 fix needs nothing — a revert to `additionalFlags` is already caught by test_fragment_sets_no_unacknowledged_scalar_list_key, since that key is a scalar list absent from _ALLOWED_SCALAR_LISTS.

### The dissenting lens, in full

All four self-marked `non-blocking`, sharing one theme: the new tests validate fragment *shape* but not fragment *values*.

1. **`tests/test_defaults_templates.py`** — no test asserts the corrected Part-1 literals (`epp-verbosity == "3"`, `disableUvicornAccessLog == false`). A future revert of either passes the whole 233-test suite silently.
2. **`templates/defaults/epponly.yaml`** — the header documents that `--concurrency 8` must track `limits.cpu: "8"`, but nothing enforces it; the comment itself calls this a coupling that can break silently.
3. **`tests/test_defaults_templates.py`** — `model-pvc-size`, `tokenizer-sidecar`, `vllm-keepalive` covered for shape only, not payload values (`1Ti`; `enabled: true` + HF_TOKEN/HF_HOME wiring; `120` on prefill only).
4. **`SKILL.md`** — the two hand-maintained inventory lists have no test tying them to the actual `templates/defaults/` contents.

Same lens, on the other side: *"The added tests are well-targeted at the specific root-cause bug class from the issue (scalar-list replacement under deep_merge) and use the real production merge function, so they would genuinely catch a reintroduction of that class of bug."*

### What the reviewers independently verified

`comment-analyzer` initialized the `llm-d-benchmark` submodule at the pin recorded in this tree (`76473d05a0f546a6bc0c4662ed7e34104da73891`) and checked every citation in the diff against it. All exact: `defaults.yaml:789` / `:51` / `:307` / `:649` / `:627` / `:691`, `_macros.j2:100` / `:154` / `:230` / `:364` / `:397`, `step.py:331`. Also confirmed all 9 guide scenarios override `modelPvc.size` to `1Ti`; the `router.proxy.args` block appears byte-for-byte in exactly 8 of 9 with `workload-autoscaling.yaml` the sole omission; `http_timeout_keep_alive` appears nowhere in the pinned tree; and the pin predates `28cfc06` per `git merge-base --is-ancestor`.

Both `comment-analyzer` and `code-reviewer` noted this **contradicts the parent issue's own "line references need re-checking against the pin" caveat** — that caveat came from checking an unpinned commit by mistake and has since been corrected on #839. The numbers in the merged fragments are right.

`silent-failure-hunter` found no error-handling surface to audit (YAML fragments, Markdown, one test file) and judged the single `except ValueError` in the new test appropriately narrow and loud.

`code-reviewer` walked #839's checklist item by item and confirmed `enforce-eager` was correctly withheld and does not leak into either SKILL.md inventory, the troubleshooting table, or the unrelated pre-existing `enforce_eager` logic in `generate_scenarios.py` / `generate_from_config.py`.

Issue #845, opened 2026-08-24
bootstrap: kvTransfer.role is shared by both roles, so P/D bundles must use the deprecated kv_both


kalantar:
## Summary

`vllmCommon.kvTransfer.role` is shared by both the prefill and decode roles, so a
P/D bundle can only state one value for both. The only value that works for both
is `kv_both`, which vLLM has deprecated for NixlConnector. Split out of #830,
which introduces the `kvTransfer` block and has to ship the deprecated value
because the correct one is not expressible today.

## What vLLM wants

NixlConnector wants the role to say which side of the transfer a worker is on:

- prefill workers produce KV blocks -> `kv_producer`
- decode workers consume them      -> `kv_consumer`

`kv_both` says "either", and vLLM warns it "is now deprecated for NixlConnector
and will be removed in a future release". So the value #830 emits is correct
today and will stop being correct.

## Why it cannot be fixed in #830

The key lives under `vllmCommon`, which by construction applies to both roles.
There is no `prefill.kvTransfer` or `decode.kvTransfer` to write a per-role value
into, so there is no way to say producer-on-prefill / consumer-on-decode from a
sim2real bundle. This is a vocabulary gap, not a bug in the generator: the
generator is emitting the only role value the schema can express.

Note the shape of the gap is the same one the prefill block already documents for
parallelism and vLLM flags (`generate_from_config.py`): those are shared rather
than per-role because no bundle to date states them per role. `kvTransfer.role`
is different in that upstream has an opinion about it being per-role, so the
shared key is actively wrong rather than merely coarse.

## What needs to happen

Roughly in order of how much has to change:

1. **Confirm the upstream shape first.** Check whether the pinned llm-d-benchmark
   has grown a per-role KV transfer key (`prefill.kvTransfer` / `decode.kvTransfer`,
   or a per-role `role` override) by the time this is picked up. If it has, the fix
   is to emit it and this issue is mostly mechanical.
2. **If upstream still has only `vllmCommon.kvTransfer`,** decide whether sim2real
   should pass the role per-role through some other route — e.g. a per-role
   `additionalFlags` entry carrying `--kv-transfer-config` directly, as
   llm-d-benchmark's own `config/scenarios/guides/pd-disaggregation.yaml` does.
   Note that path has its own hazard: `additionalFlags` is a list of scalars, so
   `pipeline/lib/values.py:_merge_lists` Tier 1 has every downstream layer replace
[...]

PR #846, opened 2026-08-24
fix(bootstrap): enable KV transfer whenever a prefill pool is emitted (closes #830)

### LINKED ISSUE #830: bootstrap: prefill pool emitted without enabling KV transfer, so P/D silently never happens
### Component

Scenarios

### Describe the bug

When `config.md` names a prefill pod count, `generate_from_config.py` emits a
prefill pool but never enables KV transfer. The result is a scenario that reads
as disaggregated and is not: vLLM starts without `--kv-transfer-config`, so no
KV connector is instantiated, the prefill pod sits idle, and the decode pods do
the prefill work themselves.

`vllmCommon.kvTransfer.enabled` defaults to `false`
(`llm-d-benchmark/config/templates/values/defaults.yaml:725-726`), and the flag
is gated on it:

```jinja
{# _macros.j2:103 #}
{% set has_kv_transfer = vllmCommon.kvTransfer is defined and vllmCommon.kvTransfer.enabled -%}
```

A bundle that states a prefill count and nothing else therefore gets the pool
without the transport. Per the upstream guide, P/D "requires a KV transfer
backend to move KV cache blocks from prefill workers to decode workers"
(llm-d/guides/pd-disaggregation/README.md).

This is the same failure class the prefill block already guards against one
layer up. Its own comment (`generate_from_config.py:794-804`) reads:

> `enabled: true` is NOT decoration. pipeline/lib/capacity.py:238-241 defaults
> prefill to ("prefill", False, 0), so a prefill block without it is skipped by
> capacity planning -- the scenario would read as disaggregated while planning
> zero prefill GPUs. config/scenarios/guides/pd-disaggregation.yaml sets it
> explicitly for the same reason.

The capacity-planning layer is guarded; the model-server layer is not. The
reference consulted for that lesson, `pd-disaggregation.yaml`, is also the file
that carries `--kv-transfer-config '{"kv_connector":"NixlConnector",
"kv_role":"kv_both"}'`.

### Steps to reproduce

1. Author a `config.md` whose `## vLLM Pod Configuration` table includes a
   prefill row, e.g. `| Number of prefill pods | 1 |`.
2. Run `/sim2real-bootstrap` to generate `baselines/baseline.yaml`.
3. Observe the generated scenario has `prefill: {enabled: true, replicas: 1}`
   and no `vllmCommon.kvTransfer`.
4. Deploy. On the decode pod:
   - `kubectl logs <decode-pod> -c vllm | grep -ic nixl` → `0`
   - vLLM's resolved engine config contains no `kv_transfer_config`
   - decode reports nonzero `Avg prompt throughput`, i.e. it is prefilling
   - the prefill pod logs zero requests for the whole run

Observed on a 1P2D Llama-3.3-70B deployment: 415 requests, all served by decode,
prefill idle throughout.

### Additional context

Proposed fix — inside the existing `if prefill_replicas > 0:` block
(~`generate_from_config.py:809`), emit:

```yaml
vllmCommon:
  kvTransfer:
    enabled: true
    connector: NixlConnector
    role: kv_both
```

Stating `connector`/`role` rather than inheriting them follows the principle the
specify skill already applies to the `blis observe` block: "Emit it even when
every value equals the pipeline default, so the values are this bundle's
decision and not a downstream fallback."

Two implementation notes:

- `scenario["vllmCommon"]` may already exist from the `enforceEager` override at
  line 789, which runs before the prefill block. Use `setdefault`/merge, or one
  assignment silently drops the other.
- Add a provenance entry for the new keys; the generator already writes
  provenance comments.

Caveat worth recording in the fix: `kv_role: kv_both` is deprecated for
NixlConnector in vLLM ("This value is now deprecated for NixlConnector and will
be removed in a future release"), which wants `kv_producer` on prefill and
`kv_consumer` on decode. That is not expressible today because `vllmCommon` is
shared by both roles and there is no `prefill.kvTransfer`. Filed separately if
wanted.

Suggested test: a `config.md` fixture with a prefill count produces a scenario
containing `vllmCommon.kvTransfer.enabled: true`; one without a prefill count
regenerates byte-identically.

kalantar:
Closes #830. Base: `main`.

A `config.md` naming a prefill pod count produced a scenario that **reads as disaggregated and is not**. `vllmCommon.kvTransfer.enabled` defaults to `false` upstream (`llm-d-benchmark config/templates/values/defaults.yaml:725-726`) and vLLM's `--kv-transfer-config` is gated on it (`_macros.j2:103` sets `has_kv_transfer` inside the single mode-parameterized macro `build_vllm_command(mode)` at `:83-188`, which emits the flag at `:111` and `:169` and is invoked for both roles from `13_ms-values.yaml.j2` — `:409` decode, `:826` prefill). So no KV connector was instantiated: the prefill pod was never routed to and logged zero requests, while the decode pods prefilled their own. Nothing errored — observed on a 1P2D Llama-3.3-70B run as **415 requests all served by decode, prefill idle throughout**.

This is the same failure class as the `enabled: true` the prefill block already carries, one layer down: an upstream default of "off" turning a stated intention into silence. That one is guarded at the capacity-planning layer; this is the model-server layer.

All line references verified against the **pinned** llm-d-benchmark (`76473d0`, per the submodule SHA), not `main`.

## Reviewer attention

**Fixed in BOTH generators — the issue named only one.** `generate_from_config.py` is what #830 cites, but `generate_scenarios.py` (the BLIS-driven path) emits its own prefill block at `:247-273`, carries the *same* "enabled: true is load-bearing" comment, and had `kvTransfer` nowhere either. Fixing one path would have left half of bootstrap emitting an inert prefill pool. A cross-generator test now pins them together.

**The issue's proposed fix was insufficient as written, in both files.** #830 says to merge into `scenario["vllmCommon"]` with `setdefault` — correct as far as it goes, and I did that. But **both generators hand-roll their YAML** rather than dumping it, and both hardcoded the `vllmCommon` subtree to exactly `flags.enforceEager`:

- `generate_from_config.py:895-902`
- `generate_scenarios.py:304-308`

So a `kvTransfer` key added to the dict would have been **silently dropped from the output**. Worse, `generate_scenarios.py:308` accessed `scenario['vllmCommon']['flags']['enforceEager']` unconditionally once `"vllmCommon" in scenario`, so a kvTransfer-only scenario (prefill pool, no `enforce_eager` override) would have raised **KeyError**. Both emitters now branch on each subtree independently, and `test_enforce_eager_alone_still_emits_flags` guards the reverse direction.

**Tests assert on emitted text, not the intermediate dict.** That is deliberate and is the point of the above: a dict-level assertion passes while the emitter drops the key, which is precisely the shape of gap that produced #830. Every new test goes through `emit()` / `write_commented_yaml()` and parses the result.

**`role: kv_both` is knowingly deprecated.** NixlConnector wants `kv_producer` on prefill and `kv_consumer` on decode, but `vllmCommon` is shared by both roles and there is no `prefill.kvTransfer`, so per-role values are not expressible today. Filed as **#845** with the trigger conditions for revisiting, and the emitted provenance comment points at it rather than staying silent.

**`connector`/`role` are stated, not inherited.** Upstream anchors already default them (`defaults.yaml:56-57` → `NixlConnector` / `kv_both`), so emitting them is redundant in effect but deliberate in intent — the same principle the specify skill applies to the `blis observe` block: the values are this bundle's decision, not a downstream fallback.

**Note on the reference scenario.** #830 cites `pd-disaggregation.yaml` as carrying `--kv-transfer-config`. It does — but by hand-writing the raw flag into a command block (`:298`, `:387`), not via the typed `kvTransfer` key. The typed key used here is the better route: a raw flag in `additionalFlags` is a scalar list, so `pipeline/lib/values.py:_merge_lists` Tier 1 has every downstream layer replace it wholesale (the #839 failure class).

## Tests added

Seven, in `test_generate_from_config_prefill.py`:

| Test | Guards |
|---|---|
| `test_prefill_pool_emits_kv_transfer` | the fix itself, in emitted YAML |
| `test_no_prefill_pool_emits_no_kv_transfer` | aggregated bundles regenerate byte-identically |
| `test_stated_zero_prefill_emits_no_kv_transfer` | a stated `0` means aggregated |
| `test_kv_transfer_and_enforce_eager_coexist` | the overwrite trap **and** the emitter-branch trap |
| `test_enforce_eager_alone_still_emits_flags` | no regression on the flags-only path |
| `test_kv_transfer_carries_provenance` | every emitted key has a `#` source comment, and `#845` is referenced |
| `test_kv_transfer_agrees_across_both_generators` | both generators emit identical transport settings |

## Stale-reference sweep

Swept `**/*.md`, `docs/`, `.claude/skills/`, and `README*` for `kvTransfer` / `kv_transfer` / "KV transfer" and for `prefill`. One hit needed updating: `sim2real-bootstrap/SKILL.md` documents the generated scenario shape and the disaggregation rules, and both were now incomplete — the shape block showed a prefill pool with no `vllmCommon.kvTransfer`, and the notes explained why `enabled: true` is load-bearing without the equivalent for the transport. Both updated, with the #845 caveat recorded.

No other doc mentions KV transfer. `pipeline/README.md` is unaffected — this is bootstrap-side generation, not a pipeline stage contract.

## Verification

```
ruff check pipeline/ .claude/skills/ --select F     ->  All checks passed
python -m pytest pipeline/ + all 5 skill test dirs  ->  2359 passed, 1 skipped, 2 xfailed
                                                        coverage 97.68% (fail-under 90)
```

Also rendered a real scenario by hand to confirm `flags` and `kvTransfer` coexist correctly under one `vllmCommon:` with provenance comments on every line.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -275,7 +275,7 @@ scenario:
       additionalFlags:
       - "--flag=value"
 
-  # Emitted ONLY when config.md names a prefill pod count (issue #824).
+  # Emitted ONLY when config.md names a prefill pod count (issues #824, #830).
   # Without those rows the output is exactly the single-decode shape above.
   prefill:
     enabled: true                      # required — see note below
@@ -286,6 +286,14 @@ scenario:
     vllm:
       additionalFlags:
       - "--flag=value"                 # shared with decode, not per-role
+
+  # Also emitted with a prefill pool, and only then (issue #830). This is what
+  # makes the pool do anything; see the note below.
+  vllmCommon:
+    kvTransfer:
+      enabled: true
+      connector: NixlConnector
+      role: kv_both
 ```
 
 **Disaggregation (issue #824).** To get a `prefill:` block, `config.md`'s vLLM
@@ -297,6 +305,16 @@ optional per-role accelerator overrides; without them both roles use the shared
 - `enabled: true` is **required**, not decorative. `pipeline/lib/capacity.py`
   defaults prefill to disabled with 0 replicas, so a `prefill:` block lacking it
   reads as disaggregated while planning zero prefill GPUs.
+- **`vllmCommon.kvTransfer` is equally required (issue #830)** — same failure
+  shape, one layer down. `vllmCommon.kvTransfer.enabled` defaults to `false`
+  upstream and vLLM's `--kv-transfer-config` flag is gated on it, so a prefill
+  pool without it gets no KV connector: the prefill pod is never routed to and
+  logs zero requests, while the decode pods prefill their own requests. Nothing
+  errors — the run completes and the results are silently not P/D. Both
+  generators emit it whenever they emit a prefill pool.
+  `role: kv_both` is deprecated for NixlConnector, which wants `kv_producer` on
+  prefill and `kv_consumer` on decode; `vllmCommon` is shared by both roles so
+  per-role values are not expressible today (tracked as #845).
 - `parallelism` and `vllm.additionalFlags` are shared across roles — `config.md`
   has no per-role form for them.
 - **One GPU type per role.** A GPU cell naming several types (`H100, A100`) emits
--- a/.claude/skills/sim2real-bootstrap/generate_from_config.py
+++ b/.claude/skills/sim2real-bootstrap/generate_from_config.py
@@ -833,8 +833,11 @@ def build_scenario(
         # KV transfer is what makes the prefill pool actually do anything (#830).
         # vllmCommon.kvTransfer.enabled defaults to false upstream
         # (llm-d-benchmark config/templates/values/defaults.yaml:725-726) and the
-        # --kv-transfer-config flag is gated on it (_macros.j2:103, emitted at
-        # :110 for decode and :354 for prefill). So a prefill pool WITHOUT this
+        # --kv-transfer-config flag is gated on it: _macros.j2:103 sets
+        # has_kv_transfer inside the single mode-parameterized macro
+        # build_vllm_command(mode) (:83-188), which emits the flag at :111 and
+        # :169 and is invoked for BOTH roles from 13_ms-values.yaml.j2 (:409
+        # decode, :826 prefill). So a prefill pool WITHOUT this
         # block reads as disaggregated and is not: no KV connector is
         # instantiated, the prefill pod is never routed to and logs zero
         # requests, and the decode pods do the prefill work themselves. Nothing
--- a/.claude/skills/sim2real-bootstrap/generate_from_config.py
+++ b/.claude/skills/sim2real-bootstrap/generate_from_config.py
@@ -835,7 +835,7 @@ def build_scenario(
         # (llm-d-benchmark config/templates/values/defaults.yaml:725-726) and the
         # --kv-transfer-config flag is gated on it: _macros.j2:103 sets
         # has_kv_transfer inside the single mode-parameterized macro
-        # build_vllm_command(mode) (:83-188), which emits the flag at :111 and
+        # build_vllm_command(mode) (:83-180), which emits the flag at :111 and
         # :169 and is invoked for BOTH roles from 13_ms-values.yaml.j2 (:409
         # decode, :826 prefill). So a prefill pool WITHOUT this
         # block reads as disaggregated and is not: no KV connector is
--- a/.claude/skills/sim2real-bootstrap/generate_from_config.py
+++ b/.claude/skills/sim2real-bootstrap/generate_from_config.py
@@ -829,6 +829,40 @@ def build_scenario(
         if additional_flags:
             prefill["vllm"] = {"additionalFlags": additional_flags}
         scenario["prefill"] = prefill
+
+        # KV transfer is what makes the prefill pool actually do anything (#830).
+        # vllmCommon.kvTransfer.enabled defaults to false upstream
+        # (llm-d-benchmark config/templates/values/defaults.yaml:725-726) and the
+        # --kv-transfer-config flag is gated on it (_macros.j2:103, emitted at
+        # :110 for decode and :354 for prefill). So a prefill pool WITHOUT this
+        # block reads as disaggregated and is not: no KV connector is
+        # instantiated, the prefill pod is never routed to and logs zero
+        # requests, and the decode pods do the prefill work themselves. Nothing
+        # errors -- observed as 415 requests all served by decode with prefill
+        # idle throughout.
+        #
+        # Same failure class as `enabled: true` above, one layer down: an upstream
+        # default of "off" turning a stated intention into silence. That one is
+        # guarded at the capacity-planning layer; this is the model-server layer.
+        #
+        # connector/role are stated rather than inherited from the upstream
+        # anchors (defaults.yaml:56-57, NixlConnector / kv_both) so the values are
+        # this bundle's decision and not a downstream fallback -- the same
+        # principle the specify skill applies to the `blis observe` block.
+        #
+        # `role: kv_both` is DEPRECATED for NixlConnector, which wants
+        # kv_producer on prefill and kv_consumer on decode. Not expressible today:
+        # vllmCommon is shared by both roles and there is no prefill.kvTransfer.
+        # Tracked as #845; kv_both works and warns until then.
+        #
+        # setdefault, NOT assignment: the enforce_eager override above may already
+        # have created scenario["vllmCommon"], and a plain assignment here would
+        # silently drop it.
+        scenario.setdefault("vllmCommon", {})["kvTransfer"] = {
+            "enabled": True,
+            "connector": "NixlConnector",
+            "role": "kv_both",
+        }
     elif "prefill_hardware" in fields:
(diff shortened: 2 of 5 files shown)


kalantar:

> **Automated review.** Generated by the `review-changes` workflow, not a human sign-off. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Treat this as one input to human review.

The behavioral fix is correct in both generators, but one of the two fixed paths is left without any regression coverage and a new upstream citation is wrong.

### Important
- **generate_scenarios.py's new vllmCommon branches have no regression tests — two injected bugs both leave the full suite green.** (`.claude/skills/sim2real-bootstrap/generate_scenarios.py:292`) — I injected the exact bug the code comment at generate_scenarios.py:290 warns against (replacing the setdefault at :292 with a plain `scenario["vllmCommon"] = {"kvTransfer": ...}`, which drops the flags dict assigned at :242) and separately narrowed the new `if "flags" in scenario["vllmCommon"]` branch at :334 to also require kvTransfer; both regressions passed all 240 tests in .claude/skills/sim2real-bootstrap/tests/. The equivalent paths in generate_from_config.py are covered by test_kv_transfer_and_enforce_eager_coexist and test_enforce_eager_alone_still_emits_flags, and the one cross-generator test never combines enforce_eager with a prefill pool, so nothing guards this generator.
- **The comment's "_macros.j2 ... :354 for prefill" citation points at the wrong macro.** (`.claude/skills/sim2real-bootstrap/generate_from_config.py:837`) — At the pinned llm-d-benchmark submodule (76473d0), _macros.j2:103 sets has_kv_transfer inside the single macro `build_vllm_command(mode)` (:83-188), whose flag emission sites are :111 and :169; that one macro is invoked for both roles from 13_ms-values.yaml.j2 (:409 decode, :826 prefill). Line 354 lives in `build_standalone_vllm_command` (:336, invoked only from 14_standalone-deployment_yaml.j2:240), which has nothing to do with the prefill role — the sibling comment in generate_scenarios.py stops at ":103" and is correct, so the two duplicated blocks also disagree.

kalantar:
Both blocking findings addressed in f769876.

**The missing tests.** You were right that `generate_scenarios.py` had no regression coverage for its two `vllmCommon` populators, and the fault injection made that undeniable. Added four tests to `test_generate_scenarios.py` — the dedicated file this PR had not touched — covering kvTransfer emitted, kvTransfer absent, `enforce_eager` + prefill coexistence, and flags-only emission.

I checked the fix the same way you found the gap rather than taking it on faith. Re-injecting the plain-assignment clobber now fails `test_kv_transfer_and_enforce_eager_coexist` with `KeyError: 'flags'`, and re-narrowing the `if "flags" in ...` branch to also require kvTransfer now fails `test_enforce_eager_alone_still_emits_flags` with a `TypeError`. Both previously left the whole suite green.

**The citation.** Wrong, and your reading of the macro structure is exactly right. At the pin, `_macros.j2:103` sets `has_kv_transfer` inside the single mode-parameterized `build_vllm_command(mode)` (`:83-188`), which emits the flag at `:111` and `:169` and is invoked for both roles from `13_ms-values.yaml.j2` (`:409` decode, `:826` prefill). The `:354` I cited is inside `build_standalone_vllm_command` (`:336`), which has nothing to do with the prefill role. I verified the macro boundaries, both emission sites, and both invocation sites against the pinned submodule before rewriting the comment. The PR description carried the same error and has been corrected too.

Good catch on the two duplicated comment blocks disagreeing — `generate_scenarios.py` stopping at `:103` was the correct one, and that asymmetry is what should have tipped me off.

Nothing skipped. Lint clean; 2363 passed, coverage 97.68%.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow, not a human sign-off. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS DISCUSSION, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Treat this as one input to human review.

The fix is correct, correctly ordered against the existing enforce_eager writer, symmetrically applied to both generators, and covered by tests that assert on emitted YAML rather than the intermediate dict — nothing blocks merge.

### Recommendations
- **The macro citation `(:83-188)` should be `(:83-180)` — the range overruns into the next macro's banner.** (`.claude/skills/sim2real-bootstrap/generate_from_config.py:833`) — In the pinned submodule (76473d0), `{% macro build_vllm_command(mode='decode') -%}` opens at _macros.j2:83 and its `{%- endmacro %}` is line 180; 181-188 are blank lines plus the doc banner for `build_ms_env_vars` (:189). Every other citation in the same comment checks out exactly (:103 has_kv_transfer, :111 and :169 --kv-transfer-config, defaults.yaml:56 kv_connector, :725-726 kvTransfer/enabled: false), so this one number is the outlier.

Issue #850, opened 2026-08-25
bootstrap: emit CPU/memory for decode and prefill — generous limits, modest explicit requests, warn that they are unmeasured


kalantar:
## Summary

Bootstrap emits no CPU/memory for the model-server pods, so bundles either inherit the
framework default (4 CPU for a pod that may hold 4 GPUs) or hand-copy numbers from an
upstream scenario. Both go wrong, in opposite directions.

Proposal: emit **generous limits with modest explicit requests**, plus a warning that
the values are unmeasured.

## What goes wrong today

**Too small.** The framework default of 4 CPU for a 4-GPU pod — sharing its cgroup with
the routing sidecar, which requests 4 more — starves vLLM. Observed signal:
`Reducing Torch parallelism from 112 threads to 1`, which surfaces as ITL noise, i.e.
it corrupts the metric arms are compared on rather than failing loudly.

**Copied from the wrong model.** `pd-infocomm-2` currently carries decode
`128Gi / 32` and prefill `16Gi / 8`, taken verbatim from
`config/scenarios/guides/pd-disaggregation.yaml`. Those were sized there for
**Qwen/Qwen3-32B** at `maxModelLen: 16384`, with `tensor: 1` on prefill and `tensor: 2`
on decode. The bundle runs **Llama-3.3-70B** at `maxModelLen: 131072`, TP=4 on both
roles. The prefill figure in particular was upstream's TP=1 number being applied to
four GPU workers.

## The requests/limits trap

`limits` without `requests` is not "no reservation" — Kubernetes copies the limit into
the request when the request is absent. So `pd-infocomm-2`'s decode block
(`limits: {128Gi, 32}`, no `requests`) effectively **requests** 128Gi and 32 CPU, times
2 replicas: 256Gi and 64 CPU reserved before prefill is scheduled at all. On a busy
cluster that is an unschedulable pod, which reads as a capacity problem rather than a
config one.

So the two halves need stating separately:

- **limits** — generous. Headroom costs nothing unless used.
- **requests** — modest and explicit. This is what the scheduler reserves.

## Warning

Two places:

1. A generation-time warning, alongside the existing
   `model '<id>' not in MODEL_METADATA, deriving values`.
2. A comment in the emitted YAML: these are generous placeholders, not measured; watch
   for `Reducing Torch parallelism from N threads to 1` as the starvation signal.

## Considered and rejected: deriving from Hugging Face

[...]

Issue #843, opened 2026-08-24
bootstrap: a model spanning multiple pods (multinode/LWS) is not expressible from config.md


kalantar:
## Problem

`config.md` cannot describe a model that spans more than one pod. Bootstrap always
emits a single-pod-per-replica topology, and there is no input row an operator can
write to say otherwise.

Three gaps, all in `.claude/skills/sim2real-bootstrap/`:

1. **No input vocabulary.** `PARAMETER_ALIASES` (`generate_from_config.py:58-104`)
   has no key for pods-per-replica. Any row naming one (`Pods per replica`,
   `Nodes per model`) fails `canonicalize_parameter` and is dropped — the same
   silent-drop class as #824/#825, except `workers`, which is specifically exempted
   from the counted-noun check by the comment at `:354-356`.

2. **`multinode` is never emitted.** Grep for `multinode` across
   `.claude/skills/sim2real-bootstrap/` returns nothing. Generated baselines
   therefore always inherit `multinode: false`
   (`llm-d-benchmark/config/templates/jinja/13_ms-values.yaml.j2:77`), so the chart
   renders a `Deployment` and `parallelism.workers` is never read. Setting `workers`
   without also emitting `multinode` is inert.

3. **The parallelism block is gated on parallelism degrees only.** Both generators
   guard emission with `if tp > 1 or dp > 1`
   (`generate_from_config.py:775`, `:822`, `:859`, `:866`;
   `generate_scenarios.py:227`, `:264`). A genuine multinode topology with `tensor: 1`
   — which is what the one real in-tree example uses — would emit no `parallelism`
   block at all.

## Prerequisite: #831

#831 fixes `workers: tp` → `workers: 1`, which is correct under any resolution of the
question below and safe to land first. This issue is the follow-on: making a
`workers > 1` topology *expressible*. #831 should not be scoped up to include this.

## Blocking question: is `tensor` whole-model or per-pod?

This must be settled before any of the changes below, because GPU accounting depends
on it. The in-tree evidence is contradictory:

- `llm-d-benchmark/config/scenarios/guides/wide-ep-lws.yaml:346-350` — the only real
  LWS scenario — puts the fan-out in `dataLocal: 8` and leaves `tensor: 1`, with
  `workers: 2`.
- `llm-d-benchmark/config/scenarios/examples/spyre.yaml:169,292-295` runs
  `multinode.enabled: true` with `tensor: 4, dataLocal: 1, workers: 1` — multinode
[...]

PR #852, opened 2026-08-25
fix(bootstrap): emit the P/D and multi-GPU pod plumbing on #846's gates (closes #848)

### LINKED ISSUE #848: bootstrap: emit the P/D and multi-GPU pod plumbing on the same gates as #846, or kvTransfer:true crashloops
## Summary

#846 made bootstrap emit `vllmCommon.kvTransfer` when a prefill pool exists. But the
pod plumbing that makes NIXL actually initialise is still not emitted, so a bundle
bootstrapped today gets `kvTransfer.enabled: true` **without** it — which is the
crash loop (`Worker proc VllmWorker-N died unexpectedly` during "Initializing NIXL
wrapper") that #830 was chasing. Enabling KV transfer without the plumbing is worse
than not enabling it.

Everything below is derivable from rows `config.md` already declares, on two gates.

## Gate 1 — prefill pods > 0 (same gate as #846)

- `decode.initContainers` / `prefill.initContainers`: the `preprocess` container
  running `set_llmdbench_environment.py -e /shared-config/llmdbench_env.sh -i`
- `vllmCommon.volumes` / `volumeMounts`: the `shared-config` emptyDir at `/shared-config`
- `vllmCommon.preprocessScript`: the `libcuda.so.1` discovery prologue, then
  `. /shared-config/llmdbench_env.sh`
- `NIXL_LOG_LEVEL`
- `routing.connector: nixlv2`

These are one unit: the init container **computes** environment values and GPU-routing
state and writes them to `/shared-config/llmdbench_env.sh`; the prologue **sources**
them; the volume is the handoff. Splitting them across layers means a `defaults.disable`
entry can break the pair.

`routing.connector` is the sidecar half of the connector decision #846 already makes
engine-side. Emitting one without the other is a half-configuration of exactly #830's
shape, so both keys should come from one "which connector" value.

## Gate 2 — tensor_parallel_size > 1

- `vllmCommon.volumes` / `volumeMounts`: the `dshm` emptyDir (`medium: Memory`) at `/dev/shm`
- `NCCL_DEBUG`, `NVSHMEM_DEBUG`

NCCL and NVSHMEM only do anything with multi-GPU collectives. The K8s default 64 MB
`/dev/shm` is a latency-jitter and hang risk for a TP>1 pod.

## Evidence for the gates

Upstream `llm-d-benchmark/config/scenarios/guides/` (9 scenarios) carries
`shared-config` 9/9, `dshm` 9/9, the `preprocess` init container 8/9, and the
`libcuda` prologue 8/9 — so it is standard boilerplate, not bundle-specific.

The gates themselves rest on mechanism, not on that count: NCCL and NVSHMEM only
exercise shared memory and collectives when there is more than one GPU per pod, and the
init-container/prologue/shared-config unit exists to set up the network and GPU-routing
state that KV transfer needs. Both conditions are already declared in `config.md`
(`tensor_parallel_size`, prefill pod count), so bootstrap can read them directly.

Note the upstream counts above are weak evidence on their own — all nine guide scenarios
derive from a common template, so they share boilerplate whether or not each one needs
it. `/dev/shm` appears in scenarios with `tensor: 1`, where it does nothing.

## One spelling note

Upstream puts the libcuda prologue inside `vllm.customCommand`, which replaces the
**entire** launch command — every flag hand-written, so nothing tracks `config.md`
any more. Use `vllmCommon.preprocessScript` instead: same prologue, command
generation preserved. (Related: the defect in the companion issue about `config.md`
rows being written where a later layer overwrites them.)

## Not in scope

Cluster-specific pieces stay in the fragment layer, tracked on #840: pod capabilities
(need an SCC grant first), the NIC exclusion list (literal device names), and the RDMA
device reservation (does not work on this cluster).

kalantar:
Closes #848

## What was wrong

#846 made both generators emit `vllmCommon.kvTransfer.enabled: true` whenever a prefill
pool exists. None of the pod plumbing that lets NIXL actually initialise was emitted, so a
bundle bootstrapped after #846 got the flag without the setup — the crashloop
(`Worker proc VllmWorker-N died unexpectedly` during "Initializing NIXL wrapper") that #830
was chasing. Enabling KV transfer without the plumbing is worse than leaving it off.

The causal chain is a bit stronger than the issue states: `set_llmdbench_environment.py`
is what computes `UCX_NET_DEVICES`, `NCCL_IB_HCA`, `NCCL_SOCKET_IFNAME`,
`NCCL_IB_GID_INDEX` and the `NVSHMEM_*` set (`:538-604`) and writes them to the env file.
Without the init container there is no UCX/NIXL network-device configuration at all.

## What changed

New `pd_plumbing.py` holds the fragments and the two gates. Both generators import it.

**Gate 1 — a prefill pool exists** (the same predicate #846 uses for `kvTransfer`):
`preprocess` init container on both roles, `shared-config` emptyDir + mount,
`vllmCommon.preprocessScript` (libcuda prologue then `. /shared-config/llmdbench_env.sh`),
`NIXL_LOG_LEVEL`, `routing.connector`.

**Gate 2 — more than one GPU per pod:** `dshm` tmpfs at `/dev/shm` + mount, `NCCL_DEBUG`,
`NVSHMEM_DEBUG`.

Both gates are additive and write into shared keys (`vllmCommon.volumes` /
`volumeMounts`, per-role `extraEnvVars`), so they are rendered in one pass per key rather
than appended per gate — a second gate clobbering the first is the obvious bug in this
shape, and there is a test for it.

### Both generators, not just the one the issue names

The issue says "bootstrap". `generate_scenarios.py` (the `top3_selection.json` path) has
its own prefill block, its own hand-rolled emitter, and had the identical hole. #846's
commit message records hitting exactly this: *"fixing one path would have left half of
bootstrap emitting an inert pool."* Both are fixed, and
`test_both_generators_emit_identical_plumbing_text` asserts the emitted fragments match
character-for-character so they cannot drift.

### One connector decision, two spellings

The issue asks that `kvTransfer.connector` (engine) and `routing.connector` (sidecar) come
from one value. They now do — `pd_plumbing.KV_CONNECTOR_ENGINE` / `KV_CONNECTOR_SIDECAR`,
with the `"NixlConnector"` literals removed from both generators. Worth noting for the
reviewer: `routing.connector: nixlv2` is *already* the upstream default
(`defaults.yaml:694`), so emitting it is "state, don't inherit" hardening rather than part
of the crash fix.

## Reviewer attention

**1. Gate 2 measures `tensor × dataLocal`, not `tensor` and `data`.** *(Revised after
review — see the commit `b6433bd` and the note at the end of this section.)*

GPUs-per-pod is `tensor × dataLocal`, which is upstream's own arithmetic:
`13_ms-values.yaml.j2:269-271` resolves the accelerator count as
`accelerator.count if explicit, else tensor * dataLocal`, commented "each DP-local rank
needs its own GPU". Bootstrap never emits `accelerator.count` (only `acceleratorType`), so
for generated scenarios that product *is* the count.

`dataLocal`, not `data`: `data` is the deployment-wide DP degree; `dataLocal` is the ranks
resident in one pod. A `data: 8, dataLocal: 1, tensor: 1` scenario is a **single-GPU pod**
whose seven sibling ranks live in other pods and communicate over the network — no
intra-pod shared-memory collectives, so no `/dev/shm` pressure and nothing for
NCCL/NVSHMEM to do locally.

This remains a deliberate superset of the issue's literal `tensor_parallel_size > 1`: it
also fires for an intra-pod data-parallel pod (`dataLocal > 1, tensor == 1`), which is
multi-GPU by the same mechanism the issue argues from. The gap between the issue's wording
and this gate is now just that case.

**What changed and why it matters.** This PR originally gated on `tp > 1 or dp > 1`,
reasoned as "GPUs-per-pod is `tensor × dataLocal` and both generators set `dataLocal: dp`."
That states the formula correctly and then keys on the wrong variable. Both generators feed
`dataLocal` from the single `data_parallel_size` input, so `data == dataLocal` today and the
defect was **latent, not live** — all 18 `tp × dp × prefill` combinations emit
byte-identically under either predicate, verified by diff across both generators. It goes
live with #843's multinode/LWS per-pod split, precisely when a gate on the wrong quantity
would begin silently over-emitting: a 16Gi tmpfs charged against the pod memory limit
(#850) plus dead env vars, for a pod needing neither.

Worth flagging for future reviews: **none of the four review lenses caught this**, and two
explicitly validated the old gate — code-reviewer wrote "justified because `dataLocal == dp`
in both generators," which is true today and is exactly the reasoning that conceals the
latent case. It was caught by human review. The new
`test_gate2_measures_data_local_not_deployment_wide_data` pins the distinction, and a
call-site fault-injection (feeding a deployment-wide degree) breaks 6 tests including the
byte-identity guard.

**2. `NIXL_LOG_LEVEL` has the weakest justification of anything here.** The issue lists it
under Gate 1 with no rationale, and it appears in only 1 of 9 upstream guides. It is pure
diagnostics — the emitted comment says so and says it is safe to drop. By contrast
`NVSHMEM_DEBUG` turns out to be load-bearing, which the issue does not mention:
`set_llmdbench_environment.py:539-541` adds `NVSHMEM_HCA_LIST` to the generated env file
only when it is not `"none"`. That is now the cited justification rather than "debug logs".

**3. `dshm sizeLimit: 16Gi` interacts with #850.** A `medium: Memory` emptyDir is tmpfs and
charges against the pod's memory limit. #850 is about emitting memory limits; a limit at or
below 16Gi would make a full `/dev/shm` an OOM. No conflict today (no limits are emitted
yet), flagged in the emitted comment, and worth a note on #850.

**4. Two out-of-scope-adjacent fixes I made anyway,** both because my change touched them:
- `sim2real-specify/SKILL.md` cited `generate_from_config.py:300-322` for "the parser
  locates the table by its heading". My import shifted that file by +2, and the citation
  was pointing at the generic table scanner, which implements neither heading-based
  selection nor first-match-wins. Repointed at `find_vllm_table:443-450`, which does both.
- `generate_scenarios.README.md` documents the JSON path's emitted shape and was never
  updated when #846 landed — it covered `prefill.enabled` but not `kvTransfer`. Added both
  that and the new gates.

## Verification

- **Byte-identity** (the primary regression contract): regenerated an ungated scenario
  (no prefill pool, tp=dp=1) through both generators before and after — `diff` is empty.
  Existing single-GPU aggregated bundles are unaffected, since both gates are conditional.
- **39/39 acceptance criteria** checked against real emitted output from both generators,
  asserted independently of the unit tests (so a shared wrong assumption between test and
  implementation would still show up): every Gate 1 and Gate 2 item, gate independence in
  both directions, no `customCommand`, the out-of-scope #840 items staying out, and
  fragment-level parity between the two paths.
- **Red-green:** reverting the fix (neutralising both gate reads) turns 12 tests red;
  restoring it returns all 307 to green.
- **Fault injection**, the method #846's review used to find the untested branches: 14
  injected defects — wrong mount path, `medium: Disk`, a truncated init-container command,
  a drifted sidecar connector, a narrowed `tp`-only gate, a wrong env-file path, a pinned
  literal image, a renamed `NVSHMEM_HCA_LIST`, a dropped `routing` emit, a Gate-2-narrowed
  volume call, the `setdefault`→plain-assignment clobber the code comments warn about, and
  a dropped prefill init container — each fails at least one test. No silent survivors.
- **CI, fresh:** `ruff check pipeline/ .claude/skills/ --select F` clean;
  `2395 passed, 1 skipped, 2 xfailed`; coverage 97.68% against the 90% gate. (`--cov=pipeline`
  does not measure `.claude/skills/`, so this change cannot move that number either way.)

## Stale-reference sweep

Swept `**/*.md`, `docs/`, `.claude/skills/`, and `README*` for the changed basenames and for
every new emitted key. The change renames nothing and alters no existing key, so most hits
were accurate as-is. Updated: the two module lists in `sim2real-bootstrap/SKILL.md`, its
emitted-YAML example and disaggregation prose, `generate_scenarios.README.md`, and the
`sim2real-specify` citation above.

I also chased the blind spot the plan called out — a consumer whose *implicit schema*
changed without any path or symbol changing. The live one is `pipeline/lib/capacity.py:98-99`,
which excludes `initContainers` from GPU accounting "on the assumption that llm-d workloads
do not use GPU-requesting init containers." I just added an init container. It declares no
`resources`, so it requests no GPU and the assumption still holds — and
`init_container_lines`'s docstring now says a future edit must not add a GPU request there,
because that would silently under-count demand in the capacity probe.

The plan is committed at `docs/superpowers/plans/2026-08-25-issue-848-pd-pod-plumbing.md`,
including a "Gotchas" section for the traps I hit (the two emitters differ on whether they
`makedirs`; the two test files' `emit` helpers have different signatures; `entry()` takes
workload-extra first).

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -207,6 +207,10 @@ Generate llm-d-benchmark scenario YAML(s) for baseline arm(s).
   scenario YAMLs. Use only when a JSON input file exists.
 - `generate_scenarios.README.md` — documents field mappings, lookup tables,
   omission rules, and coverage gaps for the JSON-input path.
+- `pd_plumbing.py` — the P/D and multi-GPU pod fragments (init container,
+  `shared-config` / `dshm` volumes, `preprocessScript`, NIXL/NCCL/NVSHMEM env
+  vars, `routing.connector`) plus the two gates that decide when they apply.
+  Imported by BOTH generators so the emitted text cannot drift between paths.
 
 **Process:**
 
@@ -287,13 +291,52 @@ scenario:
       additionalFlags:
       - "--flag=value"                 # shared with decode, not per-role
 
-  # Also emitted with a prefill pool, and only then (issue #830). This is what
-  # makes the pool do anything; see the note below.
+  # Also emitted with a prefill pool, and only then (issues #830, #848). This is
+  # what makes the pool do anything; see the note below.
   vllmCommon:
     kvTransfer:
       enabled: true
       connector: NixlConnector
       role: kv_both
+    # Sources the env file the preprocess init container writes (#848).
+    preprocessScript: |
+      export LD_LIBRARY_PATH=...libcuda.so.1 discovery...
+      . /shared-config/llmdbench_env.sh
+    volumes:
+    - name: shared-config              # prefill pool   (Gate 1)
+      type: emptyDir
+      emptyDir: {}
+    - name: dshm                       # >1 GPU per pod (Gate 2)
+      type: emptyDir
+      emptyDir:
+        medium: Memory
+        sizeLimit: 16Gi
+    volumeMounts:
+    - name: shared-config
+      mountPath: /shared-config
+    - name: dshm
+      mountPath: /dev/shm
+
+  # Per-role, on BOTH decode: and prefill: — each role is a separate pod, and
+  # upstream has no vllmCommon form for either key.
+  decode:                              # ...and the same under prefill:
+    initContainers:                    # Gate 1
+    - name: preprocess
+      imageKey: benchmark
+      command: ["set_llmdbench_environment.py", "-e", "/shared-config/llmdbench_env.sh", "-i"]
+      volumeMounts:                    # explicit — init containers do not
+      - name: shared-config            # inherit vllmCommon.volumeMounts
+        mountPath: /shared-config
+    extraEnvVars:
+    - name: NIXL_LOG_LEVEL             # Gate 1
+      value: debug
+    - name: NCCL_DEBUG                 # Gate 2
+      value: "INFO"
+    - name: NVSHMEM_DEBUG              # Gate 2
+      value: "INFO"
+
+  routing:                             # Gate 1
+    connector: nixlv2
 ```
 
 **Disaggregation (issue #824).** To get a `prefill:` block, `config.md`'s vLLM
@@ -315,6 +358,38 @@ optional per-role accelerator overrides; without them both roles use the shared
   `role: kv_both` is deprecated for NixlConnector, which wants `kv_producer` on
   prefill and `kv_consumer` on decode; `vllmCommon` is shared by both roles so
   per-role values are not expressible today (tracked as #845).
+- **The plumbing that makes `kvTransfer` work is emitted on the same gate (issue
+  #848).** `kvTransfer.enabled: true` on its own is *worse* than leaving it off:
+  the worker dies during "Initializing NIXL wrapper" because nothing configured
+  the network and GPU-routing state. Two gates, both additive, both derived from
+  rows `config.md` already declares:
+  - **Gate 1 — a prefill pool exists** (the same gate as `kvTransfer`): the
+    `preprocess` init container on *both* roles, the `shared-config` emptyDir,
+    `vllmCommon.preprocessScript`, `NIXL_LOG_LEVEL`, and `routing.connector`.
+    The first three are one unit — the init container *computes* the values and
+    writes them to `/shared-config/llmdbench_env.sh`, `preprocessScript`
+    *sources* them, and the volume is the handoff. Any one missing makes the
+    other two inert, which is why they are emitted together rather than split
+    across the fragment layer where a `defaults.disable` entry could break the
+    pair. `routing.connector` is the sidecar half of the connector decision
+    `kvTransfer.connector` makes engine-side; both spellings come from
+    `pd_plumbing.KV_CONNECTOR_*` so they cannot drift.
+  - **Gate 2 — more than one GPU per pod** (`tensor_parallel_size > 1` *or*
+    `data_parallel_size > 1`; `dataLocal` equals `dp`, so either puts several
+    GPUs in one pod): the `dshm` tmpfs at `/dev/shm` and `NCCL_DEBUG` /
+    `NVSHMEM_DEBUG`. The K8s default 64 MB `/dev/shm` is a latency-jitter and
+    hang risk for multi-GPU collectives. `NVSHMEM_DEBUG` is load-bearing rather
+    than diagnostic: `set_llmdbench_environment.py` adds `NVSHMEM_HCA_LIST` to
+    the generated env file only when it is not `"none"`.
+
+  Neither gate fires for a single-GPU aggregated bundle, so existing bundles
+  regenerate byte-identically. Both generators emit the fragments from
+  `pd_plumbing.py`, so the two paths cannot drift.
+
+  Cluster-specific pieces are deliberately *not* emitted and stay in the
+  fragment layer (tracked on #840): pod capabilities, which need an SCC grant
+  first; the NIC exclusion list, which names literal devices; and the RDMA
+  device reservation.
 - `parallelism` and `vllm.additionalFlags` are shared across roles — `config.md`
   has no per-role form for them.
 - **One GPU type per role.** A GPU cell naming several types (`H100, A100`) emits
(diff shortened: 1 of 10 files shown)


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

The gating, fragment emission, and anti-drift guarantees hold under independent verification; nothing found rises to a blocking issue.

### Recommendations
- **Header comment on the `vllmCommon:` example says "emitted with a prefill pool, and only then", which is no longer true of the `volumes`/`volumeMounts` it now heads.** (`.claude/skills/sim2real-bootstrap/SKILL.md:294`) — Generating a tp=4 config.md with no prefill rows emits `vllmCommon.volumes`/`volumeMounts` carrying only `dshm` — Gate 2 (`tp > 1 or dp > 1` in pd_plumbing.needs_multigpu_plumbing) fires independently of a prefill pool, so the "only then" claim overreaches for part of the block. The per-entry `# >1 GPU per pod (Gate 2)` tags and the gate prose later in the same section already state it correctly, so this is a comment-scope trim, not a correctness problem.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

The gating, the shared-fragment module, and the byte-identity contract for non-P/D bundles all hold up under direct verification, with only a miscounted figure in one explanatory comment worth noting.

### Recommendations
- **The dshm comment's guide count is wrong: 7 of 9, not 5 of 7.** (`.claude/skills/sim2real-bootstrap/pd_plumbing.py:56`) — `grep -rn sizeLimit llm-d-benchmark/config/scenarios/guides/` at the commit the module docstring pins (76473d05) returns 9 matches in 9 files: 7 at 16Gi, one at 20Gi, one at 32Gi. All 9 guides set a limit, so "5 of the 7 … that set a limit" also wrongly implies two set none; the emitted 16Gi is still the majority value, so this is only the justification, not the behavior.

PR #856, opened 2026-08-25
feat(bootstrap): ship three P/D cluster fragments plus preserve-request-id disabled (closes #853)

### LINKED ISSUE #853: bootstrap: ship four fragments disabled by default — three new cluster-specific P/D ones plus the existing preserve-request-id
## Summary

Four fragments should ship **disabled** rather than enabled: three new cluster-specific
pieces from the pd-infocomm-2 P/D work, plus the existing `preserve-request-id`. None can
be a correct always-on default, but all are worth shipping so an operator has something to
enable rather than reinvent.

The config.md-derivable parts of the same plumbing are #848. Broader conditional emission
of existing fragments is #840 — see the note at the end on the overlap.

## The three new ones

**Pod capabilities** — `IPC_LOCK`, `SYS_RAWIO`, `NET_ADMIN`, `NET_RAW`,
`runAsUser/Group: 0` on decode and prefill.

Must default to off because the fragment cannot grant the permission it depends on. On
OpenShift the pod will not admit until someone runs
`oc adm policy add-scc-to-user privileged -z <sa> -n <ns>`, so enabled-by-default
produces unschedulable pods. Only `IPC_LOCK` has a real justification (mlock /
`ibv_reg_mr` for NIXL memory registration) — `SYS_RAWIO` is gated on `NCCL_ACS_DISABLE`,
which nothing in the pipeline sets; `NET_ADMIN` matters only on multi-NIC same-subnet
nodes; `NET_RAW` only if the fping smoketest runs. Worth trimming to `IPC_LOCK` after
testing, which also removes a real deployment obstacle.

**NIC exclusion list** — `NCCL_EXCLUDE_IB_HCA`, plus `UCX_MAX_RMA_RAILS`.

Must default to off because the value is a literal device list for one cluster's cards.
A shipped default naming another cluster's devices would exclude the wrong ones, which is
worse than excluding none.

**RDMA device reservation** — `networkResource: "auto"`, `networkNr: "1"`.

Must default to off because it does not work on the cluster it was written for. RoCE
derives its GIDs from the associated netdev's IP; the pod's network namespace holds only
`eth0` (OVN overlay) and `lo`, so `mlx5_*` GID tables read all zeros inside the pod, and
adding `rc` to `ucxTls` fails NIXL backend creation outright
(`uct_iface_open(rc_verbs/mlx5_0:1) failed: Address not valid`). It reserves an
allocatable `rdma/roce_gdr` device to advertise a capability the pod cannot use. Using
RoCE here would need hostNetwork or an RDMA-capable CNI, neither of which is available.

## The existing one: `preserve-request-id`

Ships enabled today, and cannot work under this bundle's topology. It is an Istio
`EnvoyFilter` selecting on a hardcoded gateway name:

```yaml
workloadSelector:
  labels:
    gateway.networking.k8s.io/gateway-name: infra-llmdbench-inference-gateway
```

Under `gateway.className: epponly` with `externallyManaged: true` no Kubernetes Gateway is
deployed at all — the EPP's Envoy sidecar is the whole data plane — so the selector matches
nothing. It also requires Istio CRDs to be present.

It is additionally the **only** fragment using `extraObjects` (free-form Kubernetes
manifests), which makes it the only one that can fail at *apply* time rather than being
silently inert.

`pd-infocomm-2` already carries `defaults.disable: [preserve-request-id]`, added by hand
alongside `epponly.yaml`. Shipping it disabled makes bootstrap do what the operator
currently has to remember — and that hand edit is lost on every re-bootstrap, since
bootstrap regenerates `transfer.yaml`.

## How "ship disabled" works

No new mechanism is needed. Bootstrap writes both the fragment file and `transfer.yaml`,
so it can emit the fragment and put its stem in `defaults.disable`. Enabling it is
deleting one line.

Worth noting the alternative was considered and is not needed: the `defaults` block
accepts only `disable` (`pipeline/lib/manifest.py:142`). Adding an `enable` list would
give three-state resolution (per-fragment default, plus two lists) for something that is
currently one list.

## Source

Current hand-written versions are in `pd-infocomm-2/baselines/baseline.yaml`, inline
rather than as fragments, with the reasoning in comments. They can be lifted from there.

## Overlap with #840

#840 lists `preserve-request-id` among the fragments whose emission should be conditional.
Shipping it disabled here is the simpler answer and does not need the conditional
machinery, so whichever lands first should drop it from the other. If #840 does implement
conditional emission, it supersedes this half of the issue — the three new fragments do
not, since their disabled default is a property of the fragments themselves rather than of
the bundle's topology.

kalantar:
Closes #853

Migrates three sections of `pd-infocomm-2/baselines/baseline.yaml` into
`templates/defaults/` fragments, and adds them plus `preserve-request-id` to the
`defaults.disable` list that Task 5 writes into `transfer.yaml` — previously `[]`.

| Fragment | Source | Content |
|---|---|---|
| `pod-capabilities` | `baseline.yaml:195-204` / `265-274` (byte-identical across roles) | `extraContainerConfig.securityContext`: `capabilities.add: [IPC_LOCK, SYS_RAWIO, NET_ADMIN, NET_RAW]`, `runAsGroup/User: 0` |
| `nic-exclusion` | `baseline.yaml:176-186` / `253-262` | `NCCL_EXCLUDE_IB_HCA` + `UCX_MAX_RMA_RAILS` |
| `rdma-device-reservation` | `baseline.yaml:82-83` | `vllmCommon.networkResource: "auto"`, `networkNr: "1"` |

The source's `extraEnvVars` block holds five variables; only the two NIC keys are
here. The other three (`NVSHMEM_DEBUG`, `NCCL_DEBUG`, `NIXL_LOG_LEVEL`) come from
`config.md` via #848 and would be duplicated if the fragment carried them.

No new mechanism: bootstrap already writes both the fragment file and
`transfer.yaml`. The `defaults` block accepts only `disable`
(`pipeline/lib/manifest.py:139-160`), and nothing needed to change there.

## Two forced changes to existing tests

**`test_no_two_fragments_set_the_same_leaf_key` had a false premise.** It flagged
`prefill.extraEnvVars`, shared by `nic-exclusion` and `vllm-keepalive`, on the
grounds that "the later stem wins and the loser is inert." That is true for scalar
lists and plain scalars, and false for keyed dict-lists: `_merge_lists` Tier 2b
merges `extraEnvVars` **by `name`**, so all three variables reach the pod. The
source baseline's own comment at `:245-251` states this independently, and the
end-to-end check below confirms it. So the guard would have blocked a combination
that works.

Rewritten as `test_no_fragment_contribution_is_lost_to_another_fragment`: it merges
each pair of fragments and asserts each side's contributions survive, testing the
property the guard exists for rather than a proxy that over-rejects. **Still has
teeth** — a scalar-list clash and a plain-scalar clash both fail it, while the
keyed-list union passes.

**`capabilities.add` is a bare scalar list**, so it needs `_ALLOWED_SCALAR_LISTS`
entries. Per that allowlist's own doctrine each entry is "a standing fragility, not
an endorsement", so the fragility is documented in both the allowlist and the
fragment header.

## New drift guards

The BLIS-mode list lives in SKILL.md prose, which no test covered — the exact shape
of #550, where SKILL.md documented an `llm-d-rbac.yaml` fragment that
`templates/defaults/` never shipped. Each disabled fragment now carries a
`DISABLED BY DEFAULT` header marker, and four guards pin the two representations
together: the sets must match, the list must be sorted and non-empty, every stem
must name a real file (a stem that matches nothing disables nothing), and every
marked fragment must explain *why* under `WHY OFF BY DEFAULT` — a disabled default
is only useful if an operator can judge whether their cluster is the exception.

## Verification

- **End-to-end through the real loader**, not just the shape tests: with the four
  stems disabled, none of their keys appear in `load_defaults_overlay`'s output, and
  the eight enabled fragments still apply; with `disable=[]` all four land. 10/10.
- **Fault injection, 5/5 caught**: dropping a stem, adding one that names no file,
  unsorting the list, removing a fragment's marker, removing its rationale.
- **Fault injection on the rewritten guard, 3/3 correct**: scalar-list clash fails,
  plain-scalar clash fails, keyed-list union passes.
- **CI**: `ruff --select F` clean; 2419 passed, 9 skipped, 2 xfailed; coverage
  97.68%. (The 9 skips are `test_disabled_fragment_says_why_in_its_header`
  skipping the 8 enabled fragments, plus one pre-existing skip.)

## Reviewer attention

**The `mlxl5_7` device name is carried verbatim and is almost certainly a typo** for
`mlx5_7` — meaning that device is *not* actually excluded. Kept because it is what
ran, flagged in the fragment header, and the entire list is cluster-specific and
must be replaced before enabling. Say so if you would rather ship the corrected
spelling.

**`pod-capabilities` ships a set its own header argues is too broad.** Only
`IPC_LOCK` has a P/D justification (mlock / `ibv_reg_mr` for NIXL memory
registration). `SYS_RAWIO` is gated on `NCCL_ACS_DISABLE`, which nothing in the
pipeline sets; `NET_ADMIN` needs multi-NIC same-subnet nodes; `NET_RAW` only if the
fping smoketest runs. Carried whole because that is the configuration observed
working; trimming to `IPC_LOCK` would also shrink the SCC grant it depends on and
remove a real deployment obstacle. Upstream's shipped guides carry `IPC_LOCK` and
`SYS_RAWIO` in 8 of 9 scenarios and `NET_ADMIN`/`NET_RAW` in 4 of 9.

**Scope kept narrow deliberately.** #840 owns making existing fragments'
*emission* conditional, and also owns the broader BLIS/BYO difference (`--byo`
already disables every stem via `copy_framework_defaults`; BLIS disabled none).
This PR does not touch that difference — it only stops these four from being on by
default. SKILL.md's new section says so explicitly so the boundary is not
re-litigated later.

**Out of scope but worth a separate issue:** the working baseline sets
`NVSHMEM_DEBUG: "none"` and `NCCL_DEBUG: "WARN"`, while #848 emits `INFO` for both.
Since #848's own comment documents `NVSHMEM_DEBUG != "none"` as load-bearing (it
gates `NVSHMEM_HCA_LIST` into the generated env file), that default contradicts the
configuration actually observed working. Not folded in here.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -459,11 +459,27 @@ fragments in place to tweak them for the experiment, or list a fragment
 stem under `defaults.disable` in `transfer.yaml` to opt out entirely.
 
 **Not every fragment is universally correct.** `epponly` is a topology decision,
-`model-pvc-size` is sized for a 70B-class model, `tokenizer-sidecar` matters only
-when the arms declare a `token-producer` plugin, and `preserve-request-id` matches
-nothing under an epponly / externally-managed gateway. Each fragment's header
+`model-pvc-size` is sized for a 70B-class model, and `tokenizer-sidecar` matters
+only when the arms declare a `token-producer` plugin. Each fragment's header
 comment states its own applicability and when to disable it. Making emission
 conditional rather than leaving that to the operator is tracked as #840.
+
+**Fragments that ship disabled (issue #853).** Four are copied but listed in
+`defaults.disable` by Task 5, because none can be a correct always-on default.
+Shipping them anyway means an operator has something to enable rather than
+reinvent; enabling one is deleting its line from `defaults.disable`.
+
+| Fragment | Why it cannot default to on |
+|---|---|
+| `pod-capabilities` | Cannot grant the permission it depends on. On OpenShift the pod will not admit until someone runs `oc adm policy add-scc-to-user privileged` out of band, so enabled-by-default yields unschedulable pods. |
+| `nic-exclusion` | The value is a literal device list for one cluster's cards. A default naming another cluster's devices would exclude the wrong ones — worse than excluding none. |
+| `rdma-device-reservation` | Does not work on the cluster it was written for: the pod netns has no RDMA-capable device, so it reserves an `rdma/roce_gdr` to advertise a capability the pod cannot use. |
+| `preserve-request-id` | An Istio `EnvoyFilter` selecting on a hardcoded gateway name. Under `epponly` + `externallyManaged` no Gateway is deployed, so it matches nothing — and it is the only fragment using `extraObjects`, so it is the only one that can fail at *apply* time rather than being silently inert. |
+
+Note the two bootstrap modes reach the same place by different routes: `--byo`
+already lists **every** stem in `defaults.disable` (`byo.py`'s
+`copy_framework_defaults`), so these four are disabled there without a special
+case. Reconciling that broader BLIS/BYO difference is #840's, not this list's.
 
 ---
 
@@ -548,13 +564,24 @@ context:
   files: <list of context files>
 
 defaults:
-  disable: []
+  # These four ship DISABLED. Each is copied into baselines/defaults/ so an
+  # operator has something to enable rather than reinvent, but none can be a
+  # correct always-on default — see "Fragments that ship disabled" below.
+  # Enabling one is deleting its line.
+  disable:
+    - nic-exclusion
+    - pod-capabilities
+    - preserve-request-id
+    - rdma-device-reservation
   # Available fragments (filename stems in baselines/defaults/):
   #   - epp-verbosity
   #   - epponly
   #   - externally-managed-gateway
   #   - model-pvc-size
-  #   - preserve-request-id
+  #   - nic-exclusion               (ships disabled)
+  #   - pod-capabilities            (ships disabled)
+  #   - preserve-request-id         (ships disabled)
+  #   - rdma-device-reservation     (ships disabled)
   #   - routing-proxy-resources
   #   - tokenizer-sidecar
   #   - vllm-keepalive
@@ -626,7 +653,10 @@ Exit code 0 and all fields printed = success.
       epponly.yaml
       externally-managed-gateway.yaml
       model-pvc-size.yaml
-      preserve-request-id.yaml
+      nic-exclusion.yaml              <- in defaults.disable
+      pod-capabilities.yaml           <- in defaults.disable
+      preserve-request-id.yaml        <- in defaults.disable
+      rdma-device-reservation.yaml    <- in defaults.disable
       routing-proxy-resources.yaml
       tokenizer-sidecar.yaml
       vllm-keepalive.yaml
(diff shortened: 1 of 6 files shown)


kalantar:

Verified clean — the fragment values match their source verbatim, both bootstrap modes land the same disable set, and the rewritten merge test is a genuine strengthening rather than a loosening.

### Recommendations
- **Only one of the four places SKILL.md records "ships disabled" is cross-checked against the fragment headers.** (`.claude/skills/sim2real-bootstrap/SKILL.md:577`) — test_skill_md_disable_list_matches_the_marked_fragments only parses the `disable:` list block; the `(ships disabled)` annotations in the Available-fragments comment (SKILL.md:577-580) and the `<- in defaults.disable` markers in the Final File Tree (SKILL.md:653-656) plus the table rows are unguarded, so a future edit to one copy alone drifts silently.

PR #857, opened 2026-08-25
feat(bootstrap): emit decode/prefill CPU+memory, config.md-stated or generous defaults (closes #850)

### LINKED ISSUE #850: bootstrap: emit CPU/memory for decode and prefill — generous limits, modest explicit requests, warn that they are unmeasured

kalantar:
Closes #850

Built to the operator's restatement of the issue rather than its body: *use `config.md`'s
cpu/memory req/limits when stated; otherwise generous defaults; decode needs more than
prefill; `pd-infocomm-2/baselines/baseline.yaml` as a starting point; start fixed and
learn.*

## What changed

Bootstrap emitted **no** `resources` at all, so every bundle inherited the framework
default `limits: {memory: 40Gi, cpu: "4"}` with no `requests` (`defaults.yaml:848-851`
decode, `:994-997` prefill) — four CPU for a pod that may hold four GPUs and shares its
cgroup with the routing sidecar. It now emits per-role resources in both generators.

**Rows `config.md` may state**, resolved *per-role → shared → role default*, the same
precedence as `Decode GPU` over `GPU`:

| Shared | Per-role override |
|---|---|
| `cpu limit` | `decode cpu limit`, `prefill cpu limit` |
| `memory limit` | `decode memory limit`, `prefill memory limit` |
| `cpu request` | `decode cpu request`, `prefill cpu request` |
| `memory request` | `decode memory request`, `prefill memory request` |

Values pass through as **strings**, never parsed — `32`, `500m`, `1.5`, `128Gi`, `1536Mi`
are all valid Kubernetes quantities and re-serializing risks changing them.

**Defaults when unstated**, fixed rather than derived:

| Role | limits | requests |
|---|---|---|
| decode | `128Gi` / `32` | `64Gi` / `16` |
| prefill | `16Gi` / `8` | `8Gi` / `4` |

Limits are llm-d's `pd-disaggregation` guide verbatim (`:410-416` decode, `:321-327`
prefill), which `pd-infocomm-2` also carries. **decode > prefill is upstream's own
sizing**, not just this cluster's — worth noting since it independently corroborates the
constraint.

Requests are half those limits and **emitted explicitly**. That is the point of #850's
requests/limits section: omitting `requests` is not "no reservation" — Kubernetes copies
the limit into the request when the request is absent, so limits-only would reserve the
whole generous figure on every replica, and a 2-replica decode pool would reserve 256Gi
before prefill is scheduled at all.

Both warnings (stderr at generation, and a comment in the emitted YAML) fire only when at
least one quantity fell back to a default. State all four rows for a role and both go
quiet, because then the numbers are the operator's and calling them unmeasured would be
wrong.

## Reviewer attention

**Every generated scenario changes.** Unlike #848 and #853 there is no byte-identity
escape — this adds a `resources` block unconditionally, by design. A bundle re-bootstrapped
after this gains one. That is the fix, but it will surprise anyone expecting regeneration
to be inert.

**The golden fixture was regenerated, and here is the evidence it was safe.**
`test_single_pool_config_regenerates_byte_identically` compares bytes against a captured
golden. I diffed before replacing it: **22 lines added, 0 removed**, every added line part
of the `resources` block and its comment. Nothing else drifted. I also added
`test_single_pool_golden_carries_pod_resources` as the opposite-direction guard, so a
future version that stops emitting `resources` cannot be blessed by regenerating the
golden to match.

**Two pre-existing tests asserted `"WARNING" not in stderr`** while being specifically
about the multi-type GPU-cell warning (`test_single_gpu_cell_does_not_warn` and its JSON
twin). Any new unrelated warning made them fail spuriously. Narrowed to `"GPU types"`,
which is what they actually test. Flagging because narrowing an existing assertion
deserves a second opinion.

**The capacity probe does not model this.** `pipeline/lib/capacity.py:178-181` sums only
`requests[gpu_resource_type]`, so it plans GPUs and is blind to CPU/memory. After this
change a stack requests ~36 CPU and ~136Gi (2 decode + 1 prefill) that the probe does not
account for, so a scenario can pass the GPU check and still be unschedulable on CPU or
memory. This is the strongest argument for requests being half the limits rather than equal
to them. Left alone deliberately — `capacity.py` is in `pipeline/`, outside the bootstrap
scope set for this issue. Worth a follow-up.

**Fixed, not derived, is a deliberate choice.** `tensor × dataLocal` scaling was considered
and rejected: it reproduces one cluster's decode block at TP=4 while inventing every other
cell of the table. A wrong fixed number is visible and correctable; a wrong formula looks
principled. Recorded in the module docstring so the next person sees the reasoning rather
than re-deriving it.

## Verification

- **Fault injection, 9 distinct defects, all caught**: request == limit; prefill limit
  exceeding decode; a changed default figure; the whole `requests` block dropped; unquoted
  cpu; the starvation signal stripped from the emitted comment; the same stripped from the
  stderr text; `used_any_default` hardwired False; a dropped prefill emit. (Two apparent
  gaps on first pass turned out to be my injections hitting the docstring and a
  neighbouring line rather than the asserted strings — re-injected precisely, both caught.)
- **Cross-generator parity**: `test_both_generators_emit_identical_resources_text` asserts
  both paths emit character-identical `resources`; verified by injecting a divergence into
  the JSON path, an ignored per-role key, and an always-warn — each fails a test.
- **CI**: `ruff --select F` clean; 2457 passed, 9 skipped, 2 xfailed; coverage 97.68%.

## Sweep

Grepped `**/*.md` for `resources` — every hit outside the submodules is about *cluster*
resources (Tekton, RBAC, PVCs), unrelated to pod cpu/memory, so nothing was stale. Chased
the implicit-schema blind spot the previous PRs flagged: the only consumer that reads
`resources` is `capacity.py`, covered above. Added `pod_resources.py` to both SKILL.md
inventories (the bundled-scripts list and the file-reference table).

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -416,7 +416,17 @@ optional per-role accelerator overrides; without them both roles use the shared
 
   Values are passed through verbatim as strings — `32`, `500m`, `1.5`, `128Gi`,
   `1536Mi` are all valid Kubernetes quantities and re-serializing risks changing
-  them. Stating only some is fine; the rest take defaults.
+  them. Stating only some is fine; the rest take defaults. A per-role row left
+  **blank** counts as unset and falls through to the shared row, not to the
+  default — writing an empty row is not an instruction to ignore the shared row
+  you did fill in.
+
+  The JSON-input path (`generate_scenarios.py`) takes the same values as
+  `vllm_args` keys — `cpu_limit`, `decode_memory_request`, and so on — with the
+  same precedence, and cites `vllm_args` rather than `config.md` in its warnings
+  and source comments, since it is selected only when no `config.md` exists. Both
+  paths resolve through `pod_resources.resolve_resources`, so the precedence
+  cannot differ between them.
 
   **Defaults when a row is absent** — decode is sized above prefill, which is both
   upstream's own sizing and the observed-working configuration:
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -398,35 +398,34 @@ optional per-role accelerator overrides; without them both roles use the shared
 - `parallelism` and `vllm.additionalFlags` are shared across roles — `config.md`
   has no per-role form for them.
 - **CPU/memory is always emitted, per role (issue #850).** Bootstrap used to emit
-  none, so bundles inherited the framework default of `limits: {memory: 40Gi,
-  cpu: "4"}` with no `requests` — four CPU for a pod that may hold four GPUs and
-  shares its cgroup with the routing sidecar. That starves vLLM, and the signal
-  (`Reducing Torch parallelism from N threads to 1`) surfaces as ITL noise rather
-  than a failure, so it corrupts the metric the arms are compared on.
+  none, so bundles inherited the framework default — `limits` **and** `requests`
+  both `{memory: 40Gi, cpu: "4"}` for each role (`defaults.yaml:848-858` decode,
+  `:993-1000` prefill). Four CPU for a pod that may hold four GPUs and shares its
+  cgroup with the routing sidecar starves vLLM, and the signal (`Reducing Torch
+  parallelism from N threads to 1`) surfaces as ITL noise rather than a failure, so
+  it corrupts the metric the arms are compared on.
 
-  **Rows `config.md` may state**, each optional, resolved *per-role row → shared
-  row → role default* (the same precedence as `Decode GPU` over `GPU`):
+  **Rows `config.md` may state**, each optional, applied to **both** roles:
 
-  | Shared row | Per-role override |
+  | Row | Also accepted as |
   |---|---|
-  | `cpu limit` | `decode cpu limit`, `prefill cpu limit` |
-  | `memory limit` | `decode memory limit`, `prefill memory limit` |
-  | `cpu request` | `decode cpu request`, `prefill cpu request` |
-  | `memory request` | `decode memory request`, `prefill memory request` |
+  | `cpu limit` | `cpu_limit`, `cpu limits` |
+  | `memory limit` | `memory_limit`, `memory limits` |
+  | `cpu request` | `cpu_request`, `cpu requests` |
+  | `memory request` | `memory_request`, `memory requests` |
 
-  Values are passed through verbatim as strings — `32`, `500m`, `1.5`, `128Gi`,
-  `1536Mi` are all valid Kubernetes quantities and re-serializing risks changing
-  them. Stating only some is fine; the rest take defaults. A per-role row left
-  **blank** counts as unset and falls through to the shared row, not to the
-  default — writing an empty row is not an instruction to ignore the shared row
-  you did fill in.
+  Stating only some is fine; the rest take their per-role default. There are
+  deliberately **no per-role rows** — an operator who needs decode and prefill to
+  differ edits `baselines/baseline.yaml` after bootstrap, as they do for everything
+  else the generator does not model. The JSON-input path takes the same four names
+  as `vllm_args` keys.
 
-  The JSON-input path (`generate_scenarios.py`) takes the same values as
-  `vllm_args` keys — `cpu_limit`, `decode_memory_request`, and so on — with the
-  same precedence, and cites `vllm_args` rather than `config.md` in its warnings
-  and source comments, since it is selected only when no `config.md` exists. Both
-  paths resolve through `pod_resources.resolve_resources`, so the precedence
-  cannot differ between them.
+  Values are emitted verbatim as **quoted** strings — `32`, `500m`, `1.5`, `128Gi`,
+  `1536Mi` are all valid Kubernetes quantities, and an unquoted one is re-typed by
+  any YAML reader (`128` to an int, `yes` to a bool). A value that is **not** a
+  valid quantity — `TBD`, `16 Gi`, `32 cores`, or a bare `-` placeholder — is a hard
+  error at generation, because the cluster would otherwise reject the pod at
+  admission far from the row that caused it.
 
   **Defaults when a row is absent** — decode is sized above prefill, which is both
   upstream's own sizing and the observed-working configuration:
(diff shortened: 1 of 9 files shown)


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 2 important · 2 recommendations** across 4 lenses.

The feature is well-factored and thoroughly tested, but the config.md ingestion path silently mangles unit-bearing quantities and the defaults warning misreports numbers the operator actually set.

### Important
- **Both warnings report the full default table even when the operator overrode some quantities, so they contradict what was actually emitted.** (`.claude/skills/sim2real-bootstrap/pod_resources.py:109`) — starvation_warning() takes only `role` and formats DEFAULTS[role] unconditionally, but it is called whenever ANY of the four quantities defaulted. Reproduced: with only `decode cpu limit: 64` stated, the emitted block is cpu_limit '64' while stderr reads "limits 32 CPU / 128Gi" — and the YAML comment resource_lines() emits in the same case claims "Limits are the llm-d pd-disaggregation guide's observed-working figures", which is now false for cpu.
- **Stated CPU/memory quantities containing a space (e.g. `16 Gi`) silently lose their unit suffix.** (`.claude/skills/sim2real-bootstrap/generate_from_config.py:590`) — The 12 new canonical resource fields are not in the string-passthrough set at line 582 (`model`/`hardware`/`dtype`), so they reach `parse_numeric`, which takes only the first whitespace token. Reproduced: a row `| decode memory limit | 16 Gi |` resolves to `memory: 16` in the emitted scenario (16 bytes), and `1.50` becomes `1.5` — with no warning, since the field counts as stated. This contradicts the explicit contract in both `pod_resources.py`'s docstring and SKILL.md that these values pass through verbatim as strings; the only quantity form covered by a test (`500m`, `1536Mi`) happens to fail `int()` and so accidentally takes the string branch.

### Recommendations
- **Move `assert "labelValues" not in golden` back into `test_single_pool_golden_has_no_prefill_artifacts`.** (`.claude/skills/sim2real-bootstrap/tests/test_generate_from_config_prefill.py:479`) — The new test function was spliced between the two assertions of `test_single_pool_golden_has_no_prefill_artifacts`, so `assert "labelValues" not in golden` now trails a test whose docstring is entirely about pod resources. It still runs, but neither test matches its own docstring.
- **Consider a request-exceeds-limit sanity check at generation time.** (`.claude/skills/sim2real-bootstrap/pod_resources.py:92`) — `resolve_resources` resolves each quantity independently, so stating `decode cpu request: 40` without a matching limit emits requests.cpu '40' against the default limits.cpu '32' — a spec Kubernetes rejects at admission, far from the config.md row that caused it. The generator has all four numbers in hand and could fail loud here.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 2 important · 1 recommendation** across 4 lenses.

The shared module is sound and well tested, but its wording and the JSON generator's field lookup both assume the config.md path, so the JSON-input path emits wrong guidance and resolves shared rows differently.

### Important
- **Warning text and per-value source comments name config.md on the JSON-input path, where no config.md exists.** (`.claude/skills/sim2real-bootstrap/pod_resources.py:70`) — pod_resources.py:70 hardcodes `_STATED_SOURCE = "config.md row"` and :126/:132 tell the operator to "Add '<role> <cpu|memory> <limit|request>' rows to config.md to override", but SKILL.md:220-221 selects generate_scenarios.py only when config.md is absent (input is top3_selection.json), and every other source comment that generator emits uses the "from vllm_args.X" convention. Running build_scenario through generate_scenarios.py prints exactly that config.md remediation on stderr, so the always-taken default case gives advice pointing at a file that is not the input; no capsys test asserts the JSON path's stderr text, which is why it slipped.
- **JSON path drops the shared row when a per-role key is present but null or empty, unlike the config.md path.** (`.claude/skills/sim2real-bootstrap/generate_scenarios.py:334`) — generate_scenarios.py:334 uses `vllm_args.get(f"{role_name}_{key}", vllm_args.get(key))`, which returns None for a key present with JSON `null` (or "" for an empty string) and never consults the shared key, while generate_from_config.py:949-951 uses `fields.get(role_key) or fields.get(shared_key)` and falls through correctly. Verified by running build_scenario with `{"cpu_limit": "64", "decode_cpu_limit": null}`: decode resolves to the built-in default 32 and the warning reports 4 of 4 quantities defaulted, silently discarding the stated shared 64 — the divergence the comment at :328 claims cannot happen, and `test_both_generators_emit_identical_resources_text` only exercises the nothing-stated case.

### Recommendations
- **Test docstring cites 500m but the body exercises 2000m.** (`.claude/skills/sim2real-bootstrap/tests/test_pod_resources.py:222`) — test_pod_resources.py:222-226 says "Comparing 500m against 1 needs a real quantity parser" while asserting on `cpu_request="2000m"` against `cpu_limit="1"`; 500m would not exceed a limit of 1 even under a correct parser, so the docstring's example inverts the point the test makes.

Issue #862, opened 2026-08-26
specify/translate: treat "no extension point fits" as a core-modification path, not a HALT


kalantar:
## Ask

`/sim2real-specify` Phase 2 ends with `HALT if no extension point can express the
required shape` (`.claude/skills/sim2real-specify/SKILL.md:118`). Treat that condition as
the entry point to a **core-modification plan** rather than a dead end, and state in the
`/sim2real-translate` prompts that modifying component files is a permitted strategy whose
cost is recorded.

## Why

The pipeline already supports invasive changes end to end. Only the guidance withholds
them.

- `files_modified` is a required field in the writer's output schema
  (`sim2real-translate/prompts/agent-writer.md:205`)
- `sim2real-translate/scripts/copy_generated.py:57` *derives* `files_modified` from a
  `git diff`, so whatever the writer changed is detected automatically
- `pipeline/lib/source_toggle.py:restore_baseline` reverts those files via `git checkout`
- both paths are covered (`.claude/skills/sim2real-translate/tests/test_copy_generated.py`,
  `pipeline/tests/test_source_toggle.py`)

No guidance in bootstrap, specify or translate permits or forbids modifying core component
code. The only language of that shape is `agent-expert.md:95` ("you do not modify any
files"), which is the expert agent's read-only tool scope, not a policy.

Phase 2's vocabulary is extension-point-shaped throughout — "read the target checkout and
find an extension point that can express it" — so an algorithm requiring invasive surgery
has no representable outcome and halts instead. A halt records nothing, while
`files_modified` plus the translation hash already provide the ledger for a change that is
carried rather than avoided.

`llm-d-router` being plugin-based makes an extension point the right default. This is about
the fallback, not the default.

## Open question

An invasive change means carrying a patch that an upstream rebase can break. How should
that cost be bounded and recorded — a declared-degradation-style entry, a cap on how much
core surface one translation may touch, or the `files_modified` list plus a stated reason?

## Scope

Guidance changes to `.claude/skills/sim2real-specify/SKILL.md` Phase 2 and the
`/sim2real-translate` prompts. No pipeline behaviour change — the plumbing exists.

Surfaced while investigating #859; independent of both of that issue's constraints.

PR #863, opened 2026-08-26
docs(bootstrap,specify): make context.text state its confidence, not just its claim (closes #859)

kalantar:
Closes #859

## What changed

Task 5 specified `context.text` as `<derived summary>`. That field is substituted into
`/sim2real-translate`'s writer and reviewer prompts as `{CONTEXT_TEXT}` and treated as
authoritative for the science, and **nothing downstream checks it** — no linter, no build
gate, no pipeline validation. So the two defects #859 documents were both within spec as
written, and both were caught by hand.

Replaced with the issue's two accuracy constraints, in bootstrap Task 5 step 9 and the
`context:` output-schema block, plus specify Phase 2.

**Constraint 1 — citations must bound their construct.** Open each cited range and confirm
its first and last line are the construct's own. `lint_citations.py` cannot catch this and
says so in its own docstring: it verifies a path resolves in exactly one pinned tree and
that line numbers fall inside the file, and "cannot detect a correct citation attached to a
wrong transcription."

**Constraint 2 — separate what is ruled out from what is proposed.** Eliminations are
settled conclusions; the placement is a **declared unknown** until its method sets are
checked. Registration count still required. The text must also say what still holds if the
candidate fails.

## Reviewer attention

**1. Constraint 2 is deliberately NOT what the issue body proposed.** The body asks for the
registration count and every interface per type — more detail, stated more confidently. The
discussion on the issue (comment `#issuecomment-5431110453`) argued that aims the wrong way,
and this PR follows the comment. The defect was that the placement was *already* asserted
flatly — `"The extension point that can express it is a custom Picker ... plus a custom
ProfileHandler"` — while being unbuildable, and because the writer treats this text as
authoritative it had no licence to search elsewhere when it failed to compile. A more
detailed wrong answer is harder to escape, not easier. The count requirement is kept; the
confidence marking is what's added.

**2. No new marker convention.** This reuses what already exists rather than inventing
`PARTIAL:` as the body suggested. specify already applies exactly this rule to target-API
adapters — *"declared unknowns rather than guesses, so a wrong guess fails loudly rather
than mis-scoring silently"* (`sim2real-specify/SKILL.md:166-167`) — and its spec-layer
contract explicitly says it "invents no in-source marker conventions" (`:170`). Real bundles
already use `PLACEMENT` and `DECLARED DEGRADATION (Dn)` as prose labels in this very field.
So the fix extends an existing principle to a field it didn't cover.

**3. Scope grew to specify, deliberately, because that's where the claim is born.** Traced
by commit in the `pd-infocomm-2` bundle: `5e4cfca` (specify's output) wrote the placement
into `README.md:71-72`; `449e911` (bootstrap) copied it into `transfer.yaml` two weeks
later, near-verbatim. Phase 2's gate was expressibility-only — *"find an extension point
that can express it"* — and **expressible is not constructible**: the three citations
backing that placement all establish that data can *flow* (unfiltered profile sees both
pools, `TargetEndpoints` is a slice, `ProcessResults` splits the pair), and none asks
whether one Go type can carry both interfaces. Two interfaces declaring the same method
name with different signatures cannot coexist on one type, which is what the compiler
eventually reported. Since specify is *optional*, the constraint also has to hold in
bootstrap independently — hence both files, not just the upstream one.

**4. specify's `HALT` is untouched** — that's #862's scope. Both skills now cross-reference
#862 for the case where no extension point fits at all, since "look elsewhere" needs a
destination.

**5. Step 9 mis-described its own input.** It said "from algorithm source," but in practice
the placement claim is already written in `README.md` and bootstrap transcribes it via
`context.files`. Corrected, with the old wording noted so the change is legible.

**Deferred by agreement:** the mechanical interface-satisfiability check (extract each named
interface's method set, flag same-name/different-signature). It's decidable and worth
automating, but it belongs at translate time and wants a script with tests.

## Test

New `.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py` — four
bounded-region guards, mirroring `test_build_commands_guidance.py` from #858:

- the `context:` schema block must not regress to the retired `<derived summary>` placeholder
- it must route to derivation step 9
- step 9 must keep the citation-bounding requirement
- step 9 must keep the settled-vs-declared-unknown split and the registration count

I verified these guard content rather than restate it: run against the pre-change file from
`HEAD`, **all four fail**; after the change, all four pass.

## Verification

- `ruff check pipeline/ .claude/skills/ --select F` → All checks passed
- Full CI suite → **2618 passed, 9 skipped, 2 xfailed** (was 2614; +4 new), coverage **97.68%** (gate 90%)
- specify's own SKILL.md-asserting tests (`test_config_contract.py`, `test_skill_substitution.py`) pass against the Phase 2 edit — they run in CI via `.github/workflows/test.yml:42`

## Stale-reference sweep

Swept `context.text` / `CONTEXT_TEXT` / `derived summary` across `*.md` and `*.py`.

- **Fixed:** `test_defaults_templates.py:302` cited `<derived summary>` as its example of a
  non-YAML placeholder justifying a line-oriented parse. That placeholder no longer exists,
  so the example was stale; switched to `scenario: <derived>`. Its parser keys on
  `"  disable:"` and never touches the `context:` block, so the reshaped block does not
  affect it.
- **Left:** `sim2real-translate/SKILL.md` and `prompts/*` — consumers unchanged.
  `docs/epics/step-2/design.md:220` — an elided illustrative JSON example, no shape claim.
  `docs/superpowers/plans/2026-07-08-*.md:781` — historical plan, records what was true then.

**Flagged, not changed:** `CLAUDE.md`'s documented CI command omits
`.claude/skills/sim2real-specify/tests/`, which the real workflow does run
(`.github/workflows/test.yml:42`). The docs drifted from the workflow, not the reverse — no
workflow change needed. Left alone because CLAUDE.md is the maintainer's instruction file
with its own revise path, and it's outside this issue.

Separately noted: specify's tests fail standalone with
`ModuleNotFoundError: No module named 'pd_plumbing'` and pass when run alongside bootstrap's
tests (which is how CI invokes them). Confirmed pre-existing — identical failure on stashed
`HEAD`. Not addressed here.

## No pipeline behaviour change

`pipeline/` untouched. `manifest.py` does not validate `context.text` content, and this
changes only the guidance that produces it.

Note for future bundle fixes: `context` is inside the translation slice
(`pipeline/lib/slicer.py:58`), so editing it in an existing bundle changes that bundle's
`translation_hash` and orphans images built under the old hash. This PR changes only the
skills that generate a new bundle.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -714,7 +714,63 @@ Assemble transfer manifest from all prior task outputs.
 7. `workloads`: from task-4 output
 8. `context.files`: files in experiment root with deployment config or signal
    mapping (README.md, config.md, etc.)
-9. `context.text`: summary of target interface and placement from algorithm source
+9. `context.text`: summary of the algorithm and where it plugs into the component.
+
+   Read it from the operator-supplied context files listed in step 8 — in practice
+   `README.md`, which is where `/sim2real-specify` records its Phase 2 "Shape"
+   conclusion. Fall back to the algorithm source only when no context file states it.
+   (Earlier wording said "from algorithm source"; in practice the placement claim is
+   almost always already written in `README.md` and bootstrap is transcribing it.)
+
+   This text is substituted into the `/sim2real-translate` writer and reviewer prompts
+   as `{CONTEXT_TEXT}` and is treated as authoritative for the science. Two accuracy
+   constraints follow from that, and both are the operator's to satisfy — no gate
+   downstream checks either one.
+
+   **(a) Every `path:line` citation must bound the construct it names.** Open each
+   cited range and confirm its first and last line are the construct's own start and
+   end. `/sim2real-specify`'s `lint_citations.py` cannot do this for you — it verifies
+   that a path resolves in exactly one pinned tree and that the line numbers fall
+   inside the file, and its own docstring says it "cannot detect a correct citation
+   attached to a wrong transcription." A range that starts correctly and stops early
+   passes the linter and still misdescribes the code.
+
+   **(b) Separate what is ruled out from what is proposed.** These carry different
+   confidence and must not be written in the same voice:
+
+   - **Eliminations** — extension points shown NOT to work, each with the citation and
+     the mechanism that rules it out. These are conclusions. State them as settled;
+     they are the expensive, reusable part.
+   - **Placement** — the extension point(s) proposed to carry the algorithm. Unless
+     you have confirmed it is constructible, this is a **declared unknown**, in the
+     same sense `/sim2real-specify` already uses for target-API adapters: *declared
+     unknowns rather than guesses, so a wrong guess fails loudly rather than
+     mis-scoring silently* (`sim2real-specify/SKILL.md:166-167`). Write it as a
+     candidate, not as *the* extension point.
+
+   Confirming a placement is constructible means more than "the data can flow."
+   Citations showing that a value is reachable do not show that one type can carry
+   the interface set you are naming: two interfaces that declare the same method name
+   with different signatures cannot coexist on one Go type, and no citation about data
+   flow will reveal that. If you have not checked the method sets, the placement stays
+   a declared unknown.
+
+   State the count too — how many separate plugin registrations the placement implies.
+   "A custom X plus a custom Y" is ambiguous between one type implementing both and
+   two registrations, and a reader who resolves it the cheap way discovers the
+   difference as a compile error after designing around it.
+
+   **Say what happens if the placement fails.** The writer's build gate has a finite
+   retry budget, and a placement asserted as settled leaves it no licence to look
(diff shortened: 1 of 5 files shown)


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 0 important · 0 recommendations** across 4 lenses.

Every finding raised rests on a coverage gap that the PR head already closes, and the tests plus lint gate pass clean against it.

PR #864, opened 2026-08-27
feat(specify,translate): make a core modification a planned outcome, not a halt (closes #862)

### LINKED ISSUE #862: specify/translate: treat "no extension point fits" as a core-modification path, not a HALT

kalantar:
Closes #862

## What changed

`/sim2real-specify` Phase 2 halted whenever no extension point could express the required
shape, so an algorithm needing invasive surgery had **no representable outcome** — the
pipeline could only transfer algorithms the component already anticipated. Every piece of
plumbing for a core modification was already present and tested; only the guidance withheld
it.

**The halt is scoped, not deleted.** Its purpose — never quietly accept a weaker
decomposition that scores a different decision — is real and unchanged. Phase 2 now has
three outcomes instead of two:

1. An extension point expresses the shape → proceed
2. None does, and a core modification is planned → proceed, **declared**
3. None does, and the natural decomposition would be used instead → **HALT** (the
   silent-fallback case, and the only one that halts)

## Reviewer attention

**1. The open question is resolved as a declared entry, not a cap.** #862 asked how the
cost should be bounded and recorded. Answer: a **DECLARED CORE MODIFICATION** recorded the
way specify already records degradations — in the specification layer's own header, same
voice, no new mechanism — naming which files change, why no extension point sufficed, and
**what upstream change would invalidate it**. That last item is the real cost and the
easiest to leave implicit: the bundle now carries a patch, so the pinned ref stops being a
free variable. Deliberately **no numeric cap** on core surface — there's no principled N,
and a cap invites gaming by making fewer, larger edits. `files_modified` is the mechanical
ledger the declaration must agree with.

**2. The writer's job description was the real constraint, not a prohibition.** Nothing
anywhere forbade modifying core code. But the writer's job was *defined* as producing "a
production Go plugin", and an agent handed that deliverable produces a plugin — it never
weighs changing core code. It's now a faithful port, with a plugin as the first choice
(the target is plugin-based, and a plugin keeps the bundle patch-free) and a core
modification as a permitted fallback that must state its reason.

**3. The reviewer had to change or the two prompts deadlock — this is the non-obvious
part.** `agent-reviewer.md`'s Criterion 3 requires a five-item plugin registration chain
and mandates `NEEDS_CHANGES` if any item is missing. A port that registers no plugin fails
that **unconditionally**, so granting the writer permission without scoping the reviewer
would have produced a mandatory rejection on a criterion the port cannot satisfy — against
a finite retry budget. Criterion 3 is now scoped to the plugin the port registers, and new
**Criterion 3b** judges whether a core modification is *declared* rather than treating
`files_modified` as a defect. `agent-reviewer.md:78` already says "Apply ALL review
criteria below", so 3b is picked up with no index change.

**4. Fixes two promissory references #859 left behind.** #859 added *"see #862 — modifying
the component's own code is a legitimate outcome"* to specify Phase 2 and bootstrap step 9,
while the HALT still stood 25 lines above. The merged state was self-contradictory and
pointed at an unimplemented issue for its policy. Both now state the policy itself.

**5. A pre-existing test caught me, correctly.** #859's
`test_phase2_keeps_the_core_modification_escape_hatch` asserted a bare `"#862"` pointer —
the best available check when the policy didn't exist yet. Replacing that pointer with real
policy failed the test immediately. It now asserts the policy, which is strictly stronger
than "an issue number is mentioned."

**Verified, so the "no pipeline behaviour change" claim holds:** the writer-output check is
**presence-only** (`if f not in o`, never non-emptiness) and `plugin_type` has **no
`pipeline/` consumer at all**. So a port registering no plugin needs nothing new — the
writer is told to set `plugin_type` / `register_file` to empty strings rather than invent
values. `build-epp.sh:124` uploads the whole component checkout as build context, so
modified core files genuinely compile.

## Tests

New `.claude/skills/sim2real-translate/tests/test_writer_core_modification.py` — six guards
over **both ends of the contract**, since they only work together (writer may act + must
explain; reviewer must not mandatorily reject). Extended
`sim2real-specify/tests/test_placement_guidance.py` with the halt-scoping and rebase-cost
guards.

Verified every asserted string is **absent from the pre-change files**, so the guards fail
against the old text rather than restating the new:

| Guard | Pre-change |
|---|---|
| `Modifying component code` section | absent |
| plugin-only job statement | present (guard fires) |
| `DECLARED CORE MODIFICATION` | absent |
| `silent fallback` | absent |
| `rebase` | absent |

## Verification

- `ruff check pipeline/ .claude/skills/ --select F` → All checks passed
- Full CI suite → **2745 passed, 9 skipped, 2 xfailed**, coverage **97.73%** (gate 90%)

## Sweep

Swept every `#862` reference: no promissory `see #862` remains. The surviving mentions are
provenance in test docstrings and one in `agent-reviewer.md:117` that cites the issue and
directs the reader to Criterion 3b — which exists in the same file.

Checked the other translate prompts for plugin-only framing that would conflict:
`agent-expert.md` is read-only advisory and unaffected; the Tool Discipline bullet in
`agent-writer.md` said "editing plugin files" and was widened to match.

**Left deliberately:** the `/sim2real-translate` skill description still says "production Go
plugins". That's discovery metadata describing the common case, and a plugin remains the
right default — widening it would overstate how often core modification applies.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -774,8 +774,13 @@ Assemble transfer manifest from all prior task outputs.
    elsewhere — it will spend the budget trying to make the stated shape work. Say
    explicitly that if the candidate does not hold, the eliminations still do, and the
    search continues within them. That tells the writer where NOT to spend retries.
-   Where no extension point fits at all, see #862 — a core modification is a legitimate
-   outcome, not a dead end.
+   Where no extension point fits at all, the placement is a **declared core
+   modification** — which files must change, why no extension point sufficed, and what
+   upstream change would invalidate it. That is a legitimate outcome, not a dead end;
+   `/sim2real-specify` Phase 2 defines the declaration and `/sim2real-translate`'s writer
+   is permitted to carry it out. Copy such a declaration through intact — a core
+   modification that reaches the writer as a bare file list is indistinguishable from an
+   accident.
 
    Use plain prose labels, in the style bundles already use in this field — do not
    invent a marker syntax. `PLACEMENT` and `DECLARED DEGRADATION (Dn)` are established;
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -863,10 +863,15 @@ context:
     <ELIMINATIONS -- extension points ruled out, each with citation + mechanism.
     Settled conclusions.>
 
-    <PLACEMENT -- the candidate extension point(s), how many separate plugin
-    registrations it implies, and every interface each type must implement. A
-    declared unknown until the method sets are checked. State what still holds
-    if it fails, so the writer knows where to search next.>
+    <PLACEMENT -- either shape is valid:
+      (a) the candidate extension point(s), how many separate plugin registrations
+          it implies, and every interface each type must implement. A declared
+          unknown until the method sets are checked. State what still holds if it
+          fails, so the writer knows where to search next.
+      (b) a DECLARED CORE MODIFICATION, where no extension point can carry the
+          decision -- which component files must change, why no extension point
+          sufficed, and what upstream change would invalidate it. Copy such a
+          declaration through intact.>
 
     <DECLARED DEGRADATION (Dn) -- any substitution, with the direction of bias.>
   files: <list of context files>
--- a/.claude/skills/sim2real-bootstrap/SKILL.md
+++ b/.claude/skills/sim2real-bootstrap/SKILL.md
@@ -870,8 +870,10 @@ context:
           fails, so the writer knows where to search next.
       (b) a DECLARED CORE MODIFICATION, where no extension point can carry the
           decision -- which component files must change, why no extension point
-          sufficed, and what upstream change would invalidate it. Copy such a
-          declaration through intact.>
+          sufficed, and what upstream change would invalidate it, PLUS the plugin
+          that carries the algorithm (the core edit is the hook, the plugin is the
+          switch the treatment overlay names). Copy such a declaration through
+          intact.>
 
     <DECLARED DEGRADATION (Dn) -- any substitution, with the direction of bias.>
   files: <list of context files>
--- a/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
+++ b/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
@@ -106,11 +106,16 @@ def test_step9_carries_a_declared_core_modification_through():
     is guarded here rather than left to prose alone.
     """
     step9 = _norm(_task5_step9())
-    assert "core modification" in step9, (
-        "Task 5 step 9 no longer mentions a declared core modification as a valid "
-        "placement. #862: where no extension point fits, that is the outcome, and "
-        "bootstrap must be willing to transcribe it rather than treating the case as "
-        "a dead end."
+
+    # NOT asserted: a bare "core modification". #859 already wrote "a core
+    # modification is a legitimate outcome" here as a forward reference to #862, so
+    # that phrase passes against the pre-#862 text and discriminates nothing. What
+    # #862 added is the declaration -- the placement shape, and copying it through.
+    assert "declared core" in step9, (
+        "Task 5 step 9 no longer names a DECLARED core modification as a valid "
+        "placement shape. #862: where no extension point fits, that is the outcome, "
+        "and bootstrap must transcribe the declaration rather than treating the case "
+        "as a dead end or reducing it to a file list."
     )
     assert "intact" in step9, (
         "Task 5 step 9 no longer tells the author to copy a core-modification "
--- a/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
+++ b/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
@@ -39,8 +39,13 @@ def _norm(text: str) -> str:
     modification` is not the substring `declared core modification` -- and a guard
     that fails on reflow tells you nothing about whether the rule survived. Only
     the words matter here, never the wrapping.
+
+    Markdown emphasis is stripped for the same reason: `modifies *different*
+    files` should satisfy a guard on `modifies different files`, since the
+    asterisks are formatting, not content. Only `*` is removed -- `_` has to
+    survive for identifiers like `files_modified`.
     """
-    return " ".join(text.split())
+    return " ".join(text.replace("*", "").split())
 
 
 def _task5() -> str:
--- a/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
+++ b/.claude/skills/sim2real-bootstrap/tests/test_context_text_guidance.py
@@ -8,16 +8,18 @@ therefore within spec -- a citation range that stopped short of the construct it
 and a placement asserted as settled that could not be built at all (two interfaces
 declaring the same method name with different signatures cannot sit on one Go type).
 
-Two bounded-region guards here, mirroring test_build_commands_guidance.py (#858):
+Bounded-region guards over two areas, mirroring test_build_commands_guidance.py (#858):
 
-- the schema block must not regress to a bare placeholder, and must route to the
-  derivation step where the constraints live;
-- the derivation step must keep both constraints -- citation bounding, and the
-  settled-vs-declared-unknown split.
+- the schema block must not regress to a bare placeholder, must route to the derivation
+  step where the constraints live, and must offer both valid placement shapes;
+- the derivation step must keep both #859 constraints -- citation bounding, and the
+  settled-vs-declared-unknown split -- and must carry a declared core modification
+  through intact (#862), since bootstrap is the link between specify declaring one and
+  the writer acting on it.
 
 Prose beyond these regions is not asserted on (see test_byo.py's note on why NL
-frontends are not CI-tested); what is asserted is that the two constraints the issue
-states outright have not been quietly dropped.
+frontends are not CI-tested); what is asserted is that the constraints these issues
+state outright have not been quietly dropped.
 """
 
 import re
@@ -27,6 +29,18 @@ _SKILL_MD = Path(__file__).resolve().parents[1] / "SKILL.md"
 
 # The placeholder #859 retired. Its return would mean the constraints were dropped.
 _RETIRED_PLACEHOLDER = "<derived summary>"
+
+
+def _norm(text: str) -> str:
+    """Collapse all whitespace runs to single spaces.
+
+    Every phrase assertion below runs through this. These files are hard-wrapped
+    prose, so a guarded phrase can land across a line break -- `declared core\\n
+    modification` is not the substring `declared core modification` -- and a guard
+    that fails on reflow tells you nothing about whether the rule survived. Only
+    the words matter here, never the wrapping.
+    """
+    return " ".join(text.split())
 
 
 def _task5() -> str:
@@ -57,7 +71,7 @@ def _task5_step9() -> str:
 
 def test_context_block_is_not_a_bare_placeholder():
     """#859: `<derived summary>` imposed no accuracy requirement and was retired."""
-    assert _RETIRED_PLACEHOLDER not in _task5_context_block(), (
+    assert _RETIRED_PLACEHOLDER not in _norm(_task5_context_block()), (
         f"Task 5's context: block is back to the bare {_RETIRED_PLACEHOLDER!r} "
         "placeholder. Issue #859 replaced it because it imposed no accuracy "
         "requirement on a field /sim2real-translate treats as authoritative."
@@ -66,7 +80,7 @@ def test_context_block_is_not_a_bare_placeholder():
 
 def test_context_block_points_at_the_derivation_step():
     """The block must route the agent to where the constraints are stated."""
-    assert "step 9" in _task5_context_block(), (
+    assert "step 9" in _norm(_task5_context_block()), (
         "Task 5's context: block does not point at derivation step 9, so the "
         "accuracy constraints are reachable only by luck."
     )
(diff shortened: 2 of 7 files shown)


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 2 important · 0 recommendations** across 4 lenses.

The guidance changes hold up end to end and the suite passes, but one reviewer-prompt sentence contradicts another four lines above it, and the guard test named for the PR's central change passes against the old text too.

### Important
- **Criterion 3b's "exactly one of the three" contradicts Criterion 3's "check both", leaving the common plugin-plus-modification case ambiguous for the reviewer agent.** (`.claude/skills/sim2real-translate/prompts/agent-reviewer.md:140`) — agent-reviewer.md:119 says "Most ports that modify component code still register a plugin as well; in that case check both", while :140 says every port lands in "exactly one of the three: a registered plugin (Criterion 3), a declared core modification (the rest of 3b), or this failure" — an agent classifying a plugin-registering port as Criterion 3 could stop there and never apply 3b's declaration checks, which is exactly the undeclared-modification hazard the PR adds 3b to catch. The same overclaim is baked into test_writer_core_modification.py:147's docstring.
- **test_phase2_halts_only_on_silent_fallback cannot fail for the change it names — it passes against the pre-PR wording.** (`.claude/skills/sim2real-specify/tests/test_placement_guidance.py:104`) — The only assertion is `"silent fallback" in phase2`, and the base revision (53147e0, specify/SKILL.md: "HALT if no extension point can express the required shape. The hazard is silent fallback to the weaker natural decomposition...") already contains that substring, so the test passes on both the old blanket-HALT text and the new scoped text. Nothing in the suite would catch a regression that restores the blanket HALT alongside the new DECLARED CORE MODIFICATION section — the one semantic change this test is named for.

kalantar:
Both blocking findings fixed in 62afeeb. Both were right, and both were mine.

The contradiction was the worse of the two. Criterion 3b claimed every port lands in "exactly one of the three", which is false in the most common case: a port that registers a plugin and also modifies component files owes both criteria. As written, an agent could classify such a port under Criterion 3, watch the registration chain pass, and stop — never applying 3b's declaration checks. That is exactly the undeclared-modification hazard 3b was added to catch, so the wording defeated the criterion it introduced. It now carries an explicit table over all four combinations, with the plugin-plus-modification row called out, and says plainly that passing the registration chain tells you nothing about whether the files the port also touched were declared. The same overclaim was in the guard's docstring and is corrected there too.

The toothless test is worth explaining, because the way it happened is a trap. `test_phase2_halts_only_on_silent_fallback` asserted only `"silent fallback" in phase2`, and the pre-#862 paragraph already read "The hazard is silent fallback to the weaker natural decomposition" — so it passed against both the old blanket HALT and the new scoped one, and would not have caught a regression restoring the blanket HALT.

But the phrase was genuinely absent from the *raw* pre-change text, which is hard-wrapped as `silent\nfallback`. The verification I posted in the previous comment measured raw text and was accurate when I ran it. Adding `_norm()` to fix reflow brittleness then made the assertion run on normalized text, where the phrase *is* present. So the robustness fix silently defanged one guard, and the verification table row for "silent fallback" in the PR body is stale as a result. Flagging that explicitly rather than leaving a claim standing that no longer holds.

It now asserts what actually discriminates — that the blanket phrasing "HALT if no extension point can express" is absent — with the positive check kept alongside and a comment on why it can't stand alone.

I then re-audited every #862 guard the same way: normalized, against the post-#859 pre-#862 baseline, which is the only baseline that isolates this PR. That found one more the review hadn't flagged — bootstrap's `"core modification"` assertion was also a tautology, because #859 had already written "a core modification is a legitimate outcome" there as a forward reference to this issue. Tightened to `"declared core"`. All eight #862 guards now discriminate.

Also took the non-blocking finding: the second `.index()` in the ordering assertion had no membership check, so a rewording would have raised a bare ValueError instead of a descriptive message.

ruff clean, 2748 passed, 9 skipped, 2 xfailed, coverage 97.73%.

kalantar:
Both findings resolved in d20d8d3, by removing their cause rather than chasing them.

The review was right three passes in a row, and the third pass made the pattern clear: the plugin-less port was the single source of all of it. That branch was scope I added, not something this issue asked for. #862 asked to make a core modification a legitimate declared outcome instead of a halt; it never asked that a port be able to register zero plugins.

Following the branch through would have meant making reviewer Criteria 4, 5 and 6, the APPROVE template, and writer Phases 3 and 4 all conditional on whether a plugin exists — a large refactor of the reviewer's core checklist and the writer's main phases, changing behaviour for every ordinary plugin port, to support a case nobody needs yet.

There's also a structural reason the branch was wrong rather than merely expensive. The treatment overlay enables an algorithm by naming its plugin type. A port with no registered plugin has no way to be switched on for the treatment scenario at all — it would differ from baseline only by what's compiled into the image, which the overlay can't express. The core edit provides the hook; the plugin provides the switch.

So a core modification now supplements a registered plugin. This is a net reduction: Criterion 3 is unconditional again, which removes the fall-through hole from the first pass rather than patching it, and 3b's no-deliverable catch-all goes with it, since a port without a plugin simply fails Criterion 3. Writer Phases 3/4 and Criteria 4/5/6 now need no changes at all. What survives from the second pass is the part that matters — 3b is additional to Criterion 3, never a substitute, so a passing registration chain doesn't excuse the declaration checks on files the port also touched.

Everything the issue asked for is intact: the halt is scoped to silent fallback, a DECLARED CORE MODIFICATION is a legitimate planned outcome with its files, reason and upstream-rebase cost, the writer may make the change, and the reviewer judges whether it was declared rather than whether it exists. specify and bootstrap were updated to describe the same shape so all three skills agree.

I also took the item-4 recommendation. The "fidelity" assertion was satisfied by the closing undeclared-modification clause, so deleting item 4 — the check that modified files match a prior declaration — left the guard green. Every phrase asserted there is now unique to the item it protects.

And one more brittleness class, the same shape as the line-wrapping one from last pass: an assertion missed because the prose reads "modifies *different* files" and markdown emphasis broke the phrase. The normalizer now strips asterisks as well as collapsing whitespace, in all three guard files — only asterisks, since underscores have to survive for identifiers like files_modified.

Two obsolete guards were replaced rather than deleted, and both new ones assert the absence of the plugin-less wording so a future edit can't quietly restore it. I audited all ten #862 guards for discriminating power against the baseline that actually isolates what each protects — pre-#862 for the positive ones, this PR's plugin-less intermediate for the two negative ones. All ten discriminate.

ruff clean, 2749 passed, 9 skipped, 2 xfailed, coverage 97.73%.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 2 important · 1 recommendation** across 4 lenses.

The new policy prose is coherent and well-guarded by tests, but two untouched contract blocks inside the same two prompt files now contradict it, and neither is covered by the new guards.

### Important
- **The APPROVE verification-summary template has no line for the newly added Criterion 3b.** (`.claude/skills/sim2real-translate/prompts/agent-reviewer.md:230`) — Criterion 3b (agent-reviewer.md:133-162) applies whenever `files_modified` is non-empty and says the reviewer must "never stop at Criterion 3 because the registration passed", yet the Response Format block at lines 229-237 — "Reply in this exact format", summary "required — one line per criterion" — still enumerates exactly six bullets (Fidelity, Code quality, Registration, Config, Assembly, Treatment config). A reviewer following the literal template emits a complete-looking APPROVE with no recorded evidence that the core-modification declaration was checked, which is the enforcement this change depends on.
- **The output-schema comment for `description` still says "one-line summary" while the same file now requires four distinct items in that field.** (`.claude/skills/sim2real-translate/prompts/agent-writer.md:265`) — agent-writer.md:106-109 tells the writer to state in `{ALGO_NAME}_output.json`'s `description` that the port modifies component code, which files, why no extension point sufficed, and what upstream change would invalidate it; the Phase 4.5 schema block at line 265 still documents the field as `"<one-line summary of what was built>"`. `description` is free-form (no validation in copy_generated.py or pipeline/lib), so its only consumer is Criterion 3b, which checks for exactly the fuller content — a writer honouring the schema comment truncates the declaration into the undeclared-modification case that criterion exists to reject.

### Recommendations
- **Module docstring still scopes the file to "#859 ... the four rules" after this PR added two #862 guards and rewrote a third.** (`.claude/skills/sim2real-specify/tests/test_placement_guidance.py:1`) — The docstring (lines 1-17, untouched) frames the file as guarding #859's four rules, but it now also holds test_phase2_halts_only_on_silent_fallback, test_phase2_requires_the_rebase_cost_be_stated, and a rewritten test_phase2_keeps_the_core_modification_escape_hatch, all guarding #862 policy; the sibling test_context_text_guidance.py docstring was updated for #862 in this same PR.

kalantar:
Both fixed in 19bf088, plus a third of the same kind that I went looking for.

These two share a cause that's distinct from the earlier passes: the change added requirements without updating the literal templates and schemas the agents actually copy from. A requirement an agent never encounters in the block it's following isn't a requirement.

The APPROVE response format enumerated one line per criterion and had no line for 3b, so a reviewer following it emitted a complete-looking APPROVE carrying no evidence the declaration was ever checked — the same silent-pass shape 3b exists to close. There's now a required `Core modification` line, and its `none (files_modified empty)` form has to be written when the port changed nothing, because an omitted line and "nothing to declare" look identical to a later reader.

The Phase 4.5 schema documented `description` as a "one-line summary of what was built" while the core-modification section requires four items in that same field, and Criterion 3b reads exactly that content. The realistic failure was the writer honouring "one line" and truncating the declaration into the undeclared case. The comment now says what the field holds and that 3b reads it.

Then I swept for other blocks of the same class instead of waiting for another pass to name one, and found a third. The REVIEW REQUEST format listed `Plugin files: <files_created>` and `Test files:` — and `files_modified` is derived separately from `files_created`, so a modified component file appeared in no line of the handover at all. Since the reviewer reads only the paths the request lists, Criterion 3b item 3 ("no larger than the decision requires") was asking it to judge a change it had never been handed; it could see a filename in `files_modified` and nothing more. There's now a required `Modified component files` line with a `none` form, and a matching read step in the reviewer's Behavior list so the handover has a receiver as well as a sender. The reviewer is told to ask for the line if it's absent rather than assume there were none.

Three guards added, one per block, each mutation-tested by reverting its block to the pre-fix wording and confirming the guard fails. I also took the docstring recommendation — specify's guard file was still scoped to "#859 ... the four rules" though it now holds three #862 guards.

ruff clean, 2752 passed, 9 skipped, 2 xfailed, coverage 97.73%.

kalantar:

Nothing raised survives verification as blocking — the new guidance is internally consistent and the pipeline facts it leans on (files_modified derivation, treatment restore) actually hold.

### Recommendations
- **The `1b.` step breaks the surrounding ordered list when the file is rendered as Markdown.** (`.claude/skills/sim2real-translate/prompts/agent-reviewer.md:75`) — agent-reviewer.md:74-81 inserts a bare `1b.` line directly after item 1 with no blank line, so in CommonMark it becomes a lazy continuation of item 1 and the following `2.`–`6.` lines can no longer interrupt that paragraph (only a list starting at `1.` may). Harmless where the file is substituted as raw prompt text, but renumbering it as a plain step keeps the file readable on GitHub and matches every other numbered list in the document.

PR #866, opened 2026-08-27
fix(bootstrap): exclude bundle content-hash lines from scaffolded secret scan (closes #865)

### LINKED ISSUE #865: sim2real-bootstrap: scaffolded detect-secrets hook blocks every commit on the bundle's own content hashes
## Symptom

The pre-commit hook scaffolded by `/sim2real-bootstrap` blocks the first real commit of every bundle it scaffolds. It is not finding secrets — it is finding the bundle's own content hashes.

Observed in `kalantar-msb/pd-infocomm-3` when committing the output of `sim2real translate`:

```
Potential secrets about to be committed to git repo! Please rectify.

Secret Type: Hex High Entropy String
Location:    workspace/translations/<hash>/translation_output.json:3
Location:    workspace/translations/<hash>/skill_input.json:3
Location:    workspace/translations/<hash>/review/leastttftjoint/round_4.json:8
```

## Root cause

`.secrets.baseline` ships `HexHighEntropyString` with `hex_limit: 3`. SHA-256 hex has Shannon entropy ~3.9, so **every content hash the pipeline writes trips the detector.** This is structural, not incidental — the bundle format is built out of hashes:

| Field | Written by |
|---|---|
| `translation_hash`, `source_sha256` | `sim2real translate` → `skill_input.json`, `translation_output.json` |
| `package_manifest_sha256`, `runner_go_sha256`, `package_manifest_short` | translate review rounds → `review/<algo>/round_N.json` |
| `"hash"` (145 occurrences in one file) | `sim_results/` sample manifests |
| `ref:` | `transfer.yaml` component pin (git sha) |

Note the last two: `sim_results/counterfactual/sample_manifest.json` and `transfer.yaml` did not appear in the failure above only because they happened to be already-committed / still-unstaged at that moment. They fail on the next commit that touches them. Any bundle will hit this eventually.

Neither mitigation the hook itself prints works here:

- **`pragma: allowlist secret`** — impossible. The offending files are JSON, which has no comments, and they are machine-generated so a hand-annotation would be overwritten on the next `translate`.
- **Audit into `.secrets.baseline`** — not durable. The v0.13 baseline keys on `(filename, line_number, hashed_secret)`, and these values change on every translation. Each run strands the previous entries and fails again. This would turn the baseline into a running log of every hash the bundle ever produced.

So today the only path forward for an operator is routine `git commit --no-verify`, which defeats the hook the template exists to install.

## Fix (verified working in pd-infocomm-3)

Line-level exclusion keyed on the **field name**, not on the value and not on the file. Two template files:

**1. `templates/precommit/pre-commit-config.yaml`** — replace the flow-style `args` (line 45) with:

```yaml
        args:
          - --baseline
          - .secrets.baseline
          - --use-all-plugins
          - --exclude-lines
          - '(?i)(?:^[\s"-]*[a-z0-9_.]*(?:sha1|sha256|sha512|hash|digest|checksum|fingerprint|manifest_short)"?\s*[:=])|(?:^\s*-?\s*"?(?:ref|commit|revision)"?\s*[:=]\s*"?[0-9a-f]{7,64}"?\s*,?\s*(?:#.*)?$)'
```

**2. `templates/precommit/secrets.baseline`** — set `exclude.lines` to the same regex (currently `null`), so that a later `detect-secrets scan --baseline .secrets.baseline` regeneration agrees with the hook instead of silently re-introducing the findings.

The CLI arg is the load-bearing one. Under `--use-all-plugins`, `initialize.merge_plugins_from_baseline()` rebuilds every plugin with `exclude_lines_regex=args.exclude_lines`, so the baseline's own `exclude.lines` is **not** honored by the hook. Both must be set, and the config's arg is what actually takes effect. Worth a comment in the template so the next person doesn't "simplify" it to baseline-only.

### Why field-name keying rather than excluding the files

Excluding `workspace/translations/**` wholesale would also stop scanning the paths, env, and config those same files carry — which is exactly the blind spot the whole-repo hook was added to cover (#822 rationale vs. the collect-time redactor #819). Keying on the field name keeps the file in scope and drops only the hash line.

### Regex design notes

- Matches a key **ending** in `sha1|sha256|sha512|hash|digest|checksum|fingerprint|manifest_short`. The "ending" anchor is deliberate: `hashicorp_vault_token:` and `refresh_token:` do **not** match and stay scanned.
- Matches `ref|commit|revision` **only** when the value is a bare 7–64 char git sha. `ref: refs/heads/main` and any opaque token under `ref:` stay scanned.

## Verification

Whole staged set in pd-infocomm-3 (346 files) went from Failed to `Detect secrets....Passed`.

Canary file mixing hashes with real secret shapes — the hashes pass through, everything else still blocks the commit:

| Line | Result |
|---|---|
| `translation_hash`, `source_sha256` | skipped (intended) |
| `hf_token` | caught — Base64 High Entropy |
| `aws_secret_access_key` | caught — Base64 High Entropy + Secret Keyword |
| `password` | caught — Secret Keyword |
| `refresh_token` | caught — Hex High Entropy |
| `some_hash: <48-hex>` | **skipped** |

## Known cost, stated plainly

That last row is the tradeoff and it is the whole tradeoff: **a credential parked under a key ending in `hash`/`digest`/`checksum`/`sha256`/`fingerprint` would now slip past the hook.** Judged acceptable because those key names denote content-addressed values by convention and the alternative is habitual `--no-verify`, which disables everything. If that is judged too broad, the tighter variant is to enumerate the six field names the pipeline actually emits instead of the `hash|digest|checksum|...` family — at the cost of needing a template edit each time a new hash field is added to an output schema.

## Suggested tests

`.claude/skills/sim2real-bootstrap/tests/test_precommit_scaffold.py` already has the right shape for this (`test_hook_blocks_high_entropy_token_value`, `test_tuned_limit_is_load_bearing`, `test_hook_allows_token_tuning_fields`). Add alongside them:

- scaffolded hook **passes** on a JSON file containing `translation_hash` / `source_sha256` / `package_manifest_sha256` / `runner_go_sha256` / `package_manifest_short`, a `"hash"` field, and a `transfer.yaml`-style `ref: <sha>`;
- scaffolded hook still **blocks** `hf_token`, `aws_secret_access_key`, `password`, and `refresh_token` in that same file (the exclusion must not become a blanket bypass);
- `exclude.lines` in the scaffolded baseline equals the `--exclude-lines` arg in the scaffolded config (they drift silently otherwise).

## Provenance

Discovered and fixed locally in `kalantar-msb/pd-infocomm-3`; that repo now carries the change on both files. Filing here because the template is the source and every bundle `/sim2real-bootstrap` scaffolds inherits the failure.

kalantar:
Closes #865

## Problem

The `detect-secrets` hook that `/sim2real-bootstrap` scaffolds blocked the first real commit of every bundle it scaffolds — not on secrets, on the bundle's own content hashes. SHA-256 hex has Shannon entropy ~3.9, above the shipped baseline's `hex_limit: 3`, so every hash the pipeline writes trips `HexHighEntropyString`: `translation_hash` / `source_sha256` in `skill_input.json` + `translation_output.json`, `package_manifest_sha256` / `runner_go_sha256` / `package_manifest_short` in translate review rounds, the bare `"hash"` key in `sim_results/` sample manifests, and `transfer.yaml`'s bare-sha component `ref:`.

Neither mitigation the hook prints applies. `pragma: allowlist secret` is impossible — the files are machine-generated JSON, which has no comments. Auditing into `.secrets.baseline` isn't durable — the baseline keys on `(filename, line_number, hashed_secret)`, all of which change every translation, so each run strands the previous entries and fails again. That left routine `git commit --no-verify`, which disables the very scan the template exists to install.

## Fix

Line-level exclusion keyed on the **field name** — not the value, not the file — in both template files.

Field-name keying rather than excluding `workspace/translations/**` wholesale: those same files also carry paths, env, and config, and keeping them in scope is the whole #822 rationale for a whole-repo hook over the collect-time redactor (#819). The exclusion drops only the hash line.

- The key must **end** in `sha1|sha256|sha512|hash|digest|checksum|fingerprint|manifest_short`, so `refresh_token:` and `hashicorp_vault_token:` — which merely *contain* `ref`/`hash` — keep being scanned.
- `ref|commit|revision` is dropped only when the value is a bare 7–64 char git sha, so `ref: refs/heads/main` and any opaque token under `ref:` keep being scanned.

## Reviewer attention

**The CLI arg is load-bearing, and this is the non-obvious part.** The hook does **not** honor the baseline's own `exclude.lines` — verified directly: with the regex set only in `.secrets.baseline` and no `--exclude-lines`, all 13 canary findings came back. Both files carry the regex for different reasons: the config's arg is what actually takes effect, and the baseline's copy is so a later `detect-secrets scan --baseline .secrets.baseline` regeneration agrees with the hook instead of silently re-finding what the hook skips. A test asserts they're byte-identical, and the template comment says explicitly not to "simplify" this to baseline-only.

One correction to the issue's stated mechanism, which does not change the fix: the issue attributes the non-honoring to `merge_plugins_from_baseline()` under `--use-all-plugins`. Empirically the baseline's `exclude.lines` is ignored by `detect-secrets-hook 0.13.1+ibm.64.dss` **either way**, so the template comment states the broader fact rather than pinning it on that flag.

**Known cost, carried over from the issue and accepted there.** A credential parked under a key ending in `hash`/`digest`/`checksum`/`sha*`/`fingerprint` would now slip past the hook. Those names denote content-addressed values by convention, and the alternative is habitual `--no-verify`, which disables everything. The template documents the tighter variant (enumerate the exact field names the pipeline emits) and its cost (a template edit each time an output schema gains a hash field) for anyone who judges the word family too broad.

**Test-harness change worth a look.** `_hook_blocks` previously hardcoded `["--baseline", ".secrets.baseline", "--use-all-plugins"]`. That invocation is no longer the one operators run, and a hardcoded list would have silently omitted `--exclude-lines` — the new tests would have exercised an invocation that doesn't exist. It now reads the args out of the scaffolded `.pre-commit-config.yaml` via `_shipped_hook_args()`, so every behavioral test (including the four pre-existing #819/#822 ones) exercises the shipped invocation. `_hook_blocks` is kept as a thin single-file wrapper over a new multi-file `_run_hook`, so existing call sites are unchanged.

Also note: `detect-secrets-hook` writes its findings report to **stderr**, not stdout. `_flagged_lines` joins both streams.

## Tests

Four new tests alongside the existing ones in `test_precommit_scaffold.py`:

| Test | Asserts |
|---|---|
| `test_hook_allows_bundle_content_hashes` | hook passes on a `translation_output.json`-shaped canary of all six hash fields plus a `transfer.yaml` bare-sha `ref:` |
| `test_hook_blocks_real_secrets_beside_content_hashes` | in that same file, flagged lines equal **exactly** the `hf_token` / `aws_secret_access_key` / `password` / `refresh_token` / `hashicorp_vault_token` lines — proves the exclusion is not a blanket bypass and the hash lines are still skipped |
| `test_exclude_lines_is_load_bearing` | dropping the arg restores every false positive |
| `test_exclude_lines_agrees_between_config_and_baseline` | the two regex copies are byte-identical (unguarded, so CI without `detect-secrets` still catches a removal) |

Canary line numbers are derived from the fixture, not hand-counted, so adding a field can't silently shift an assertion.

Verified by mutation, not just by green:

- reverting the config to the old flow-style `args` → all 4 new tests fail
- loosening the "ending in" anchor so `refresh_token`/`hashicorp_vault_token` match → `test_hook_blocks_real_secrets_beside_content_hashes` fails

Local `detect-secrets` is exactly the pinned `0.13.1+ibm.64.dss`, so the behavioral tests ran against the shipped version rather than skipping.

Full CI: `ruff check pipeline/ .claude/skills/ --select F` clean; **2719 passed, 9 skipped, 2 xfailed**, coverage 97.73%.

## Stale-reference sweep

Grepped `*.md`, `docs/`, `.claude/skills/`, and `README*` for `detect-secrets`, `secrets.baseline`, `use-all-plugins`, `no-verify`, `pre-commit`. Updated two: `sim2real-bootstrap/SKILL.md` (the Task 0b scaffold description, plus a line in the operator-facing activation text noting content hashes are already excluded so a finding deserves a real look before `--no-verify`) and `CLAUDE.md`'s bootstrap bullet. Left alone: the retrofit guidance at `SKILL.md:1031`, which is still correct — `detect-secrets scan --baseline <file>` reads the existing baseline's settings, so a regeneration preserves `exclude.lines`, which is precisely why the regex is mirrored there. No pipeline code changed, so `pipeline/README.md` needed nothing.


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `comment-analyzer` said NEEDS DISCUSSION; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 0 important · 0 recommendations** across 4 lenses.

The exclusion regex behaves exactly as its comments and tests claim, the two copies are pinned in sync by an unguarded test, and nothing raised survives verification as a merge blocker.

PR #868, opened 2026-08-27
fix(bootstrap): correct invalid HCA name in nic-exclusion default (closes #867)

### LINKED ISSUE #867: bootstrap: defaults/nic-exclusion.yaml ships an invalid HCA name (mlxl5_7), excluding nothing at that slot
## Summary

`templates/defaults/nic-exclusion.yaml` ships a device list containing `mlxl5_7`, which
is not a valid HCA name. At that slot the fragment excludes nothing, so an operator who
enables it believing the list is complete gets eight exclusions where the list claims
nine — and `mlx5_7`, if present on their cluster, stays in NCCL's candidate set.

Lines 38 and 44 (decode and prefill, kept in sync):

```yaml
value: "mlx5_0,mlx5_2,mlx5_4,mlx5_8,mlxl5_7,mlx5_10,mlx5_12,mlx5_14,mlx5_16"
```

The header already flags this, at line 12:

> `mlxl5_7` IS CARRIED VERBATIM AND IS ALMOST CERTAINLY A TYPO for mlx5_7, which
> would mean that device is NOT actually excluded. Kept as-is because it is what
> ran in the configuration this was lifted from; fix it deliberately, together with
> replacing the rest of the list, rather than by assuming the correction is safe.

So this issue is that deliberate fix. The note is accurate provenance and the decision to
preserve it was defensible; what makes it worth closing now is the drift below.

## Why now: the note is not holding

`pd-infocomm-3`'s copy of the fragment has already been hand-edited to `mlx5_7` on both
lines, diverging from this template. That is the exact move the header argues against, and
the result is a fragment whose body now contradicts its own line 12. Nothing has rendered
from it — the stem is in `defaults.disable` per #853 — so the blast radius is still zero,
which is the right time to settle it rather than after a run.

The prediction is that every P/D bundle repeats this: an operator reads line 12, agrees the
correction is obvious, fixes it locally, and the template stays wrong.

## Upstream origin — not ours to fix alone

The typo is inherited, not introduced here. It appears 4 times in the `llm-d-benchmark`
submodule's own shipped guides at the current pin (`76473d0`):

- `config/scenarios/guides/pd-disaggregation.yaml:334,423`
- `config/scenarios/guides/wide-ep-lws.yaml:392,530`

`llm-d-benchmark` is a third-party submodule (`https://github.com/llm-d/llm-d-benchmark.git`,
branch `main`), so fixing this template does not fix upstream, and a later template resync
against those guides could reintroduce it. Reporting it upstream is a separate track — I
have not opened anything there, and this issue does not depend on it.

## Options

**A — Correct the typo, retire the note (minimum, recommended).** Set both lines to
`mlx5_7` and rewrite the line 12 paragraph, which becomes false the moment the value
changes. Keep the "replace with your own `ibv_devinfo` output" warning at lines 9-10; that
one is still true and is the actually load-bearing caution. Smallest diff, removes the
foot-gun, keeps fragment shape and existing tests valid.

**B — Make the example unmistakably an example.** The header says at line 9 "THE LIST
BELOW IS AN EXAMPLE, NOT A DEFAULT." A real-looking nine-device list does not read that
way, and A leaves it looking *more* authoritative by fixing it. Replace with an obvious
placeholder (e.g. `REPLACE_WITH_YOUR_IBV_DEVINFO_OUTPUT`) so enabling the stem without
supplying a list fails loudly instead of excluding another cluster's cards. Larger
behavioural change for anyone who enables the fragment expecting it to work as-shipped,
and may need `tests/test_defaults_templates.py` updated.

A and B are not exclusive — B with the corrected list in a commented example block gets
both properties. Maintainer's call; A alone already closes the defect as stated.

## Suggested test

`tests/test_defaults_templates.py` already covers this fragment's merge behaviour
(`:305`, the `prefill.extraEnvVars` Tier 2b case). Add a cheap guard alongside it:

- every comma-separated entry in `NCCL_EXCLUDE_IB_HCA` matches `^mlx5_\d+$` (or, under
  option B, equals the placeholder sentinel) — this catches `mlxl5_7` and any future
  transposition on resync;
- the decode and prefill values are identical, since the header requires the two copies
  stay in sync and nothing currently enforces it.

## Provenance

Found while semantically diffing `pd-infocomm-2`'s hand-authored
`baselines/baseline.yaml` (which carries the original `mlxl5_7`) against `pd-infocomm-3`'s
fragment-based baseline, where the same settings moved into `baselines/defaults/` and ship
disabled per #853.

kalantar:
Closes #867.

## What changed

`templates/defaults/nic-exclusion.yaml` shipped `mlxl5_7` in `NCCL_EXCLUDE_IB_HCA` on both the decode and prefill copies. Not a valid HCA name, so that slot excluded nothing — eight exclusions from a list of nine, with `mlx5_7` left in NCCL's candidate set.

Two hunks:

1. Both values corrected to `mlx5_7`.
2. The header paragraph that documented the typo as deliberately preserved is replaced with a three-line note. The correction makes the old paragraph false, so it could not stay.

Issue Option A, minimum scope. Option B (placeholder sentinel so an unconfigured enable fails loudly) is not implemented.

## Not in scope

- **The suggested regex/parity test.** The issue proposes asserting every entry matches `^mlx5_\d+$` and that decode/prefill are identical. Deliberately left out as beyond "just fix the typo" — worth a follow-up, since the resync risk below is otherwise unguarded.
- **Upstream.** The typo is inherited, not introduced here: 4 occurrences in `llm-d-benchmark`'s own guides at the current pin `76473d0` (`config/scenarios/guides/pd-disaggregation.yaml:334,423`, `wide-ep-lws.yaml:392,530`). That submodule is third-party, so this does not fix upstream, and a later template resync against those guides will reintroduce it. The new note says so. No upstream issue filed.

## Reviewer attention

- **Existing bundles are not retroactively fixed.** `copy_framework_defaults` (`.claude/skills/sim2real-bootstrap/byo.py:533`) copies templates at bootstrap time, so already-bootstrapped repos keep their own copies. `pd-infocomm-2` still carries the original typo; `pd-infocomm-3` was hand-patched to `mlx5_7`, which is what motivated the issue. Only future bootstraps (or `--force` re-bootstraps) pick this up. Not addressed here and arguably not worth addressing — the fragment ships in `defaults.disable` per #853 and nothing has rendered from it.
- **Note wording keeps `mlxl5_7` once**, as the historical referent. A grep for the bad token still hits this file at line 12 by design.

## Sweep

Grepped for `mlxl5_7`, `NCCL_EXCLUDE_IB_HCA`, and `nic-exclusion` across `.claude/`, `pipeline/`, and `docs/`. No doc or skill prompt quotes the device list, so nothing else needed updating. The only remaining `mlxl5_7` in the repo outside the submodule is the intentional mention at line 12.

## Verification

- `pytest .claude/skills/sim2real-bootstrap/tests/` — 542 passed, 9 skipped.
- `ruff check pipeline/ .claude/skills/ --select F` — clean.

No test changes were needed: the fragment tests at `test_defaults_templates.py` iterate all fragments and assert merge properties (Tier 2b merge-by-`name` on `extraEnvVars`), never the device string.

### DIFF
--- a/.claude/skills/sim2real-bootstrap/templates/defaults/nic-exclusion.yaml
+++ b/.claude/skills/sim2real-bootstrap/templates/defaults/nic-exclusion.yaml
@@ -9,10 +9,9 @@
 # to a slower transport, silently. THE LIST BELOW IS AN EXAMPLE, NOT A DEFAULT:
 # replace it with your own `ibv_devinfo` output before enabling.
 #
-# `mlxl5_7` IS CARRIED VERBATIM AND IS ALMOST CERTAINLY A TYPO for mlx5_7, which
-# would mean that device is NOT actually excluded. Kept as-is because it is what
-# ran in the configuration this was lifted from; fix it deliberately, together with
-# replacing the rest of the list, rather than by assuming the correction is safe.
+# `mlxl5_7` was not a valid HCA name, so that slot excluded nothing -- eight
+# exclusions from a list of nine. Corrected to `mlx5_7` (#867). Upstream's guides
+# still carry the typo, so a template resync will reintroduce it.
 #
 # Duplicated across decode and prefill because `extraEnvVars` is a per-role key
 # with no vllmCommon equivalent. Keep the two copies in sync.
@@ -35,12 +34,12 @@ scenario:
   decode:
     extraEnvVars:
       - name: NCCL_EXCLUDE_IB_HCA
-        value: "mlx5_0,mlx5_2,mlx5_4,mlx5_8,mlxl5_7,mlx5_10,mlx5_12,mlx5_14,mlx5_16"
+        value: "mlx5_0,mlx5_2,mlx5_4,mlx5_8,mlx5_7,mlx5_10,mlx5_12,mlx5_14,mlx5_16"
       - name: UCX_MAX_RMA_RAILS
         value: "1"
   prefill:
     extraEnvVars:
       - name: NCCL_EXCLUDE_IB_HCA
-        value: "mlx5_0,mlx5_2,mlx5_4,mlx5_8,mlxl5_7,mlx5_10,mlx5_12,mlx5_14,mlx5_16"
+        value: "mlx5_0,mlx5_2,mlx5_4,mlx5_8,mlx5_7,mlx5_10,mlx5_12,mlx5_14,mlx5_16"
       - name: UCX_MAX_RMA_RAILS
         value: "1"


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `pr-test-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

The one-character correction and the rewritten header are right, but the change ships with nothing that would catch the same typo coming back.

### Important
- **No test guards the corrected HCA list or the decode/prefill parity the fragment's header requires.** (`.claude/skills/sim2real-bootstrap/tests/test_defaults_templates.py:305`) — The suite is entirely structural and glob-driven — its own docstring says "Nothing else validates their *content*" — and the only nic-exclusion mention is a comment at line 305 about merge behaviour; no assertion reads NCCL_EXCLUDE_IB_HCA's value or compares the decode/prefill copies, so reverting this fix leaves all 77 tests passing (verified locally). The file itself declares the invariant "Keep the two copies in sync" with nothing enforcing it, and issue #867 asked for exactly this guard; the existing per-fragment test `test_disabled_fragment_says_why_in_its_header` is the precedent for adding one.

PR #874, opened 2026-09-01
fix(health): stop triaging unschedulable pods as pod-health failures

### LINKED ISSUE #873: Unschedulable pods are cancelled before --pending-threshold can elapse (two classifiers conflict)
## Summary

Two independent subsystems classify an unschedulable pod, they run in the same
orchestrator cycle, and the second one cancels the PipelineRun before the first
one's deliberate grace period can elapse. The effect is that
`--pending-threshold` (default 600 s) is **unreachable for an unschedulable
pod**, and a cell blocked only by transient resource scarcity is recorded as
`failed` roughly one poll interval after the pod goes Pending.

Both subsystems are individually intentional and individually tested. Nothing
tests the interaction, which is why this has gone unnoticed.

## The two classifiers

**System 1 — `_handle_pending_pods` (`pipeline/deploy.py:713`), designed to wait.**

`pipeline/lib/pod_pending.py:classify_pending_reason` sorts scheduling-failure
messages into `recoverable` (scarcity) and `non_recoverable` (config error):

```python
_RECOVERABLE_PATTERNS = [
    re.compile(r"Insufficient\s+\S+", re.IGNORECASE),
    re.compile(r"\d+/\d+ nodes are available:.*Insufficient", re.IGNORECASE),
]
```

`Insufficient nvidia.com/gpu` matches both patterns, so GPU scarcity is
`recoverable`. For a recoverable pending, `_handle_pending_pods` stamps
`pending_since`, returns `False`, and only reclaims the slot once
`elapsed > pending_threshold`. The docstring states the intent explicitly:

> Unrecognized non-empty messages -> recoverable (waiting up to the configured
> threshold is cheaper than irreversible cancellation).

**System 2 — `_check_pod_health` (`pipeline/deploy.py:854`), no grace period.**

`pipeline/lib/health.py:triage_pod` has its own scheduling branch at
`health.py:184-196`: phase `Pending` plus a `FailedScheduling` event whose
message contains `insufficient` returns tier 2, and `_check_pod_health`
escalates every tier-2 finding to a PipelineRun cancellation
(`deploy.py:894`). The scheduling branch consults no duration and no
`RemediationTracker`, so it fires on the first check.

## Why the wait never happens

The run loop, `deploy.py:2887-2929`:

```
:2889   reclaimed = _handle_pending_pods(..., pending_threshold=600, ...)
        if reclaimed: ... continue          # only short-circuits on RECLAIM
:2909   _handle_timeout(...)
:2926   escalate = _check_pod_health(...)   # cancels tier-2 immediately
```

While System 1 is *waiting* it returns `False`, which is indistinguishable from
"nothing wrong". Control therefore falls through to `_check_pod_health` in the
same cycle, which cancels. The `pending_since` timestamp is written to an entry
whose PipelineRun has already been deleted, so `elapsed` never grows toward the
threshold.

## Observed impact

During a period of cluster-wide GPU contention (77 of 96 GPUs held by other
tenants, no node with 4 free GPUs for a TP=4 pod), four consecutive cells were
cancelled 2-3 minutes after dispatch:

```
16:24:13  PipelineRun "leastttftjoint-deep-research-single-turn-..."   was cancelled
16:26:18  PipelineRun "causalsloexternality-deep-research-single-..."  was cancelled
16:28:21  PipelineRun "baseline-deep-research-single-turn-..."         was cancelled
```

Each had a decode pod Pending with
`0/20 nodes are available: 14 Insufficient nvidia.com/gpu`. An unschedulable pod
is not terminal -- the kube-scheduler re-queues it and it will schedule once
capacity frees -- and free capacity was observably churning minute to minute, so
some of these would likely have started had the 600 s window been honoured.
Escalation also marks the pair `failed` (`deploy.py:2937`), so a
capacity-only stall is recorded identically to a real defect.

## Proposed fix

Remove the scheduling branch from `triage_pod` (`health.py:184-196`) so that
pods which are Pending/Unschedulable are owned solely by `pod_pending`, which
is the classifier written for this condition and the one that distinguishes
scarcity from misconfiguration. `health.py` keeps everything else: Evicted,
OOMKilled, image-pull failures, startup-probe, CrashLoopBackOff, tier-3.

Rationale for one owner rather than reconciling both: the two disagree about the
same condition, and `pod_pending`'s judgement is strictly more informed
(`Insufficient` -> wait; node-affinity / PVC-missing / untolerated-taint ->
cancel now). Keeping both and short-circuiting between them preserves a
duplicate classifier that can drift.

### On the quota arm

The branch being removed also matches `quota` / `exceeded`
(`health.py:188-193`). This appears not to fire in practice: ResourceQuota
exhaustion is rejected at **admission**, so no pod is created and there is no
Pending pod with `PodScheduled=False` for either classifier to inspect -- the
event is `FailedCreate` on the ReplicaSet. Both `triage_pod` and
`parse_pod_conditions` require an existing Pending pod, so neither sees it.

If a scheduler-level quota plugin ever does produce such a `FailedScheduling`
message, `classify_pending_reason` currently falls through to `recoverable`
with a warning (verified), so it would wait `--pending-threshold` and then
reclaim rather than cancelling immediately. That is a behaviour change worth
noting but not a regression in correctness.

## Acceptance criteria

- [ ] `triage_pod` no longer returns a result for a pod that is merely
      Pending/Unschedulable; `pod_pending` handles it.
- [ ] A regression test covering the **interaction**: a pod Pending with
      `Insufficient nvidia.com/gpu` does not cause a PipelineRun cancellation
      before `pending_threshold` elapses, and does cause reclamation after it.
- [ ] Existing `pipeline/tests/test_health.py`, `test_pod_pending.py`, and
      `test_deploy_run.py` still pass; any test asserting the removed branch is
      updated with a note explaining why.
- [ ] Non-recoverable pendings (node affinity, missing PVC, untolerated taint)
      still cancel promptly via `pod_pending`.
- [ ] Tier-1 and other tier-2 health findings (Evicted, OOM, ImagePullBackOff)
      are unchanged.

---

## Relationship to prior work, and why this is a reclassification

**Provenance.** #62 (*Early reclaim: detect Pending pods and reclaim slots*)
introduced the pending/reclaim path that became `pod_pending.py` +
`_handle_pending_pods`. #228 (*Integrate pod health monitoring into orchestrator
for remote mode*) later added `_check_pod_health` into the same run loop. The
overlap between the two dates from that point: nothing reconciled the new health
path against the pending path that was already there.

**#387 is the same bug class, and it was fixed with a tolerance budget.**
#387 (*health: startup probe failures escalate to PipelineRun cancellation on
first Unhealthy event*) is the same function, the same tier-2 mechanism, and the
same complaint — escalation on *existence* rather than *persistence*. It was
closed by adding a count budget, still present today:

```python
_OOM_MAX_ATTEMPTS = 2                # health.py:115
_STARTUP_PROBE_MIN_EVENT_COUNT = 10  # health.py:123
```

The scheduling branch is the only tier-2 branch with no such budget, so the
"consistent" change would be to add a third constant.

**That is deliberately NOT what this issue proposes, and the reason is that a
tolerance budget is the wrong shape for this condition.**

A budget encodes "this is eventually fatal; tolerate N first." That premise is
true for OOMKilled (a pod that OOMs repeatedly is genuinely broken). It is
**false** for an unschedulable pod:

- The kube-scheduler keeps an unschedulable pod in its queue and retries it on
  cluster-state changes and on backoff. It is not a failure state and has no
  natural count — it is one condition that persists, not N events.
- Nothing about the deployment is wrong. The pod is correct; the cluster is
  full. Waiting is the correct response for as long as the operator is willing
  to wait, which is exactly what `--pending-threshold` expresses.
- A budget would therefore have to be a *duration* budget, and one already
  exists in `pod_pending` — purpose-built, configurable, and with the
  scarcity/misconfiguration distinction that a count cannot express
  (`Insufficient` -> wait; node-affinity / missing-PVC / untolerated-taint ->
  cancel now).

Adding `_UNSCHEDULABLE_MIN_EVENT_COUNT` would create a second timer for one
condition, in the subsystem that classifies it less precisely, while the
better-specified one sits unreachable in the same loop.

So the defect is a **classification** error, not a tolerance error: `health.py`
treats a capacity condition as a pod-health failure. The fix is to stop
classifying it there and let the owner handle it.

**On #387 itself:** by the same reasoning, a startup probe failing while a large
model loads from a cold PVC is also not a fatal pod-health condition, and
`_STARTUP_PROBE_MIN_EVENT_COUNT = 10` may be a threshold papering over the same
misclassification. That is explicitly **out of scope here** and is not changed by
this issue — noted only so the precedent is understood as prior art with a known
limitation rather than as a rule this issue disregards.

## Related

- #567 (open) — *classify infra-caused PipelineRun Failed and route to retry
  queue*. Overlaps on the consequence: escalation marks the pair `failed`
  (`deploy.py:2937`), so a capacity-only stall is indistinguishable from a real
  defect in `deploy.py status`. This issue removes one source of those
  mislabelled failures; #567 addresses what to do with the ones that remain.
- #62, #228 — provenance, above.
- #387 — precedent, above.

kalantar:
Closes #873. Base: `main`.

## What changed

Removed the `# Tier 2: Scheduling failure` branch from `triage_pod`
(`pipeline/lib/health.py`), so a pod that is merely `Pending`/`Unschedulable` is
no longer a pod-health finding. Ownership of that condition sits solely with
`pipeline/lib/pod_pending.py` + `_handle_pending_pods`, which already implement
the intended behaviour.

Two files, +87/-27. No production code outside `triage_pod` is touched.

## Why the threshold was unreachable

Both subsystems run in the same orchestrator cycle:

```
deploy.py:2889   _handle_pending_pods(..., pending_threshold=600, ...)
                 if reclaimed: ... continue      # only short-circuits on RECLAIM
deploy.py:2909   _handle_timeout(...)
deploy.py:2926   _check_pod_health(...)          # escalates tier-2 immediately
```

While System 1 is *waiting* it returns `False`, which is indistinguishable from
"nothing wrong", so control falls through to `_check_pod_health`, which cancels
on the first check (`deploy.py:895`). `pending_since` was written to an entry
whose PipelineRun had already been deleted, so `elapsed` never grew toward 600 s.

Observed in practice: four consecutive cells cancelled 2-3 minutes after
dispatch during cluster-wide GPU contention (77 of 96 GPUs held by other
tenants, no node with a free 4-GPU block for a TP=4 pod), each with a decode pod
`Pending` on `0/20 nodes are available: 14 Insufficient nvidia.com/gpu`.

## Reviewer attention

**1. This is a reclassification, not a tolerance change — and it deliberately
departs from #387's fix.** #387 was the same bug class in the same function and
was closed by adding `_STARTUP_PROBE_MIN_EVENT_COUNT = 10`, alongside the
existing `_OOM_MAX_ATTEMPTS = 2`. The scheduling branch was the only tier-2
branch without such a budget, so the "consistent" change would have been a third
constant. It isn't, because a count encodes *"eventually fatal, tolerate N"* —
true of a repeatedly-OOMing pod, false of a full cluster. An unschedulable pod
has no natural count (one persistent condition, not N events) and no fatality;
the correct budget is a **duration**, and `pod_pending` already has a
configurable one with a scarcity-vs-misconfiguration distinction a count cannot
express. Adding a budget here would have created a second timer for one
condition in the subsystem that classifies it less precisely. Full argument in
the issue.

**2. `pipeline/README.md:598` already documented the behaviour this restores:**

> Recoverable reasons (e.g. `Insufficient nvidia.com/gpu`) trigger early reclaim
> after `--pending-threshold` seconds. Non-recoverable reasons (e.g. node
> affinity mismatch, PVC not found) fail the pair immediately.

So the documented contract was correct and the code had diverged from it. No
README change was needed — the fix makes it true. Worth verifying you agree with
that reading.

**3. Behaviour change on quota, stated rather than hidden.** The removed branch
also matched `quota`/`exceeded`. `pod_pending`'s `classify_pending_reason` does
not recognise quota messages and falls through to `recoverable` (verified), so
such a pod now waits `--pending-threshold` before reclamation instead of being
cancelled on the first check. I believe this is near-unreachable in practice:
standard `ResourceQuota` exhaustion is rejected at **admission**, so no pod is
created and there is no Pending pod with `PodScheduled=False` for either
classifier to inspect — the event is `FailedCreate` on the ReplicaSet, and both
`triage_pod` and `parse_pod_conditions` require an existing Pending pod. If you
know of a scheduler-level path that does produce such a `FailedScheduling`
message, this is the assumption to challenge.

**4. Incidental robustness improvement.** `pod_pending` keys off the
`PodScheduled=False, reason=Unschedulable` pod **condition**; the removed branch
keyed off a `FailedScheduling` **event**. Conditions persist, events expire
(~1 h TTL), so a long-unschedulable pod whose event has aged out is now still
detected where previously it was not.

## Tests

- `test_triage_ignores_unschedulable_pod_scarcity` / `..._quota` — the two tests
  that asserted the removed branch, **inverted rather than deleted**, so the
  quota behaviour change stays visible in the suite with its rationale.
- `test_triage_still_reports_non_scheduling_pending_failures` — guards against
  the removal blinding triage to other Pending pods (image-pull is matched
  earlier by the reason-based branch).
- `test_check_pod_health_does_not_escalate_unschedulable` — **the coverage gap
  that let this through.** `test_health.py` asserted `triage_pod`'s output and
  `test_deploy_run.py` asserted `_handle_pending_pods`' timer, but nothing
  asserted the interaction: that an unschedulable pod does not trigger the
  escalation that cancels the run. The wait/reclaim half remains covered by the
  existing `_handle_pending_pods` tests in `test_deploy_run.py`, cited rather
  than duplicated.

`python -m pytest pipeline/` → **2057 passed, 2 xfailed**, no failures.
`ruff check pipeline/ --select F` → clean.

## Stale-reference sweep

Grepped `**/*.md`, `docs/`, `.claude/skills/`, `README*` for `FailedScheduling`,
`unschedulable`, the removed suggestion strings, `pending-threshold`, and
`pod health`. Hits: `pipeline/README.md:598` (becomes accurate, see above);
`CLAUDE.md:108` and `pipeline/README.md:881` (generic one-line descriptions of
`health.py`'s role — still accurate); `capacity.py` cordon/taint text and
`sim2real-bootstrap` SKILL.md (unrelated uses of "unschedulable"); two
historical plan docs under `docs/superpowers/` (left as history). Nothing
required updating.

## Not addressed here

By the same reasoning, a startup probe failing while a large model loads from a
cold PVC is arguably also not a fatal pod-health condition, so
`_STARTUP_PROBE_MIN_EVENT_COUNT = 10` may be a threshold papering over the same
misclassification. Explicitly out of scope; noted in the issue so the precedent
is understood as prior art with a known limitation.

### DIFF
--- a/pipeline/lib/health.py
+++ b/pipeline/lib/health.py
@@ -180,28 +180,28 @@ def triage_pod(
             ),
         )
 
-    # Tier 2: Scheduling failure
-    if pod.phase == "Pending":
-        sched = next((e for e in pod_events if e.reason == "FailedScheduling"), None)
-        if sched:
-            msg_lower = sched.message.lower()
-            if "quota" in msg_lower or "exceeded" in msg_lower:
-                return TriageResult(
-                    tier=2, action="suggest", needs_logs=False,
-                    message=f"{pod.name}: Pending (resource quota exceeded)",
-                    suggestion=f"Resource quota exhausted: {sched.message}",
-                )
-            if "insufficient" in msg_lower or "nodes available" in msg_lower:
-                return TriageResult(
-                    tier=2, action="suggest", needs_logs=False,
-                    message=f"{pod.name}: Pending (no nodes match GPU affinity)",
-                    suggestion=(
-                        f"No schedulable nodes: {sched.message}\n"
-                        "Check nodeAffinity in scenario bundle → "
-                        "model.helmValues.decode.extraConfig.affinity"
-                    ),
-                )
-            # unrecognized scheduling message — falls through to None
+    # NO SCHEDULING BRANCH HERE, DELIBERATELY -- see issue #873.
+    #
+    # A pod that is merely Pending/Unschedulable is NOT a pod-health failure.
+    # The kube-scheduler keeps it queued and retries it, so it recovers on its
+    # own the moment capacity appears; nothing about the deployment is wrong.
+    # Classifying it here made it a tier-2 finding, and every tier-2 finding
+    # escalates to a PipelineRun cancellation (deploy.py:895) with no grace
+    # period -- which pre-empted the deliberate --pending-threshold wait in
+    # _handle_pending_pods (deploy.py:713) that runs earlier in the same
+    # orchestrator cycle, making that threshold unreachable.
+    #
+    # Unschedulable pods are owned solely by pipeline/lib/pod_pending.py, which
+    # is purpose-built for the condition: it distinguishes scarcity (Insufficient
+    # -> wait up to the threshold) from misconfiguration (node affinity, missing
+    # PVC, untolerated taint -> reclaim now). It also keys off the PodScheduled=
+    # False/Unschedulable pod CONDITION rather than a FailedScheduling EVENT, so
+    # it still sees a pod whose scheduling event has aged out of the event TTL.
+    #
+    # Do not reinstate a branch here, and do not add a count budget in the shape
+    # of _OOM_MAX_ATTEMPTS / _STARTUP_PROBE_MIN_EVENT_COUNT: a count encodes
+    # "eventually fatal, tolerate N", which is true of OOM but false of a full
+    # cluster. The correct budget is a DURATION, and pod_pending already has one.
 
     # Tier 2: Startup probe timeout
     if pod.phase == "Running" and not pod.ready:
--- a/pipeline/tests/test_health.py
+++ b/pipeline/tests/test_health.py
@@ -217,24 +217,84 @@ def test_triage_image_pull_backoff():
     assert "run_metadata.json" in result.suggestion
 
 
-def test_triage_failed_scheduling_gpu():
+def test_triage_ignores_unschedulable_pod_scarcity():
+    """Unschedulable-for-scarcity is NOT a pod-health finding (issue #873).
+
+    triage_pod used to return tier 2 here, and every tier-2 finding escalates to
+    a PipelineRun cancellation with no grace period, which pre-empted the
+    --pending-threshold wait in deploy.py's _handle_pending_pods. Ownership of
+    Pending/Unschedulable now sits solely with pipeline/lib/pod_pending.py.
+    """
     from pipeline.lib.health import triage_pod, RemediationTracker
     pod = _make_pod(phase="Pending")
     events = [_make_event(reason="FailedScheduling",
                           message="0/5 nodes available: 5 Insufficient nvidia.com/gpu")]
     result = triage_pod(pod, events, RemediationTracker())
-    assert result.tier == 2
-    assert "affinity" in result.suggestion.lower() or "nvidia" in result.suggestion.lower()
+    assert result is None
 
 
-def test_triage_quota_exceeded():
+def test_triage_ignores_unschedulable_pod_quota():
+    """Quota-exceeded scheduling messages are also no longer triaged here.
+
+    Behaviour change recorded deliberately (issue #873): pod_pending's
+    classify_pending_reason does not recognise quota messages and falls through
+    to "recoverable", so such a pod now waits --pending-threshold before the
+    slot is reclaimed instead of being cancelled on the first health check.
+    Note that standard ResourceQuota exhaustion is rejected at admission, so no
+    Pending pod exists for either classifier to inspect in that case.
+    """
     from pipeline.lib.health import triage_pod, RemediationTracker
     pod = _make_pod(phase="Pending")
     events = [_make_event(reason="FailedScheduling",
                           message="exceeded quota: requests.nvidia.com/gpu=4")]
     result = triage_pod(pod, events, RemediationTracker())
+    assert result is None
+
+
+def test_triage_still_reports_non_scheduling_pending_failures():
+    """Removing the scheduling branch must not blind triage to other Pending pods.
+
+    A Pending pod with an image-pull failure is matched by the reason-based
+    branch earlier in triage_pod, so it is unaffected by issue #873.
+    """
+    from pipeline.lib.health import triage_pod, RemediationTracker
+    pod = _make_pod(phase="Pending", reason="ImagePullBackOff")
+    result = triage_pod(pod, [], RemediationTracker())
+    assert result is not None
     assert result.tier == 2
-    assert "quota" in result.suggestion.lower()
+
+
+def test_check_pod_health_does_not_escalate_unschedulable(monkeypatch):
+    """The interaction this issue exists for: no cancellation for scarcity.
+
+    This is the coverage gap that let #873 through -- test_health.py asserted
+    triage_pod's output and test_deploy_run.py asserted _handle_pending_pods'
+    timer, but nothing asserted that an unschedulable pod does not trigger the
+    escalation that cancels the PipelineRun. _check_pod_health returning False
+    is what keeps the run alive long enough for pod_pending's threshold (whose
+    wait/reclaim behaviour is covered by test_deploy_run.py's
+    _handle_pending_pods tests) to govern the outcome.
+    """
+    import pipeline.deploy as deploy
+    from pipeline.lib import health as health_mod
+
+    pod = _make_pod(phase="Pending")
+    events = [_make_event(reason="FailedScheduling",
+                          message="0/20 nodes are available: 14 Insufficient nvidia.com/gpu")]
+
+    monkeypatch.setattr(health_mod, "get_all_pods", lambda ns: [pod])
+    monkeypatch.setattr(health_mod, "get_events", lambda ns: events)
+
+    def _fail_delete(ns, name):  # must not be reached
+        raise AssertionError(f"delete_pod called for {name}; scarcity is not remediable here")
+
+    monkeypatch.setattr(health_mod, "delete_pod", _fail_delete)
+
+    escalate = deploy._check_pod_health(
+        namespace="kalantar-0", pair_key="wl-x|baseline|i1",
+        tracker=health_mod.RemediationTracker(), skip_teardown=False,
+    )
+    assert escalate is False
 
 
 def test_triage_crash_loop_tier3():


kalantar:
## Automated review — needs human judgment

**1 must-fix · 1 important · 0 recommendations** across 4 lenses.

The deletion removes the only classifier that observes GPU-requesting pods, and the duration budget it hands off to runs against a disjoint pod set, so the promised grace period still cannot apply.

### Must fix
- **pod_pending never sees the pods this branch was classifying, so unschedulable model pods are now owned by nobody and hold a slot until the 4-hour timeout.** (`pipeline/deploy.py:736`) — _handle_pending_pods queries only `-l tekton.dev/pipelineRun=<pr_name>` (deploy.py:736) while get_all_pods deliberately excludes every pod carrying tekton.dev/{pipelineRun,taskRun,pipelineTask} (health.py:259-278, and the run-loop call site at deploy.py:2922 is commented "non-Tekton pods"); the two pod sets are complementary by construction. The issue's own field evidence — a Pending *decode* pod triggering health escalation — proves the decode pod is in get_all_pods' set and therefore cannot be in _handle_pending_pods' set (test_get_all_pods_excludes_tekton confirms vllm-decode-0 is the non-Tekton case). So the comment's claim that "Unschedulable pods are owned solely by pipeline/lib/pod_pending.py" is false for exactly the pods that request GPUs: after this change a scarcity-Pending decode pod matches no classifier, --pending-threshold is still unreachable (AC2's "does cause reclamation after it" is unmet), and the slot is held until timeout_hours = 4 (deploy.py:2814) instead of 600 s, with pending_stalls never incrementing so --max-pending-stalls never applies.

### Important
- **AC2's "interaction" test is satisfied only by composing two tests whose pod sources are separately mocked, so nothing can detect that the two classifiers never see the same pod.** (`pipeline/tests/test_health.py:267`) — test_check_pod_health_does_not_escalate_unschedulable monkeypatches health_mod.get_all_pods to return the Pending pod, while test_early_reclaim_recoverable_* (test_deploy_run.py:2163-2260) monkeypatches deploy.run to return a pods_json with no metadata/labels at all — so neither test ever exercises the real `-l tekton.dev/pipelineRun=` selector or the real Tekton-label filter, which is precisely why the disjoint-pod-set gap above survives a green suite. A test that feeds one namespace pod list (decode pod labelled as in test_get_all_pods_excludes_tekton) through both functions across two cycles would fail today.

kalantar:
Merging with the `must_fix` unresolved, as a deliberate call by the repo owner. The finding is confirmed, not disputed — I independently verified both selectors (`deploy.py:736`, `health.py:259-278`). Follow-up filed as #875 with the verification, the 4 h vs 2 min regression table, three candidate fix directions, and the missing interaction test from the `important` finding.

Issue #877, opened 2026-09-01
assemble: --replicas grow re-resolves overlays, so i2..iN can carry a different config than i1


kalantar:
## Problem

`sim2real assemble --replicas N` on a run that already has fewer replicas takes the additive-grow path (`pipeline/lib/assemble_run.py:766`). It preserves `i1..i_prior` byte-for-byte and generates `i_{prior+1}..i_N`.

To generate the new ones it **re-runs the whole resolution pipeline** (`_resolve_packages`, called at `:804-812`). So the new iterations are built from whatever the overlays say *now*, while the preserved iterations keep whatever they were built from earlier. If any resolution input changed in between, `i2..iN` silently carry a different plugin config than `i1`.

Nothing detects this. The drift check that gates the grow path hashes only `slicer.assembly_slice(manifest)` — transfer.yaml's own fields — which does not cover the translation-generated overlays, the baseline scenario, `baselines/defaults/*`, or the workload YAML contents (see #876 for the full table). `_additive_grow`'s docstring already states the hole:

> the caller's drift check hashes only `slicer.assembly_slice(manifest)` — transfer.yaml content — so the translation-generated overlays this re-resolution merges may have changed since the first assemble ... while the manifest hash still matches

The docstring treats this as a reason to re-report scalar-list conflicts. The larger consequence — that the emitted replicas may not be replicas — is not addressed.

## Why it matters

Replicas exist to measure variance under a **fixed** condition. If `i1` and `i2` differ in configuration, the resulting variance figure is not a measurement of anything, and nothing on disk records why. This is a silent misattribution rather than a visible failure.

## Repro

```
sim2real assemble --translation <T> --cluster <C> --run r1              # i1
# edit workspace/translations/<hash>/generated/<algo>/<algo>_config.yaml
sim2real assemble --translation <T> --cluster <C> --run r1 --replicas 3 # i2, i3
```

`params_hash` still matches (the overlay is not in the assembly slice), so the grow path is taken rather than refused. `cluster/pipelinerun-<wl>|<algo>|i1.yaml` embeds the old config; `i2` and `i3` embed the new one. Exit code 0, no warning.

This is reachable today with no new features. It becomes considerably easier to hit if #876 lands, since scoped assemble makes per-workload overlay edits a normal workflow.

## Proposed fix

Do not re-resolve. **Derive each new iteration from the pair's own existing PipelineRun** by copying it and rewriting the iteration identifier.

[...]

PR #878, opened 2026-09-01
fix(assemble): scoped re-assembly, and stop --force from destroying collected results

### LINKED ISSUE #876: assemble: no way to re-assemble a subset, and --force rmtrees the whole run dir including measured results
## Problem

`sim2real assemble` has no way to regenerate part of a run, and its only escape hatch destroys measured data. These are two defects that compose into a trap.

### 1. Assemble is all-or-nothing

`generate_pipelineruns` (`pipeline/lib/assemble_run.py:456`) emits one PipelineRun per `(workload, package, iteration)` tuple, always the full cross product. There is no `--workload` / `--package` scoping, unlike `deploy collect|status|run|wipe`, which all accept it (`pipeline/deploy.py:3632-3656`).

### 2. The only way to re-assemble destroys `results/`

Re-assembling an existing run requires `--force`, and every `--force` branch calls `shutil.rmtree(run_dir)` — seven sites, `assemble_run.py:936` through `:1007`. But `results/` lives **inside** the run dir (`pipeline/lib/resolve.py:239`: `results_dir = run_dir / "results"`), holding the collected `trace_data.csv` per `(phase, workload, iteration)`.

So `--force` does not just discard regenerable YAML — it destroys measured cluster results, which are expensive and unrecoverable without re-running.

`pipeline/tests/test_assemble_run.py:1350` (`test_force_overwrites_existing_run`) currently asserts `not (run_dir / "sentinel").exists()`, so the wipe of unrelated files is a *tested guarantee*. That test encodes the bug and would need to change.

### 3. The no-op message is inaccurate

`params_hash` is SHA-256 over `slicer.assembly_slice(manifest)` (`pipeline/lib/slicer.py:88`) — transfer.yaml's **own fields**. `workloads` is a list of *paths*, and the translation-generated overlays are not in the slice at all. But resolution reads far more than that:

| input | read at | in `params_hash`? |
|---|---|---|
| `baselines/<name>.yaml` | `assemble_run.py:686` | no |
| `baselines/defaults/*.yaml` | `:641` | no |
| `generated/baselines/<n>/baseline_config.yaml` | `:694` | no |
| `generated/<algo>/<algo>_config.yaml` | `:720` | no |
| `treatment.yaml` diffs | `:716` | no |
| `workloads/*.yaml` contents | `:738` | no |
| `translation_output.json` image_refs | `:733` | no |

So editing an overlay or a workload YAML leaves the hash matching, and assemble takes the true-no-op path (`:988-996`) and reports:

```
run '<R>' is up to date — inputs and translation are unchanged since <ts>
```

That statement is false: the inputs did change. `_additive_grow`'s own docstring already acknowledges the hole ("the drift check hashes only `slicer.assembly_slice(manifest)` — transfer.yaml content — so the translation-generated overlays this re-resolution merges may have changed ... while the manifest hash still matches"), and `pipeline/deploy.py:298-310` works around the no-op path when backfilling `scenario`.

### How they compose

Defect 3 sends the operator to `--force`, and defect 2 makes `--force` eat their results. Defect 1 means there is no narrower option. The observed workflow that motivated this: a plugin config key that must differ per workload (`activeWorkload` selecting one SLO-target triple) cannot be varied without either a silent no-op or wiping the whole grid.

## Proposed change

**Add `--workload` / `--package` to `assemble`,** matching `deploy`'s flag names and its `_report_filter_mismatch` habit of printing valid values on a miss (`deploy.py:2396-2430`). Scope must be derived from the manifest cross product, **not** from `_resolve_scope` (`deploy.py:2376`), which resolves pairs from the ConfigMap-backed progress store and therefore needs a cluster namespace — unavailable and irrelevant at assemble time.

**Split the current overloaded `--force` into two orthogonal axes**, keeping its existing meaning:

| axis | question | flag |
|---|---|---|
| regenerate | this PipelineRun already exists — redo it? | `--force` (unchanged meaning) |
| results | results exist for a pair being regenerated — keep them? | `--no-wipe` (new) |

Giving four cases, evaluated **per pair in scope**:

| PipelineRun exists | results exist | behavior |
|---|---|---|
| no | no | proceed |
| no | yes | proceed; wipe results unless `--no-wipe` |
| yes | no | nothing unless `--force` |
| yes | yes | nothing unless `--force`; wipe results unless `--no-wipe` |

Both predicates are `Path.exists()` — on `cluster/pipelinerun-<wl>|<pkg>|iN.yaml` and `results/<pkg>/<wl>/iN/` respectively. **No new state, no content hashing, no schema change to `run_metadata.json`.**

**Replace the no-op message** with one that does not claim inputs are unchanged, e.g. `N pair(s) already assembled; pass --force to re-assemble`. This is what actually fixes defect 3 — the inaccuracy was in the wording, not in the absence of a digest.

### Row 2 deserves care

Row 2 (no PipelineRun, results present) wipes without the operator having typed any flag, since `--force` was not required. Measured data should not die silently in an unexplained state. `deploy wipe` already has the right pattern — enumerate targets, prompt, `--yes` to skip, `EOFError` → abort (`deploy.py:3232-3243`) — and it should be reused here, honoring `--no-wipe`.

## Out of scope

- **Per-workload config as a first-class feature.** Scoped assemble makes a run dir *able* to hold divergent per-workload configs; it does not add a config axis to transfer.yaml. Tracked separately if wanted.
- **Detecting "operator edited an overlay and never ran assemble."** No assemble invocation exists to catch this; it would need a deploy-time staleness check.

## Tests

- results survive a scoped re-assemble with `--force` (fails today)
- `--no-wipe` preserves results while regenerating
- scoped assemble leaves out-of-scope PipelineRuns byte-identical and mtime-unchanged
- `--workload` / `--package` mismatch prints valid values and exits non-zero
- existing `test_force_overwrites_existing_run:1350` rewritten: `--force` no longer destroys unrelated run-dir contents

kalantar:
Closes #876

## What changed

`--force` no longer `shutil.rmtree`s the run directory. It did so on all six branches of the existing-run decision tree, and `results/` lives inside that directory — so the documented escape hatch for "re-assemble this run" also destroyed measured cluster data that cannot be recovered without re-running.

Re-assembly is now decided per `(workload, package, iteration)` triple, over a scope selected by two new flags, along two orthogonal axes:

| axis | question | flag |
|---|---|---|
| regenerate | this PipelineRun already exists — redo it? | `--force` (unchanged meaning) |
| results | results exist for a pair being regenerated — keep them? | `--no-wipe` (new) |

| PipelineRun exists | results exist | behavior |
|---|---|---|
| no | no | generate |
| no | yes | generate; wipe results unless `--no-wipe` — **prompts first** |
| yes | no | nothing unless `--force` |
| yes | yes | nothing unless `--force`; when `--force`, wipe unless `--no-wipe` |

Both predicates are plain `Path.exists()`. No new state, no content hashing, no `run_metadata.json` schema change.

New flags on `assemble`: `--workload`, `--package`, `--no-wipe`, `-y/--yes`.

The no-op message is replaced. It used to say *"manifest, replicas, and translation are unchanged since &lt;ts&gt;"*. `params_hash` covers only `transfer.yaml`'s own fields — not the overlays or workload YAMLs that resolution reads — and the translation is never compared at all, so the wording asserted more than the code checks. It now reports the pair count and the flags that would change it.

## Reviewer attention

**Three design decisions the issue left open**, resolved with the issue author:

1. **A scoped assemble rewrites `cluster/<pkg>.yaml`, and leaves `manifest.assembly.yaml` and `params_hash` untouched.** The scenario YAMLs are advisory copies (the authoritative scenario is inlined into each PipelineRun) read by `deploy` for GPU-config derivation and by `/sim2real-check` for parity, so refreshing them is safe. Leaving `params_hash` alone is the load-bearing half: it keeps meaning *"the last state in which the whole run was consistent"*, so a later unscoped assemble still detects drift instead of reporting nothing to do. The alternative — advancing it from a partial invocation — would create a fresh instance of the same class of bug this issue is about.

2. **No `--iteration`, no `--only`.** The iteration count is fixed across pairs, and `--only` is slated for deprecation.

3. **No special handling for workload names containing `_`.** The `_` → `-` substitution is slated for deprecation. The two predicates mirror their producers faithfully (`cluster/pipelinerun-*` substituted, `results/<pkg>/<wl>/` raw) and `pipelinerun_filename()` is now the single source of truth for the substitution. Filter values are compared normalized, so `--workload wl_a` and `--workload wl-a` both work rather than the operator having to know which producer they're naming.

**Two additions beyond the issue text**, both deliberate:

- **`pipeline/lib/scope.py`** (a pure move) — `_parse_list`, `_is_glob` and `_expand_glob_values` lifted out of `deploy.py` so assemble's `--workload`/`--package` grammar *is* deploy's rather than a lookalike. `deploy.py` imports them under their existing private names, so no call site changed. The glob tests moved to `test_scope.py` alongside the functions.

- **`prune_orphan_cluster_files`** — required by removing `rmtree`, not optional polish. `deploy` discovers pairs by globbing `cluster/pipelinerun-*.yaml`, so without pruning, a `transfer.yaml` edit that drops a workload would leave that pair's PipelineRun on disk and `deploy` would still execute it. Unscoped assembles prune and warn; the pruned pairs' `results/` are left in place.

**Prompt scope.** Only row 2 prompts — the case that regenerates without the operator having typed any flag. `--force` stays non-interactive because discarding results is its long-documented behavior and prompting would break scripted use; `--force --no-wipe` regenerates YAML while keeping data. Worth a second opinion on: this means `--force` alone still silently discards the results of pairs it regenerates. That is the issue's specified design (row 4 has no prompt), and it is now scoped to the pairs being redone rather than the whole run dir.

**Refusals a scoped assemble adds**, each pointing at the unscoped path: no such run; `--replicas` given (it reuses the recorded count, so narrowing scope can't trip the shrink guard); `transfer.yaml` drift, even with `--force`; and missing / corrupt / legacy run metadata.

## Tests

`pipeline/tests/test_assemble_scope.py` is new: scope resolution, all four table rows (each × `--no-wipe`), the prompt's accept / decline / EOF / `--yes` / `--no-wipe` paths, scoped-mode refusals, byte- and mtime-identity of out-of-scope PipelineRuns, `replicas=None` defaulting, and pruning.

Four existing tests asserted the destructive behavior and are rewritten, with the reason recorded in each docstring:

- `test_assemble_run.py::test_force_overwrites_existing_run` → `test_force_rebuilds_without_destroying_unrelated_run_dir_contents`
- `test_assemble_replicas.py::test_reassemble_at_same_replica_count_with_force_rebuilds` and `test_grow_with_force_rebuilds_instead_of_additive_grow` — both proved the rebuild via a destroyed sentinel; they now prove it via PipelineRun mtime and assert the sentinel *survives*
- `test_sim2real.py::test_force_overwrites` → `test_force_rebuilds_and_keeps_unrelated_run_dir_contents`

`test_drift_with_force_rmtree_rebuilds` is renamed (body still valid; only the name referenced the removed call), and the CLI no-op test is rewritten against the new wording.

CI-equivalent run: `ruff check pipeline/ .claude/skills/ --select F` clean, 2794 passed / 9 skipped / 2 xfailed, coverage 97.76%. The new code in `assemble_run.py` and `scope.py` is at 100% — the 11 remaining uncovered lines in `assemble_run.py` are pre-existing paths this change does not touch.

## Docs swept

`pipeline/README.md` re-assembly section rewritten (four-case table, prompt rationale, scoped rules, pruning, updated failure modes); `CLAUDE.md` assemble paragraph updated; `scope.py` added to both library-module tables. Swept `*.md`, `docs/`, `.claude/skills/` and `README*` for the changed symbols and behavior — the only other live hit was a now-vacuous `"No change needed" not in out` assertion, retargeted. No skill prompt or doc referenced assemble's `--force` semantics.

## Out of scope

Per the issue: per-workload config as a `transfer.yaml` axis; deploy-time staleness detection for un-assembled overlay edits; and #877's replica-grow overlay divergence. Also not done here: deprecating the workload `_` → `-` substitution and `deploy --only`, both noted during vetting.

Base: `main`


kalantar:

The implementation, tests, and docs hold up under verification — no blocking issues survived.

### Recommendations
- **The "pass --force to rebuild" messages on the metadata-repair path could mention --no-wipe.** (`pipeline/lib/assemble_run.py:1222`) — The three repair-path guards at assemble_run.py:1219-1224, 1235-1240 and 1253-1259 tell the operator to "pass --force to rebuild" a run with missing/corrupt/legacy metadata; --force then makes every pair regenerate, so step 6 wipes all collected results non-interactively. Adding "(add --no-wipe to keep collected results)" to those three messages closes the last door to an unannounced results wipe.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

The implementation holds up under line-by-line tracing and the full suite is green, but the headline safety invariant is only proven by composition — one missing test stands between this and merge.

### Important
- **No test covers the issue's first listed acceptance test: out-of-scope collected results surviving a scoped `--force` re-assemble.** (`pipeline/tests/test_assemble_scope.py:486`) — Read all of `TestScopedAssemble` (pipeline/tests/test_assemble_scope.py:486-621) — it asserts out-of-scope PipelineRun YAMLs and `cluster/<pkg>.yaml` are byte/mtime-identical, but never seeds `results/<other-pkg>/...` and asserts survival; `_seed_results` is only used in the unscoped `TestForceNoLongerDestroysResults` / `TestRowTwoPrompt` classes. I confirmed the behavior itself is correct by writing an ad-hoc test (seed `results/sr/wl_a/i1/` + `results/baseline/wl_a/i1/`, then `--force --package baseline`: `sr` survives, `baseline` is wiped — passes), so this is a test gap rather than a bug, but it is the one invariant the whole change exists to establish and issue #876 lists it explicitly under Tests.

kalantar:

None of the raised findings survive verification as blocking — the destructive paths are ordered build-then-delete, scoped correctly, reported to the operator, and covered by tests that pass locally.

### Recommendations
- **prune_orphan_cluster_files' docstring promises a provenance guarantee it implements by file extension.** (`pipeline/lib/assemble_run.py:606`) — Lines 602-607 say "Anything else an operator has left in cluster/ is theirs and is never touched," but the predicate at 629-634 keeps only `pipelinerun-*` in `expected_pipelineruns` and `*.yaml` whose stem is a current package name — any other `.yaml` is unlinked. In practice cluster/ is wholly assemble-generated (deploy.py only reads it) and every removal is warned at sim2real.py:2612, so this is wording, not behavior: narrowing the sentence to "any non-YAML file" would match the code.

Issue #851, opened 2026-08-25
assemble: scalar-list merge discards every layer but the last, so multi-layer CLI-flag lists lose values silently


kalantar:
## Summary

`pipeline/lib/values.py:_merge_lists` replaces scalar lists wholesale (Tier 1), so a
CLI-flag list set by more than one layer keeps only the last writer's copy. Every
flag the earlier layer stated is discarded silently. Supersedes #849, which described
the symptom in `bootstrap`-emitted `vllm.additionalFlags` and proposed moving those
values to typed keys instead; see "Why fix the merge and not the writers" for why the
merge layer is the better place.

## The bug, concretely

`resolve_baseline` merges `framework defaults -> baselines/<name>.yaml -> registered
overlay`. `*.vllm.additionalFlags` is a list of strings, so each layer that sets it
throws away the previous one's entire list.

Today there are exactly two writers: `baselines/baseline.yaml` (written by bootstrap
from `config.md`) and the translate-generated baseline overlay. No defaults fragment
writes it any more -- that was fixed in #842.

Observed in a real bundle
(`workspace/translations/<hash>/generated/baselines/baseline/baseline_config.yaml`):

```yaml
additionalFlags:
- "--max-num-seqs=256"             # restating the baseline's
- "--max-num-batched-tokens=2048"  # restating the baseline's
- "--enable-chunked-prefill"       # restating the baseline's
- "--enable-prefix-caching"        # restating the baseline's
- "--enable-force-include-usage"   # the overlay's OWN flag
```

The four `config.md`-derived flags reach the cluster **only** because the overlay
happens to repeat them verbatim. Regenerate the overlay without one and that
`config.md` decision vanishes with nothing failing. For `--enable-prefix-caching`
the framework default is `false`, so the run would quietly have prefix caching OFF
while `config.md` says ON.

## Why fix the merge and not the writers

The restating above is not sloppiness in the translate skill -- it is **compelled** by
the merge rule. The overlay has one genuine flag of its own to contribute
(`--enable-force-include-usage`), and under replace semantics contributing one flag
forces it to restate all four. Any writer that needs to add a single flag must first
reproduce every flag every other layer set, which is exactly the coupling that makes
the loss invisible.

#849's proposal -- move the values to the framework's typed `vllmCommon.flags.*` keys,
which are dict-merged -- also runs into an availability problem. Of the four affected
rows, checked against the pinned llm-d-benchmark (`76473d0`):

| `config.md` row | typed key |
[...]

PR #881, opened 2026-09-02
fix(assemble): grow replicas from each pair's own i1, not by re-resolving overlays

### LINKED ISSUE #877: assemble: --replicas grow re-resolves overlays, so i2..iN can carry a different config than i1

kalantar:
Closes #877

## What changed

`--replicas N` on a run with fewer replicas took the additive-grow path, which re-ran the whole resolution pipeline to build the new iterations while preserving `i1..i_prior` byte-for-byte. The new iterations therefore came from whatever the overlays said *now*, the preserved ones from whenever they were built. The drift check gating that path hashes only `slicer.assembly_slice(manifest)`, so an edited overlay left the hash matching: `i2..iN` silently carried a different plugin config than `i1`, exit code 0, no warning.

`grow_pair_iterations()` now derives each new iteration from that pair's own `i1`. Exactly two fields carry the iteration — the `-i<N>` suffix on `metadata.name` and the `replica` param — and `resultsDir` threads `i$(params.replica)`, so it follows with no edit of its own.

Replicas are now replicas **by construction**, so there is no drift to detect. And legitimate per-pair differences are preserved for free: each pair grows from itself, which matters now that #876 makes per-workload overlay edits a normal workflow.

## Reviewer attention

**The rewrite is structural, not textual — deliberately.** The same document holds three unrelated `replicas:` keys inside the inlined scenario (vLLM and deployment pod counts), and a run name may itself contain `-i<digits>`. I verified both against a real generated PipelineRun before writing the code:

```
4:   name: causalsloexternality-deep-research-single-turn-try3-i1
555: - name: replica
556:   value: '1'
```

`test_unrelated_replicas_keys_are_untouched` and `test_run_name_containing_i_digits_is_not_confused` pin the two traps.

**Beyond the issue:** `metadata.name` is re-validated against the 253-char DNS-subdomain limit on grow. `i9` → `i10` adds a character, so a name that fit can stop fitting — the issue does not mention this and nothing else would have caught it.

**Four decisions confirmed with the issue author before implementing:**

1. Structural rewrite, per above.
2. **A pair with other iterations but no `i1` is refused**, naming it. Falling back to resolution for that pair would silently reintroduce the exact divergence this removes.
3. **Pre-existing divergence is not repaired.** New iterations are made to match `i1`; an already-mismatched `i2` from an earlier grow stays mismatched. Stated in the docstring.
4. **`generate_pipelineruns` is deleted.** #876 split it into `build_pipelineruns` + `write_pipelineruns`, leaving `_additive_grow` its only production caller; removing that call orphaned it.

**Two behaviors this removes — worth a second opinion:**

- **Scalar-list conflicts (#851) are no longer reported on the grow path.** That reporting existed *because* grow re-merged overlays; with the merge gone there is nothing to conflict. The information is not lost — the next full assemble still reports it — and the rewritten test asserts both halves, so a future change that stops reporting it anywhere still fails.
- **The input validation that came free with `_resolve_packages` is gone.** A deleted baseline scenario file no longer fails a grow, which I think is correct: the grown iterations were derived from an `i1` built while it existed, so nothing about them is wrong. The counterpart guarantee still holds and is now pinned by `test_grow_still_refuses_a_deleted_translation_dir` — step-1 validation runs ahead of the decision tree. Note also that `sim2real assemble` itself still refuses a corrupt `translation_output.json`, because `_cmd_assemble` reads it for the unbuilt-image check before calling the library; only the library contract loosened.

## Tests

`pipeline/tests/test_assemble_grow.py` is new (31 tests): the retarget helper in isolation including every refusal, filename parsing including the junk it must ignore, and the integration behaviors — grow after an overlay edit produces `iN` equal to a retargeted `i1`, each pair grows from its own `i1`, `i1..i_prior` stay byte- and mtime-identical, `replica` tracks `N`, missing-`i1` refuses and writes nothing.

Three tests in `test_assemble_replicas.py` encoded the old behavior and are rewritten with the reason in each docstring; the four `generate_pipelineruns` call sites in `test_assemble_run.py` retarget onto the two-phase pair rather than being dropped.

Rewriting those tests briefly cost coverage of two resolution error paths that had only ever been exercised *through* the grow path. I moved them to the resolution path where they still apply, which brings `assemble_run.py` back to the same 11 uncovered lines as `main` — all pre-existing paths this change does not touch. New code is at 100%.

CI-equivalent: `ruff check pipeline/ .claude/skills/ --select F` clean, 2851 passed / 9 skipped / 2 xfailed, coverage 97.79%.

## Docs swept

`pipeline/README.md`'s additive-grow bullet now documents copy-from-`i1`, the two rewritten fields, and all five consequences; `CLAUDE.md`'s assemble paragraph likewise. Swept `pipeline/README.md`, `CLAUDE.md`, `docs/` and `.claude/skills/` for `generate_pipelineruns` and the grow semantics — the only remaining hits are in a July design spec, left as the point-in-time record it is. Also corrected a now-false comment at the surviving `_resolve_packages` call site, which claimed the grow path shared it.

## Unchanged, confirmed rather than altered

Nothing derives a seed from `iteration` — `trace_shuffle_seed` comes from the workload's declared `sample.shuffle_seed` (default 42) at `tekton.py:178`. Replicas already share a seed and issue identical request streams, measuring scheduler/cluster nondeterminism only. Copy-and-rewrite preserves that exactly.

Base: `main`


kalantar:

The copy-from-i1 rewrite is correct and well tested, but it gives up the build-before-write ordering the rest of this file relies on, so a per-pair refusal can leave a half-grown run behind.

### Important
- **grow_pair_iterations writes each grown PipelineRun as it goes, so a refusal on a later pair leaves earlier pairs' new iterations on disk while run_metadata.json still records the old replica count.** — pipeline/lib/assemble_run.py:949-965 interleaves _load_yaml + _retarget_pipelinerun_iteration (six reachable AssembleError raise sites: parse/mapping errors at 165-179, and name-suffix / 253-char / missing-params / missing-replica-param refusals at 841-893) with _write_text_or_raise inside the same loop, and _additive_grow only rewrites manifest.assembly.yaml and run_metadata.json after the whole loop returns (1256-1272). Pairs sort alphabetically, so a corrupt or over-long 'sr' i1 aborts after 'baseline' i2..iN are already written; deploy.py:1944 discovers iterations by globbing cluster/pipelinerun-*.yaml, so the run then has iterations metadata does not claim. The pre-PR grow path did not have this hazard — it went through generate_pipelineruns, i.e. build-all-then-write-all — and the surviving sibling pair build_pipelineruns/write_pipelineruns (724-812) is split with a docstring naming exactly this bug class. Only the missing-i1 pre-check is covered pre-write (test_refusal_writes_nothing, pipeline/tests/test_assemble_grow.py:304); no test covers a mid-batch retarget refusal. Mirror the sibling split: build every (path, yaml-text) pair first, write only after all succeed.

kalantar:

**0 must-fix · 0 important · 2 recommendations** across 4 lenses.

The fix is correct, tightly scoped, and well covered by tests; nothing blocks merge.

### Recommendations
- **`_additive_grow`'s return annotation still says `-> list[str]` though it now returns a `GrowPlan`.** (`pipeline/lib/assemble_run.py:1259`) — Verified at assemble_run.py:1259 — the signature reads `) -> list[str]:` while the body ends `return plan` (a GrowPlan NamedTuple) and the docstring already says "Returns the :class:`GrowPlan`."; CI runs only `ruff --select F`, so nothing catches the mismatch.
- **No test distinguishes "always source i1" from "source the highest existing iteration".** (`pipeline/tests/test_assemble_grow.py:253`) — In every multi-stage grow test (e.g. test_prior_iterations_stay_byte_and_mtime_identical, replicas 2 then 4) i1 and i2 are already identical apart from name/replica, so swapping `_GROW_SOURCE_ITERATION = 1` for max-existing would pass the suite unchanged; the "pre-existing divergence is NOT repaired" contract stated in grow_pair_iterations' docstring is therefore unpinned. Mutating i2 before growing to 4 and asserting i3/i4 match i1 would close it.

PR #884, opened 2026-09-03
docs(sim2real-check): allow the report to be persisted to check/<timestamp>/

### LINKED ISSUE #882: sim2real-check: read-only constraint prevents persisting the report; add a durable review/ write target
## Problem

`/sim2real-check`'s operating constraint forbids the thing the check most needs to do with its output — persist it.

`.claude/skills/sim2real-check/SKILL.md`, "Operating Constraints":

> **STRICTLY READ-ONLY**: Do **not** modify any files. Output a structured report.

So the report exists only in conversation scrollback. For a 9-cell run the three section reports run to ~50 KB each; Section 1 alone was 691 lines / 287 table rows. That is not reviewable in a terminal, does not survive the session, and cannot be diffed against the next run — which matters because several of the check's own sub-sections are explicitly cross-run comparisons (§5c.6's "delta vs prior run" trend diagnostic, §5h's "persistent hotspots ... across >= 2 distinct (phase, workload) pairs").

Two secondary problems compound it:

1. **Agent results are truncated in transit.** The skill's Parallelization section directs the work to three subagents. On a real 9-cell run, each agent's final report exceeds the inter-agent message limit and is silently cut mid-sentence. In this run Section 1 was lost once that way and had to be re-requested.
2. **Subagents cannot use the `Write` tool for report files.** A harness guard rejects it: *"Subagents should return findings as text, not write report files."* The working path is a Bash heredoc, which is also what the auto-mode session guidance prescribes for file changes — but the skill says nothing about this, so each agent rediscovers it.

## Proposed change

Replace the Operating Constraints block:

```markdown
## Operating Constraints

**READ-ONLY except the report.** Do not modify any bundle input, config,
translation output, or result artifact. The one permitted write is the report
itself, to `<run_dir>/review/` (resolve-mode) or `<REAL>/review/`
(legacy-mode), created if absent. Never write under `results/`, and never
modify a file the check reads as evidence.

Subagents must write their section via a Bash heredoc — the harness guard
blocks the `Write` tool for subagent report files. Append in chunks
(`cat >> ...`) for large sections and verify the file on disk afterwards
(line count, first line, last line complete).
```

### Why `review/` specifically

- **Sibling of `results/`, not under it.** The run enumerator and legacy-mode phase discovery both scan for phase directories; anything written under `results/` risks being read as a phase.
- **The name is already reserved.** `review` is in the skill's own legacy-mode non-phase denylist: `generated|workloads|results_charts|snapshots|review|cluster|plans`. So it is consistent with existing convention rather than a new one.
- **Committable.** In a typical experiment repo `.gitignore`, `workspace/runs/` excludes only `run_metadata.json` and `manifest.assembly.yaml`, so a report under `workspace/runs/<run>/review/` is version-controllable — which is what makes the cross-run diagnostics usable.

Suggested filenames, one per parallel agent so there is no write contention:

```
<run_dir>/review/section1-workload-parity.md
<run_dir>/review/section234-config-signal-policy.md
<run_dir>/review/section5-runtime-pd.md
<run_dir>/review/parity-report.md          # assembled summary + rollup
```

### Also update "Output Format"

That section currently describes only the markdown shape. It should state where the report is written and that the assembled `parity-report.md` carries the Summary table, the Action Items, and the per-iteration verdict rollup.

## Acceptance criteria

- [ ] Operating Constraints permits exactly one write target and names it; everything else stays read-only, with `results/` called out as off-limits.
- [ ] The skill tells subagents to use a Bash heredoc, not `Write`, and to verify the file afterwards.
- [ ] Output Format states the report path and the per-agent filenames.
- [ ] Resolve-mode and legacy-mode paths are both specified (`<run_dir>/review/` and `<REAL>/review/`).
- [ ] A run of `/sim2real-check` leaves a durable, diffable report and modifies nothing else.

## Provenance

Found while running `/sim2real-check` against `pd-infocomm-3` run `try3` (translation `cee8bb60cbc3`, 9 PRESENT cells, 3 arms x 3 workloads). Worked around in-session by creating `workspace/runs/try3/review/` and instructing the agents that the single report file is an authorized exception. The workaround needs to be the documented behaviour.

kalantar:
Closes #882

## What changed

`/sim2real-check`'s Operating Constraints said **"STRICTLY READ-ONLY: Do not modify any files"**, which forbade the one write the check most needs — saving its own report. It therefore lived only in conversation scrollback: ~50 KB per section on a 9-cell run (Section 1 alone was 691 lines / 287 table rows), gone when the session ended, impossible to diff.

The last part is the sharp end. Two of the skill's own subsections are cross-run comparisons — §5c.6's "delta vs prior run" and §5h's persistent hotspots across ≥2 pairs — so the skill carried instructions it was **structurally unable to follow**.

Three edits to `.claude/skills/sim2real-check/SKILL.md`, plus one line in `CLAUDE.md`:

1. **Operating Constraints** — one permitted write target, named; everything else read-only, with `results/` and any file read as evidence called out explicitly.
2. **Step 0 phase-discovery denylist** — `check` added.
3. **Output Format + Parallelization** — where the report goes, which agent owns which file, and how agents write it.

## Reviewer attention

**The constraint is made precise, not loosened.** The original rule exists for a real reason: a parity checker that edits the artifacts it judges invalidates its own evidence. The new wording preserves that and adds two prohibitions the old blanket rule only implied — never under `results/`, never a file read as evidence.

**The denylist edit is mandatory, and the reason is non-obvious.** Legacy mode has a fallback at `SKILL.md:104`:

```bash
if [ -d "$REAL/results" ]; then RESULTS_DIR="$REAL/results"; else RESULTS_DIR="$REAL"; fi
```

In a flat legacy bundle there is no `results/`, so `RESULTS_DIR` becomes `$REAL` — and the phase scan then covers exactly the level the report directory sits on. Without `check` on the denylist, a bundle checked once grows a bogus `check` **phase** on the next run. That is why `review` was already on that list, and it's why the directory name and the denylist have to move together; the code comment now says so, so a future rename doesn't silently reintroduce the bug.

Resolve-mode is immune either way, which I verified rather than assumed: `enumerate_iterations.py:115` builds `phases` from `baseline_names + algo_names` (declared names), and its only `iterdir()` is at `:239`, inside a workload directory well below phase level. So this could never have bitten `--run` mode.

**Two deviations from the issue's proposal, both agreed with the issue author:**

- **`check/` rather than `review/`.** The issue picked `review` because it was already denylisted; `check/` is more obviously this skill's own output, at the cost of the one-word denylist edit above.
- **A fresh `check/<UTC timestamp>/` per invocation rather than a fixed path.** The issue proposed fixed filenames, which means a re-check overwrites the verdict the cross-run diagnostics exist to compare against. Timestamped directories never clobber, keep each invocation's four files together, make "latest" the last sorted entry, and make comparing two invocations a directory diff.

**Also fixed, because they're downstream of the same cause:**

- Subagent section reports exceed the inter-agent message limit on a real run and are **silently truncated mid-sentence** — this already cost Section 1 once. Agents now write to files and return only a path plus a verdict rollup, so nothing large is in transit to truncate. Fixing persistence is what dissolves this.
- A harness guard rejects the `Write` tool for subagent report files, so the working path is a Bash heredoc — which the skill never said, leaving each agent to rediscover it. Now documented, with `cat >` for the first chunk and `cat >>` only after: **starting with `>>` makes a retry after a partial write silently append a second copy of the section.** The issue's "verify afterwards" catches that after the fact; truncate-first prevents it.

**One claim I could not verify from the repository:** the harness guard's exact text (*"Subagents should return findings as text, not write report files"*, quoted in the issue as encountered at runtime). It is not in the tree — it's harness behavior. The heredoc instruction is correct whether or not the guard fires, so nothing depends on it.

**Scope.** `sim2real-check` was the only read-only skill of the four — `bootstrap`, `translate` and `analyze` all write, and `analyze` has `Write` in its `allowed-tools` frontmatter. So the blanket constraint was a local inconsistency rather than a cross-skill policy, and this change is correctly confined to one skill.

## Verification

Every acceptance criterion in the issue is mechanically checkable against the file, and all five pass:

| Criterion | Where |
|---|---|
| Exactly one write target, named; `results/` off-limits | `SKILL.md:353`, `:364` |
| Heredoc not `Write`; verify afterwards | `:379-386` |
| Output Format states path + per-agent filenames | `:876-895` |
| Both modes specified | `:357-358` |
| A run leaves a durable report and modifies nothing else | the constraint text; `git status` after a real run |

`ruff check pipeline/ .claude/skills/ --select F` clean. `.claude/skills/sim2real-check/tests/` — 18 passed. No test changes needed: the suite covers `enumerate_iterations.py`, and this PR changes prose plus one shell `case` arm.

## Note for whoever lands #487

#487 is a separate quality-review pass on this same file (silent-fallback predicates, per-workload completeness visibility). It will edit `SKILL.md` too — whichever lands second gets a conflict. Not a duplicate; the concerns don't overlap.

Base: `main`


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 1 important · 2 recommendations** across 4 lenses.

The change itself is sound and correctly scoped, but the new persistent artifact is missing from the CLAUDE.md table that the same file designates as the authoritative record of what each stage writes.

### Important
- **New `runs/<run>/check/<timestamp>/` artifact is not added to the Workspace Artifacts table.** (`CLAUDE.md:154`) — CLAUDE.md:211 states "The workspace artifact table above is the definitive map of what each stage reads and writes. Use it to determine the blast radius of any change," and rows 137-154 give every other file under `runs/<run>/` its own row (run_metadata.json, manifest.assembly.yaml, cluster/*.yaml, results/**, plans/**). The PR updates only the prose line at :82, so anyone tracing writers from the table alone will not see that `/sim2real-check` now owns a directory name under the run dir — which matters precisely because that name is coupled to the Step 0 phase-discovery denylist.

### Recommendations
- **Resolve-mode `$REPORT_DIR` re-derives the run directory instead of reading `.run_dir` from the resolve JSON.** (`.claude/skills/sim2real-check/SKILL.md:362`) — `pipeline/lib/resolve.py:134` already emits a top-level `"run_dir"` field, and Step 0 pulls every other path out of that same JSON with `jq -re` so schema drift fails loudly (SKILL.md:58-83). Operating Constraints instead tells the reader `<run dir>` is `<experiment-root>/workspace/runs/<RUN>/`, a second hand-maintained copy of `layout.runs_dir()`; capturing `RUN_DIR=$(jq -re '.run_dir' "$RESOLVED_JSON")` would match the file's own pattern.
- **No mechanical guard keeps `check` in the Step 0 phase-discovery denylist.** (`.claude/skills/sim2real-check/SKILL.md:121`) — The directory name in Operating Constraints and the `case` arm at SKILL.md:121 must move together — the diff's own comment says so — and the repo already uses bounded-region regex tests over SKILL.md for exactly this class of invariant (`.claude/skills/sim2real-bootstrap/tests/test_build_commands_guidance.py`), with `.claude/skills/sim2real-check/tests/` already wired into CI.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS DISCUSSION, `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 1 important · 2 recommendations** across 4 lenses.

One retracted sentence survived in CLAUDE.md and now contradicts the corrected SKILL.md text; everything else the specialists raised is either accurate as written or too thin to hold up.

### Important
- **CLAUDE.md keeps the cross-run rationale that the same PR removed from SKILL.md as false.** (`CLAUDE.md:82`) — CLAUDE.md:82 says the timestamped directory exists "so the cross-run diagnostics (§5c.6 delta-vs-prior-run, §5h persistent hotspots) have something to compare against", while SKILL.md:381-383 in the same PR states §5c.6 and §5h "are both computed entirely from the current run's own evidence ... and neither reads a previous report"; commit bd20c81's own message says both files carried this claim and that it is wrong, but the diff only corrected SKILL.md, leaving the retracted rationale live in the table CLAUDE.md itself calls the definitive map.

### Recommendations
- **The "entirely from the current run's own evidence" phrasing is broader than §5c.6 actually is.** (`.claude/skills/sim2real-check/SKILL.md:381`) — SKILL.md:751's secondary trend diagnostic says "Report as `delta vs prior run` if a prior run is available", so the within-run framing at :381 is only true of §5c.6's gating criteria; the load-bearing half ("neither reads a previous report") is correct, but a one-clause carve-out would stop the two passages reading as contradictory.
- **`run_dir` is missing from the documented `resolve --run` field list this PR now depends on.** (`CLAUDE.md:86`) — Step 0 now hard-fails when `.run_dir` is absent, and `pipeline/lib/resolve.py` emits it, but the `resolve --run` bullet's "fields include" list does not name it — a one-word addition in a file this PR already edits.

kalantar:

Documentation-only change whose factual claims check out against the code, with the one coupling invariant that could silently corrupt phase discovery actually guarded by a passing test.

### Recommendations
- **The pre-assembly gate names truncation as a re-run trigger but only enumerates exists-and-non-empty as the test.** (`.claude/skills/sim2real-check/SKILL.md:970`) — SKILL.md:969-972 says "confirm all three section files exist and are non-empty; if one is missing or was truncated, re-run that agent" — the stated failure condition includes truncation, but the enumerated check cannot detect it; mirroring the three checks already required of writers (line count, first line, complete last line, SKILL.md:414) would make the two halves of the sentence agree.

Issue #886, opened 2026-09-03
collect: resources/ is redacted only on the --skip-logs path, not on the default full copy


kalantar:
`redact_yaml_tree` runs on `i<N>/resources/` only when `_copy_workload_iterations_full` is called with `redact_resources=True`, which only the `--skip-logs` branch of `_extract_phases_from_pvc` does (`pipeline/deploy.py`). The default full-copy path pulls `resources/` without redacting, so the redactor meant to stub sensitive fields before they reach `results/` does not run on the path operators normally use.

Found while implementing #885. Not changed there to keep that PR scoped to the copy mechanism; #885 turned the difference into an explicit `redact_resources` argument so it is greppable rather than buried in two divergent copy loops.

Acceptance: `resources/` is redacted on every collect path, with a test that fails if either path skips it.

kalantar:
Adjacent gap in the same call site, surfaced by review of #887: the guard around `redact_yaml_tree(iN_dest / "resources")` in `_copy_workload_iterations_full` catches only `OSError`.

`redact_yaml_file` guards its own reads and writes, but `yaml.safe_dump_all` on its write path is not wrapped, and `redact_yaml_tree`'s own `is_dir`/`rglob` walk is unguarded. A non-`OSError` failure from either therefore escapes the best-effort guard and escalates one workload's redaction hiccup into a whole-slot failure — the blast-radius problem #885 fixed at every other layer.

Not changed in #887 to keep that PR's final state identical to what was reviewed. Whoever fixes this issue will be in that exact code; broadening the guard to `(OSError, yaml.YAMLError)` there is the small half, hardening `redact.py`'s serialization path is the rest.

Issue #883, opened 2026-09-03
deploy.py: --skip-logs docstring says it skips EPP logs, but the code copies them (skips server_logs only)


kalantar:
## Problem

`_extract_phases_from_pvc`'s docstring says `--skip-logs` omits EPP logs. The code copies them in full.

Docstring (`pipeline/deploy.py`, `_extract_phases_from_pvc`):

> When *skip_logs* is True, only trace files are copied (skipping vLLM and EPP log files which typically account for the bulk of the data).

The inline comment immediately below it says the opposite, and matches the code:

```python
# Selective copy: trace files + epp_logs via kubectl cp per
# iteration; skips vLLM server_logs which dominate data volume.
```

And the code does copy `epp_logs/`:

```python
# Copy epp_logs directory
epp_dest = iN_dest / "epp_logs"
epp_dest.mkdir(exist_ok=True)
r = run(
    ["kubectl", "cp", f"{remote_prefix}/epp_logs/",
     str(epp_dest), "--retries=3"],
    check=False, capture=True,
)
```

`skip_logs` skips `server_logs/` only. `epp_logs/`, `gpu_logs/`, `metrics/` and `resources/` are all still copied.

## Why it matters

The docstring's claim is the reason an operator reaches for the flag, and on EPP-heavy runs it is backwards. Measured on `pd-infocomm-3` run `try3`, cell `causalsloexternality/reasoning-single-turn/i1` (total 433 MB):

| subdirectory | size | `--skip-logs` behaviour |
|---|---|---|
| `epp_logs/` | 324 MB | **copied** |
| `metrics/` | 82 MB | copied |
| `server_logs/` | 24 MB | skipped |
| `gpu_logs/` | 2.4 MB | copied |
| `resources/` | 836 KB | copied |

So the flag skips 5.5% of the payload while the docstring implies it skips the bulk. An operator hitting a transfer timeout (see the companion collect-timeout issue) reads the docstring, sets `--skip-logs`, and fails at the same place — because 324 MB of EPP logs still go over the wire in a single `kubectl cp`.

This is also user-visible beyond the docstring: `_cmd_collect`'s own help text and the runtime hint printed on the >1 GB size probe (`"Logs make up most of the size. Re-run with --skip-logs to collect traces only."`) carry the same implication. "Traces only" is not what the flag does.

## Fix options

Either is fine; the point is that the three statements agree.

[...]

PR #887, opened 2026-09-03
fix(collect): inventory-driven incremental copy with a real completeness marker (#885)

### LINKED ISSUE #885: collect: iterations over ~420MB fail on a 120s guard, retry cannot converge, and partial copies are reported as complete
## Summary

`sim2real collect` cannot copy a result iteration larger than roughly 420 MB, retrying does not make progress, and a partially-copied iteration is subsequently reported as complete and skipped. The operator is told none of this.

Four defects, in increasing order of severity.

## 1. A whole-iteration `kubectl cp` runs under a 120 s hung-cluster guard

`_copy_workload_iterations_full` (`pipeline/deploy.py`) copies an entire `iN/` directory in a single `kubectl cp` with no `timeout=` override, so it inherits `deploy.run`'s default:

```python
def run(cmd, *, check=True, capture=False, cwd=None, timeout: int = 120):
    """``timeout`` defaults to 120 s so kubectl/helm calls against a hung or
    unreachable cluster do not block the deploy process indefinitely (#694)."""
```

That guard is correct for a control-plane call and wrong for a bulk transfer. Measured throughput through the API server on `pokprod001` is ~3.5 MB/s, so 120 s caps an iteration at ~420 MB. `--retries=3` does not help — it is a kubectl-internal flag, and the Python-side timeout kills the process.

Observed on `pd-infocomm-3` run `try3`, reconstructed from copy mtimes:

| cell | size | outcome |
|---|---|---|
| causalslo/deep-research | 209 MB | ok (~60 s) |
| baseline/reasoning | 402 MB | ok (~120 s, marginal) |
| leastttft/reasoning | 489 MB | ok (~120 s, marginal) |
| **causalslo/reasoning** | **433 MB** | **timeout, twice** |

Several cells squeaked under the wire; one sat just over it.

## 2. Retry restarts from zero instead of continuing

```python
if iN_dest.exists():
    shutil.rmtree(iN_dest)          # discards everything that landed
iN_dest.mkdir(parents=True, exist_ok=True)
result = run(["kubectl", "cp", f"{namespace}/{pod_name}:/data/{run_name}/{phase}/{wl_name}/{iN}/",
              str(iN_dest), "--retries=3"], check=False, capture=True)
```

Each attempt wipes the destination and re-streams the whole directory, so every attempt gets the same ~120 s budget and dies at roughly the same place. Two real attempts on the failing cell truncated at `..._0010.log` (9 files landed) and `..._0020.log` (10 files landed) respectively — different cut points, no accumulation. The operation cannot converge by repetition.

## 3. A partial iteration is reported complete and then skipped — silent data loss

```python
def _is_iteration_up_to_date(iN_dir, remote_mtime):
    csv = iN_dir / "trace_data.csv"
    return csv.exists() and csv.stat().st_mtime >= remote_mtime
```

Completeness is inferred from `trace_data.csv` alone. That file is copied early in the tar stream, so it survives a timeout that kills the rest. Consequence: after a failed copy, the next `collect` sees the iteration as up to date, prints `up to date — skipping`, and reports success — permanently retaining a truncated log set.

This is the worst of the four, because it turns a loud failure into a clean-looking run. In this instance a 41,497,088-byte EPP log ending mid-JSON-record survived two subsequent "successful" collects. The truncation was only found by checking for a trailing newline on every log file.

## 4. The failure message carries almost no actionable information

```
09:13:13 [WARN]  Extractor pod failed in kalantar-0: Command '['kubectl', 'cp',
  'kalantar-0/sim2real-extract:/data/try3/causalsloexternality/reasoning-single-turn/i1/',
  '/.../results/causalsloexternality/reasoning-single-turn/i1', '--retries=3']'
  timed out after 120 seconds
09:13:13 [WARN]  causalsloexternality/reasoning-single-turn    (kalantar-0)
```

It reports the raw `TimeoutExpired` argv. It does not say that data was partially written, which files are incomplete, that `trace_data.csv` survived, or that re-running will silently skip the cell rather than retry it. All four are known at the point of the warning. The operator's reasonable reading — "most of it copied, I will just run it again" — leads directly to defect 3.

## Proposed fix

**Copy per file, not per iteration.** The largest single file anywhere in `try3` is 141,525,689 bytes (~40 s at 3.5 MB/s), comfortably inside the existing 120 s budget. With per-file granularity no individual copy can reach the ceiling, and the 433 MB cell completes in one pass. The `skip_logs` path already copies per-subdirectory, so the pattern exists in the file.

**Replace `rmtree` with an inventory check.** Enumerate the remote iteration (`ls -l`), compare name and size against local, and copy only what is missing or short. A retry then costs only the delta. `kubectl cp` cannot resume mid-file, so a truncated file is re-copied whole — acceptable at a 141 MB worst case.

**Write a real completeness marker.** Only after every file in the remote inventory is present at the correct size, write e.g. `iN/.collect_complete` recording the remote inventory hash and timestamp. `_is_iteration_up_to_date` should require that marker; `trace_data.csv`'s mtime is not evidence of anything beyond `trace_data.csv`.

**Tell the operator what actually happened.** On a partial copy, report the partial state and the concrete next step. Something like:

```
[WARN] causalsloexternality/reasoning-single-turn/i1 — PARTIAL COPY
       transferred 9 of 13 files (410 MB of 433 MB) before the 120 s limit
       incomplete: epp_logs/..._0010.log (truncated at 41,497,088 B)
       missing:    epp_logs/..._0016.log, epp_logs/..._0022.log, metrics/raw/...
       trace_data.csv is present and complete — the workload's trace is safe
       ACTION: re-run `sim2real collect --run try3` to fetch only the missing files
```

The three things the operator needs are: this is partial and not a total loss; the primary artifact is or is not safe; and re-running will resume rather than no-op. A summary line at the end of a multi-cell collect should also list every cell left partial, so a failure in cell 3 of 9 is not lost in the scrollback.

**No `--timeout` flag is required.** Per-file granularity puts every copy inside the existing default. A flag would only become useful if some future artifact were itself larger than a 120 s budget; it should not be the primary mechanism, because the operator cannot know the right value before the copy starts.

## Acceptance criteria

- [ ] An iteration of arbitrary total size copies successfully, provided no single file exceeds the per-file budget.
- [ ] A partial copy followed by `collect` re-run converges: only missing or short files are transferred.
- [ ] A partially-copied iteration is never reported as complete, and never skipped by the up-to-date check.
- [ ] The completeness marker is what the up-to-date check consults; `trace_data.csv` mtime alone is not sufficient.
- [ ] On partial copy the operator is told: that it is partial, byte/file counts, which files are incomplete or missing, whether `trace_data.csv` is intact, and the exact command to resume.
- [ ] A multi-cell collect ends with a summary listing all partial cells.
- [ ] Regression test: simulate a timeout mid-iteration, assert the marker is absent, assert the next collect transfers only the delta and then writes the marker.

## Provenance

Found while running `/sim2real-check` against `pd-infocomm-3` run `try3` (translation `cee8bb60cbc3`, 9 cells). The operator hit defect 1, retried per defect 4's implication, and landed in defect 3 — two subsequent collects reported success while leaving a truncated 41 MB EPP log in place. Related: #883 (`--skip-logs` does not skip EPP logs, so it is not a viable workaround for this).

kalantar:
Closes #885

## What changed

Each `i<N>/` copy now inventories the remote iteration once (`find . -type f -exec stat -c '%s|%n' {} +` in the extractor pod), diffs it against local files by name and size, and transfers only the delta — one `kubectl cp` per remote top-level entry when nothing has landed yet, then one per individual file for anything still outstanding. The destination is never wiped, so a retry costs only the delta.

`i<N>/.collect_complete` is written only after every inventory entry is present locally at the matching size, and is the only evidence the up-to-date gate accepts.

Both copy paths (default and `--skip-logs`) now share one helper; the `--skip-logs` branch loses ~90 lines of duplicated per-subdirectory `cp` blocks.

| Defect (#885) | Fix |
|---|---|
| 1. Whole-iteration cp under the 120 s guard | per-entry then per-file granularity |
| 2. Retry `rmtree`s and restarts from zero | inventory delta; destination never wiped |
| 3. Partial iteration reported complete | `.collect_complete` gates the skip |
| 4. Failure message is a raw argv | `PARTIAL COPY` report + `Partial:` run summary |

## Reviewer attention

**Three things the issue didn't name, which I found while vetting and fixed here.**

1. **The timeout aborted the whole slot, not just the cell.** `lib/proc.py:run` passes `timeout=` straight to `subprocess.run`, which raises regardless of `check=False`. So `TimeoutExpired` never reached the per-iteration error append at the old `deploy.py:1075` — it unwound past every remaining iteration *and* workload to the slot-level handler, and later cells never got an `on_workload_done` callback. Every copy is now wrapped individually (`_cp_one`).

2. **`--skip-logs` had both defects too.** It used the same `_is_iteration_up_to_date` and copied `epp_logs/` as one directory, so the same ceiling applied. Now routed through the shared helper with `exclude_subdirs={"server_logs"}`.

3. **Pre-#885 runs migrate for free.** Requiring the marker means every already-collected iteration is re-examined once. That costs one remote inventory and *zero* transfers, because the delta against a complete local tree is empty — the marker is then just written. Test: `test_zero_transfer_when_local_already_matches_remote`.

**A design decision beyond the issue's proposal: stale-file pruning.** Dropping the `rmtree` would have let files from an earlier collect of the same iteration pile up beside new ones (e.g. `epp_logs/` from a re-run workload), which analysis tools glob over. So files absent from the remote inventory are deleted before copying. This keeps the wipe's one useful effect without its data loss: **a file the remote still holds is never touched, fetched or not.** Pruning compares against the *unfiltered* inventory, so `--skip-logs` still clears a stale `server_logs/` the PVC dropped while leaving one it still has (rather than deleting what an earlier full collect fetched). A failed probe prunes nothing — a transient error must not become data loss.

**Behavior I deliberately did not change.** `redact_yaml_tree` still runs on `resources/` only for `--skip-logs`; the default full-copy path leaves it unredacted. That predates this PR and is now filed as #886. Rather than leave the difference buried in two divergent loops, it is an explicit `redact_resources=` argument, scoped to iterations the call actually copied (the intermediate version I wrote globbed every local `i*`, which would have rewritten other slots' iterations). #883 (`--skip-logs` docstring claims it skips EPP logs) is also untouched — the exclusion set is still `server_logs` only.

**Verified the remote command against the real image** rather than assuming BusyBox semantics — `alpine:3.19` is what the extractor pod runs, and it has no GNU `find -printf`:

```
$ podman run --rm alpine:3.19 sh -c 'cd /d && find . -type f -exec stat -c "%s|%n" {} +'
2|./epp_logs/a.log
0|./metrics/raw/m.log
4|./trace_data.csv
```

Missing directory → rc=2 (becomes an error string, no copy attempted). Empty directory → rc=0 with empty output (empty inventory, no error). The `%s|%n` format is also what test mocks key on to tell this probe from the mtime probe's `%Y %n`.

## Test changes

New `pipeline/tests/test_collect_incremental.py` (54 tests) covers the inventory parser, delta, two-pass copy, timeout containment, marker write/gate, pruning, and convergence-on-retry. New code in `deploy.py` is at full line coverage.

Existing tests needed updating in three ways, all noted inline with `#885`:

- 48 `mock_extract` signatures gained `partials=None`.
- Mocks that answered the copy path now answer the inventory probe. Most are not about the copy set and return an empty inventory; the ones that assert on it return a canned list and materialize files via a new `_materialize_cp` helper.
- Four `_is_iteration_up_to_date` tests asserted the `trace_data.csv` semantics and three asserted the wipe. Rewritten against the marker and the delta copy. The #564 cross-slot preservation tests still pass on their original assertions — that invariant is unchanged, only the evidence for "complete" moved.

`test_collect_idempotent_no_stale_accumulation` previously held because of the `rmtree`; it now holds because of pruning, and its docstring says so.

## Stale-reference sweep

Grepped `*.md`, `docs/`, `.claude/skills/`, `README*` for `trace_data.csv`, `_is_iteration_up_to_date`, `up to date — skipping`, `iteration granularity`, `collect_complete`.

- Updated: `pipeline/README.md` (collect section rewritten; `.collect_complete` added to the results-layout list), `CLAUDE.md` (artifact table row), `.claude/skills/sim2real-analyze/SKILL.md` (layout tree, with a note that an absent marker means the cell may be truncated).
- Checked and left alone: `sim2real-check/SKILL.md` uses `trace_data.csv` existence as a *presence* predicate for enumeration, which is still correct for that purpose. Its phase/workload discovery uses plain `ls` and `ls -d */`, and iteration discovery uses `_ITER_DIR_RE` + `is_dir()` — none of which see a dotfile, so `.collect_complete` is inert there. Same for `lib/resolve.py:_workload_has_data` and `deploy.py:_has_iN_subdirs`.
- Left alone: `docs/superpowers/plans/*` and `docs/epics/*` are records of past work.


kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES, `comment-analyzer` said NEEDS DISCUSSION; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 3 important · 0 recommendations** across 4 lenses.

The incremental-copy design is sound and well tested, but three narrow gaps around the degraded paths — a false \"trace is safe\" claim, a full re-download when the mtime probe fails, and a doc line that condemns every legacy run — should be closed first.

### Important
- **`trace_ok` reports the trace safe on the inventory-probe-failure path, where nothing was ever examined.** (`pipeline/deploy.py:1187`) — `_copy_iteration_incremental` returns at deploy.py:1376-1378 when `_remote_file_inventory` errors, before the verification loop that populates `missing`/`short`; `trace_ok` (deploy.py:1187) is therefore True by default and `_report_partial` (deploy.py:1235) prints "trace_data.csv is present and complete — the workload's trace is safe" alongside "transferred 0 of 0 files" and the inventory error. Issue #885 defect 4 is specifically about the failure report telling the operator the truth; `test_inventory_error_yields_no_copy_and_no_marker` (test_collect_incremental.py:445) asserts on `complete` and `errors` but not `trace_ok`, so the false-positive is unexercised.
- **A failed mtime probe forces a full re-download of every already-complete iteration, not the delta the README promises.** (`pipeline/deploy.py:1401`) — `_is_iteration_up_to_date` returns False whenever `remote_mtime is None` (deploy.py:1018), regardless of the marker, so a failed `_probe_remote_mtimes` sends every iteration into `_copy_iteration_incremental`, where `stale_marker = _read_collect_marker(...) is not None` (deploy.py:1401) sets `delta = sorted(remote_inv)` — the whole inventory. That contradicts the new README sentence "If the mtime probe fails (e.g., pod not running), ... it now costs a delta rather than a full re-download", and it re-exposes the ~420 MB timeout ceiling in exactly the degradation path the docs claim is now cheap. `test_stale_marker_forces_a_full_refetch` only covers a genuinely newer `remote_mtime=999.0`; the None case is untested. Either narrow the stale-marker branch to a real mtime comparison or correct the README claim.
- **The new tree comment says an absent marker means the iteration was copied partially, which is false for every run collected before this change.** (`.claude/skills/sim2real-analyze/SKILL.md:190`) — The marker is absent for an iteration never collected and for any iteration collected pre-#885 — deploy.py:1015-1017's own docstring says as much ("Iterations collected before #885 have no marker and are therefore not skipped"), and the CLAUDE.md row added in this same diff phrases it correctly ("Absent ⇒ the iteration is resumed, never skipped"). As written, SKILL.md tells an analyst that every existing run's iterations may be truncated mid-record.

kalantar:
Pushed `dc6f543` addressing the review. All three points were valid — I confirmed each against the code rather than taking them as given — and I fixed a fourth that a lens raised but the verdict didn't carry.

The trace verdict is now three-state instead of two. `missing` and `short` only get populated by the post-copy verification loop, and an inventory-probe failure returns before that loop ever runs, so both were empty because nothing had been looked at — not because everything was fine. The report was printing "trace_data.csv is present and complete — the workload's trace is safe" right next to "transferred 0 of 0 files" and the probe error. An `IterationCopy` now records whether verification actually ran, and reports UNKNOWN when it didn't.

The refetch decision now needs positive evidence that the remote moved on. The old check was just "does a marker exist", which meant a failed mtime probe got read as "remote advanced" and refetched the entire inventory — and a failed mtime probe is exactly what routes every iteration into the copy in the first place, since it blocks the skip. So the one path the README described as cheap was the one that re-downloaded everything and re-exposed the ~420 MB ceiling. The mtime comparison now lives in a single `_marker_covers` helper shared by the up-to-date gate and the copy, so those two can't drift on what "stale" means. A marker with no comparable mtime no longer forces a refetch; an unparseable one is no longer trusted.

An unparseable `stat` line now fails the whole inventory instead of being dropped with a warning. This is the one the verdict didn't include. I fixed it because a dropped entry can never show up in `missing` or `short`, so the iteration could be marked complete and skipped permanently while that file was never fetched — the same silent data loss as defect 3, just moved down into the parser. Three lines, and it closes the failure mode this issue is about.

Fixed the `sim2real-analyze` layout note too. It claimed an absent marker means the iteration was copied partially; it actually means completeness is unproven, which is true of every run collected before this change and of any iteration that was never collected. The `CLAUDE.md` row in the same diff already had this right.

Two doc claims are now corrected to match the code: the degradation-path sentence (one inventory plus the size delta, and nothing at all for an already-complete iteration) and the partial-copy report description (three trace states, not two).

Nothing was skipped. Tests: 2931 passed, 9 skipped, 2 xfailed; `ruff --select F` clean; coverage 97.82%, with the new code still at full line coverage. Seven new tests cover the specific gaps — the UNKNOWN trace state at both the record and report level, a failed mtime probe transferring nothing for a complete iteration, a corrupted marker mtime forcing a refetch, the shared staleness predicate, and the parser failure both in isolation and end-to-end.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `silent-failure-hunter` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**1 must-fix · 0 important · 1 recommendation** across 4 lenses.

Solid work overall, but the new completeness marker still makes an unqualified claim it cannot support on the `--skip-logs` path, which reintroduces silent permanent data loss for `server_logs`.

### Must fix
- **A marker written under `--skip-logs` satisfies a later full collect, so `server_logs` is silently never fetched.** (`pipeline/deploy.py:1481`) — At deploy.py:1410-1412 the remote inventory is filtered by `exclude_subdirs={"server_logs"}` (set at deploy.py:1697 for the `--skip-logs` path), `res.complete` is computed over that filtered set, and deploy.py:1481 writes the marker from it with no record of what was excluded; `_is_iteration_up_to_date` (deploy.py:997, called at 1526) decides purely from `_marker_covers` and skips the iteration before the copy helper ever runs. A `collect --skip-logs` followed by a plain `collect` on the same run therefore prints "up to date — skipping" and never transfers `server_logs`, with no warning — while pipeline/README.md states the marker's presence is "the only evidence `collect` accepts that an iteration transferred in full." test_collect_incremental.py:496 pins the current behavior ("must not block the marker"), and the >1 GB prompt at deploy.py:1643-1646 actively steers operators into exactly this sequence. Recording the excluded set in the marker and refusing to honor a narrower-scope marker closes it.

### Recommendations
- **The end-of-collect `Partial:` summary is never exercised through `_cmd_collect`.** (`pipeline/deploy.py:1936`) — `_print_partial_summary` and the `partials.append` in `_copy_workload_iterations_full` are each unit-tested, but no test drives `_cmd_collect`/`_extract_phases_from_pvc` with an actual incomplete iteration (no test asserts on the `Partial:` line), so dropping `partials=partials` from any of the ~8 call sites would pass the suite while losing the summary.

kalantar:
Pushed `6da4c05`. The must-fix was real and I'd seen it coming — a lens raised it as non-blocking last round and I noted it and moved on because the verdict hadn't promoted it. That was the wrong call; it's a silent permanent data loss and I should have fixed it then.

The marker now records what it did not look at. A `--skip-logs` collect writes `excluded_subdirs: ["server_logs"]`, and a marker covers a request only when everything it omitted is also omitted now. A full collect's marker still satisfies a later `--skip-logs` one, since the data is already there; the reverse no longer holds. A marker with no scope field isn't honoured — one inventory, zero transfers, then it gets rewritten with the field. Worth noting the >1 GB size prompt actively tells operators to run `--skip-logs` first, so the losing sequence was the recommended one.

While verifying that fix I found the test I'd written for it was passing for the wrong reason, which is worth flagging because it would have hidden a regression. `_FakePVC` answered both `find`+`stat` probes with inventory-format lines, so `_probe_remote_mtimes` received unparseable input and returned an empty dict — which makes `remote_mtime` None and disables the up-to-date gate entirely. Any test depending on that gate was vacuous. The fake now tells the two probes apart by stat format (`%s|%n` vs `%Y %n`). I then confirmed the guard actually bites by temporarily defeating `_marker_scope_covers` and checking the tests fail — three did before the fake was fixed, four after, including the end-to-end one.

I also went after something the verdict didn't include. `_copy_iteration_incremental`'s docstring says "Never raises", and that was only true for `TimeoutExpired`. `check=False` rules out `CalledProcessError` but not the `OSError` family — an unresolvable kubectl, a fork failure, a full disk on the marker write. Each would have unwound past every remaining iteration and workload to the slot handler, and in the no-progress fallback branch (which caught only `RuntimeError`) crashed the entire collect with no summary and no per-iteration report. That's the exact propagation this PR exists to stop, so I'd rather enforce the contract than keep asserting it. The inventory probe, `_cp_one`, the destination mkdir and the marker write now all record and return. A failed marker write drops the completeness claim rather than failing the copy — the files did land; what's lost is the proof, so the next collect re-examines instead of skipping.

Took the recommendation too. The partials accumulator and the summary printer were each unit-tested but nothing drove the wiring between them, so dropping `partials=partials` from any of the call sites would have passed the suite while losing the summary. There's now a test that goes through `_extract_phases_from_pvc` with a genuinely incomplete iteration.

Docs updated to match: the README's marker description is scope-qualified, and the `CLAUDE.md` row mentions `excluded_subdirs`.

Nothing skipped. `ruff --select F` clean, 2942 passed / 9 skipped / 2 xfailed, coverage 97.81%.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `code-reviewer` said NEEDS CHANGES, `silent-failure-hunter` said NEEDS CHANGES, `pr-test-analyzer` said NEEDS CHANGES; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 1 important · 1 recommendation** across 4 lenses.

The implementation is solid and well-tested, but one exception handler this PR deliberately widened has no test driving anything through it, so a future narrowing would silently reintroduce the crash it was widened to prevent.

### Important
- **The no-progress fallback's handler was widened from `except RuntimeError` to `except Exception` with no test exercising the new path.** (`pipeline/deploy.py:2362`) — Commit 6da4c05 changes `except RuntimeError` to `except Exception` at deploy.py:2362 and its message names the defect it fixes (a non-RuntimeError surviving the copy layer crashed the whole collect with no summary). Running the exact CI suite with coverage shows lines 2362-2365 — the entire handler body — uncovered; the only test on this branch, `test_collect_fallback_no_progress` in pipeline/tests/test_deploy_collect.py:91, uses a mock that never raises, while the sibling branches did get `test_collect_unscoped_parallel_non_runtime_error` (OSError) and `test_collect_scoped_runtime_error`. A revert to `except RuntimeError` would keep all 2942 tests green.

### Recommendations
- **`_prune_absent_locals`'s empty-directory cleanup is the one remaining hole in the "Never raises" contract this PR enforced everywhere else.** (`pipeline/deploy.py:1430`) — deploy.py:1430-1434 walks `iN_dest.rglob("*")` and calls `path.iterdir()` unguarded, and the call site at line 1494 is unwrapped, so an OSError there escapes `_copy_iteration_incremental` (whose docstring promises it never raises) up to the slot-level handler — while commit 6da4c05 explicitly wrapped the inventory probe, `_cp_one`, the mkdir, and the marker write for exactly this reason. The `except OSError: pass` on `rmdir` in the same loop is also the only silent catch in a function whose sibling unlink failure warns. The trigger is a local-filesystem error on a directory this process just created, so this is a contract-tidiness note, not a live bug.

kalantar:
Pushed `74152ca`.

The important item was right and the catch was a good one: I widened that fallback handler from `except RuntimeError` to `except Exception` and never tested it, so the whole handler body was uncovered and narrowing it back would have kept the suite green. There's now a test that raises an OSError through `_cmd_collect` and asserts both the warning and that the summary still prints. I confirmed it bites by temporarily narrowing the handler again and watching it fail.

I took the recommendation as well. `_prune_absent_locals`'s directory cleanup was the last hole in the never-raises contract the previous commit enforced everywhere else, and its `except OSError: pass` on rmdir was the only silent catch in a function whose sibling unlink failure warns. Both fixed — the walk and each rmdir are guarded, and the failure is logged, because a recurring permission problem or a pile of undeleted empty dirs otherwise leaves nothing behind for whoever is debugging disk or inode growth.

Writing that test turned up a wider version of the same hole than was reported: `_local_file_inventory`'s `rglob` and `is_dir` are unguarded too, and it's called three times from `_copy_iteration_incremental`. Rather than patch the two spots that were named, I swept the whole #885 region mechanically for unguarded filesystem and subprocess calls. That found three more: that `is_dir`, the workload-directory mkdir in `_copy_workload_iterations_full`, and the `redact_yaml_tree` call site — the callee guards its own reads and writes but not its directory walk, and I'm the one who moved that call into the shared loop. All of them now report instead of raising, and the sweep comes back clean.

Worth stating explicitly since it's a design choice rather than a fix: under-reporting the local inventory is the safe direction. A missing entry is treated as needing a copy, so a failed walk can only cost an extra transfer — it can never cause a file to be skipped. That's documented at the guard.

One note on this review round for anyone reading the thread. The code-reviewer lens reported that the diff "contains no implementation of any acceptance criterion — it only edits two documentation files" and that `pipeline/deploy.py` is unchanged. That's not so; it appears to have diffed a single commit rather than the branch. The adversarial synthesizer correctly did not carry that finding into the verdict. Separately, the harness reported that the safety classifier was unavailable when reviewing the comment-analyzer agent's output — nothing I acted on this round came from that lens.

Nothing skipped. `ruff --select F` clean, 2950 passed / 9 skipped / 2 xfailed, coverage 97.83%, new code at full line coverage.

kalantar:

> **Automated review.** Generated by the `review-changes` workflow from the authoring session, not a human sign-off. No human has reviewed this change. 4 specialist lenses ran in isolated fresh contexts (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`), then an adversarial verifier-synthesizer re-read the code before ruling. Lens verdicts were not unanimous — `silent-failure-hunter` said NEEDS CHANGES, `comment-analyzer` said NEEDS DISCUSSION; the synthesizer's ruling is below and the per-lens findings are returned to the caller. Recommendation only — the merge decision is the reviewer's.

**0 must-fix · 1 important · 1 recommendation** across 4 lenses.

The incremental-copy machinery and its tests hold up under scrutiny, but the marker's handling of an unknown remote mtime reintroduces a permanent silent skip that should be settled before merge.

### Important
- **A marker recording `remote_mtime: null` disables staleness detection for that iteration permanently, so a later genuine remote rewrite is skipped silently.** (`pipeline/deploy.py:1243`) — `_marker_covers` (deploy.py:1243-1247) returns True unconditionally when the recorded value is None, and `_is_iteration_up_to_date` (1025-1030) then skips the iteration for every subsequent collect no matter how new the current probe's mtime is. A null value is written whenever `wl_remote_mtimes.get(iN)` was None at copy time (1642) — the phase-wide probe failed, or the iteration dir exists on the PVC without a `trace_data.csv` (`_probe_remote_mtimes`, 946-977, keys only off trace files). Re-running that iteration afterwards rewrites the same remote path, and collect now skips it and reports success, whereas main's gate (git show main:pipeline/deploy.py, old `_is_iteration_up_to_date`) re-copied. The choice is deliberate and tested (test_up_to_date_when_marker_has_no_recorded_remote_mtime), but its stated rationale — avoiding "re-copying forever" — is undercut by the module's own docstring at 1022-1024, which notes that re-examining an unskippable complete iteration costs one inventory and zero transfers.

### Recommendations
- **Copy errors that a later pass recovers from are discarded, so a bulk-pass timeout leaves no trace anywhere.** (`pipeline/deploy.py:1661`) — `res.complete` is set from the missing/short verification only (1581), and `_copy_workload_iterations_full` does `if res.complete: continue` at 1661 before touching `res.errors` — so a Pass-1 bulk `kubectl cp` that times out (recorded by `_cp_one` at 1400) and is then recovered file-by-file in Pass 2 is never logged at any level. Logging non-empty `res.errors` even on a complete iteration would give the operator the one signal that says the 120s guard is being hit at bulk granularity.

kalantar:
Pushed `1992942`. This one was a real regression I introduced, not a polish item, and the reviewer's reasoning was better than mine.

`_marker_covers` treated `remote_mtime: null` as proof of coverage, on the grounds that the marker "still proves the transfer completed". It proves the transfer completed *at that time* and says nothing about whether the remote has moved since. So once an iteration picked up a null-mtime marker, every later collect skipped it and reported success no matter how new the remote got — a genuine re-run of that cell would be silently discarded. That's strictly worse than the code this issue set out to fix: the old gate re-copied whenever it had no mtime to compare. And my stated rationale for it, avoiding a re-copy loop, was already contradicted by the docstring three lines above, which says re-examining a complete iteration costs one inventory and zero transfers. A null gets written whenever the phase-wide probe produced no mtime for the iteration — the probe failed, or the iteration exists on the PVC with no `trace_data.csv` for the probe to key off.

Fixing it meant splitting the two decisions that read the marker, because they turn out to ask different questions. "Skip this iteration?" needs a comparable mtime, so absent and corrupt both decline. "Refetch everything instead of the size delta?" needs positive evidence the size diff can't be trusted — an mtime older than the remote's, or a corrupt marker — and absence of information is neither, so that case stays on the cheap path. Collapsing both into one boolean is what let the bug hide. `_marker_mtime` now classifies the field once as ok / absent / corrupt and each decision reads what it actually needs.

Took the recommendation too, and it's a good catch about the shape of the code rather than a missing guard: `complete` is decided purely from the final missing/short check, so a bulk first-pass copy that hit the 120s limit and was then rebuilt file-by-file got dropped at `if res.complete: continue` and left no trace at any level. That's the only signal that would tell you a single top-level subtree has grown big enough to cost a wasted timeout on every cold copy, so it now prints as `recovered after N copy failure(s)` with the causes.

I verified both by reverting each in turn and confirming the new tests fail — three for the mtime gate including the end-to-end null-marker-then-real-mtime sequence, one for the recovery report. I also tightened the recovery test after writing it, because my first version's mock condition was convoluted enough that I couldn't tell whether it was passing for the right reason; it now asserts the bulk copy was actually attempted and actually failed and that both per-file retries ran.

The README's marker section had quietly grown into one unreadable paragraph as these rules accumulated. It's now split by the question each rule answers, and the mtime claim in it is corrected.

Nothing skipped. `ruff --select F` clean, 2954 passed / 9 skipped / 2 xfailed, coverage 97.85%, new code at full line coverage.

kalantar:

The implementation and its test suite hold up under verification; one docstring left asserting the opposite of the safety rule the final commit established needs correcting first.

### Important
- **`_is_iteration_up_to_date`'s docstring states a marker with a null recorded mtime "is trusted", the opposite of what the code now does.** (`pipeline/deploy.py:1012`) — deploy.py:1012-1014 says "A marker whose own remote_mtime is None was written when the probe had failed; it still proves completion, so it is trusted", while `_marker_covers` (deploy.py:1254-1270) and its test (test_collect_incremental.py:693, `_marker_covers({"remote_mtime": None}, 500.0) is False`) establish the reverse. Commit 1992942 deliberately reversed that semantics — it rewrote `_marker_covers`'s docstring and README:716 but missed this one — so the file now documents, at the entry point to the gate, the exact permanent-silent-skip behavior that commit removed.

kalantar:
Pushed `7bf09e6` — docstring only, one hunk, no behavior change.

The finding was right and it's the kind that causes the next regression rather than being one: commit `1992942` reversed the null-mtime semantics and updated `_marker_covers`'s docstring and the README, but left `_is_iteration_up_to_date` — the entry point to the gate — still asserting that a marker with no recorded mtime "still proves completion, so it is trusted." That is exactly the permanent silent skip that commit removed, sitting where a maintainer would read it instead of the test. It now points at `_marker_covers` as the owner of the rule rather than restating it, so the two can't drift again. I swept the rest of `pipeline/` and the skills for the same claim; no other occurrence.

Three non-blocking items from the silent-failure lens that the verdict did not carry, and my reasoning for each:

The `"no such file"` substring tolerance in `_cp_one` stays. It's pre-existing behavior carried forward, and the reviewer's own note says it cannot cause data loss because the post-copy inventory verification catches a genuinely missing file. Recording those as errors would also make the new `recovered after N copy failure(s)` line fire on every absent optional sentinel, which would bury the signal it exists to give.

Missing tracebacks on the `except Exception` sites is a fair point but spans six pre-existing call sites, five of which this PR never touched. That's its own change, not a rider on this one.

The `except OSError` around `redact_yaml_tree` genuinely is narrower than my own comment implies — `yaml.safe_dump_all` and the `rglob` walk aren't `OSError`. I've noted it on #886 rather than fixing it here, because that issue is already about this exact call site and I'd rather the final state of this PR be identical to what was reviewed than land an unreviewed logic change on the last iteration.

`ruff --select F` clean, 2954 passed / 9 skipped / 2 xfailed, coverage 97.85%.

